PR #76, opened 2026-05-05
feat(group): approver UI for incoming join requests

rinat-enikeev:
## Summary

PR 2 of the new-member announcement stack. Stacked on #75 (member-profile data model).

Wires a SwiftUI surface on top of the existing `JoinRequestApprover` so admins can drive the approve / decline pipeline by hand. The backend transport doesn't change — this PR just exposes what's already there. Later PRs in the stack add wire format + fanout.

### Pieces

- **`ApproveRequestsFlow`** — `@Observable @MainActor` wrapper around `JoinRequestApprover`. Mirrors `IdentitiesFlow`'s posture: single shared instance, idempotent `start()`, fire-and-forget intents. Maps `ApproveOutcome` to user-facing reason strings.
- **`JoinRequestApproving` protocol** — small test seam over the approver actor so the flow's bookkeeping can be unit-tested without standing up the full keychain + transport stack.
- **`ApproveRequestsView`** — modal sheet with Approve / Decline rows. Each row shows the joiner alias + group name + inbox-pubkey hex prefix as an out-of-band fingerprint, per the trust framing already documented on `JoinRequestPayload`. Empty state is the steady state.
- **`ApproveRequestsToolbarButton`** — `ChatsView` toolbar entry-point. Always rendered for discoverability; a numeric badge appears when `pending.count > 0`.
- **Wiring**: `OnymIOSApp.init` constructs the shared approver + flow, threads them through `AppDependencies`, and the WindowGroup `.task` calls `start()` so the collector runs from app boot regardless of which tab is mounted.

### Out of scope (later PRs in the stack)

- PR 3 — `MemberAnnouncementPayload` wire format
- PR 4 — Approver fanout to existing members on approve (also append joiner's `GovernanceMember` + profile to local roster)
- PR 5 — Inbox dispatcher merges incoming announcements into `ChatGroup.memberProfiles`
- PR 6 — UI: \"X joined\" + member roster

### Test plan

- [x] 7 new `ApproveRequestsFlowTests` covering: stream propagation, idempotent `start()`, approve/decline routing, error mapping for `unknownGroup` / `transportFailed`, dismiss-error
- [x] Full unit suite — 451/451 pass (3 skipped, pre-existing)
- [x] Existing UI test page objects don't reference the toolbar trailing items by accessibility id, so no regression
- [ ] Manual: launch on simulator, confirm toolbar button renders + badge appears when `IntroRequestStore` has entries, confirm modal sheet drives approve/decline. (Not run in this session — flagged for reviewer.)

### DIFF
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -18,4 +18,9 @@ struct AppDependencies {
     /// Settings → Identities screen observe the same state, so a
     /// factory closure here would split them.
     let identitiesFlow: IdentitiesFlow
+    /// Single shared instance — the toolbar badge on Chats and the
+    /// modal `ApproveRequestsView` observe the same `pending` list,
+    /// and the underlying collector should run for the app's
+    /// lifetime regardless of which surface is mounted.
+    let approveRequestsFlow: ApproveRequestsFlow
 }
--- a/Sources/OnymIOS/Chats/ChatsView.swift
+++ b/Sources/OnymIOS/Chats/ChatsView.swift
@@ -8,6 +8,7 @@ import SwiftUI
 struct ChatsView: View {
     let flow: ChatsFlow
     let identitiesFlow: IdentitiesFlow
+    let approveRequestsFlow: ApproveRequestsFlow
     let makeCreateGroupFlow: @MainActor () -> CreateGroupFlow
     let makeShareInviteFlow: @MainActor () -> ShareInviteFlow
 
@@ -28,6 +29,12 @@ struct ChatsView: View {
             ToolbarItem(placement: .topBarLeading) {
                 IdentityPickerMenu(flow: identitiesFlow)
             }
+            // Pending join requests — always rendered so the surface
+            // is discoverable even before the first request lands;
+            // the badge only appears when `pending.count > 0`.
+            ToolbarItem(placement: .topBarTrailing) {
+                ApproveRequestsToolbarButton(flow: approveRequestsFlow)
+            }
             // Plus button mirrors iOS Mail / Messages — useful once
             // the user already has at least one chat. Hidden in the
             // empty state because the central CTA already covers it.
@@ -44,6 +51,7 @@ struct ChatsView: View {
         }
         .task { flow.start() }
         .task { await identitiesFlow.start() }
+        .task { await approveRequestsFlow.start() }
         .fullScreenCover(isPresented: $showCreateGroup) {
             CreateGroupViewHost(
                 makeFlow: makeCreateGroupFlow,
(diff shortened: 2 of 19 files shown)


dev-onym:
Skimmed the diff against `feat/group-member-profiles`. Three things caught my eye on the new surface — happy to drill in further if you kick off the request-review flow for a structured pass, but worth flagging now:

**1. Decline is disabled exactly when its helper text tells the user to tap it.** `ApproveRequestsView.swift` row body:

```swift
.accessibilityIdentifier("approve_requests.decline_button.\(request.id)")
.disabled(request.groupName == nil)
…
if request.groupName == nil {
    Text("This request is for a group that isn't on this device. Decline to clear it.")
}
```

Approve disabling on `groupName == nil` is correct, but Decline should remain enabled — the actor's decline path doesn't need the group at all (it just drops the request + revokes the intro key). Right now an unknown-group request gets stuck on screen with no UI exit.

**2. Hardcoded strings, no `String(localized:)`.** The Russian pass landed in 59b97f5 and the rest of the app reaches for `String(localized:)` (Recovery, Settings, Chain). The new modal hardcodes "Join requests", "No pending requests", "People who tap one of your invite links show up here.", "Decline", "Approve", the unknown-group helper, and all five `failureReason` strings. Adding them to `Resources/Localizable.xcstrings` is cheap and avoids a regression on the localized build.

**3. `ApproveRequestsFlow.start()` has a concurrent-call race.** Both `WindowGroup.task` (in `OnymIOSApp.swift`) and `ChatsView.task` call it. `start()` is `@MainActor` but suspends on `await approver.start()` *before* assigning `streamingTask`, so two interleaved callers can both pass the `guard streamingTask == nil` check and end up with two streaming tasks subscribed to `approver.pending`. The second one wins the `streamingTask` slot, the first leaks. The existing idempotency test calls `start()` sequentially (`await flow.start(); await flow.start()`) so it doesn't catch this. Easy fix: set a `private var didStart = false` synchronously before the first await, or assign `streamingTask = Task { … }` first and have it await `approver.start()` from inside.

For inline patches, drop a `@<me> <instruction>` comment on the specific line. For the deeper review pass (test stubs, error-banner UX, accessibility identifier shapes for UI tests), `/review` from the PR will give you structured feedback.

dev-onym:
Hi @rinat-enikeev — I see I'm tagged on the PR body but don't have a specific instruction to act on. A couple of options:

- For a full structured review of the diff, request a review from me through the PR's review-request flow and I'll go through it properly.
- For a targeted change on a specific hunk (e.g. tweak `ApproveRequestsFlow`'s outcome→reason mapping, adjust the toolbar badge threshold, or factor `JoinRequestApproving` differently), drop an inline comment on the file/line with `@claude <instruction>` and I'll push a commit.

One thing worth flagging from the description while I'm here: the manual sim check is the only unchecked item in your test plan — happy to draft a short repro script (seed `IntroRequestStore` with a fake entry, walk approve + decline) if that would speed up the manual pass. Just say the word.

PR #87, opened 2026-05-05
feat(transport): Nostr relays config + Settings → Transport section

rinat-enikeev:
## Summary

PR 12 of the new-member announcement stack. Stacked on #86. **Closes the \"Couldn't send: not connected\" runtime failure** by giving the inbox transport actual relays to connect to and surfacing config in Settings.

The original gap: nothing in the app was calling `NostrInboxTransport.connect(to:)`. `send` threw `TransportError.notConnected` and `subscribe` silently no-op'd because `connections` was empty. Both legs of the inbox were dead from boot.

### What changes

**Data layer** (mirrors `RelayerRepository` shape, simpler — no fetch, no primary, no strategy):

- `NostrRelayEndpoint` — value type, `wss://` URL + display name + `isDefault` flag for UI badging. Distinct from `RelayerEndpoint` (chain relayer at `relayer.onym.chat`); Nostr relay is a separate service at `nostr.onym.chat` (strfry behind nginx).
- `NostrRelaysConfiguration` — `{ endpoints, hasUserInteracted }`. Sticky interaction flag prevents re-seeding after the user clears the list.
- `NostrRelaysSelectionStore` protocol + `UserDefaultsNostrRelaysSelectionStore` (production) + `InMemoryNostrRelaysSelectionStore` (tests).
- `NostrRelaysRepository` actor — first-launch seed with the Onym Official relay (`wss://nostr.onym.chat`); add / remove / clear / reset intents; `snapshots` stream.

**Wiring**:

- `OnymIOSApp` constructs the repo and reads `currentEndpoints()` in the WindowGroup `.task` **before** `InboxFanoutInteractor.run()`. The single missing call.
- `AppDependencies` + `RootView` + `SettingsView` pass through a `makeNostrRelaySettingsFlow` factory.

**UI**:

- `NostrRelaySettingsFlow` `@Observable` view-model (add / remove / reset).
- `NostrRelaySettingsView` — list with `DEFAULT` badge on app-shipped entries, custom-URL field with inline `wss://`/`ws://` validation, \"Restore default\" affordance.
- New **TRANSPORT** section in Settings between NETWORK and APP:
  - **Nostr Relays** (working) → drills into the new screen.
  - **Tor (Hidden Service)** — TBA placeholder row, disabled, \"Coming soon\" subtitle.

### V1 limitation

Changes apply on **next app launch** only — the inbox transport reads endpoints once at boot. Footnote on the screen calls this out. Live re-connect on config change lands in a follow-up alongside WebSocket reconnect-on-graph-change.

### Test plan

- [x] 13 new tests:
  - 8 `NostrRelaysRepositoryTests` — first-launch seed, sticky interaction flag, idempotent add, remove, reset, snapshots stream.
  - 5 `NostrRelaySettingsFlowTests` — `wss`/`ws` accept, `http`/bare reject, invalid-URL keeps draft + sets error, valid-URL clears + dispatches, snapshots drain into `state`.
- [x] Build green; full unit suite — 496/496 pass (3 skipped, pre-existing).
- [ ] **Manual smoke** (the original problem): with this PR installed, run the original Phase 2 of the smoke plan again — joiner taps invite link, taps Send, no \"not connected\" error, sheet flips to \"Waiting for the host to approve\". Admin sees the request, taps Approve, joiner gets \"You're in!\".
- [ ] Manual: open Settings → Transport → Nostr Relays. Verify Onym Official appears with DEFAULT badge. Add a custom relay (e.g. `wss://relay.damus.io`), confirm it lands. Remove it. Tap Restore default to confirm reset.

### DIFF
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -9,6 +9,7 @@ import Foundation
 struct AppDependencies {
     let makeRecoveryPhraseBackupFlow: @MainActor () -> RecoveryPhraseBackupFlow
     let makeRelayerSettingsFlow: @MainActor () -> RelayerSettingsFlow
+    let makeNostrRelaySettingsFlow: @MainActor () -> NostrRelaySettingsFlow
     let makeAnchorsPickerFlow: @MainActor () -> AnchorsPickerFlow
     let makeCreateGroupFlow: @MainActor () -> CreateGroupFlow
     let makeShareInviteFlow: @MainActor () -> ShareInviteFlow
--- a/Sources/OnymIOS/OnymIOSApp.swift
+++ b/Sources/OnymIOS/OnymIOSApp.swift
@@ -8,6 +8,7 @@ struct OnymIOSApp: App {
     private let contractsRepository: ContractsRepository
     private let groupRepository: GroupRepository
     private let inboxTransport: any InboxTransport
+    private let nostrRelaysRepository: NostrRelaysRepository
     private let incomingInvitations: IncomingInvitationsRepository
     private let introKeyStore: any IntroKeyStore
     private let introRequestStore: any IntroRequestStore
@@ -85,12 +86,25 @@ struct OnymIOSApp: App {
         }
         self.groupRepository = groupRepository
 
-        // Inbox transport for invitation send. Connects on first use
-        // via `RootView.task`.
+        // Inbox transport for invitation send. The endpoint list comes
+        // from `NostrRelaysRepository` — read at app boot in the
+        // WindowGroup .task before the fanout interactor starts. The
+        // transport itself is just a Nostr-WebSocket client; nothing
+        // here picks the URL.
         let inboxTransport = NostrInboxTransport(
             signerProvider: OnymNostrSignerProvider()
         )
         self.inboxTransport = inboxTransport
+
+        // Nostr-relays config — drives the inbox transport's
+        // connections + the Settings → Transport → Nostr screen.
+        // First-launch path inside the actor seeds with the Onym
+        // Official relay; subsequent launches honor the user's
+        // persisted list.
+        let nostrRelaysRepository = NostrRelaysRepository(
+            store: UserDefaultsNostrRelaysSelectionStore()
+        )
+        self.nostrRelaysRepository = nostrRelaysRepository
 
         // Incoming invitations repository — falls back to in-memory if
         // the on-disk store can't open (same protection-class concerns
@@ -137,6 +151,9 @@ struct OnymIOSApp: App {
             },
             makeRelayerSettingsFlow: { @MainActor in
                 RelayerSettingsFlow(repository: relayerRepository)
+            },
+            makeNostrRelaySettingsFlow: { @MainActor in
+                NostrRelaySettingsFlow(repository: nostrRelaysRepository)
             },
             makeAnchorsPickerFlow: { @MainActor in
                 AnchorsPickerFlow(repository: contractsRepository)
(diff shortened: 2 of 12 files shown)


dev-onym:
Structure reads clean and the fix is precise — the missing `connect(to:)` before `fanout.run()` is exactly what was failing (`send` threw `notConnected`, `subscribe` no-op'd on an empty `connections`). Reading `currentEndpoints()` once at boot puts both legs back online without restructuring the transport.

A few observations worth thinking about:

- **`hasUserInteracted` is the right shape** for "sticky empty list." Without it, a user clearing every relay would silently get the default re-seeded on the next launch — exactly the regression the flag prevents. Worth a one-line comment on the field itself though; the intent isn't obvious from the name alone.
- **The V1 limitation is the riskiest UX surface here.** Settings → Add relay → nothing visibly happens until relaunch. The footnote helps, but I'd watch during manual smoke whether the relaunch requirement is obvious enough that a tester doesn't think Add is broken. A "Restart required" inline banner that appears after the first mutation might be cheaper than waiting on the live-reconnect follow-up.
- **`isDefault` on the endpoint** persists into UserDefaults as part of `NostrRelayEndpoint`. If you ever change which relay is the shipped default (e.g. `nostr.onym.chat` → something else), older installs will keep showing the old URL with a `DEFAULT` badge. Consider deriving `isDefault` at read-time by comparing URLs against the current shipped list rather than persisting it — same UX, no stale-badge failure mode.
- **Manual smoke is the gating item.** The unit suite can't tell you the `connect` call lands before the first `send`; only the device can. The original Phase-2 repro is the load-bearing check.

If you want a formal pass over the wiring + tests, request a review and I'll do a structured one. For a specific change, drop `@claude <instruction>` on the file/line.

dev-onym:
Looked over the diff. Quick takeaways:

**Root cause / fix shape**

Verified the actual fix is the four-line block in `OnymIOSApp.swift:261-270`:

```swift
let endpoints = await nostrRelaysRepository
    .currentEndpoints()
    .map { TransportEndpoint(url: $0.url) }
await inboxTransport.connect(to: endpoints)
```

Placed inside the `WindowGroup .task` and before `InboxFanoutInteractor.run()` — order is correct, since `subscribe()` no-ops on empty `connections` and `send()` throws `notConnected` without it. The ~830 lines around it are the supporting cast (repo, store, config, flow, view, settings entry, tests). That's a good ratio for a fix that needs to ship with a UI seam.

**Observations worth a thought**

- **V1 boot-only read is asymmetric for remove.** The footnote correctly says "Changes apply on the next app launch." But that means removing a relay in Settings doesn't tear down the live connection — outbound traffic keeps flowing over a relay the user just deleted until relaunch. Worth keeping in mind for the follow-up: a "live re-connect" only fixes adds; removes need an explicit `disconnect(from:)` call too. Probably fine for V1, but the footnote phrasing might mask it.
- **`resetToDefault` doc vs. behavior.** Comment at `NostrRelaysRepository.swift:99-101` says it "resets `hasUserInteracted = false` so the seed sticks even if the user later clears + relaunches." After reset, endpoints = `[seed]` and `hasUserInteracted = false`. If the user then calls `clearAll()`, that flips `hasUserInteracted = true`, which means the next-launch seed-on-empty branch in `init` will NOT fire. So the "sticks across clear+relaunch" claim isn't quite accurate. Minor doc nit, not a behavior bug — `clearAll`'s sticky-empty semantic is intentional and probably the right call.
- **Custom-URL display name drops the port.** `url.host()` strips the port, so `wss://relay.example.com:7777` shows the display name as `relay.example.com`. The full URL is rendered on the row below, so it's recoverable, but `url.host(percentEncoded:false)` + ":\(port)" would be friendlier. Cosmetic.
- **`URL(string:)!` in `.seed`.** Acceptable — static literal, validated at build time effectively.
- **Smoke tests still pending.** Both unchecked items in the test plan are the manual ones; the unit suite (496/496) doesn't exercise the actual `notConnected` failure mode. Worth running the original Phase 2 smoke before merge so the "this closes the runtime failure" claim is observed end-to-end and not just argued from the call graph.

**Next steps**

- For a structured pass, kick off the request-review flow.
- For inline code changes, comment `@claude <instruction>` on the specific file/line and I'll pick it up there.

Otherwise this looks ready once the manual smoke clears.

Issue #97, opened 2026-05-05
ux: align Create Group Step 1 with Android — move cancel to top-left cross, free up vertical space


dev-onym:
## Background

Android shipped [onym-android#91](https://github.com/onymchat/onym-android/pull/91) (fixes [#87](https://github.com/onymchat/onym-android/issues/87)) — the soft keyboard was covering the **Next** button on Step 1 of Create Group. The fix landed as a series of compaction tweaks during PR review. iOS `CreateGroupView.swift` (originally pixel-ported in PR #26) still has the older layout and should be aligned.

### 1. Move the cancel button to the top-left as a cross (primary item)

- **Current iOS**: Step 1 has a quiet `Cancel` button in the bottom footer below `Next`.
- **Target**: Replace the bottom `Cancel` with a `✕` close affordance at the top-left of the Step 1 header (above/left of the nav title), 32pt tap target. Remove the bottom `Cancel`.
- The cross calls the same close action that the bottom `Cancel` did.
- Visual: monochrome glyph in `text2` token color, ~18pt, medium weight.

### 2. Free up vertical space on Step 1 so the keyboard does not cover Next

The Android fix removed two decorative surfaces from Step 1 to keep `Next` reachable when the IME is up. iOS should do the same:

- Remove the **\"Selected governance explanation\" card** (the accent-tinted block with the governance subtitle + tooltip text under the governance picker on Step 1). The Step 2 type-banner already conveys this.
- Remove the **\"🔒 Encrypted\" footer card** from Step 1.

The brand avatar and accent-color picker stay on Step 1 — earlier review iterations moved them to Step 2 then put them back; final layout keeps them on Step 1.

### 3. Group Name field — single-line + Done dismisses keyboard

- Make the name `TextField` single-line.
- IME submit label = `Done`. Tapping `Done` (or the equivalent return key) resigns first responder and dismisses the keyboard.
- This complements the layout compaction so users can finish typing and tap `Next` without an extra tap-outside-to-dismiss step.

### 4. Confirm sticky footer respects the keyboard

- The `Next` button container should be pinned and use `.ignoresSafeArea(.keyboard, edges: .bottom)` *inverted* — i.e. the footer should ride above the keyboard (the Android equivalent is `Modifier.imePadding()`). Verify on a small device (iPhone SE 3rd gen / iPhone 13 mini) that `Next` stays visible while the name field is focused.

## Reference — Android final state

The Android Step 1 layout after this PR series:

\`\`\`
[...]

PR #98, opened 2026-05-05
ux(create-group): align Step 1 with Android — top-left ✕, free vertic…

rinat-enikeev:
…al space (#97)

Mirrors onym-android#91: ports the keyboard-room compaction so Next stays reachable on small devices when the IME is up.

- Replace the bottom Cancel quiet-button with a 32pt ✕ close affordance at the top-left of the Step 1 header (calls the same `flow.onClose`).
- Remove the accent-tinted governance explanation card and the 🔒 encrypted footer card from Step 1 (Step 2's type banner already conveys the governance summary).
- Name TextField now uses `.submitLabel(.done)` and dismisses the keyboard `onSubmit`; embedded newlines from a paste are stripped so the field stays single-line.

Step 2, InviteByKey, Creating, and Success screens are unchanged.

### DIFF
--- a/Sources/OnymIOS/Group/CreateGroupView.swift
+++ b/Sources/OnymIOS/Group/CreateGroupView.swift
@@ -158,6 +158,24 @@ private struct OnymQuietButton: View {
     }
 }
 
+/// Top-left close affordance — 32pt tap target rendering an ✕ glyph
+/// in `text2`. Mirrors the Android `IconButton(Icons.Close)` used at
+/// the start of the Create Group top bar.
+private struct OnymTopBarCloseButton: View {
+    var action: () -> Void
+
+    var body: some View {
+        Button(action: action) {
+            Image(systemName: "xmark")
+                .font(.system(size: 18, weight: .medium))
+                .foregroundStyle(OnymTokens.text2)
+                .frame(width: 32, height: 32)
+                .contentShape(Rectangle())
+        }
+        .buttonStyle(.plain)
+    }
+}
+
 // MARK: - Step 1: name + accent + governance
 
 private struct CreateGroupStep1View: View {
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
APPROVE

- `Sources/OnymIOS/Group/CreateGroupView.swift:170-174` — `OnymTopBarCloseButton` lacks an explicit `.accessibilityLabel("Close")`. VoiceOver will fall back to the SF Symbol name ("xmark"), which is functional but worse than the prior labeled "Cancel" button — consider adding a label.
- `Sources/OnymIOS/Group/CreateGroupView.swift:187-192` — ZStack with `.leading` alignment overlays the close button on a centered `OnymNavTitle`. With a 32×32 hit target at `.leading + 8` it won't collide with the title text, but verify visually on the smallest device width and on iPad in compact width that the title doesn't crowd the ✕.
- `Sources/OnymIOS/Group/CreateGroupView.swift:251-256` — newline-strip logic mutates `flow.name` then `return`s, relying on the re-entrant `onChange` to apply the 32-char cap on the next pass. Correct, but worth a quick check that pasting a multiline string >32 chars truncates as expected (two onChange cycles).
- `Sources/OnymIOS/Group/CreateGroupView.swift:259` — `.onSubmit { nameFocused = false }` matches the new `.submitLabel(.done)` — good.
- Removal of `selectedExplanation`/`encryptedFooterCard` is consistent with the PR description (governance summary lives on Step 2's type banner). No dangling references — both helpers were private and only used here.
- The footer simplification (drop `VStack(spacing: 4)` + `OnymQuietButton`) is clean; `flow.onClose` is preserved via the new top-bar button.

Overall looks good — small a11y polish (explicit Close label) is the only nit worth addressing.

PR #117, opened 2026-05-09
feat(create-group): scan QR on Invite by Inbox Key (closes #115)

### LINKED ISSUE #115: [BUG] No QR code scanner on the invitation/add people screen
# [BUG] No QR code scanner on the invitation/add people screen

## Environment

| Field            | Value                                                                 |
|------------------|-----------------------------------------------------------------------|
| **App version**  | 0.0.14 (build 14)                                                    |
| **Platform**     | iOS                                                                   |
| **Devices**      | iPhone 16 Pro Max (iOS 18) ; iPhone 12 (iOS 26)                      |
| **Profile**      | Fresh install, new profile                                            |

## Summary

The app generates QR codes for invitations (visible on the Invite Key screen in Settings), but there is **no QR code scanner** available on the Add People / invitation screen during or after group creation. If one user has a QR code to share, the other user has no way to scan it within the app's invitation flow.

## Steps to reproduce

1. Open the app.
2. Create a group and navigate to the **Add People** screen (Step 2 of creation or via "Add people" inside the group).
3. Look for a QR code scanner option.

## Expected result

The invitation screen provides a **"Scan QR code"** button that opens the device camera, allowing the user to scan another user's invite QR code directly from the Add People screen.

## Actual result

There is no QR scanner on the invitation screen. The only invitation methods available are sharing/copying a link and pasting a 64-char inbox key. Despite the app generating QR codes (Invite Key in Settings), there is no corresponding scanner to read them in the invitation flow.

## Suggested implementation

Add a **"Scan QR code"** button on the Add People screen alongside existing invite methods:
- Share invite link
- Copy invite link
- Paste inbox key
- **Scan QR code** ← missing

This creates a complete loop: User A shows QR → User B scans QR → join request is sent.

## Impact

- **Severity:** Medium
- The app generates QR codes but provides no way to scan them during the invitation flow. This breaks the QR-based invitation loop and forces users to rely solely on link sharing or manual key pasting — defeating the purpose of having QR codes in the first place.

dev-onym:
## Summary

- Add a **Scan QR** button next to **Paste** on the Create Group → *Invite by Inbox Key* screen so the inviter can capture an invitee's inbox key with the camera instead of typing 64 hex chars by hand.
- New `QRCodeScannerView` wraps an `AVCaptureSession` with QR-only metadata output, requests camera permission on first use, and surfaces a friendly inline message on denial / unavailable hardware.
- New `CreateGroupFlow.canonicalizeInviteKey(_:)` parses scan payloads — bare hex, the legacy iOS `?payload=<hex>` URL, and the cross-platform Android `/i?k=<urlsafe-b64>` shape — into the same 64-char hex string the existing paste path validates.
- `settingsInviteURL` now emits the **full** 32-byte X25519 inbox key (was truncated to 22 bytes / 44 hex chars), so the Settings → Invite Key QR actually round-trips through the new scanner.
- `NSCameraUsageDescription` added to `Info.plist` + `project.yml`.

## Test plan

- [ ] Unit: `CreateGroupFlowTests` — canonicalisation of bare hex, `payload=` URL, and Android `/i?k=` URL; garbage falls through unchanged.
- [ ] Unit: `SettingsQRCodeTests` — generated URL contains all 64 hex chars and round-trips through `canonicalizeInviteKey`.
- [ ] Manual (device): Settings → Invite Key shows QR; on a second device, Create Group → Add People → Invite by Inbox Key → Scan QR → permission prompt → scan → inbox key prefilled in the field → Add to group succeeds.
- [ ] Manual (denied path): Deny camera in iOS Settings → reopen scanner → inline "Open Settings" affordance shown, no crash.
- [ ] Manual (simulator): Scanner falls back to "Camera unavailable" message instead of hanging on a black preview.

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -33,6 +33,8 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>NSCameraUsageDescription</key>
+	<string>Onym uses the camera to scan invite-key QR codes when you add someone to a group.</string>
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
--- a/Sources/OnymIOS/Group/CreateGroupFlow.swift
+++ b/Sources/OnymIOS/Group/CreateGroupFlow.swift
@@ -202,6 +202,49 @@ final class CreateGroupFlow {
         route = .step2
     }
 
+    /// Drop a scanned QR payload into `inviteeInput`. Strips known URL
+    /// wrappers so the user sees the raw 64-char hex (or whatever the
+    /// payload canonicalises to) and can review before tapping
+    /// "Add to group". Validation is left to the existing pipeline —
+    /// a malformed scan surfaces the same inline error as a malformed
+    /// paste.
+    func tappedScannedKey(_ raw: String) {
+        inviteeInput = Self.canonicalizeInviteKey(raw)
+        inviteeError = nil
+    }
+
+    /// Pull a candidate inbox key out of a raw scanned/pasted string.
+    /// Recognises:
+    ///  - bare hex (returned unchanged, lowercased)
+    ///  - `https://onym.chat?payload=<hex>` (legacy iOS settings QR)
+    ///  - `https://onym.chat/i?k=<urlsafe-base64>` (Android identity
+    ///    invite — `IdentityInviteUrl.kt`)
+    /// Falls back to the trimmed raw input otherwise so the existing
+    /// validation surfaces a meaningful error.
+    static func canonicalizeInviteKey(_ raw: String) -> String {
+        let trimmed = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+        if let comps = URLComponents(string: trimmed),
+           let items = comps.queryItems {
+            if let payload = items.first(where: { $0.name == "payload" })?.value,
+               !payload.isEmpty {
+                return payload.lowercased()
+            }
+            if let k = items.first(where: { $0.name == "k" })?.value,
+               let bytes = urlSafeBase64Decode(k) {
+                return bytes.map { String(format: "%02x", $0) }.joined()
+            }
+        }
+        return trimmed
+    }
+
+    private static func urlSafeBase64Decode(_ s: String) -> Data? {
+        var t = s.replacingOccurrences(of: "-", with: "+")
+        t = t.replacingOccurrences(of: "_", with: "/")
+        let pad = (4 - t.count % 4) % 4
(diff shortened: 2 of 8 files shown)


rinat-enikeev:
@lead-onym make a build

rinat-enikeev:
@lead-onym make a build again

rinat-enikeev:
@lead-onym make a build

ux-onym (commented):
APPROVE

- `Sources/OnymIOS/Group/QRCodeScannerView.swift:152-186`: PR test plan promises an "Open Settings" affordance on denial, but the failure UI only renders text — no button calling `UIApplication.openSettingsURLString`. Either drop the claim from the description or add a Button that opens Settings.
- `Sources/OnymIOS/Group/QRCodeScannerView.swift:191-227`: `configureAndStart()` runs on the main actor, so `session.beginConfiguration` / `addInput` / `addOutput` and the AVCaptureDeviceInput init all block the UI thread (Apple specifically recommends doing this off the main queue; you already have `sessionQueue` for `startRunning`). Likely a 100–300ms hitch when the scanner first opens — minor, but worth fixing while you're here.
- `Sources/OnymIOS/Group/CreateGroupFlow.swift:226-235`: `URLComponents(string:)` succeeds on inputs like `payload=&k=` (empty values), and an empty-but-present `k` decodes to `""`, which `tappedScannedKey` then writes into `inviteeInput`, silently clearing the field. Low-risk (only on malformed scans) but consider rejecting empty decodes so the user sees the original payload and the existing inline error.
- `Sources/OnymIOS/Group/CreateGroupFlow.swift:228-230`: Bare-hex test (`test_canonicalizeInviteKey_bareHex_returnsUnchanged`) at `Tests/OnymIOSTests/CreateGroupFlowTests.swift:142` passes lowercase `"ab"*32`. There's no test for an *uppercase* bare-hex scan; it would currently bypass the `payload.lowercased()` branch and fall through to `return trimmed` (still uppercase). The downstream `Data(hex:)` validator presumably handles mixed case, but if you're standardising on lowercase elsewhere it's a small consistency gap — add a test or strip case here.
- `Info.plist:36-37` + `project.yml:91`: Both files duplicate `NSCameraUsageDescription`. Consistent with the existing pattern (`CFBundleDisplayName`, `CFBundleURLTypes` are also duplicated), so xcodegen merges cleanly — flagging only so a future xcodegen regen doesn't surprise anyone.
- Settings round-trip (`Sources/OnymIOS/Settings/SettingsQRCode.swift:55-61`): the truncation fix is the right call and the `SettingsQRCodeTests` round-trip assertion is exactly the load-bearing test you want here. Nice.

Net: the parser, dedup latch, permission flow, and tests all look solid. Ship it after deciding on the "Open Settings" button (either build it or trim the test-plan line).

PR #122, opened 2026-05-10
fix(identities): restore swipe-to-Remove via styled List

rinat-enikeev:
## Summary

`IdentityManagementUITests.test_removeIdentity_nameConfirmGate_blocksThenAllows` was the last red on Release run [25622030590](https://github.com/onymchat/onym-ios/actions/runs/25622030590). Root cause: `.swipeActions(edge: .trailing) { Remove }` was already declared on the identity row, but the Settings redesign moved the row into a custom `SettingsCard { VStack }` — and `.swipeActions` is a List-only modifier. SwiftUI silently dropped it, so swiping a row revealed nothing for users **and** the test.

## Fix

Wrap the row block in a `List` with system chrome suppressed so `.swipeActions` activates without changing the surrounding Identities chrome (title, intro footnote, Add button, closing footnote stay in the existing ScrollView/VStack):

- `.listStyle(.plain)` + `.scrollContentBackground(.hidden)` — kill the system List background
- `.listRowBackground(OnymTokens.surface2)` + `.listRowSeparator(.hidden)` + `.listRowInsets(EdgeInsets())` — recreate the SettingsCard row look; we draw our own hairlines via `SettingsRow`
- `.scrollDisabled(true)` + a measured `.frame(height: ...)` — the outer ScrollView still drives scrolling; the inner List just provides swipe behavior
- `.clipShape(RoundedRectangle(cornerRadius: 14))` + `.padding(.horizontal, 16)` — preserve the rounded-card shell

The `.contextMenu` block is kept so long-press → Remove still works as a parity affordance.

## Test plan

Verified locally on iPhone Air (iOS 26 simulator) — all four IdentityManagementUITests pass:

- [x] `test_identitiesList_showsBootstrappedIdentityAsActive` (9.5s)
- [x] `test_addIdentity_sheetSubmits_growsList` (18.8s)
- [x] `test_chatsPicker_switchesActiveIdentity_titleFlips` (25.6s)
- [x] `test_removeIdentity_nameConfirmGate_blocksThenAllows` (32.2s) — previously failing at IdentitiesScreen.swift:65 with "swipe Remove action never appeared"

The first three confirm that row identifiers + Active badge id + Add sheet survived the refactor.

Reviewer should:
- [ ] Visual sanity-check on a device — rounded card look should match the rest of the Settings tree.
- [ ] Confirm a swipe-Remove with multiple identities behaves correctly (list shrinks, no orphan swipe state).
- [ ] Re-run the Release workflow and confirm UI tests are green end-to-end.

### DIFF
--- a/Sources/OnymIOS/Identity/IdentitiesView.swift
+++ b/Sources/OnymIOS/Identity/IdentitiesView.swift
@@ -16,51 +16,7 @@ struct IdentitiesView: View {
                 SettingsFootnote("Tap an identity to open it. Each identity has its own keys, chats, and recovery phrase.")
 
                 SettingsSectionLabel("YOUR IDENTITIES")
-                SettingsCard {
-                    let summaries = flow.identities
-                    ForEach(Array(summaries.enumerated()), id: \.element.id) { idx, summary in
-                        NavigationLink {
-                            IdentityDetailView(flow: flow, summary: summary)
-                        } label: {
-                            SettingsRow(
-                                title: LocalizedStringKey(summary.name),
-                                subtitle: "BLS \(flow.blsPrefix(of: summary))…",
-                                subtitleMono: true,
-                                inset: 68,
-                                last: idx == summaries.count - 1
-                            ) {
-                                IdentityRingTile(active: summary.id == flow.currentID, size: 40)
-                            } right: {
-                                if summary.id == flow.currentID {
-                                    Text("Active")
-                                        .font(.system(size: 11, weight: .semibold))
-                                        .padding(.horizontal, 8)
-                                        .padding(.vertical, 3)
-                                        .background(OnymTokens.green.opacity(0.18),
-                                                    in: Capsule())
-                                        .foregroundStyle(OnymTokens.green)
-                                        .accessibilityIdentifier("identities.active_badge.\(summary.id)")
-                                }
-                            }
-                        }
-                        .buttonStyle(.plain)
-                        .accessibilityIdentifier("identities.row.\(summary.id)")
-                        .swipeActions(edge: .trailing) {
-                            Button(role: .destructive) {
-                                flow.startRemoval(of: summary)
-                            } label: {
-                                Label("Remove", systemImage: "trash")
-                            }
-                        }
-                        .contextMenu {
-                            Button(role: .destructive) {
-                                flow.startRemoval(of: summary)
-                            } label: {
-                                Label("Remove", systemImage: "trash")
-                            }
-                        }
-                    }
-                }
+                identitiesList
 
                 Button {
                     showAddSheet = true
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
COMMENT
- `IdentitiesView.swift:67` (rowHeight: CGFloat = 64) — hard-coded row height clips with Dynamic Type / accessibility text sizes. `SettingsRow` uses default-sized fonts (16.5pt title + 12.5pt subtitle) without `.dynamicTypeSize` clamps, so at XL+/AX1+ the row needs more vertical room and content will be cut. Consider measuring per-row (`GeometryReader` or `onPreferenceChange`) or clamping Dynamic Type for parity with SettingsCard, since SettingsCard's `VStack` lets rows grow naturally.
- `IdentitiesView.swift:131` (`.frame(height: max(rowHeight, CGFloat(summaries.count) * rowHeight))`) — the `max(rowHeight, ...)` floor reserves a 64pt empty row when `summaries` is empty. In practice always ≥1 identity, but worth a comment or `summaries.isEmpty ? 0 : ...` to avoid a phantom strip if that invariant ever loosens.
- `IdentitiesView.swift:130` (`.scrollDisabled(true)`) — List-in-ScrollView with `.scrollDisabled` is a known workaround; combined with the measured frame it should behave, but please verify on iOS 17+ simulators in addition to iOS 26: in earlier releases nested-List chrome (selection highlight, list-edge insets) sometimes leaks through `.listStyle(.plain)`.
- `IdentitiesView.swift:128-129` (`.listStyle(.plain)` + `.scrollContentBackground(.hidden)`) — both are iOS 16+; matches the rest of the project's deployment target, fine.
- Visual parity: the new wrapper applies `.clipShape(RoundedRectangle(14))` + per-row `.listRowBackground(surface2)` + outer `.padding(.horizontal, 16)`, which reproduces `SettingsCard`'s shape (SettingsDesign.swift:132-140). Looks equivalent on light/dark.
- Behavior: `.swipeActions` + `.contextMenu` + `accessibilityIdentifier("identities.row.\(id)")` and `"identities.active_badge.\(id)"` are preserved, so the four UI tests' selectors still resolve.
- Suggest adding a small UI test at AX1 Dynamic Type to lock in the row-height assumption before this regresses silently again.

rinat-enikeev:
Addressed review comments in `ae8fa42`:

- **Hard-coded 64pt → measured.** New `IdentityRowsHeightKey` PreferenceKey: each row publishes its rendered height through a hidden `GeometryReader` background; parent sum-reduces and feeds the total into the List frame. Matches the natural growth `SettingsCard`'s VStack would give at AX1+/XL.
- **`max(count*estimate, measured)` floor.** First version gated on `measured > 0`, which broke `Add Identity → list grows` because a stale 1-row measurement clipped the newly-added second row. The per-count estimate is now the floor; measurement only raises the ceiling for AX1+ taller rows.
- **Empty list branch.** Returns `EmptyView()` instead of reserving a 64pt phantom strip, so the layout is honest if the ≥1-identity invariant ever loosens.

Skipping the iOS 17 verification per project policy — we're iOS 18+ only. Skipping the AX1 UI test for now to keep scope focused; happy to file it as a follow-up if you want it tracked.

Re-ran `IdentityManagementUITests` locally on iPhone Air (iOS 26 sim, default Dynamic Type) — all four pass, including `test_removeIdentity_nameConfirmGate_blocksThenAllows`.

rinat-enikeev:
@dev-onym do you think https://github.com/onymchat/onym-ios/pull/122/commits/ae8fa423c806234556acd3e314127366e52f1a49 is an over engineering?

PR #123, opened 2026-05-10
feat(release-issue): auto-bump patch tag when Tag field is blank

rinat-enikeev:
## Summary

File a Release issue without filling in the **Tag** field and the `release-from-issue` workflow now resolves the next patch version itself — `v0.0.40 → v0.0.41` — and dispatches `Release` with that.

### Issue template
- `Tag` is no longer required. New help text + placeholder explain the auto-bump behaviour. The non-empty regex pattern stays so an explicit-but-malformed value (e.g. `0.0.10` without the `v`) still gets caught at submit time by GitHub's form validator.

### `release-from-issue.yml`
- `actions/checkout@v4` now uses `fetch-tags: true` so the runner sees `v*` tags (default depth-1 fetch wouldn't).
- **Resolve release tag** step (renamed from "Extract tag…"):
  - Treats GitHub's empty-input rendering (`_No response_`, which collapses to `_Noresponse_` after whitespace strip) as missing.
  - On missing: picks the latest `v[0-9]+\.[0-9]+\.[0-9]+` reachable tag, bumps the patch, falls back to `v0.0.1` on a tagless repo.
  - Refuses to dispatch on a tag that already exists — protects a hand-typed value that collides with a previous release. The auto-bump path is collision-free by construction.
  - Emits an `auto_bumped` step output (0/1).
- **Body-rewrite step** now does two replacements in one `gh issue edit`:
  1. Overwrites the `### Tag` value with the resolved tag — idempotent for the explicit case, meaningful for the auto-bumped case so the issue itself shows the actual tag being released.
  2. Hydrates the `<!-- UI_TESTS -->` block as before.
- When auto-bumped, leaves a one-line comment on the issue noting the chosen version so the maintainer can correct it (close + reopen with a different value) if needed.

## Test plan

Verified locally:
- [x] Auto-bump arithmetic: `git tag --sort=-v:refname | head -1` returns `v0.0.40`; the bash IFS-split + `$((PATCH+1))` produces `v0.0.41`.
- [x] Body-rewrite Python on a synthetic body with `### Tag\n\nv0.0.45\n` (typed) → unchanged tag, UI section hydrated.
- [x] Body-rewrite Python on a synthetic body with `### Tag\n\n_No response_\n` (blank) → tag overwritten with the resolved value, UI section hydrated.

Reviewer checklist:
- [ ] File a release issue with the Tag field blank — confirm the workflow rewrites the body to show the auto-bumped tag, leaves the explanation comment, and dispatches `Release` for the right tag.
- [ ] File a release issue with a colliding tag (e.g. `v0.0.40`) — confirm the workflow comments + fails before dispatch.
- [ ] File a release issue with a properly-typed new tag — confirm behaviour matches the previous workflow.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.yml
+++ b/.github/ISSUE_TEMPLATE/release.yml
@@ -16,12 +16,16 @@ body:
     id: tag
     attributes:
       label: Tag
-      description: Semver, prefixed with `v` (e.g. `v0.0.10`). The workflow refuses anything else.
-      placeholder: v0.0.10
+      description: |
+        Semver, prefixed with `v` (e.g. `v0.0.10`). Leave blank to auto-bump
+        the patch version from the latest existing `v*` tag — the
+        `release-from-issue` workflow rewrites this field in the issue body
+        with the resolved tag before dispatching `Release`.
+      placeholder: v0.0.10 (or leave blank to auto-bump)
     validations:
-      required: true
-      # Issue forms accept a regex on `pattern`; the workflow re-validates
-      # server-side as a defence-in-depth check.
+      required: false
+      # Issue forms validate non-empty values against this pattern; the
+      # workflow re-validates + auto-bumps when the field is empty.
       pattern: '^v[0-9]+\.[0-9]+\.[0-9]+$'
 
   - type: textarea
--- a/.github/workflows/release-from-issue.yml
+++ b/.github/workflows/release-from-issue.yml
@@ -33,14 +33,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Need the v* tag history so the auto-bump step can resolve
+          # the latest patch version when the issue's Tag field is left
+          # blank. `fetch-tags` is lighter than `fetch-depth: 0`.
+          fetch-tags: true
 
-      - name: Extract tag from issue body
+      - name: Resolve release tag
         id: tag
         env:
           BODY: ${{ github.event.issue.body }}
+          ISSUE: ${{ github.event.issue.number }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           # Issue-form `Tag` field renders as `### Tag\n\n<value>\n` in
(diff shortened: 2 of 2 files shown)


dev-onym:
Read through the diff. The shape is clean — `fetch-tags: true`, single-pass Python doing both rewrites, `auto_bumped` flag threaded through to the comment, and the collision check that only matters on the manually-typed path. Comments explain *why* at every non-obvious step (the `_Noresponse_` collapse, the heredoc-fed `IFS` split, the order of the two replacements). Nice.

A few observations worth considering, none blocking:

1. **`_Noresponse_` literal compare is brittle.** If GitHub ever rephrases the placeholder, the check silently falls through to the regex validator and the issue fails instead of auto-bumping. A more forgiving check — e.g. `case "${TAG,,}" in ""|_no*response*) TAG="" ;; esac` — would survive cosmetic changes to the placeholder text without losing strictness.

2. **TOCTOU between resolve and dispatch.** The `git rev-parse "$TAG"` check protects against historical collisions, but two release issues filed back-to-back could both auto-bump to the same patch (each sees `v0.0.40`, both resolve `v0.0.41`). The first `Release` run wins on tag creation; the second fails later with a less obvious error. Probably not worth solving — release issues aren't a high-concurrency surface — but worth knowing the failure mode.

3. **Comment-then-dispatch ordering.** The "auto-bumped to X" comment lands a beat before the dispatch call in the same step's successor. The comment's "close + reopen with a different value" wording reads like a pre-dispatch escape hatch but in practice `Release` is already in flight by the time anyone sees it. Either reword ("Released as X — open a new issue if you wanted a different version") or move the comment to after the dispatch step so its timing matches its semantics.

4. **Tag-section regex assumes single-line input.** `[^\n]+` is correct for the current `input`-type field but would silently mis-rewrite if the form ever switched to `textarea`. Worth a one-line comment pinning that assumption to the form schema, since the rewrite and the form spec live in different files.

For a formal pass, hit **Request review** on the PR — that opens the structured review flow. For any of the above you want as an actual code change, drop `@claude <instruction>` on the specific line and I'll push a commit.

PR #126, opened 2026-05-10
feat(release): publish signed IPA to onym.app/qa/ alongside GitHub Release

rinat-enikeev:
## Summary

Until now a successful tag-gated `Release` attached the signed ad-hoc IPA to a GitHub Release and stopped — testers had to download it manually. The QA dashboard at https://onym.app/qa/ was populated only by the manually-dispatched `QA Build` workflow, so users couldn't tap "Install on iOS" for an actual release.

This mirrors `qa-build.yml`'s publish chain in the `build` job of `release.yml`, immediately after the GitHub-Release upload. A successful Release now:

- Stages the IPA + writes `manifest.plist` under `dist/qa/ios/<tag>/` (same plist shape as the manual QA dispatch).
- Builds a dashboard JSON entry (`kind: tag`, `branch: null`, `description: "Release vX.Y.Z"`).
- Fetches the live `builds.json`, merges with `unique_by(platform+@+version)` so a re-run for the same tag cleanly overwrites, caps at 50.
- rsyncs via `onymchat/onym-website/.github/actions/deploy@main` with `delete_stale: false`.
- Writes a step-summary table with the install / manifest / IPA URLs.
- When dispatched from a release issue, posts a comment on the issue with the same URLs (best-effort — a `gh` failure leaves a `::warning::` but never fails the release run).

### Concurrency

Added a job-level `qa-build-ios` group on the release `build` job — same group `qa-build.yml` uses at workflow level. Tag-gated releases and manual QA dispatches now serialize on `builds.json` writes; lint/tests still run in parallel as before.

### Why inline rather than a composite action

The QA-publish chain is duplicated between `qa-build.yml` and (now) `release.yml`. Refactoring into `.github/actions/qa-publish-ios/action.yml` would deduplicate but adds an indirection layer for two callers. Kept inline so both flows stay readable side-by-side; easy refactor opportunity if a third caller ever appears.

## Test plan

- [x] `python3 -c 'import yaml; yaml.safe_load(open(...))'` parses the workflow cleanly.
- [ ] Reviewer dispatches `Release` for a fresh tag (e.g. via filing a release issue with the auto-bump path) and confirms:
  - The IPA lands as a GitHub Release asset (existing behavior).
  - `https://onym.app/qa/ios/<tag>/OnymIOS.ipa` is reachable.
  - `https://onym.app/qa/ios/<tag>/manifest.plist` is reachable and has `bundle-identifier=chat.onym.ios`.
  - The QA dashboard at `https://onym.app/qa/?platform=ios` shows the new entry with `kind: tag`.
  - Tapping "Install on iOS" on a UDID-registered device installs the IPA via OTA.
  - The release issue receives a comment with the install URL.
- [ ] Reviewer triggers a manual `QA Build` dispatch while a release is publishing — confirms the second waits on the `qa-build-ios` lock instead of racing on `builds.json`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -272,6 +272,13 @@ jobs:
   build:
     runs-on: [self-hosted, macOS, ARM64]
     needs: [lint, create-release, unit-tests, ui-tests]
+    # Share the lock with qa-build.yml so a tag-gated release and a
+    # manual ad-hoc QA dispatch don't race on /qa/builds.json. Only the
+    # build job (where the QA publish runs) takes the lock; lint /
+    # tests run in parallel as before.
+    concurrency:
+      group: qa-build-ios
+      cancel-in-progress: false
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


dev-onym:
LGTM overall — the inline mirror of `qa-build.yml`'s publish chain looks faithful and the job-level `qa-build-ios` concurrency group correctly serializes against `qa-build.yml`'s workflow-level group on `builds.json` writes while leaving lint/tests parallel.

A few small observations:

1. **`tag` field isn't null-guarded.** `qa-build.yml` does `tag: (if $tag != "" then $tag else null end)`; the inline version uses `tag: $tag`. Safe in practice because `inputs.tag` is required, but worth noting if you ever wire a non-tag caller in.

2. **`description: "Release ${VERSION}"` is a bit thin.** When dispatched from a release issue, you already fetch the issue title for `ISSUE_OBJ`. Consider folding it in (e.g. `"Release ${VERSION} — ${TITLE}"` when `TITLE` is non-empty) so the dashboard row carries the same context the issue does. Not blocking.

3. **GitHub Release upload happens before the QA chain.** If the deploy step fails, the GitHub Release exists but the dashboard entry doesn't. Re-running is fine (`softprops/action-gh-release` updates in place, and `unique_by(.platform + "@" + .version)` overwrites the dashboard entry), but it's worth a one-line comment near the QA-stage step so a future debugger isn't surprised that a "failed" release still has an attached IPA.

4. **Issue-comment step uses `if: success() && inputs.issue_number != ''`.** Since `success()` is the implicit default, this is equivalent to just the issue check — fine, but a tiny readability nit.

5. **Test plan check #2 (concurrent QA dispatch during release):** worth eyeballing the Actions UI shows the second run as "Pending — waiting on `qa-build-ios`" rather than just sitting idle, since GH's queue UI for shared groups can be confusing.

None of the above blocks merging from my reading — the schema match against `qa-build.yml`, the heredoc/`stat -f%z` macOS-isms, and the `MARKETING="${VERSION#v}"` derivation all line up. The reviewer-driven test plan is the right gate here.

PR #135, opened 2026-05-15
fix(create-group): render generated name as placeholder, not real input (closes #112)

### LINKED ISSUE #112: [BUG] Group name input field displays placeholder text as if it were actual user input
# [BUG] Group name input field displays placeholder text as if it were actual user input

## Environment

| Field            | Value                                                                 |
|------------------|-----------------------------------------------------------------------|
| **App version**  | 0.0.14 (build 14)                                                    |
| **Platform**     | iOS                                                                   |
| **Devices**      | iPhone 16 Pro Max (iOS 18) ; iPhone 12 (iOS 26)                      |
| **Profile**      | Fresh install, new profile                                            |

## Summary

On the group creation screen (Step 1 of 2), the group name input field displays a **pre-filled random name** (e.g., "Silver Atlas") that visually appears as actual user input rather than a placeholder. The text uses the same color and style as real input, making it indistinguishable from a user-entered name. Only when the user taps the field to type, the text behaves as a **placeholder mask** — it disappears or becomes editable, revealing it was never actual input.

## Steps to reproduce

1. Open the app.
2. Tap **"Create a group"**.
3. Observe the group name input field on the "New Group — Step 1 of 2" screen.
4. Note the pre-filled text (e.g., "Silver Atlas") — it looks like real input.
5. Tap on the input field to start typing.
6. Observe that the text behaves as a placeholder (disappears or transforms into a mask).

## Expected result

The placeholder text in the name field should be **visually distinct** from real user input — using a lighter/greyed-out color and standard iOS placeholder styling, so the user immediately understands it is a suggestion or hint, not pre-filled data.

## Actual result

The placeholder text is displayed in the **same style as actual input** (white/full-opacity text), misleading the user into thinking a name has already been entered. This may cause users to skip naming their group, assuming the displayed text is the group name.

## Impact

- **Severity:** Low
- The bug does not block functionality but creates a confusing UX. Users may unintentionally create groups with random placeholder names (e.g., "Silver Atlas") if they don't tap the field and assume the name is already set.

dev-onym:
The Step 1 group-name field was seeded with the generated "Adjective
Noun" suggestion as the actual TextField value, so it rendered in the
full-opacity input color and looked indistinguishable from text the
user had typed. Users could miss that they were supposed to (or could)
change it, and unknowingly create groups called "Silver Atlas".

Leave `name` empty at init and reset so the TextField's `prompt` —
already styled with the muted `OnymTokens.text3` color — surfaces as a
proper iOS placeholder. The existing `effectiveName` fallback already
substitutes `generatedName` at submit when the user leaves the field
blank, so the "hit Create immediately" path still works.

The first-focus-clears-placeholder dance is no longer needed and is
removed along with its tests.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/Sources/OnymIOS/Group/CreateGroupFlow.swift
+++ b/Sources/OnymIOS/Group/CreateGroupFlow.swift
@@ -39,17 +39,10 @@ final class CreateGroupFlow {
     var invitees: [OnymInvitee] = []
 
     /// Friendly placeholder generated on init / reset (e.g. "Maple
-    /// Garden"). The TextField pre-fills with this so the user can
-    /// hit Create immediately without typing — first focus on the
-    /// field clears it (see `tappedNameFieldFocused`). Submit also
-    /// falls back to this if the user emptied the field and didn't
-    /// retype.
+    /// Garden"). Rendered as the TextField's `prompt` so it appears
+    /// in the muted placeholder style — never as real input. Submit
+    /// falls back to this when the user leaves the field empty.
     private(set) var generatedName: String
-
-    /// Goes true on the first focus event of the name field. Used to
-    /// distinguish "user accepted the placeholder" from "user wants
-    /// to type their own".
-    private var nameFieldHasBeenFocused = false
 
     /// Bound to the InviteByKey screen's TextField.
     var inviteeInput: String = ""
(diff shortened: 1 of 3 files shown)


rinat-enikeev:
@lead-onym make a build

dev-onym:
QA Build workflow dispatched for branch `agent/issue-112` (PR #135): https://github.com/onymchat/onym-ios/actions/runs/25926709845

No code change needed for this request, so no commit was made.

PR #138, opened 2026-05-15
feat(chats): surface invite QR on Chats toolbar; fix dead-end hint (closes #137)

### LINKED ISSUE #137: Lift invite-key sharing to the Chats tab and fix the dead-end "Settings → Advanced" hint
# Lift invite-key sharing to the Chats tab and fix the dead-end "Settings → Advanced" hint

## Summary

The "Invite by Inbox Key" flow currently tells the user to find their inbox key in **Settings → Advanced** — a location that does not exist as a top-level path and does not display the inbox key anywhere even at its actual per-identity nesting (see #113). This issue replaces that dead-end with a discoverable, one-tap entry point on the Chats tab: a QR toolbar icon that opens the existing `ShareKeyView`. It also fixes the upstream naming drift (`blsPublicKey:` parameter that actually carries `inboxPublicKey`) which is the most plausible source of the original spec divergence.

Three coordinated changes, single PR.

### Two distinct keys per identity

Per `Sources/OnymIOS/Identity/IdentitySummary.swift:11-21`:

- `blsPublicKey` — 48-byte BLS12-381 G1, rendered as a `BLS xxxx…` fingerprint. Governance / signing.
- `inboxPublicKey` — 32-byte X25519, rendered as 64 hex characters. **This is what an inviter shares and what the receiver pastes / scans into Create Group → Step 2.**

### The receive side already handles both raw hex and URL form

`CreateGroupFlow.canonicalizeInviteKey` (`Sources/OnymIOS/Group/CreateGroupFlow.swift:224-238`) strips `https://onym.chat?payload=<hex>` and the Android-style `https://onym.chat/i?k=<urlsafe-base64>` wrappers before validation. **Therefore the `Copy link` affordance on `ShareKeyView` is already sufficient for the paste path** — the receiver pastes the URL, the validator unwraps it. No raw-hex copy affordance is required.

### Where the inbox key surfaces today

The inbox key is only reachable through Settings:

- `SettingsView.qrHero` (`Sources/OnymIOS/Settings/SettingsView.swift:274-315`) — large "INVITE KEY" card; opens `ShareKeyView`.
- `IdentityDetailView.inviteCard` (`Sources/OnymIOS/Identity/IdentityDetailView.swift:149-211`) — same content embedded in the per-identity drill-down.
- `ShareKeyView` (`Sources/OnymIOS/Identity/ShareKeyView.swift`) — full-screen sheet with QR + `Copy link` + `Share`.

The Chats tab has no entry point at all — `Sources/OnymIOS/Chats/ChatsView.swift:26-51` shows the toolbar carries only the identity picker, the approve-requests button, and a compose button.

### Why this is wrong product positioning

Sharing one's invite is a core messenger action, not a configuration step. Every comparable messenger surfaces it on or near the main surface:

- **WhatsApp** — chat list toolbar offers "Scan / show QR" one tap from the main screen.
- **Signal** — profile pill in the chat list opens directly to the QR.
- **Telegram** — Contacts tab surfaces "Invite friends" with QR.

Onym buries this two taps deep into Settings. Result: users who never trigger Create Group never discover their invite QR exists, and the Step 2 hint in the invite-by-key flow has to reference a multi-level Settings path that is easy to misremember (and was, in fact, misremembered — that is #113).

### The naming drift behind #113

`Sources/OnymIOS/Settings/SettingsQRCode.swift:58` declares:

```swift
func settingsInviteURL(blsPublicKey: Data) -> String
```

But every call site passes `summary.inboxPublicKey`:

- `Sources/OnymIOS/Identity/ShareKeyView.swift:14`
- `Sources/OnymIOS/Identity/IdentityDetailView.swift:155, 171, 193, 199`
- `Sources/OnymIOS/Settings/SettingsView.swift:280`

The parameter type-checks (both are `Data`) but actively lies about its contents. The original hint in `CreateGroupView` ("Settings → Advanced") is most plausibly explained by a writer following the parameter name and landing in Advanced — which is the only place a `Copy public key (BLS)` row currently lives (`IdentityDetailView.swift:65-76`). Fixing the hint without fixing the parameter leaves the trap armed for the next contributor.

## Proposed changes

Three changes, all in the same PR.

### 1. Add a QR toolbar icon to `ChatsView`

`Sources/OnymIOS/Chats/ChatsView.swift` — add a new `ToolbarItem` in `topBarTrailing`, positioned between `ApproveRequestsToolbarButton` and the existing `compose` button.

- **Icon:** `qrcode` (SF Symbol).
- **Action:** present `ShareKeyView` for the active identity in a sheet, mirroring `SettingsView.qrHero`'s presentation pattern.
- **Visibility:** shown regardless of `flow.groups.isEmpty` — unlike `compose`, which is hidden in the empty state. The invite QR is at least as discoverable in the empty state, where users are most likely to want it.
- **Disabled state:** hidden when `activeSummary` is nil (pre-bootstrap), matching the `qrHero` guard in `SettingsView`.
- **Accessibility identifier:** `chats.share_invite_qr`.

### 2. Update the Create Group invite hint

`Sources/OnymIOS/Group/CreateGroupView.swift:682-696` — replace:

> "Ask for their **inbox key** — they can find it in **Settings → Advanced**, or share a QR code from there."

with:

> "Ask for their **invite key** — they can tap the **QR icon on their Chats screen**, then share the QR or copy the link."

The new hint points at a path that exists and renders the key in one tap. The wording "invite key" is also more user-friendly than "inbox key" (which is implementation jargon — see Question 3 below).

### 3. Rename `blsPublicKey:` → `inboxPublicKey:`

`Sources/OnymIOS/Settings/SettingsQRCode.swift:58` plus the five call sites listed in Background. Pure rename, no behavioural change. Add a doc-comment to the new signature clarifying the X25519 vs BLS distinction so the trap stays disarmed:

```swift
/// The invite URL embeds the 32-byte X25519 inbox public key (raw hex),
/// **not** the BLS12-381 governance key. See `IdentitySummary.swift:11-21`
/// for the distinction.
func settingsInviteURL(inboxPublicKey: Data) -> String { ... }
```

## Out of scope

- Adding a separate `Copy inbox key` raw-hex button to `ShareKeyView`. The `Copy link` affordance is already sufficient because the receiver's paste validator unwraps URL form. A second copy button doing essentially the same job adds friction without payoff. Re-evaluate only if telemetry shows users hitting the paste flow with non-URL content.
- Removing the existing `Copy public key (BLS)` row from `IdentityDetailView`'s Advanced section. It serves a small but real audience (out-of-band identity verification). Separate discussion.
- Changing the URL format (`https://onym.chat?payload=<hex>`), the key derivation, or the cryptographic material.
- Android parity. Android already accepts iOS's URL form via its own `IdentityInviteUrl.kt` path (referenced in `canonicalizeInviteKey`'s comment); placement on the Android home screen is a separate decision for that platform.

## Acceptance criteria

- [ ] Tapping the new QR icon on the Chats tab presents `ShareKeyView` for the active identity.
- [ ] The QR icon is visible in both the empty state and the populated chat list.
- [ ] The QR icon is hidden when no active identity exists (fresh install, pre-bootstrap).
- [ ] `CreateGroupView` → Step 2 → Invite by Inbox Key shows the updated hint pointing at the Chats QR icon.
- [ ] End-to-end manual: tap Chats QR → land on `ShareKeyView` → tap `Copy link` → paste into Step 2 invite-by-key field on a second device → `64/64` counter lights up, accent border, "Add to group" enabled.
- [ ] End-to-end manual: tap Chats QR → land on `ShareKeyView` → scan the QR from a second device → same successful invitee resolution.
- [ ] `git grep 'blsPublicKey:'` returns no hits referring to `settingsInviteURL`'s parameter. (BLS-related uses elsewhere — `IdentitySummary.blsPublicKey`, etc. — are unrelated and stay.)
- [ ] Existing tests (`SettingsView`, `IdentityDetailView`, `ShareKeyView`, `CreateGroupFlowTests`) pass unchanged after the rename.
- [ ] New UI test: `ChatsView` toolbar exposes accessibility identifier `chats.share_invite_qr` and tapping it presents the share sheet.
- [ ] Dynamic Type at largest accessibility size: toolbar items fit or wrap gracefully on the smallest supported device (iPhone 12 mini / SE-class width).

## Risk and mitigation

- **Toolbar crowding on small devices.** `ChatsView` already carries up to three trailing items (`ApproveRequestsToolbarButton`, `compose`, plus the leading identity picker on the other side). Adding a fourth is consistent with iOS Mail / Messages density, but worth confirming at the largest Dynamic Type setting. Mitigation: if the row overflows, `compose` is the candidate for relegation to an empty-state CTA, since QR sharing is a higher-frequency action than composing a new group.
- **Re-introduction of the naming drift.** A future contributor could add a second URL builder using the old `blsPublicKey:` label. Mitigation: the doc-comment on the renamed signature is the durable guard.

## Open questions

1. **Icon choice.** `qrcode` is the obvious SF Symbol; `qrcode.viewfinder` reads more like "scan" than "show", so `qrcode` is correct. Confirm.
2. **Hint copy.** Proposed: "they can tap the QR icon on their Chats screen, then share the QR or copy the link." Open to shorter / different phrasing.
3. **Terminology — "inbox key" vs "invite key".** The UI on `ShareKeyView` already says "Invite Key" (title, line 94 of `ShareKeyView.swift`). The Step 2 screen and `CreateGroupFlow` use "inbox key" throughout. These are the same artefact in two voices — implementation calls it inbox key (because that's what the bytes are), UI calls it invite key (because that's what the user uses it for). Worth aligning on one. Suggestion: keep "Invite Key" in user-facing strings, "inbox key" only in code / comments / property names.

## References

- Original bug: #113 — this issue subsumes the fix and broadens scope.
- `Sources/OnymIOS/Identity/IdentitySummary.swift:11-21` — canonical distinction between the two keys.
- `Sources/OnymIOS/Group/CreateGroupFlow.swift:224-238` — `canonicalizeInviteKey` URL-unwrap, demonstrating the `Copy link` path already works end-to-end.
- `Sources/OnymIOS/Settings/SettingsView.swift:274-315` — existing `qrHero` → `ShareKeyView` presentation pattern to mirror on Chats.
- `Sources/OnymIOS/Chats/ChatsView.swift:26-51` — current toolbar definition; the new item slots here.

dev-onym:
Adds a QR toolbar icon to the Chats tab that opens ShareKeyView, so users
can share their invite without drilling into Settings. Updates the Invite
by Key hint in Create Group to point at the new entry point instead of the
non-existent "Settings -> Advanced" path. Renames settingsInviteURL's
parameter from blsPublicKey: to inboxPublicKey: to match what it actually
carries (X25519 inbox key, not the BLS governance key) and adds a doc note
to keep the next contributor from re-introducing the drift.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/Sources/OnymIOS/Assembly.swift
+++ b/Sources/OnymIOS/Assembly.swift
@@ -1,0 +1,10 @@
+import Foundation
+
+/// Seam over the app's composition root. The production conformer is
+/// `OnymIOSApp` itself (whose `init` wires the live graph); UI-test and
+/// preview hosts can adopt this to vend a hand-built `AppDependencies`
+/// without dragging in the full `WindowGroup` lifecycle.
+@MainActor
+protocol Assembly {
+    func assemble() -> AppDependencies
+}
--- a/Sources/OnymIOS/Chats/ChatsView.swift
+++ b/Sources/OnymIOS/Chats/ChatsView.swift
@@ -13,6 +13,7 @@ struct ChatsView: View {
     let makeShareInviteFlow: @MainActor () -> ShareInviteFlow
 
     @State private var showCreateGroup = false
+    @State private var showShareKey = false
 
     var body: some View {
         Group {
@@ -34,6 +35,21 @@ struct ChatsView: View {
             // the badge only appears when `pending.count > 0`.
             ToolbarItem(placement: .topBarTrailing) {
                 ApproveRequestsToolbarButton(flow: approveRequestsFlow)
+            }
+            // Show invite QR — primary discovery point for the
+            // user's own invite key. Hidden pre-bootstrap (no active
+            // identity yet); shown in both empty and populated
+            // states because surfacing one's invite is at least as
+            // useful before the first chat exists.
+            if activeSummary != nil {
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button {
+                        showShareKey = true
+                    } label: {
+                        Image(systemName: "qrcode")
+                    }
+                    .accessibilityIdentifier("chats.share_invite_qr")
+                }
             }
             // Plus button mirrors iOS Mail / Messages — useful once
             // the user already has at least one chat. Hidden in the
@@ -59,6 +75,21 @@ struct ChatsView: View {
                 onClose: { showCreateGroup = false }
             )
         }
+        .sheet(isPresented: $showShareKey) {
+            if let active = activeSummary {
+                NavigationStack {
+                    ShareKeyView(identity: active, blsPrefix: identitiesFlow.blsPrefix(of: active))
+                }
+            }
+        }
+    }
+
+    /// Active identity, mirroring `SettingsView.activeSummary` — used
+    /// both to gate the invite-QR toolbar item and to feed
+    /// `ShareKeyView` when the sheet presents.
+    private var activeSummary: IdentitySummary? {
+        guard let id = identitiesFlow.currentID else { return nil }
+        return identitiesFlow.identities.first { $0.id == id }
     }
 
     /// "Chats" when no identity exists yet (pre-bootstrap), otherwise
--- a/Sources/OnymIOS/Group/CreateGroupView.swift
+++ b/Sources/OnymIOS/Group/CreateGroupView.swift
@@ -682,10 +682,10 @@ private struct CreateGroupInviteByKeyView: View {
     private var explanation: some View {
         (
             Text("Ask for their ").foregroundColor(OnymTokens.text2)
-            + Text("inbox key").font(.system(size: 13, weight: .semibold)).foregroundColor(OnymTokens.text)
-            + Text(" \u{2014} they can find it in ").foregroundColor(OnymTokens.text2)
-            + Text("Settings \u{2192} Advanced").font(.system(size: 13, weight: .semibold)).foregroundColor(OnymTokens.text)
-            + Text(", or share a QR code from there.").foregroundColor(OnymTokens.text2)
+            + Text("invite key").font(.system(size: 13, weight: .semibold)).foregroundColor(OnymTokens.text)
+            + Text(" \u{2014} they can tap the ").foregroundColor(OnymTokens.text2)
+            + Text("QR icon on their Chats screen").font(.system(size: 13, weight: .semibold)).foregroundColor(OnymTokens.text)
+            + Text(", then share the QR or copy the link.").foregroundColor(OnymTokens.text2)
         )
         .font(.system(size: 13))
         .lineSpacing(1.5)
--- a/Sources/OnymIOS/Identity/IdentityDetailView.swift
+++ b/Sources/OnymIOS/Identity/IdentityDetailView.swift
@@ -152,7 +152,7 @@ struct IdentityDetailView: View {
             SettingsCard {
                 VStack(spacing: 14) {
                     SettingsQRCode(
-                        value: settingsInviteURL(blsPublicKey: summary.inboxPublicKey),
+                        value: settingsInviteURL(inboxPublicKey: summary.inboxPublicKey),
                         size: 200
                     )
                     .padding(12)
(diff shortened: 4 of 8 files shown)


innovesss:
@dev-onym make build for iOS

dev-onym:
QA Build workflow dispatched against `agent/issue-137`: https://github.com/onymchat/onym-ios/actions/runs/25922151059 — in_progress. It will build + sign the ad-hoc IPA, publish to `onym.app/qa/`, and comment on PR #138 with the `itms-services://` install URL when done.

No code change needed for this comment, so the working tree is clean and there's nothing to commit.

PR #148, opened 2026-05-19
feat(chats): message persistence layer (PR 2/11 chat stack)

lead-onym:
Second PR in the chat stack. Adds on-disk persistence + a reactive snapshots stream for chat messages. **Mirrors `GroupStore` / `GroupRepository` exactly** so the chat thread UI (later PRs) can subscribe to one group's messages the same way `ChatsView` subscribes to the group list.

## What's in here

| File | Role |
|---|---|
| `ChatMessage` | Domain type (id, groupID, owner, sender BLS hex, body, sentAt, direction, status, groupType) |
| `MessageDirection` / `MessageStatus` | enums: `.incoming` / `.outgoing` ; `.pending` / `.sent` / `.failed` / `.received` |
| `PersistedMessage` | `@Model`. Plain columns (id / groupID / sentAt / direction / status / groupType) for querying + sort; sender BLS hex and body encrypted via `StorageEncryption` |
| `MessageStore` | Async protocol seam |
| `SwiftDataMessageStore` | Actor impl. **Separate `Messages.store` file** next to `Groups.store` so a schema bump in either domain doesn't force a wipe of the other |
| `MessageRepository` | Actor with per-group `snapshots(groupID:)` streams. **Lazy per-group cache** — chat thread reads one group at a time, so loading the whole messages table at launch would be wasted I/O |

## Design notes
- Encryption split matches `PersistedGroup`: queryable / sort columns stay plain, content is encrypted.
- `MessageRepository.removeForOwner` refreshes each cached group from the store rather than wiping in-memory — other-identity rows in the same group must survive (tested in `test_removeForOwner_emptiesAllCachedGroups`).
- `updateStatus` is a hot-path API for the outgoing pipeline (pending → sent/failed) so we don't round-trip the whole row through the encryption boundary.

## Out of scope (deferred)
- No wiring to `IncomingMessageDispatcher` or any send path — that lands in PR 4.
- No app-shell wiring (the production `init()` initializer exists but isn't called anywhere yet).

## Test plan
- [x] `SwiftDataMessageStoreTests` — 10 tests (roundtrip, in-place update, sort, filter, updateStatus, cascade deletes)
- [x] `MessageRepositoryTests` — 10 tests (initial snapshot, broadcast on mutate, per-group isolation, cascade behavior)
- [x] Full `OnymIOSTests` suite — 538 / 538 pass (3 pre-existing skips, 0 regressions)

## Plan context
PR 2 of 11. **PR 1** (#147) added the wire-format payload. **PR 3 next** extends `MemberProfile` with `sendingPubkey` so PR 4 can do Ed25519 envelope-signature verification on incoming chat messages.

### DIFF
--- a/Sources/OnymIOS/Chats/ChatMessage.swift
+++ b/Sources/OnymIOS/Chats/ChatMessage.swift
@@ -1,0 +1,69 @@
+import Foundation
+
+/// One persisted chat message — what the UI renders and what the
+/// store hands back. Mirrors `ChatMessagePayload` but flattens the
+/// `variant` into a `body` + `groupType` pair because every variant
+/// has a body (today) and a single type column is enough for the
+/// store-level layer.
+///
+/// Stable across direction: incoming and outgoing messages share the
+/// same shape, distinguished by `direction` + `status`. Each row
+/// belongs to exactly one `ownerIdentityID` so cascade-delete on
+/// identity removal is a `WHERE owner = ?` over the messages table —
+/// same pattern as `ChatGroup`.
+struct ChatMessage: Equatable, Sendable, Identifiable {
+    /// Stable per-message UUID. Matches `ChatMessagePayload.messageID`
+    /// for incoming rows; minted locally at send time for outgoing.
+    /// Receive-side dedup is a uniqueness check on this ID.
+    let id: UUID
+
+    /// 64-char hex of the parent `ChatGroup.id`. Plain (queryable) on
+    /// disk so `list(groupID:)` is a single `#Predicate`.
+    let groupID: String
+
+    /// Identity that owns this row. Inherits from
+    /// `ChatGroup.ownerIdentityID` of the parent group at write time.
+    let ownerIdentityID: IdentityID
+
+    /// Lowercase BLS pubkey hex (96 chars) of the sender. Matches the
+    /// `memberProfiles` keying — the dispatcher will use this to look
+    /// up the sender's `sendingPubkey` for envelope-signature
+    /// verification in PR 4.
+    let senderBlsPubkeyHex: String
+
+    let body: String
+
+    /// Sender's claim of when they sent. For incoming, taken from
+    /// the payload's `sent_at_millis`. For outgoing, the local clock
+    /// at submit time. Drives chronological display order.
+    let sentAt: Date
+
+    let direction: MessageDirection
+
+    /// Mutable for outgoing (pending → sent / failed). For incoming,
+    /// always `.received` at insert and never changes.
+    let status: MessageStatus
+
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
APPROVE — clean mirror of `GroupStore`/`GroupRepository`, well-tested, scope-limited as promised.

- `MessageRepository.swift:62, 75` — doc comments say `removeForGroup` and `removeForOwner` are "wired into the group-delete path"/"`IdentityRepository.identityRemoved`", but no such wiring exists in this diff (PR notes confirm app-shell wiring is deferred). Trim the "wired into..." phrasing to "will be wired into..." or it'll read as a lie until PR 4+.
- `MessageRepositoryTests.swift:78-94` — `test_snapshots_insertIntoOtherGroup_doesNotEmit` cannot actually prove silence; an implementation that yielded stale empties for A before the A-insert would pass. The inline comment already admits this. Consider a small `Task.sleep` window + counting yields, or accept the limitation and note it once at file top.
- `SwiftDataMessageStore.swift:36-44` — recovery loop redundantly removes `Messages.store` twice (once via `suffix=""` in the loop, again after). Harmless, but the explicit `removeItem(at: url)` after the loop is dead.
- `SwiftDataMessageStore.swift:84, 110, 119, 128, 137` — `try? context.save()` silently swallows save failures across every mutation. Matches the `GroupStore` pattern, but consider centralizing through a helper that at least logs on failure — a silent persistence drop in the chat hot-path will be miserable to diagnose later.
- `SwiftDataMessageStore.swift:78` — `list` uses `compactMap(Self.decode)`, so a decryption failure on one row silently disappears from the snapshot without surfacing. A single corrupted/key-mismatched row will make the UI look intermittently broken with no signal. Same concern as `SwiftDataGroupStore` presumably, so OK to defer — but worth a TODO at minimum.
- `ChatMessage.swift:31, 39` — doc claims `groupID` is "64-char hex" and `senderBlsPubkeyHex` is "96 chars"; nothing enforces it. Init-time precondition or a typed wrapper would catch upstream bugs at the boundary (cheap insurance before this becomes the canonical row shape).
- `MessageRepository.swift:79-92` — `snapshots()` is `nonisolated` and spawns a `Task` for `subscribe`. The first yield is therefore deferred by at least one actor hop. Fine in practice (AsyncStream buffers), worth knowing if any UI ever races subscribe vs. an immediate mutation it just issued.
- Nit `PersistedMessage.swift` — `directionRaw`/`statusRaw`/`groupTypeRaw` as `String` is consistent with `PersistedGroup`, but if `MessageStatus` ever gains a case, old rows with an unknown raw silently vanish from `list` via `compactMap`. Same family of concern as the decryption nit above.

PR #149, opened 2026-05-19
feat(identity): plumb Ed25519 sender pubkey through member profiles (PR 3/11 chat stack)

lead-onym:
**Insider-spoofing fix.** Plumbs the per-member Ed25519 sending pubkey (== `Identity.stellarPublicKey`, the key that already signs every `SealedEnvelope`) end-to-end through the join + invite flows. **Sets up PR 4's chat dispatcher to verify a chat message's envelope signature against the claimed BLS sender.** Without this, a group member could forge another member's `senderBlsPubkeyHex` claim — the BLS pubkey alone is a claim, not a proof.

## Wire compatibility — hard cutover, no migration window

Pre-release, no installed base of join requests or member announcements. All three new fields are **required** on the wire and at construction; receivers throw at decode if missing or wrong-sized.

| Type | New field | Wire key |
|---|---|---|
| `MemberProfile` | `sendingPubkey: Data` | `sending_pubkey` |
| `MemberAnnouncementPayload.AnnouncedMember` | `sendingPub: Data` | `sending_pub` |
| `JoinRequestPayload` | `joinerSendingPublicKey: Data` | `joiner_sending_pub` |
| `IdentitySummary` | `sendingPublicKey: Data` (internal projection only) | n/a |

All three wire-shipped types validate `count == 32` at decode (`MemberProfile.init(from:)`, `AnnouncedMember.init(from:)`, `JoinRequestPayload.init(from:)`). Wrong-sized or missing values throw a `*Error.shape(...)`. No `nil` fallback in PR 4's dispatcher — wire shapes that decode produce a real 32-byte verification key by construction.

## Population sites
- **`JoinRequestSender`** — joiner ships their own Ed25519 pubkey to the admin.
- **`JoinRequestApprover`** — admin reads from request, stores on local roster (`recordJoiner`), broadcasts via the fan-out announcement (`broadcastJoin`).
- **`IncomingMessageDispatcher`** — joiner reads from incoming `MemberAnnouncementPayload` + the inviting `GroupInvitationPayload.memberProfiles`. Self-profile populated via `IdentitySummary.sendingPublicKey`.
- **`CreateGroupInteractor`** — creator stamps their own Ed25519 pubkey into their own profile so it rides on the `GroupInvitationPayload` to every joiner.
- **`JoinRequestApprover.PendingRequest`** — new field carries it through approval UI plumbing.

## Out of scope
Verification logic — that lands in PR 4 along with the chat dispatcher branch.

## Test plan
[...]

PR #150, opened 2026-05-19
feat(chats): wire chat-message transport directions (PR 4/11 chat stack)

lead-onym:
**Both directions wired in one PR — symmetric, paired because they share `ChatMessagePayload`.** With PRs 1-3 underneath, two users in a Tyranny group can now exchange encrypted text end-to-end at the data + transport layer. No UI yet.

## Stacked on
This PR's branch is stacked on PRs #147, #148, #149 (chat-stack PRs 1, 2, 3). Once those merge to main, rebase this onto main and the predecessor commits drop out.

## Receive — `IncomingMessageDispatcher` chat-message branch

New third fast path on `dispatch()`: try-decode plaintext as a `ChatMessagePayload`. On match, run the trust chain:

1. **Envelope must have been signed** — guard `senderEd25519PublicKey` is non-nil. Anonymous chat messages aren't part of the v1 trust model.
2. **Group lookup** — payload's `group_id` must match a local `ChatGroup` owned by the receiving identity. Multi-identity correct: `groupRepository.currentGroups()` returns all cached groups across identities; the filter uses the `ownerIdentityID` parameter passed to `dispatch` (each identity's `InboxFanoutInteractor` subscription supplies the right value).
3. **Sender lookup** — payload's `sender_bls_pubkey_hex` must be in the group's `memberProfiles`.
4. **Insider-spoof check** — envelope's verified Ed25519 signer must equal that member's stored `MemberProfile.sendingPubkey`. Closes the gap PR 3 (#149) set up; without this, any group member could write another's BLS hex into the payload and the receiver would mis-attribute.
5. **Variant gate** — variant must match the group's governance type.
6. **Persist** via `messageRepository.insert` (idempotent on `messageID` so Nostr re-delivery is a no-op).

Dispatcher gains a `messageRepository: MessageRepository` field; production wiring in `OnymIOSApp` instantiates one alongside `GroupRepository` with the same on-disk-or-in-memory fallback.

## Send — `SendMessageInteractor`

Optimistic-insert pipeline mirroring `CreateGroupInteractor.sendInvitations`:

1. Validate (identity loaded, group exists, current identity is in roster, body non-empty, variant matches governance).
2. Mint `messageID` + payload, persist locally as `.pending` so the chat thread shows the bubble immediately.
3. Fan out one sealed envelope per **other** member's inbox key — best-effort, swallow per-recipient failures.
4. Flip status: `.sent` if any relay accepted (or no recipients), `.failed` if all sends rejected.

Today only the `.tyranny` variant ships; other group types throw `unknownGroup` until their variants come online.

## Tests
- **`IncomingMessageDispatcherChatMessageTests`** (8): happy path, unknown group / sender / signature mismatch / missing signature / wrong owner identity / **non-active identity (proves multi-identity)** / duplicate messageID idempotency.
- **`SendMessageInteractorTests`** (8): persist + fan out, skip self recipient, zero peers, all-fail / partial-fail status, error paths (unknown group, not a member, empty body).
- Existing dispatcher tests adapted to pass a `MessageRepository`.
- Full suite: **573 / 573**, 0 regressions.

## Multi-identity correctness (called out)
A specific test (`test_chatMessage_persistsForNonActiveIdentity`) seeds two identities, leaves identity A active, and dispatches a message addressed to identity B's inbox. Asserts the message lands in identity B's group regardless. The dispatcher reads `groups.currentGroups()` (unfiltered) and routes by the `ownerIdentityID` parameter, so inbox fan-out per identity does the right thing.

## Out of scope
UI — that's PR 5+ (`ChatThreadViewController`, UIKit shell, input panel, etc).

## Plan context
PR 4 of 11. **PR 1** (#147) — payload type. **PR 2** (#148) — persistence. **PR 3** (#149) — Ed25519 sender pubkey plumbing. **PR 5 next**: UIKit `ChatThreadViewController` skeleton + SwiftUI bridge + replace `ChatMembersView` entry point.

### DIFF
--- a/Sources/OnymIOS/Chats/SendMessageInteractor.swift
+++ b/Sources/OnymIOS/Chats/SendMessageInteractor.swift
@@ -1,0 +1,197 @@
+import Foundation
+import CryptoKit
+
+/// Outgoing-message pipeline: persist a `.pending` row locally,
+/// seal + ship one envelope per other group member, then flip the
+/// status to `.sent` (at least one relay accepted) or `.failed`
+/// (every send threw or no relay accepted).
+///
+/// Mirrors `CreateGroupInteractor.sendInvitations` shape — same
+/// `sealInvitation` + `InboxTransport.send` per recipient, same
+/// inbox-tag derivation. The local insert happens *before* the fan-
+/// out so the chat thread sees the bubble immediately; the status
+/// flip is the UI's signal that the network is done.
+actor SendMessageInteractor {
+    private let identity: IdentityRepository
+    private let inboxTransport: any InboxTransport
+    private let messageRepository: MessageRepository
+    private let groupRepository: GroupRepository
+
+    enum SendError: Error, Equatable {
+        case noIdentityLoaded
+        case unknownGroup
+        /// Current identity isn't in the group's `memberProfiles`. The
+        /// roster lookup is by BLS pubkey hex, so this fires when the
+        /// creator's own profile was never recorded (shouldn't happen
+        /// post-PR-3) or when the active identity switched mid-flight.
+        case senderNotAMember
+        /// Caller-side guard. Empty bodies are technically valid on
+        /// the wire (the payload encodes them) but the interactor
+        /// refuses to ship them — a no-op send would burn a relay
+        /// publish for nothing.
+        case emptyBody
+    }
+
+    init(
+        identity: IdentityRepository,
+        inboxTransport: any InboxTransport,
+        messageRepository: MessageRepository,
+        groupRepository: GroupRepository
+    ) {
+        self.identity = identity
+        self.inboxTransport = inboxTransport
+        self.messageRepository = messageRepository
+        self.groupRepository = groupRepository
+    }
+
+    /// Returns the locally-persisted `ChatMessage` after the fan-out
+    /// completes. The returned row always reflects the final status
+    /// (`.sent` or `.failed`) — callers that want the optimistic
+    /// `.pending` view subscribe to `MessageRepository.snapshots`.
+    @discardableResult
+    func send(
+        groupID: String,
+        body: String,
+        now: Date = Date()
+    ) async throws -> ChatMessage {
+        guard !body.isEmpty else { throw SendError.emptyBody }
+
+        guard let active = await identity.currentIdentity() else {
+            throw SendError.noIdentityLoaded
+        }
+        let groups = await groupRepository.currentGroups()
+        guard let group = groups.first(where: { $0.id == groupID }) else {
+            throw SendError.unknownGroup
+        }
+
+        let myBlsHex = active.blsPublicKey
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
COMMENT — solid PR with sound trust chain and good test coverage; a few polish issues worth surfacing.

- `Sources/OnymIOS/Chats/SendMessageInteractor.swift:107` — `SendError.unknownGroup` is overloaded for two distinct cases: group truly missing (`groups.first(where:)` fails) AND group exists but variant is unsupported (`.oneOnOne/.anarchy/...`). Caller can't distinguish "no such group" from "chat not supported for this group type" — add a dedicated `.unsupportedVariant` case to keep error semantics honest.
- `Sources/OnymIOS/Chats/SendMessageInteractor.swift:79` — interactor finds groups via unfiltered `groupRepository.currentGroups()`, then validates via `active.blsPublicKey`. If active identity ≠ group owner but the active identity happens to also be a member (multi-identity sharing a group), the send would succeed under the wrong identity. Worth either filtering by `active`'s identity ID or accepting `ownerIdentityID` as a parameter (mirroring the dispatcher's pattern).
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:457,464,471,479` — every drop in `persistChatMessage` is a silent `return`. The signature-mismatch case at :471 is the insider-spoof defense — a successful active attack against this check is exactly the event you'd want a log/metric for. Consider at least `os_log` on that branch.
- `Sources/OnymIOS/Chats/MessageRepository.swift:18` — `cached: [String: [ChatMessage]]` grows monotonically per group touched this session; no eviction. Fine for v1's group counts but worth a note that the comment "Kept narrow on purpose" doesn't actually bound the cache.
- `Tests/OnymIOSTests/SendMessageInteractorTests.swift` — no coverage for the unsupported-variant throw path (seed a `.oneOnOne` group, assert it throws). Easy add and would catch the overloaded-error issue above if you split it.
- `Sources/OnymIOS/Chats/SendMessageInteractor.swift:135-152` — encoding-failure branch is unreachable in practice (`ChatMessagePayload` has no failing fields), so the elaborate fallback path is dead code. Either `try!` and remove or trim the rebuilt-`ChatMessage` boilerplate.
- `Sources/OnymIOS/Group/MemberProfile.swift:51` — manual `init(from:)` with no matching `encode(to:)`; relies on synthesized encode. Works because `CodingKeys` is fully specified, but a future field added only to one half would silently diverge. Minor — pinning encode explicitly mirrors how `MemberAnnouncementPayload.AnnouncedMember` does it.

Trust-chain logic at `IncomingMessageDispatcher.swift:480` (`senderEd25519PublicKey == senderProfile.sendingPubkey`) looks correct, and the multi-identity test (`test_chatMessage_persistsForNonActiveIdentity`) is a nice catch.

PR #151, opened 2026-05-19
feat(chats): UIKit chat-thread shell + SwiftUI bridge (PR 5/11 chat stack)

lead-onym:
**First user-visible PR of the chat stack.** Tapping a group in the Chats tab now opens a UIKit `ChatThreadViewController` instead of the members roster. The members list moves one tap deeper, behind the thread's info button.

## Stacked on
This PR's branch is stacked on #150 (PR 4). Base is `pr-chat-4-dispatcher-and-send`.

## What's in here

**`ChatThreadViewController` (UIKit)** — `Sources/OnymIOS/Chats/ChatThreadViewController.swift`
- Custom top bar: back chevron (left) + centered group name + info button (right).
- Empty `UITableView` placeholder. PR 6 attaches a diffable data source + custom bubble cell.
- Empty `UIView` input-panel placeholder. PR 6/7 build the UITextView + send button + keyboard handling.
- Fired-via-closure: `onBack` and `onShowMembers`. No internal navigation — the SwiftUI wrapper handles all routing.

**`ChatThreadView` (SwiftUI bridge)** — `Sources/OnymIOS/Chats/ChatThreadView.swift`
- `UIViewControllerRepresentable` wrapping the controller.
- `.toolbar(.hidden, for: .navigationBar)` so SwiftUI's nav bar disappears — the controller's custom top bar is the only one drawn.
- `Environment(\.dismiss)` wired to the back chevron.
- `navigationDestination(isPresented:)` pushes `ChatMembersView` when info is tapped.
- Group name is computed from `chatsFlow.groups.first { $0.id == groupID }?.name` and passed in every render — `updateUIViewController` keeps the title in sync as the group renames (PR 9 polish) without recreating the controller.

**`ChatsView` rewire** — `Sources/OnymIOS/Chats/ChatsView.swift`
- `.navigationDestination(for: String.self)` body switched from `ChatMembersView` to `ChatThreadView`.
- One-line change. Members surface is still reachable from the thread's info button.

## What's NOT in here
- No message rendering (PR 6).
- No input UI or send wiring (PR 6 / PR 7).
- No keyboard avoidance (PR 6).
- No styling polish (PR 9).
- The input panel shows placeholder text "Message input lands in PR 6" so the UI region is visible during this slice.

## Tests
- `ChatThreadViewControllerTests` (5 new): viewDidLoad doesn't crash, `update(groupName:)` writes through, empty name falls back to "Chat", back/info button taps invoke their closures.
[...]

PR #152, opened 2026-05-19
feat(chats): message list rendering (PR 6/11 chat stack)

lead-onym:
**Read-only chat scrollback.** Messages from `MessageRepository` flow through SwiftUI into the UIKit controller and render as direction-styled bubbles. No input or send yet — that's PR 7.

## Stacked on
This PR's branch is stacked on #151 (PR 5). Base is \`pr-chat-5-uikit-shell\`.

### `ChatBubbleCell` — `Sources/OnymIOS/Chats/ChatBubbleCell.swift`
Custom `UITableViewCell`. Two visual modes toggled by `ChatMessage.direction`:
- `.outgoing` → accent fill (`OnymAccent.blue`), trailing-aligned, `onAccent` text.
- `.incoming` → `surface2` fill, leading-aligned, text color.

Bubble max width capped at 75% of the cell so long messages don't reach the opposite edge. The two alignment constraints (leading vs trailing) are stored as properties; `configure(message:)` swaps which is active so reused cells flip sides correctly.

### `ChatThreadViewController` gains
- `UITableViewDiffableDataSource<Section, UUID>` keyed by message id — swapping cells across renders doesn't churn rows that didn't actually change.
- `update(messages:)` — called by the SwiftUI bridge on every render. First apply is non-animated and **always scrolls to bottom** (cold open lands on the latest message). Subsequent applies animate and **only auto-scroll when the user was already near the bottom** (reading older messages doesn't get hijacked).
- `isNearBottom` heuristic with a 100pt threshold from the content end. Exposed for tests.

### `ChatThreadView` (SwiftUI bridge)
- New `messageRepository` parameter.
- `task(id: groupID)` subscribes to `messageRepository.snapshots(groupID:)`. Per-group restart on navigation so leaving a chat cancels its stream.
- Snapshot pushed into the controller via `updateUIViewController`.

### `AppDependencies` — adds `messageRepository`
Shared singleton, plumbed through `RootView` → `ChatsView` → `ChatThreadView`.

## What's NOT in here
- Input UI or send wiring (PR 7).
- Keyboard avoidance (PR 7).
- "↓ new messages" pill (deferred to PR 9 polish — for now, messages that arrive while the user is scrolled up simply appear without disrupting their position).
- Avatars, timestamps, status glyphs, emoji rendering (PR 8 + later).

## Tests (11 new, suite 589 / 589)

**`ChatBubbleCellTests`** (4):
- outgoing pins trailing, releases leading.
- incoming pins leading, releases trailing.
- reconfigure flips alignment (cell reuse path).
- body text writes through to the inner label.

**`ChatThreadViewControllerTests`** +7:
- `update(messages:)` renders 0 / N rows.
- subsequent add reflects diff (N → N+1).
- subsequent remove shrinks table.
- `isNearBottom`: empty content → true.
- `isNearBottom`: at content end → true.
- `isNearBottom`: scrolled up → false.

## Visual correctness — NOT covered by tests
Bubble shape, colors, scroll behavior: verify in the simulator. Easiest end-to-end demo: two devices in the same Tyranny group, send a message from one, watch it land on the other. The receive path is already covered by the dispatcher tests in #150.

## Plan context
PR 6 of 11. **PR 7 next**: input panel + send wiring — `UITextView` with auto-height + send button + keyboard handling, hook the Send button to `SendMessageInteractor` (from #150).

### DIFF
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -24,4 +24,9 @@ struct AppDependencies {
     /// and the underlying collector should run for the app's
     /// lifetime regardless of which surface is mounted.
     let approveRequestsFlow: ApproveRequestsFlow
+    /// Single shared instance — the chat-thread screen subscribes to
+    /// per-group message snapshots, and the receive-side dispatcher
+    /// writes into the same actor. Constructed once in
+    /// `OnymIOSApp.init` and threaded down.
+    let messageRepository: MessageRepository
 }
--- a/Sources/OnymIOS/Chats/ChatBubbleCell.swift
+++ b/Sources/OnymIOS/Chats/ChatBubbleCell.swift
@@ -1,0 +1,126 @@
+import SwiftUI
+import UIKit
+
+/// One message bubble row. Two visual styles toggled by
+/// `ChatMessage.direction`:
+///
+///   - `.outgoing` — accent-filled bubble pinned to the trailing edge.
+///   - `.incoming` — surface-tinted bubble pinned to the leading edge.
+///
+/// Layout is a single `UILabel` inside a `UIView` bubble container.
+/// The bubble's leading/trailing edge is swapped via two stored
+/// constraints; the `lessThanOrEqualTo` width cap keeps long
+/// messages from filling the full row.
+///
+/// PR 6 scope: no avatars, no timestamps, no status glyphs, no
+/// emoji rendering. Status indicator lands in PR 8.
+final class ChatBubbleCell: UITableViewCell {
+    static let reuseID = "ChatBubbleCell"
+
+    /// Max bubble width as a fraction of the cell's content width.
+    /// Matches the usual messenger convention — long messages wrap
+    /// but never reach the opposite edge.
+    private let maxWidthFraction: CGFloat = 0.75
+
+    private let bubble = UIView()
+    private let bodyLabel = UILabel()
+
+    // Direction-dependent constraints — only one of these pair is
+    // active at a time.
+    private var leadingAlignConstraint: NSLayoutConstraint!
+    private var trailingAlignConstraint: NSLayoutConstraint!
+
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        backgroundColor = .clear
+        selectionStyle = .none
+        contentView.backgroundColor = UIColor(OnymTokens.bg)
+        buildBubble()
+        layoutBubble()
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) not implemented")
+    }
+
+    /// Swap colors + alignment for the message's direction. Caller
+    /// (the diffable data source's cell provider) invokes this on
+    /// every dequeue.
+    func configure(message: ChatMessage) {
+        bodyLabel.text = message.body
+        switch message.direction {
+        case .outgoing:
+            bubble.backgroundColor = UIColor(OnymAccent.blue.color)
+            bodyLabel.textColor = UIColor(OnymTokens.onAccent)
+            leadingAlignConstraint.isActive = false
+            trailingAlignConstraint.isActive = true
+        case .incoming:
+            bubble.backgroundColor = UIColor(OnymTokens.surface2)
+            bodyLabel.textColor = UIColor(OnymTokens.text)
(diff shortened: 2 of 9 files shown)


dev-onym (commented):
REQUEST_CHANGES

- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:98-119` — **Auto-Layout conflict on every cell**. The `greaterThanOrEqualTo … +56` (leading) and `lessThanOrEqualTo … -56` (trailing) are *always* active, but the equal-to alignment constraint is pinned at ±12. For an incoming bubble, `leading == +12` and `leading >= +56` are mutually unsatisfiable; for outgoing, `trailing == -12` vs `trailing <= -56` is the same conflict. UIKit will log a constraint break each row and silently disable one. Fix: pair each opposite-edge `>=`/`<=` with the *opposite* alignment direction (activate only the one that isn't currently the equal-pin), or relax the equal constraint's priority below required. Tests pass because they only inspect `isActive`, not layout validity — please reproduce in a sim and watch the console.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:96-103` — Diffable identity is just `UUID`, so a re-render with the same id set but a mutated `ChatMessage` (status flip, future body edit) won't reconfigure visible cells; `messagesByID` is updated, but the cell only re-runs `configure` on dequeue. Fine for PR 6's read-only/no-status scope, but flag for PR 8 — consider `snapshot.reconfigureItems(changedIDs)` then.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:89-91` — `update(messages:)` documents but doesn't enforce ascending-`sentAt` order. Since `MessageRepository.snapshots` is the sole caller, an unsorted snapshot would silently land messages in the wrong row order. Either sort here defensively or assert/document the repository's ordering contract.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:71` — `bodyLabel.font = .systemFont(ofSize: 15)` is fixed; no Dynamic Type. Probably fine as a PR-6 stub, but worth tracking for the accessibility pass.
- Nice touches: first-apply non-animated + always-scroll, subsequent-applies near-bottom gate, per-`groupID` `task(id:)` to cancel streams on nav — all behave the way the description claims.

dev-onym (commented):
**COMMENT** — solid PR with thorough tests; one real behavior bug worth fixing before PR 7.

- **First-apply animation is bypassed in practice.** `ChatThreadView.swift:25` initializes `@State messages: []`, so `makeUIViewController` calls `update(messages: [])` first — `hasAppliedFirstSnapshot` flips to `true` on an empty snapshot. The first *real* snapshot then takes the `animatingDifferences: true` path with `wasNearBottom == true` (empty table), so the cold-open "fly-in" the comment at `ChatThreadViewController.swift:96-100` claims to prevent will actually animate. Fix: gate `hasAppliedFirstSnapshot = true` on `!sorted.isEmpty`, or skip applying empty snapshots when nothing has been applied yet.
- **`allConstraintsStillActive(_:)` magic number is off** (`ChatBubbleCellTests.swift:117-122`). Counting from `layoutBubble()`: 4 bubble pins (top/bottom/width-cap is 3; bodyLabel pins are between bubble and label, so attached to bubble) + 4 bodyLabel insets + 2 direction = 9 on contentView+bubble combined, but the comment says "8 + 2 == 10". The `>= 10` threshold may overshoot — recount or relax the comment so the next reader doesn't trust a wrong invariant.
- **PR description test counts mismatch the diff.** Body says "11 new" — actual diff adds 5 (`ChatBubbleCellTests`) + 8 (`ChatThreadViewControllerTests`) = 13. Body also lists 4 cell tests but `test_layoutResolves_withoutConstraintConflicts` is a 5th, and the controller block adds `test_updateMessages_unsortedInput_isSortedAscending` (the PR body's own "defensive sort" rationale) without listing it. Update the PR body so the suite total claim (589/589) actually matches.
- **`scrollToBottom(animated:)` is internal, not private** (`ChatThreadViewController.swift:144`). No test uses it; tighten visibility to `private` to keep the surface area small (companion to the `isNearBottom` "exposed for tests" doc).
- **Re-entering a chat flashes empty.** `@State messages = []` is rebuilt on each navigation; until the first stream tick lands, the controller renders 0 rows. Not a blocker for PR 6 but worth a TODO alongside the deferred "↓ new messages" pill.

Nice touches: `wasNearBottom` is captured before the diff apply (`ChatThreadViewController.swift:90`), `task(id: groupID)` correctly restarts on navigation, and the constraint-pairing rationale comment at `ChatBubbleCell.swift:21-27` accurately calls out the prior conflict shape.

PR #153, opened 2026-05-19
feat(chats): input panel + keyboard avoidance (PR 7/11 chat stack)

lead-onym:
**Composer is real now.** A `UITextView` that grows to a 3-line cap, a Send button, bottom-anchored to `view.keyboardLayoutGuide.topAnchor` so the panel rides with the keyboard. **No real send wiring yet** — tap is a no-op that clears the field. PR 8 swaps the closure body for `SendMessageInteractor`.

## Stacked on
Base: \`pr-chat-6-message-list\` (#152).

## `ChatInputPanelView`
- `UITextView` with body-font + Dynamic Type, rounded fill, inset 12pt on the leading; Send button on the trailing.
- Placeholder `UILabel` overlay ("Message") that hides on first keystroke.
- **Auto-height**: text view grows up to `intrinsicHeight(forLines: 3)`, then scrolls internally. Below 3 lines it sizes to the actual content, never below the one-line floor. The floor + cap are derived from `font.lineHeight + textContainerInset.top + textContainerInset.bottom` so Dynamic Type re-flows correctly.
- `onSendTapped: (String) -> Void` fires on tap with the body trimmed of leading/trailing whitespace. Whitespace-only input is a no-op (the button's disabled anyway; this is belt-plus-braces).
- `refreshAfterTextChange` calls `textView.layoutIfNeeded()` before measuring — **TextKit 2 (iOS 18+ default) returns stale `sizeThatFits` results immediately after a programmatic `.text = ...` write**. The forced flush makes the height update correctly in both interactive and programmatic paths.

## `ChatThreadViewController` wire-up
- Replaced the PR-5 placeholder UIView with a real `ChatInputPanelView`.
- Bottom anchor switches from `safeArea.bottomAnchor` to `view.keyboardLayoutGuide.topAnchor`. **No manual `keyboardWillShow` notifications or frame math.** With no keyboard up, the guide equals the safe-area bottom; when the keyboard rises, the panel slides automatically.
- Dropped the fixed 56pt height — the panel sizes intrinsically from its text-view height + 16pt vertical padding.
- For PR 7, `onSendTapped` just clears the field. PR 8 will replace this closure body with `SendMessageInteractor.send`.

## Tests (9 new, full suite 600 / 600)

**`ChatInputPanelViewTests`** (8):
- initialState: send disabled
- setText nonEmpty: enables send; emptyAgain: disables
- tap send → closure fires with trimmed body
- tap send on whitespace-only → no-op
- single-line text: same height as empty
- multi-line text: grows beyond single-line height
- text growth caps at `maxLineCount` (3 vs 10 lines = same height)

**`ChatThreadViewControllerTests`** +1:
- inputPanel is hosted; tapping send clears the field (PR-7 closure contract).

One previously-passing controller test (`test_updateMessages_unsortedInput_isSortedAscending`) needed a real `UIWindow` because the PR 7 layout uses `keyboardLayoutGuide`, which only resolves inside a window. Switched that test to host the controller in a window.

## Visual correctness — NOT covered by tests
Bubble layout while typing, keyboard slide animation, Dynamic Type scaling: verify in the simulator. Tap a group, tap the field, type, hit Send → the field should clear and the keyboard stays up.

## Plan context
PR 7 of 11. **PR 8 next**: replace the placeholder `onSendTapped` closure with a real `SendMessageInteractor.send` call — optimistic insert as `.pending`, status flip on completion. Will need to plumb `SendMessageInteractor` through `AppDependencies` → `ChatThreadView`.

### DIFF
--- a/Sources/OnymIOS/Chats/ChatInputPanelView.swift
+++ b/Sources/OnymIOS/Chats/ChatInputPanelView.swift
@@ -1,0 +1,212 @@
+import SwiftUI
+import UIKit
+
+/// Bottom input panel for the chat thread. Hosts the message
+/// composer's `UITextView`, a Send button, and the hairline divider
+/// that separates the panel from the message list.
+///
+/// Height behavior — the text view grows with content up to a
+/// 3-line cap (`maxLineCount`); beyond that it scrolls internally.
+/// Single-line state hugs `minTextViewHeight` so the panel never
+/// shrinks below a thumb-tappable target.
+///
+/// PR 7 scope: layout + keyboard avoidance + enabled-state toggle.
+/// **No real send wiring** — `onSendTapped` clears the text but the
+/// host doesn't do anything else with the body. PR 8 hooks the
+/// closure to `SendMessageInteractor`.
+final class ChatInputPanelView: UIView {
+    /// Invoked when the user taps Send. Receives the current text
+    /// trimmed of leading/trailing whitespace. PR 8 will wire this
+    /// to `SendMessageInteractor.send`.
+    var onSendTapped: ((String) -> Void)?
+
+    var text: String {
+        get { textView.text ?? "" }
+        set {
+            textView.text = newValue
+            // Re-evaluate height + send-button state + placeholder
+            // so external writes look the same as user typing.
+            refreshAfterTextChange()
+        }
+    }
+
+    /// Cap on the auto-grow behavior. Past this many lines the
+    /// text view scrolls internally instead of pushing the
+    /// message list further up.
+    static let maxLineCount = 3
+
+    private let topDivider = UIView()
+    private let textView = UITextView()
+    private let placeholderLabel = UILabel()
+    private let sendButton = UIButton(type: .system)
+    private var textViewHeightConstraint: NSLayoutConstraint!
+
+    /// Height for exactly `lines` worth of text plus the text
+    /// view's vertical insets. Drives both the natural empty-state
+    /// floor (`lines = 1`) and the auto-grow cap
+    /// (`lines = maxLineCount`). Computed dynamically so Dynamic
+    /// Type changes the floor too.
+    func intrinsicHeight(forLines lines: Int) -> CGFloat {
+        let font = textView.font ?? .preferredFont(forTextStyle: .body)
+        let lineHeight = ceil(font.lineHeight)
+        let insets = textView.textContainerInset.top + textView.textContainerInset.bottom
+        return lineHeight * CGFloat(lines) + insets
+    }
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        backgroundColor = UIColor(OnymTokens.surface2)
+        buildSubviews()
+        layoutSubviewsLocal()
+        refreshAfterTextChange()
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) not implemented")
+    }
+
+    // MARK: - Build
+
+    private func buildSubviews() {
+        topDivider.backgroundColor = UIColor(OnymTokens.hairline)
+        topDivider.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(topDivider)
+
+        textView.font = .preferredFont(forTextStyle: .body)
+        textView.adjustsFontForContentSizeCategory = true
+        textView.textColor = UIColor(OnymTokens.text)
+        textView.backgroundColor = UIColor(OnymTokens.bg)
+        textView.layer.cornerRadius = 18
+        textView.layer.cornerCurve = .continuous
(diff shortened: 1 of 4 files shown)


dev-onym (commented):
REQUEST_CHANGES — the documented "non-whitespace" enable contract isn't actually implemented; PR 8 plans to trust it.

- **`ChatInputPanelView.swift:188`** — `sendButton.isEnabled = !body.isEmpty` uses the raw text, but the PR description, the type-level doc comment, and the test-file header all claim the button is "enabled iff the text view has non-whitespace content." With `panel.text = "   "` the button is enabled, yet `tappedSend` no-ops. The "(button's disabled anyway; belt-plus-braces)" comment on the no-op path is false. Fix: gate on `!body.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty`. Add a test that asserts `isEnabled == false` for whitespace-only — `test_tappingSend_emptyText_isNoOp` only checks the closure, not the button state, so this regressed silently.
- **`ChatInputPanelViewTests.swift:91-99`** — `test_singleLineText_doesNotGrowBeyondEmpty` reads the height pre-text via `textViewHeight(...)`, but the panel has no superview, so the height constraint hasn't been driven to a "real" value yet — both reads will return the initial `36` constant, making the assertion pass trivially. Putting the panel in a window (as the controller test already does) or asserting against `intrinsicHeight(forLines: 1)` would make the test meaningful.
- **`ChatInputPanelView.swift:124`** — `UIScreen.main.scale` is soft-deprecated in iOS 16+ for multi-scene apps; prefer `traitCollection.displayScale` (resolved per-trait-collection-change). Minor; same pattern exists elsewhere in the repo, so flagging only.
- **`ChatInputPanelView.swift:139-146`** — placeholder constraints capture `textContainerInset` constants at constraint-creation time. Not a bug today (inset is constant), but if/when inset becomes Dynamic-Type-driven the placeholder won't move. Worth a one-line comment noting the coupling, or just constrain to `textView.layoutMarginsGuide`-style anchors.
- Nit: `text` getter `textView.text ?? ""` is dead defense — `UITextView.text` is `String!`, never `nil` once initialized. Same at lines 158 and 201.

Tests, structure, keyboard-guide approach, and the controller test that switched to a real window all look correct. Fix the trim gap and I'd approve.

dev-onym (commented):
Solid PR. A few minor observations only.

APPROVE

- ChatInputPanelView.swift:128 — `1.0 / UIScreen.main.scale`: `UIScreen.main` is deprecated in iOS 16+ and gives wrong scale on external/connected displays. Matches the pre-existing pattern in `ChatThreadViewController`, so fine to defer, but worth a follow-up to derive from `traitCollection.displayScale` (or the view's `window?.screen`).
- ChatInputPanelView.swift:140-141 — Send button is a 36×36 hit target; HIG recommends ≥44pt for primary controls. Consider enlarging the button (the symbol can stay 28pt) or adding `hitTest` slop. Minor a11y polish.
- ChatInputPanelView.swift:139 — Send button bottom-aligns to `textView.bottom`. With internal scrolling at the 3-line cap the button still anchors to the visible bottom, which is correct; just confirm visually that it doesn't visually crowd the scroll indicator on a maxed-out composer.
- ChatInputPanelView.swift:191-195 — `layoutSubviews` mutates `textViewHeightConstraint.constant`. The 0.5pt guard prevents loops, but if width is unstable across passes (e.g., container animations) this can flicker; consider invalidating intrinsic size and letting Auto Layout drive it instead. Non-blocking for PR 7.
- ChatInputPanelView.swift:69 — `UIColor(OnymTokens.surface2)` snapshots the SwiftUI color at init; dark↔light mode switches while mounted won't re-resolve. Same pattern as elsewhere in this stack, so not a PR-7 regression.
- ChatInputPanelViewTests.swift:166-172 — `textViewHeight` reads `tv.constraints.first { firstAttribute == .height }`. Brittle: any future intrinsic/content-hugging change would silently return `nil` and fall back to `bounds.height`. Consider exposing `textViewHeightConstraint` internally for tests or asserting the constraint is found.
- ChatInputPanelViewTests.swift:174-179 / sendButton:155 — `fatalError` in test helpers loses the line of the failing assertion. `XCTFail` + `nil`-return is friendlier on CI.
- ChatInputPanelViewTests.swift:142 — `makePanelInWindow` builds a window per test with no teardown. Swift will release it, but stashing in an `addTeardownBlock { window.isHidden = true }` reads cleaner and avoids surprises if a test holds a reference accidentally.
- ChatThreadViewControllerTests.swift:139 — `test_inputPanel_isHosted_andClearsTextOnSendTap` uses `loadViewIfNeeded()` without a window; `keyboardLayoutGuide` constraints can log warnings here. Functionally OK because the assertion is on closure behavior, but consistent with the sibling fix at line 96-104 it could also host in a window.

PR #154, opened 2026-05-19
feat(chats): wire Send button to SendMessageInteractor (PR 8/11 chat stack)

lead-onym:
**End-to-end.** Type a message, tap Send, see it appear in the thread immediately and propagate to the other group members.

## Stacked on
Base: \`pr-chat-7-input-panel\` (#153).

## What's in here

PR 4 (#150) already built `SendMessageInteractor` with the full optimistic-insert / fan-out / status-flip lifecycle. PR 8 plumbing is thin:

- **`AppDependencies.sendMessageInteractor`** — shared actor, constructed in `OnymIOSApp.init` alongside the existing repositories.
- **`RootView` → `ChatsView` → `ChatThreadView`** thread it through.
- **`ChatThreadView`** constructs the fire-and-forget closure:
  ```swift
  Task { try? await interactor.send(groupID: groupID, body: body) }
  ```
  Errors are swallowed because the `SendError` cases (no identity loaded, unknown group, not a member, empty body) all represent invariants that should hold mid-chat-screen. Network / fan-out failures flip the message status to `.failed` inside the interactor — PR 9 surfaces that visually.
- **`ChatThreadViewController`** exposes `onSendTapped: ((String) -> Void)?`. The input panel forwards its trimmed body to this dispatcher *before* clearing the field, so the host sees the typed content even though the composer empties immediately.

## UX flow
1. User types "hi" → Send button enables (PR 7 trim gate).
2. Tap Send → controller forwards "hi" to the interactor.
3. Interactor inserts a `.pending` row into `MessageRepository` **before** the await.
4. `MessageRepository.snapshots` emits → PR 6 diffable data source reconciles → bubble appears + table auto-scrolls.
5. Input clears immediately (snappier than waiting for the await).
6. Network round-trip resolves → interactor flips status to `.sent` / `.failed`. PR 9 will style the bubble accordingly.

## Tests (2 new, suite 603 / 603)

**`ChatThreadViewControllerTests`** +2:
- `test_inputPanel_send_invokesOnSendTapped_withTrimmedBody` — the controller must forward the trimmed body to the host's dispatcher closure.
- `test_inputPanel_send_doesNotFire_forWhitespaceOnlyBody` — belt-plus-braces; even if a future caller bypasses the input panel's disabled-state gate, the controller forwarder must not invoke the dispatcher with an empty body.

PR 4's `SendMessageInteractorTests` already cover the optimistic-insert / fan-out / status-flip behavior; no new tests needed there.

## Visual / end-to-end correctness — NOT covered by tests
Bubble appearance timing, status flip animation, table auto-scroll on send: **verify in the simulator with two devices in the same Tyranny group** to confirm the full round-trip works.

## Plan context
PR 8 of 11. **PR 9 next**: message status indicator (clock / check / red bang) + retry-on-failed. Needs the PR 6 diffable data source's `reconfigureItems` switch to land too.

### DIFF
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -29,4 +29,9 @@ struct AppDependencies {
     /// writes into the same actor. Constructed once in
     /// `OnymIOSApp.init` and threaded down.
     let messageRepository: MessageRepository
+    /// Stateless façade over identity + transport + repositories
+    /// for outgoing chat messages. Safe to share across screens
+    /// (it's an actor); each chat-thread view captures it and
+    /// dispatches `send(groupID:body:)` on the send-button tap.
+    let sendMessageInteractor: SendMessageInteractor
 }
--- a/Sources/OnymIOS/Chats/ChatThreadView.swift
+++ b/Sources/OnymIOS/Chats/ChatThreadView.swift
@@ -20,6 +20,7 @@ struct ChatThreadView: View {
     @Bindable var chatsFlow: ChatsFlow
     @Bindable var identitiesFlow: IdentitiesFlow
     let messageRepository: MessageRepository
+    let sendMessageInteractor: SendMessageInteractor
     let makeShareInviteFlow: @MainActor () -> ShareInviteFlow
 
     @Environment(\.dismiss) private var dismiss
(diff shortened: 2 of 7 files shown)


dev-onym (commented):
APPROVE

- `Sources/OnymIOS/Chats/ChatThreadView.swift:51-52` — explicit `let interactor = sendMessageInteractor; let groupID = groupID` before the `Task` is the right call: it avoids capturing `self` (which holds `@Bindable` observables that aren't unconditionally Sendable in Swift 6 strict mode). The shadowed `groupID` reads awkwardly but is intentional. Optional: rename to `let capturedGroupID = groupID` for clarity, but not blocking.
- `Sources/OnymIOS/Chats/ChatThreadView.swift:43` — "stream we're subscribing to above" is mildly misleading since the `.task(id:)` modifier appears below the bridge construction in source order; reads fine in practice. Nit.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:251-255` — forward-then-clear ordering matches the PR description and keeps the composer responsive. The forwarder is correctly unconditional; the whitespace gate lives in `ChatInputPanelView.tappedSend` (`ChatInputPanelView.swift:219-224`), which short-circuits before invoking `onSendTapped`, so `test_inputPanel_send_doesNotFire_forWhitespaceOnlyBody` is verifying the panel's contract rather than a check at the controller layer — the test's "if anything ever bypasses that" framing is slightly off but harmless.
- `Sources/OnymIOS/OnymIOSApp.swift:223-228` — interactor constructed once with the same `repository` / `inboxTransport` / `messageRepository` / `groupRepository` used elsewhere; shared actor semantics intact. Good.
- `Sources/OnymIOS/Chats/ChatThreadView.swift:39-56` — swallowing errors with `try?` is justified by the PR description (precondition violations only; network failures flip to `.failed` inside the interactor). Worth confirming during the manual two-device simulator pass that no `SendError` case can fire mid-chat-screen for legitimate reasons (e.g., race between leaving the group and tapping Send) before PR 9 lands.
- Tests in `Tests/OnymIOSTests/ChatThreadViewControllerTests.swift:159-186` cleanly pin the new contract; existing PR 7 test trimmed to its actual invariant.

PR #155, opened 2026-05-19
feat(chats): message status indicator + retry on failed (PR 9/11 chat stack)

lead-onym:
**Status glyphs on outgoing bubbles + tap-to-retry on failed.** Tapping a failed bubble retries the send with the original message id so receivers dedup against any prior delivery.

## Stacked on
Base: \`pr-chat-8-wire-send\` (#154).

## `ChatBubbleCell`
- **Status indicator** — `UIImageView` below the bubble's trailing edge (outgoing only). Glyphs:
  | Status | SF Symbol | Color |
  |---|---|---|
  | `.pending` | `clock` | muted (`text3`) |
  | `.sent` | `checkmark` | muted (`text3`) |
  | `.failed` | `exclamationmark.circle.fill` | red |
  Incoming rows hide the indicator (the message arriving is proof of delivery).
- **Conditional cell-bottom anchor** — outgoing pins to the status icon's bottom (extra room for the glyph); incoming pins to the bubble's bottom (no extra height for the hidden indicator).
- **Retry tap** — a `UITapGestureRecognizer` is installed on the bubble **only** when `status == .failed && direction == .outgoing`. Cell reuse forgets the closure on each `configure`, so a bubble that flips out of `.failed` loses its tap target.
- `simulateBubbleTapForTest()` (DEBUG-only) — small internal test seam because `UITapGestureRecognizer` has no clean public "fire" API.

## `ChatThreadViewController`
- **Diffable now reconfigures on content change.** Before this PR, the diff identity was just `UUID`, so a status flip (same id, different content) didn't repaint. Now we detect changed messages (`prior != msg`) and call `snapshot.reconfigureItems(changedIDs)` — the cell stays at the same index, just re-runs the cell provider against the updated `messagesByID`. Closes the PR 6 review TODO.
- `onRetryRequested: ((UUID) -> Void)?` exposed; cell provider passes a per-message retry closure to `ChatBubbleCell` only for failed outgoing rows.

## `ChatThreadView` (SwiftUI bridge)
- Constructs the retry closure as fire-and-forget `Task { await interactor.retry(groupID:messageID:) }` — same shape as the send dispatcher.

## `SendMessageInteractor`
- **Extracted `fanOut(payload:recipients:)`** shared between send + retry. Same encode + per-recipient seal + transport flow.
- **New `retry(groupID:messageID:)`** — no-op (silent) for unknown / non-failed / non-outgoing rows (prevents double-delivery on `.sent` / `.pending`). On a valid retry:
  1. Flips status to `.pending` immediately so the glyph swaps to the in-flight clock before any network work.
  2. Re-fans out using the **original payload fields** (same UUID + body + sentAt) so receivers dedup against any prior delivery.
  3. Flips to `.sent` / `.failed` on completion.

## Tests (14 new, suite 617 / 617)

**`ChatBubbleCellTests`** +6:
- outgoing `.pending` / `.sent` / `.failed` each show the correct glyph
- incoming hides the indicator
- reconfigure flips the glyph (cell reuse)
- failed tap fires the retry closure
- non-failed tap is a no-op
- reconfigure from `.failed` to `.sent` drops the retry closure

**`ChatThreadViewControllerTests`** +2:
- failed-bubble tap invokes `onRetryRequested` with the message id
- status flip on the same UUID reconfigures the visible cell (proves `reconfigureItems` lands)

**`SendMessageInteractorTests`** +4:
- retry on a failed message flips to `.sent` on relay acceptance
- retry stays `.failed` if relays still reject
- retry on an unknown messageID is a no-op
- retry on a `.sent` message is a no-op (prevents double-delivery)

## Plan context
PR 9 of 11. **PR 10 next**: polish — nav title with group avatar + member count, empty-state copy, chats-list last-message preview + timestamp + unread dot.

### DIFF
--- a/Sources/OnymIOS/Chats/ChatBubbleCell.swift
+++ b/Sources/OnymIOS/Chats/ChatBubbleCell.swift
@@ -7,13 +7,12 @@ import UIKit
 ///   - `.outgoing` — accent-filled bubble pinned to the trailing edge.
 ///   - `.incoming` — surface-tinted bubble pinned to the leading edge.
 ///
-/// Layout is a single `UILabel` inside a `UIView` bubble container.
-/// The bubble's leading/trailing edge is swapped via two stored
-/// constraints; the `lessThanOrEqualTo` width cap keeps long
-/// messages from filling the full row.
-///
-/// PR 6 scope: no avatars, no timestamps, no status glyphs, no
-/// emoji rendering. Status indicator lands in PR 8.
+/// Status indicator (PR 9): outgoing rows show a tiny glyph below
+/// the bubble's trailing edge that reflects `ChatMessage.status`
+/// (clock / check / red bang). Incoming rows hide the indicator —
+/// the message arriving is the proof. Tapping a `.failed` bubble
+/// invokes `onRetryRequested`, which the controller wires to
+/// `SendMessageInteractor.retry`.
 final class ChatBubbleCell: UITableViewCell {
     static let reuseID = "ChatBubbleCell"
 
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
APPROVE

- `SendMessageInteractor.swift:165-175` (retry path) — the early flip to `.pending` before `await fanOut(...)` is doubly load-bearing: it updates the UI and, via actor serialization, makes any concurrent `retry(messageID:)` call no-op on the `.failed`-guard. Nicely defensive against double-tap during the in-flight window.
- `SendMessageInteractor.swift:144-158` — retry uses `message.senderBlsPubkeyHex` (snapshot) as `myBlsHex` rather than re-deriving from `currentIdentity()`. Correct for dedup (same payload), but if the user has switched identities since the original send, the `group.memberProfiles[myBlsHex] != nil` guard silently no-ops and the user has no feedback. Probably fine for PR 9; worth a follow-up if identity-switching mid-thread becomes a real path.
- `SendMessageInteractor.swift:189-194` — non-tyranny groups return *before* the `.pending` flip, so a misrouted retry doesn't strand the row mid-state. Good ordering.
- `ChatThreadViewController.swift:115-122` — `changedIDs` via `prior != msg` requires `ChatMessage: Equatable`; that's pre-existing but worth confirming the synth covers every field that should trigger a repaint (it should — status, body, sentAt are all stored).
- `ChatBubbleCell.swift:96-108` — recognizer cleanup order is right: remove old, null out closure, then re-install only on `.failed && .outgoing && onRetry != nil`. The reconfigure-from-failed-to-sent test (`ChatBubbleCellTests.swift:139-157`) pins this.
- `ChatBubbleCell.swift:50` — `simulateBubbleTapForTest()` is `#if DEBUG` and used by `ChatThreadViewControllerTests.swift:230` which is also debug-only; fine, just keep them aligned if a release-config test target ever lands.
- `SendMessageInteractorTests.swift:228-236` — `test_retry_failedMessage_staysFailedIfRelaysStillReject` only asserts final status; could also assert `recordedSends.count == 2` to prove the retry actually fanned out (vs. silently bailing on a guard). Minor.
- Nit: `ChatBubbleCell.swift:111-114` — `.received` on an outgoing row is treated as defensive-hide; comment says "outgoing rows never carry .received" — that invariant lives in the model layer and isn't checked here, just trusted. OK.

Tests cover the right surface (status flips, tap installation/removal, reconfigure path, retry no-op cases). Ship it.

PR #156, opened 2026-05-19
feat(chats): chat-thread polish — nav subtitle + empty state (PR 10/11 chat stack)

lead-onym:
Two view-only improvements that make the chat thread feel complete.

## Stacked on
Base: \`pr-chat-9-status-indicator\` (#155).

## What's in here

**Nav title subtitle.** Replaces the single centered title with a vertical stack: name on top, "N members" below. Hidden when `memberCount <= 1` (singleton groups don't need the count; nothing-known groups would render "0 members" awkwardly).

**Empty state.** Centered "No messages yet. Say hi." label overlaid on the table view. Toggled on when `update(messages:)` lands an empty array and off as soon as the first row arrives. Lives behind the table (added as a sibling) so the keyboard layout guide affects it the same way as the message list.

## Wire-up
- `ChatThreadViewController.update(groupName:memberCount:)` replaces the old single-arg overload.
- The SwiftUI bridge computes `memberCount` from `chatsFlow.groups.first { $0.id == groupID }?.memberProfiles.count` so an admin admitting a new joiner updates the bar live as the announcement lands.
- The `emptyStateLabel` toggle lives inside the existing `update(messages:)` flow — no new hook surface.

## What's NOT in here — deferred to a focused follow-up

The plan's polish bucket also included:
- **Group avatar in the nav.** Would need a UIKit equivalent of `OnymGroupAvatar` or a `UIHostingController` bridge — both heavier than the rest of PR 10.
- **`ChatsView` last-message preview + timestamp + unread dot.** Needs a per-group "last message" query in `MessageRepository` *and* a `PersistedGroup.lastReadMessageID` schema bump. Worth a focused follow-up rather than absorbing here.

PR 10 stays small + view-only so PR 11 (E2E + CI smoke) can ship on the same chain.

## Tests (5 new, suite 622 / 622)

**`ChatThreadViewControllerTests`** +5:
- `memberCount` 0 / 1 → subtitle hidden
- `memberCount >= 2` → subtitle visible, pluralized as "N members"
- empty list → empty state visible
- non-empty list → empty state hidden
- list cleared again → empty state re-appears (symmetric toggle)

Updated two pre-existing tests for the `update(groupName:memberCount:)` signature.

## Plan context
PR 10 of 11. **PR 11 next**: E2E integration test + CI smoke — two-identity round-trip through real `IdentityRepository` + in-memory fake transport, plus a UI smoke that opens a chat and verifies a typed message round-trips.

### DIFF
--- a/Sources/OnymIOS/Chats/ChatThreadView.swift
+++ b/Sources/OnymIOS/Chats/ChatThreadView.swift
@@ -33,6 +33,7 @@ struct ChatThreadView: View {
     var body: some View {
         ChatThreadControllerBridge(
             groupName: currentGroupName,
+            memberCount: currentMemberCount,
             messages: messages,
             onBack: { dismiss() },
             onShowMembers: { showMembers = true },
@@ -90,10 +91,18 @@ struct ChatThreadView: View {
     private var currentGroupName: String {
         chatsFlow.groups.first { $0.id == groupID }?.name ?? "Chat"
     }
+
+    /// Drives the title subtitle ("N members"). Reads from the same
+    /// `chatsFlow.groups` source as the name so an admin admitting
+    /// a new joiner updates the bar live as the announcement lands.
+    private var currentMemberCount: Int {
+        chatsFlow.groups.first { $0.id == groupID }?.memberProfiles.count ?? 0
+    }
 }
 
 private struct ChatThreadControllerBridge: UIViewControllerRepresentable {
     let groupName: String
+    let memberCount: Int
     let messages: [ChatMessage]
     let onBack: () -> Void
     let onShowMembers: () -> Void
@@ -107,7 +116,7 @@ private struct ChatThreadControllerBridge: UIViewControllerRepresentable {
         vc.onSendTapped = onSendTapped
         vc.onRetryRequested = onRetryRequested
         vc.loadViewIfNeeded()
-        vc.update(groupName: groupName)
+        vc.update(groupName: groupName, memberCount: memberCount)
         vc.update(messages: messages)
         return vc
     }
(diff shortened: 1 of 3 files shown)


rinat-enikeev:
@lead-onym make a build

dev-onym (commented):
APPROVE — clean, well-tested view-only polish. Two minor nits, neither blocking.

- `ChatThreadViewController.swift:278` — `emptyStateLabel` is `addSubview`ed *after* `tableView`, so it sits **in front of** the table, not "behind" as the comment at `:265-268` (and the PR body) state. Functionally fine because it's hidden whenever the table has rows, but consider either fixing the comment or using `insertSubview(..., belowSubview: tableView)` to match intent.
- `ChatThreadView.swift:99` — `memberProfiles.count` collapses two cases into one: (a) the group truly has ≤1 known member and (b) the SwiftUI source hasn't materialized the group yet (`first { … } == nil` → 0). Both currently hide the subtitle, which is the safer outcome, so this is acceptable; worth confirming `memberProfiles` includes self so a true 2-person chat doesn't render as "1 member" (hidden). If it doesn't, the threshold is off-by-one for DMs — but that's a follow-up question, not a blocker for view-only polish.
- Pluralization "N members" is hard-coded; no localization hook. Consistent with surrounding strings in the chat surface, so fine for now.
- Tests cover the symmetric toggle (empty → non-empty → empty) and the 0/1/≥2 subtitle thresholds — exactly the edges I'd want. Signature-migration updates to the two pre-existing tests look correct.
- API break `update(groupName:)` → `update(groupName:memberCount:)` is contained to the bridge (`ChatThreadView.swift:119,133`); no other call sites in the diff.

dev-onym:
Done. 

- Dispatched `QA Build` workflow on `pr-chat-10-thread-polish` → run [26121956620](https://github.com/onymchat/onym-ios/actions/runs/26121956620). Builds an ad-hoc IPA via fastlane match + gym and publishes it to `onym.app/qa/` for OTA install.
- Recorded an empty commit `b9eff05` with the required message. Not pushed — framework will handle that.

PR #158, opened 2026-05-22
feat(groups): invite-offer handshake for create-time members

lead-onym:
## Problem

Create-time "add member via QR" shipped a `GroupInvitationPayload` (a *membership grant*) to a non-member's inbox. The invitee was never added to the on-chain roster, never recorded in the inviter's `memberProfiles`, and the epoch-pinned snapshot went stale the moment any other invitee anchored. Observed: **inviter saw 1 member, invitee saw 2, and messages never flowed back.**

## Approach

Replace the broken send with a **push handshake** that funnels into the existing on-chain join machinery (intro keys → `JoinRequestSender` → `IntroInboxPump` → `JoinRequestApprover` → `update_commitment`). Membership stays on-chain (preserving the admin's anonymous-ownership proof), and roster anchoring is gated behind the admin's **explicit Approve** — no auto-anchor, no silent add of an unintended party.

Flow: scan QR → admin sends a durable offer → invitee **explicitly accepts** (ships a join request) → admin **explicitly approves** → `update_commitment` anchors + ships the real invitation.

## Changes

- **`GroupInviteOfferPayload`** — durable, epoch-free offer carrying a freshly-minted per-invitee intro key + display context. Grants nothing, never expires.
- **`CreateGroupInteractor` (Tyranny)** — mint an intro key per invitee and send an offer instead of a snapshot; creator stays the only on-chain member until an approval.
- **`IncomingMessageDispatcher`** — route offers to `PendingInvitesStore`; never materialize a group from an offer.
- **`PendingInvitesStore` / `PendingInvitesFlow` / `PendingInvitesView`** — invitee-side Invitations surface (Chats toolbar) with explicit Accept → ships a `JoinRequestPayload` to the offer's intro key.
- **Bounded converge-forward verification** — the receiver byte-verifies the on-chain commitment at an *exact* epoch match (the strong gate: reproducing it requires the random `salt`, which is never on chain). When the chain is *ahead*, the snapshot can't be byte-verified, so it's accepted only within a small staleness window (`maxInvitationStaleEpochs = 16`) — enough to absorb a burst of serialized approvals during delivery latency, while bounding the salt-free forgery surface to young groups. (Initial revision relaxed this unbounded; see security note below.)
- **Serialize `JoinRequestApprover.approve`** — concurrent approvals no longer both read the same stale epoch (which lost one to a stale-epoch replay).

The existing admin approval UI (`ApproveRequestsView`) is reused unchanged.

## Security note

The chain-ahead branch trades some forgery resistance for late-join delivery: within the window, a party who knows a recipient's inbox key + a real groupID can still materialize a self-consistent fake of a *young* group (the forged envelope's signer becomes the local admin, so the real admin's announcements won't reconcile). The clean fix is to bind the snapshot's `adminPubkeyHex` to the on-chain `admin_pubkey_commitment` regardless of epoch — but `SEPCommitmentEntry` (the `get_commitment` read) doesn't expose it today, so that's a follow-up gated on relayer + SDK support. Announcement verification stays unbounded `>=` because it's already gated by the stored-admin Ed25519 signature check.

## Tests

`GroupInviteOfferPayload` codec + wire-disambiguation; dispatcher offer-route (no materialize, queued for accept); converge-forward materialization when the chain is ahead within the window, and rejection beyond it. Full `OnymIOSTests` suite green on iOS Simulator.

### DIFF
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -24,6 +24,11 @@ struct AppDependencies {
     /// and the underlying collector should run for the app's
     /// lifetime regardless of which surface is mounted.
     let approveRequestsFlow: ApproveRequestsFlow
+    /// Single shared instance — the invitee-side push-invitation
+    /// surface. Backs the Chats toolbar "Invitations" badge + modal,
+    /// and its store watcher runs for the app's lifetime like
+    /// `approveRequestsFlow`.
+    let pendingInvitesFlow: PendingInvitesFlow
     /// Single shared instance — the chat-thread screen subscribes to
     /// per-group message snapshots, and the receive-side dispatcher
     /// writes into the same actor. Constructed once in
--- a/Sources/OnymIOS/Chats/ChatsView.swift
+++ b/Sources/OnymIOS/Chats/ChatsView.swift
@@ -9,6 +9,7 @@ struct ChatsView: View {
     let flow: ChatsFlow
     let identitiesFlow: IdentitiesFlow
     let approveRequestsFlow: ApproveRequestsFlow
+    let pendingInvitesFlow: PendingInvitesFlow
     let messageRepository: MessageRepository
     let sendMessageInteractor: SendMessageInteractor
     let makeCreateGroupFlow: @MainActor () -> CreateGroupFlow
@@ -37,6 +38,11 @@ struct ChatsView: View {
             ToolbarItem(placement: .topBarTrailing) {
                 ApproveRequestsToolbarButton(flow: approveRequestsFlow)
             }
+            // Invitations received by this identity (push offers). Same
+            // always-rendered + badge-on-nonempty treatment.
+            ToolbarItem(placement: .topBarTrailing) {
+                PendingInvitesToolbarButton(flow: pendingInvitesFlow)
+            }
             // Plus button mirrors iOS Mail / Messages — useful once
             // the user already has at least one chat. Hidden in the
             // empty state because the central CTA already covers it.
@@ -54,6 +60,7 @@ struct ChatsView: View {
         .task { flow.start() }
         .task { await identitiesFlow.start() }
         .task { await approveRequestsFlow.start() }
+        .task { await pendingInvitesFlow.start() }
         .fullScreenCover(isPresented: $showCreateGroup) {
             CreateGroupViewHost(
                 makeFlow: makeCreateGroupFlow,
--- a/Sources/OnymIOS/Group/CreateGroupInteractor.swift
+++ b/Sources/OnymIOS/Group/CreateGroupInteractor.swift
@@ -38,6 +38,11 @@ struct CreateGroupInteractor: Sendable {
     let networkPreference: any NetworkPreferenceProviding
     let proofGenerator: any GroupProofGenerator
     let inboxTransport: any InboxTransport
+    /// Mints a per-invitee intro key for each create-time invite offer
+    /// (Tyranny). That intro key is the reply channel the invitee seals
+    /// their `JoinRequestPayload` to when they accept — the same
+    /// machinery the deeplink share-invite flow uses, just pushed.
+    let introducer: InviteIntroducer
     /// Builds a `SEPContractTransport` from the relayer URL chosen
     /// per-call. Injected so tests can swap in a fake without
     /// touching `URLSession`.
@@ -51,6 +56,7 @@ struct CreateGroupInteractor: Sendable {
         networkPreference: any NetworkPreferenceProviding = UserDefaultsNetworkPreference(),
         proofGenerator: any GroupProofGenerator = OnymGroupProofGenerator(),
         inboxTransport: any InboxTransport,
+        introducer: InviteIntroducer,
         makeContractTransport: @escaping @Sendable (URL) -> any SEPContractTransport = { url in
             URLSessionSEPContractTransport(
                 endpoint: url,
@@ -65,6 +71,7 @@ struct CreateGroupInteractor: Sendable {
         self.networkPreference = networkPreference
         self.proofGenerator = proofGenerator
         self.inboxTransport = inboxTransport
+        self.introducer = introducer
         self.makeContractTransport = makeContractTransport
     }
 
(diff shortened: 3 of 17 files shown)


dev-onym (commented):
COMMENT

- **Converge-forward gate weakens forgery protection** at `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:355-372`: when `onchain.epoch > invitation.epoch`, neither the snapshot's `commitment` nor its `adminPubkeyHex` is cross-checked against chain — only chain existence is. Combined with the internal Poseidon recompute (`Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:331-345`), an attacker who knows a recipient's inbox X25519 key and a real `groupID` can craft an internally-consistent epoch-0 invitation with arbitrary `members` / `adminPubkeyHex` and have it materialize as a fake group. The PR description claims announcements then "catch up," but the fake will fail Ed25519 admin-signature checks on later announcements (`adminEd25519PubkeyHex` is stamped from the forged envelope's signer). Worth either binding the snapshot's `adminPubkeyHex` to a chain lookup, or restricting the relaxation to a narrow window (e.g. `onchain.epoch <= invitation.epoch + N`).
- **Wire-disambiguation tests are incomplete** at `Tests/OnymIOSTests/GroupInviteOfferPayloadTests.swift:55-72`: only `JoinRequestPayload` is verified not to decode as an offer. The dispatcher (`Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:97-108`) puts offers ahead of `MemberAnnouncementPayload` and `GroupInvitationPayload` — add negative-decode tests for both to lock the ordering invariant.
- **`var pendingInvites` with default** at `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:64`: comment-justified by memberwise-init behavior, but allows accidental mutation post-construction. Explicit init overloads would preserve the test ergonomics without giving up immutability.
- **`sendOffers` is sequential** at `Sources/OnymIOS/Group/CreateGroupInteractor.swift:712-783`: a slow relay or mint call blocks all subsequent invitees and propagates the first failure as `invitationSendFailed(index:)`. Acceptable for small N at create-time, but worth a `TaskGroup` once group-size grows.
- **Dismiss is local-only** at `Sources/OnymIOS/Inbox/PendingInvitesFlow.swift:127-131`: the admin's freshly-minted intro key is never revoked or signaled — long-lived dangling capability. Minor, but flag in the design.
- **`JoinRequestApprover` serialization** at `Sources/OnymIOS/Group/JoinRequestApprover.swift:121-128, 211-219`: chaining via `approvalChain` with synchronous read+assign is correct and the comment explains the stale-epoch hazard precisely. Nice fix.
- **Offer payload + handshake plumbing** otherwise looks sound: shape validation in `GroupInviteOfferPayload.init`, idempotent store keyed by Nostr event id, `consumeForMaterializedGroups` cleanup, and identity-switch cascade all match the existing `IncomingInvitationsRepository` patterns.

dev-onym (commented):
COMMENT
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:65` — `var pendingInvites: any PendingInvitesRecording = PendingInvitesStore()` is a `var`-with-default hack to keep the synthesized memberwise init backward compatible. Works, but production breakage is silent: if `OnymIOSApp` ever forgets to pass the shared store, offers vanish into a per-dispatcher orphan store with no warning. Consider a small factory or an explicit init.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:352-385` — the chain-ahead branch trusts an internally-consistent snapshot with no commitment binding to chain (only the staleness-window gate). PR acknowledges this; the comment is clear and the bound is sized appropriately, but worth tracking the `admin_pubkey_commitment` follow-up as an issue so the gap isn't lost.
- `Sources/OnymIOS/Inbox/PendingInvitesStore.swift:42` — `consumeForMaterializedGroups` keys solely on `groupID`. If two identities on the same device receive offers for the same `groupID`, materializing under one consumes the other's pending invite too. Probably theoretical, but a one-line `ownerIdentityID` filter would close it.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:99-105` — fast-path-0 offer decode runs before the announcement / invitation paths on every inbound message. Disambiguation test (`test_joinRequestPayload_doesNotDecodeAsOffer`) only covers join requests; a snapshot test against `GroupInvitationPayload` and `MemberAnnouncementPayload` would be cheap insurance against future schema drift.
- `Sources/OnymIOS/Group/CreateGroupInteractor.swift:707-784` — `sendOffers` fails-fast on first invitee error after the group is already published on chain. Leaves the group anchored with a partial invitee set and no resend UI surfaced here. Pre-existing posture for `sendInvitations`, but worth a follow-up — re-invite via the new offer path is now cheap (no chain write).
- `Sources/OnymIOS/Group/JoinRequestApprover.swift:120-218` — global serialization across all groups for one admin is correct but coarse; under load a queue of approvals serializes through multi-second proves. Acceptable for now; comment captures the why well.
- `Sources/OnymIOS/Inbox/PendingInvitesFlow.swift:103-111` — `IntroCapability` is rebuilt at accept time instead of via `offer.introCapability()` even though that helper exists on the payload. Minor — collapse to the helper for one source of truth on shape validation.

ux-onym (commented):
**COMMENT** — Approach is sound and the security trade-offs are clearly documented; a few concerns worth addressing before/after merge.

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:378-384` — chain-ahead branch is a real soft spot as the PR notes: within 16 epochs, anyone with a victim's inbox key + a real groupID can materialize a fake young group where they become the local admin. Trade-off seems reasonable, but the follow-up to bind `adminPubkeyHex` to chain-side `admin_pubkey_commitment` deserves a tracking issue, not just a comment.
- `Sources/OnymIOS/Group/CreateGroupInteractor.swift:709-783` (`sendOffers`) — loop throws on the first failure after earlier iterations already minted intro keys (persisted in `IntroKeyStore`) and shipped offers. Leaves the system in a partially-completed invite state with no rollback/retry; later invitees never get the offer and the admin has no signal to re-invite just the failed ones. Consider per-invitee best-effort + aggregate report, or at least a "first N succeeded" failure mode.
- `Tests/OnymIOSTests/` — no coverage of the new `JoinRequestApprover.approve` serialization (`approvalChain`). This was a real bug ("loser's join silently fails"); an actor-level overlap test (two `approve` calls awaiting the same gated `update_commitment`) would prevent regression.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:411-439` — `verifyTyrannyAnnouncement` got the same converge-forward relaxation but no test covers the chain-ahead branch (only the invitation verifier has `acceptedWhenChainEpochAhead…` / `rejectsWhen…TooFarAhead`). Announcement path is signature-gated, but the change is non-trivial.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:66` — `var pendingInvites: any PendingInvitesRecording = PendingInvitesStore()` defaulted to a fresh store is a footgun: any future dispatcher construction that forgets to thread the shared store silently swallows offers (no compile error, no runtime warning). Prefer making it required and adding a `makeForTests(...)` helper.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:63-65` — the rationale "a `let` with a default is omitted from the memberwise init entirely" is incorrect for Swift structs (let-with-default *is* included with its default). Doc nit, but it justifies the unusual `var`.
- `Sources/OnymIOS/Group/GroupInviteOfferPayload.swift:55-77` — no upper bound on `inviterAlias` length; an attacker who knows an inbox key could ship MB-sized offers that get queued. Cap to e.g. 256 bytes during init.

PR #159, opened 2026-05-22
feat(inbox): verify-at-current group state (Option 2) — replaces staleness window

lead-onym:
Stacked on #158. Replaces the bounded staleness-window stopgap from that PR with the **verify-at-current-state** design (Option 2), which closes the young-group forgery window entirely and adds an explicit admin-offline notification.

## Problem with the window (from #158)
The bounded `maxInvitationStaleEpochs` window still materialized a snapshot the chain had advanced past, trusting only internal-consistency + chain existence. Within the window that let an outsider (who knows a recipient's inbox key + a real groupID) forge a self-consistent fake of a *young* group.

## This PR
A Tyranny snapshot materializes **only** when it byte-verifies against the chain at an **exact epoch** — the salt-gated commitment check that a forger can't reproduce (salt is random and never on chain). When the chain is ahead, the receiver refuses to trust the snapshot and instead **fetches the current state from the admin**.

- **Dispatcher** — `verifyTyrannyInvitation` → `verified | staleNeedsRefresh | reject`. Stale snapshots are deferred to the verifier, not materialized. `maxInvitationStaleEpochs` removed.
- **`GroupStateVerifier` (invitee)** — sends a `GroupStateRefreshRequest` to the admin, marks the group **verifying**, and on timeout / unknown admin inbox marks it **unreachable**. Resolves when the fresh snapshot materializes the group.
- **`GroupStateVerifier` (admin)** — replies with the current `GroupInvitationPayload`, **gated on current membership + Ed25519-signer match**, sealing to the member's *stored* inbox (a forged request can't redirect the salt — the reply carries it).
- **Hidden until verified** — an unverifiable group never enters the chats list; it shows in the Invitations sheet as "Verifying…" or "Couldn't verify — admin offline" with **Retry**.

## Trade-off
Verification now depends on the admin being reachable for the current salt — exactly the liveness the user asked to surface. A long-offline invitee on a fast-growing group falls back to Retry rather than trusting an unverifiable snapshot.

## Tests
Refresh-request codec + wire disambiguation; dispatcher defers stale snapshots and routes refresh requests to the verifier; verifier marks `.unreachable` on unknown admin inbox and refuses a non-member (salt non-disclosure). Full `OnymIOSTests` suite green on iOS Simulator.

### DIFF
--- a/Sources/OnymIOS/Group/GroupStateRefreshRequest.swift
+++ b/Sources/OnymIOS/Group/GroupStateRefreshRequest.swift
@@ -1,0 +1,96 @@
+import Foundation
+
+/// Member → admin "send me the current group state" request, sealed to
+/// the admin's inbox and signed by the requester's Ed25519.
+///
+/// ## Why this exists
+///
+/// A Tyranny invitation is a snapshot pinned to one epoch. The receiver
+/// verifies it by recomputing `Poseidon(Poseidon(merkleRoot(members),
+/// epoch), salt)` and matching the on-chain commitment — but the chain
+/// only stores the *latest* `(commitment, epoch)`. If another invitee
+/// was anchored between the snapshot being sealed and it landing, the
+/// snapshot is from a past epoch the chain no longer holds, so it can't
+/// be verified (and we refuse to materialize an unverifiable group —
+/// see `GroupStateVerifier`).
+///
+/// To recover, the receiver asks the admin for the *current* state. The
+/// admin replies with a fresh `GroupInvitationPayload` at the current
+/// `(epoch, salt, members, commitment)`, which the receiver verifies at
+/// an exact epoch match. This is the "verify at current state" leg of
+/// the converge-forward design.
+///
+/// ## Privacy
+///
+/// The reply carries `salt` (the on-chain roster-privacy blinding
+/// factor), so the admin MUST only answer requesters that are in the
+/// *current* roster — `GroupStateVerifier.handleRefreshRequest` gates on
+/// membership + an Ed25519-signer check before replying. The request
+/// itself reveals only that the requester wants to resync a group it was
+/// invited to.
+///
+/// ## Wire disambiguation
+///
+/// `refresh_group_id` + `requester_inbox_pub` are required and unique to
+/// this type; no other inbox payload decodes as one.
+struct GroupStateRefreshRequest: Codable, Equatable, Sendable {
+    let version: Int
+    /// 32-byte on-chain `group_id` the requester wants the current
+    /// state for.
+    let groupID: Data
+    /// Requester's 32-byte X25519 inbox key — where the admin seals the
+    /// fresh snapshot reply.
+    let requesterInboxPublicKey: Data
+    /// Requester's 48-byte compressed BLS pubkey — used by the admin to
+    /// confirm the requester is in the current roster before disclosing
+    /// the salt.
+    let requesterBlsPublicKey: Data
+
+    enum CodingKeys: String, CodingKey {
+        case version = "refresh_version"
+        case groupID = "refresh_group_id"
+        case requesterInboxPublicKey = "requester_inbox_pub"
+        case requesterBlsPublicKey = "requester_bls_pub"
+    }
+
+    init(
+        version: Int = 1,
+        groupID: Data,
+        requesterInboxPublicKey: Data,
+        requesterBlsPublicKey: Data
+    ) throws {
+        guard groupID.count == 32 else {
+            throw GroupStateRefreshRequestError.shape(
+                "groupID: expected 32 bytes, got \(groupID.count)"
+            )
+        }
+        guard requesterInboxPublicKey.count == 32 else {
+            throw GroupStateRefreshRequestError.shape(
+                "requesterInboxPublicKey: expected 32 bytes, got \(requesterInboxPublicKey.count)"
+            )
+        }
+        guard requesterBlsPublicKey.count == 48 else {
+            throw GroupStateRefreshRequestError.shape(
+                "requesterBlsPublicKey: expected 48 bytes, got \(requesterBlsPublicKey.count)"
+            )
+        }
+        self.version = version
+        self.groupID = groupID
+        self.requesterInboxPublicKey = requesterInboxPublicKey
+        self.requesterBlsPublicKey = requesterBlsPublicKey
+    }
+
+    init(from decoder: Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        try self.init(
+            version: c.decode(Int.self, forKey: .version),
+            groupID: c.decode(Data.self, forKey: .groupID),
+            requesterInboxPublicKey: c.decode(Data.self, forKey: .requesterInboxPublicKey),
+            requesterBlsPublicKey: c.decode(Data.self, forKey: .requesterBlsPublicKey)
+        )
+    }
+}
+
+enum GroupStateRefreshRequestError: Error, Equatable {
+    case shape(String)
+}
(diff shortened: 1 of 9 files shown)


dev-onym (commented):
Based on the diff and source verification, the design is sound — the salt-gated commitment check is the right anti-forgery primitive, the admin-side gates (current-member + Ed25519 signer + inbox-key pinned to profile) are layered correctly, and the UI surfaces the new liveness dependency honestly.

COMMENT

- **Admin-side missing "am I the admin?" check** at `GroupStateVerifier.swift:191-206`: `handleRefreshRequest` gates on the requester being a current member but never verifies that *this* device is the group's admin. Today the request only reaches admins (sealed to `adminInbox`), but if another member's inbox were ever cached/reused as `adminInbox` in some snapshot, they would happily reply with the salt. A `group.adminPubkeyHex == myBls` guard is cheap defense-in-depth and matches the threat model the comment claims.
- **Deferral is sticky across re-deliveries** at `GroupStateVerifier.swift:74` (`guard await !store.contains else { return }`): once a group is marked `.unreachable`, the in-session re-delivery of the same stale snapshot is dropped silently — only manual Retry recovers. If that's intentional (avoid send-storms), worth a one-liner; if not, consider re-arming on `.unreachable`.
- **Retry has no backoff / cap** at `PendingInvitesFlow.swift:118-121` + `GroupStateVerifier.retry`: rapid taps fire one sealed send per tap. Low blast radius (one envelope each), but a tight loop on `.unreachable` could needlessly burn relayer budget. A simple in-flight guard would suffice.
- **`markUnreachableIfStillVerifying` doesn't actually check `.verifying`** at `GroupStateVerifier.swift:169-172` — name vs. behavior mismatch. Race window is narrow (resolve removes the entry first), but renaming or actually checking status would prevent a stale timeout from flipping a just-resolved-then-re-recorded entry.
- **Wire-disambiguation tests are thin** in `GroupStateRefreshTests.swift`: only verifies `GroupInviteOfferPayload` doesn't decode as a refresh. Since fast-path 0.5 runs *before* `MemberAnnouncementPayload` / `GroupInvitationPayload` / `JoinRequestPayload` / `ChatMessagePayload`, add negative-decode tests for those too — cheap insurance against future schema drift.
- **No happy-path / timeout coverage**: the deferral and admin-membership-reject cases are tested, but there's no test that a fresh snapshot reply actually resolves the pending entry (`resolve` + `resolveMaterialized`), nor one that the 30s timer transitions `.verifying → .unreachable`. The state machine is the load-bearing change here; worth a test that exercises the full loop with an injected clock.
- **Hardcoded 30s timeout** at `GroupStateVerifier.swift:39`: defaulted-parameter so easy to tune, but on slow relays / cold inbox subscriptions, 30s is aggressive. Consider 60-90s for V1 and surface the time-to-unreachable in a metric so it can be tuned from telemetry.
- **DoS / UI-clutter surface**: a forger that knows recipient inbox key + a real on-chain groupID can now spawn `.verifying`-then-`.unreachable` cards at will (each costs the receiver one sealed send to the forger's claimed admin inbox). Out of scope for this PR but worth a follow-up tracking issue — rate-limiting deferrals per (ownerIdentity, groupID) or per peer would close it.

lead-onym:
Thanks — addressed in 5cc52fd. Point by point:

1. **Admin "am I the admin?" check** ✅ — `handleRefreshRequest` now guards `currentIdentity().blsPublicKey == group.adminPubkeyHex` before the membership gate. A member whose inbox was ever reused as an admin inbox can no longer disclose the salt.
2. **Sticky deferral across re-deliveries** ✅ — changed to **re-arm**: `deferVerification` retries a `.unreachable` entry on re-delivery of the retained snapshot (auto-recovery once the admin is back); a `.verifying` entry is left alone.
3. **Retry backoff/cap** ✅ — added an in-flight guard; `retry` also no-ops unless the entry is `.unreachable` (a `.verifying` one already has a request outstanding), so taps/loops can't fan out sends.
4. **`markUnreachableIfStillVerifying` name vs behavior** ✅ — moved to `PendingVerificationStore.markUnreachableIfVerifying`, atomic on the store actor and actually checks `.verifying`, so a late timer can't clobber a resolved/re-armed entry.
5. **Thin disambiguation tests** ✅ — added negative-decode for `MemberAnnouncementPayload` / `GroupInvitationPayload` / `JoinRequestPayload` against both the offer and the refresh request.
6. **Happy-path / timeout coverage** ✅ — added: `resolve` clears the pending entry when the group materializes; and the timer transitions `.verifying → .unreachable` (via the injectable `refreshTimeoutSeconds`). Plus admin-not-self and non-member reject tests.
7. **Hardcoded 30s** — bumped default to **60s**. Pushing back on the telemetry-tuned value: there's no metrics seam in the app yet, so I left it as a documented injectable rather than invent one. Follow-up when telemetry lands.
8. **DoS / UI-clutter** — agreed it's a follow-up. Worth noting the bound is tighter than it reads: deferrals are **idempotent per `groupIDHex`**, so a forger gets **one** card + **one** sealed send *per distinct real on-chain groupID they know* — not an unbounded stream. Per-peer rate-limiting of deferrals is the right follow-up; left a code comment marking it.

Full suite green (35 tests in the affected classes, 0 failures).

dev-onym (commented):
APPROVE

- `Sources/OnymIOS/Inbox/GroupStateVerifier.swift:35-43,182-200` — admin-side gates (current-member + signer-Ed25519 + stored-inbox sealing target) are correctly layered; combined with `MemberProfile`'s 32/48-byte decode validation, no empty-Data spoof via `requesterEd25519 == profile.sendingPubkey` is possible. Salt-disclosure is well-bounded.
- `Sources/OnymIOS/Inbox/GroupStateVerifier.swift:62-67` — `targets`/`timeouts` are never cleared on identity removal. `PendingVerificationStore.removeForOwner` is wired in `OnymIOSApp.swift:351`, but no matching cleanup hook on the verifier; per-identity targets accumulate until app restart. Minor leak; consider adding a `removeForOwner` on the verifier and calling it alongside the store.
- `Sources/OnymIOS/Inbox/GroupStateVerifier.swift:88-101` — `groupName` on `PendingGroupVerification` is taken from the untrusted invitation. A forger who knows recipient inbox + real `groupID` can spawn one "Verifying…/Couldn't verify" card with an attacker-chosen name (`PendingInvitesView.swift:101`). Idempotency caps it at one card per `groupID`, but the surface area is attacker-controlled text. Worth flagging in UI copy or sanitizing — non-blocking.
- `Sources/OnymIOS/Group/GroupStateRefreshRequest.swift:33` — version field is decoded but not enforced (`version != 1` still accepted). Cheap to add a guard; future-proofs against a v2 receiver silently accepting v1-only fields or vice versa.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:117-131` — fast-path-0.5 ordering is safe: `refresh_group_id`/`requester_inbox_pub` are disjoint from offer/announcement/invitation/join required keys; test `test_otherPayloads_doNotDecodeAsOfferOrRefresh` locks this in.
- `Sources/OnymIOS/Inbox/GroupStateVerifier.swift:138-156` — `inFlight` set is inserted before any `await`, so actor reentrancy across awaits is correctly guarded; `scheduleTimeout` cancels the prior pending timer; `markUnreachableIfVerifying` is atomic on the store. Race analysis checks out.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:395-419` — verify-at-current logic correctly funnels chain-ahead to `staleNeedsRefresh`; tests cover both "no admin inbox → unreachable" and "non-member → no salt disclosure". `maxInvitationStaleEpochs` window fully retired — closes the young-group forgery surface from #158.
- `Sources/OnymIOS/Inbox/PendingInvitesFlow.swift:51-67` — constructor breaking-change is contained to the single call site in `OnymIOSApp.swift:186`; no test/preview constructors to update.

Net: solid, security-positive replacement of the staleness window. The leak + version + attacker-chosen-name notes are follow-ups, not blockers.

PR #162, opened 2026-05-23
feat(chats): differentiate message senders without avatars

lead-onym:
## Why

Group chats showed **no per-message sender identity** — only left/right alignment and a fixed blue/grey color. With no avatars, consecutive incoming messages from different people were indistinguishable.

## What

Sender differentiation built on the **load-bearing identity (BLS pubkey)** rather than the self-asserted, spoofable `MemberProfile.alias`:

- **`OnymAccent.forSender(blsPubkeyHex:)`** — deterministic accent hashed (FNV-1a) from the BLS pubkey. Stable per-person across devices, groups, and launches. Keys on the pubkey, never the alias, so an alias-impostor can't steal the original's color — the color doubles as a cheap **visual fingerprint**.
- **Incoming runs** get an accent-colored **name header** at the start of each consecutive same-sender run; mid-run messages stay bare. The incoming bubble carries a low-opacity (20%) tint of the sender's accent; body text stays on the regular text token for readability. Name = alias, falling back to `BLS xxxxxxxx` when unset (mirrors `ChatMembersView`).
- **Outgoing bubbles** now fill with the user's **own** hashed accent instead of hardcoded blue, so "your color" is consistent app-wide.
- **1-on-1 groups** suppress the header (only one other person — naming them every run is noise).

## Design

`ChatThreadViewController` owns run-grouping + alias resolution and feeds each cell a `ChatSenderDisplay` (name + accent + showNameHeader); the cell stays a dumb renderer. Member profiles flow in from the SwiftUI wrapper (`update(memberProfiles:)`) and repaint headers live as joiners land or aliases change.

## Files

- `Group/OnymBrand.swift` — `OnymAccent.forSender(blsPubkeyHex:)`
- `Chats/ChatBubbleCell.swift` — `ChatSenderDisplay`, name-header label + top-constraint toggle, per-sender coloring
- `Chats/ChatThreadViewController.swift` — member-profile state, run-grouping, profile-change repaint
- `Chats/ChatThreadView.swift` — pipes `memberProfiles` from the group into the controller

## Tests

49 chat-related tests green (build clean). New coverage: accent determinism + alias-independence; cell header show/hide + reuse; controller run-grouping (run start, mid-run, sender change, outgoing, 1-on-1, fingerprint fallback, live profile refresh).

### DIFF
--- a/Sources/OnymIOS/Chats/ChatBubbleCell.swift
+++ b/Sources/OnymIOS/Chats/ChatBubbleCell.swift
@@ -1,11 +1,47 @@
 import SwiftUI
 import UIKit
 
+/// Resolved sender presentation for one bubble — computed by
+/// `ChatThreadViewController` (which alone knows the group's member
+/// profiles + run grouping) and handed to the cell. The cell stays a
+/// dumb renderer: it doesn't look up aliases or hash pubkeys itself.
+///
+/// `accent` is derived from the sender's BLS pubkey via
+/// `OnymAccent.forSender(blsPubkeyHex:)`, so it's a stable per-person
+/// color rather than anything tied to the (spoofable) alias.
+struct ChatSenderDisplay {
+    /// Alias if the member set one, else a short BLS-fingerprint
+    /// fallback. Only rendered when `showNameHeader` is true.
+    let name: String
+    /// Per-sender color. Tints the incoming bubble; fills the outgoing
+    /// bubble; colors the name header.
+    let accent: OnymAccent
+    /// Show the name header above this bubble. The controller sets this
+    /// only at the start of a run of consecutive same-sender incoming
+    /// messages, and never in 1-on-1 groups (one other person — naming
+    /// them on every run is noise).
+    let showNameHeader: Bool
+
+    /// Neutral default used by the cell when no sender info is supplied
+    /// (and by older call sites). Blue, no header — matches the
+    /// pre-sender-differentiation look.
+    static let unknown = ChatSenderDisplay(name: "", accent: .blue, showNameHeader: false)
+}
+
 /// One message bubble row. Two visual styles toggled by
 /// `ChatMessage.direction`:
 ///
-///   - `.outgoing` — accent-filled bubble pinned to the trailing edge.
-///   - `.incoming` — surface-tinted bubble pinned to the leading edge.
+///   - `.outgoing` — filled with the sender's accent, pinned to the
+///     trailing edge.
+///   - `.incoming` — tinted with the sender's accent, pinned to the
+///     leading edge, optionally topped by a colored name header.
+///
+/// Sender differentiation: there are no avatars, so consecutive
+/// incoming messages from one person are grouped under a single
+/// accent-colored name header (`ChatSenderDisplay.showNameHeader`), and
+/// the bubble itself carries that sender's accent tint. The accent is a
+/// hash of the BLS pubkey, not the alias, so it doubles as a cheap
+/// visual fingerprint that an alias-spoofer can't forge.
 ///
 /// Status indicator (PR 9): outgoing rows show a tiny glyph below
 /// the bubble's trailing edge that reflects `ChatMessage.status`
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
APPROVE — clean implementation with strong test coverage; a few minor notes below.

- `Sources/OnymIOS/Group/OnymBrand.swift:101` — FNV-1a is keyed off `blsPubkeyHex.utf8`, so case differences produce different colors. `ChatMessagePayload.swift:36` documents the field as lowercase, so this works *if* all senders honor that convention. Consider hashing `blsPubkeyHex.lowercased().utf8` (cheap defense) or asserting the precondition, so a stray uppercased value can't fracture identity-color stability.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:267-289` — `rebuildSenderDisplays()` is O(messages) on every snapshot apply *and* every `update(memberProfiles:)` invocation, and the wrapper calls the latter on every SwiftUI render. Probably fine at chat-thread scale, but worth noting that an alias edit triggers a full rebuild + `reconfigureItems(all)` rather than just the affected sender's rows.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:60-65` — `incomingTintAlpha = 0.20` is uniform across all 5 palette accents over `OnymTokens.bg`; light-mode contrast for warmer accents (e.g. yellow/orange if in palette) plus `OnymTokens.text` may be borderline. Worth a visual check in both light & dark.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:309-312` — `nameLabel`'s leading/trailing constraints stay active when hidden (label pinned to `bubble.leadingAnchor`). Harmless because `bubble`'s position is direction-driven and a hidden empty-text label has zero intrinsic width, but the always-on leading anchor to `bubble` would render mis-aligned on the (currently impossible) outgoing-with-header path. Defensive note only.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:197-213` — docstring says "Must be called before `update(messages:)` on first render", but the controller is actually order-independent now (both paths rebuild displays). The bridge does call in the correct order; consider softening the "must" so a future caller isn't misled. `test_profileUpdate_refreshesRenderedHeaderName` already exercises the reverse order successfully.
- Tests look thorough: determinism + alias-independence at the palette level, run-grouping/1:1/outgoing/fingerprint-fallback at the controller level, header show/hide/reuse at the cell level. Nice coverage.

ux-onym (commented):
APPROVE — clean implementation, well-documented, solid test coverage. A few non-blocking observations.

- `Sources/OnymIOS/Group/OnymBrand.swift:96` — hash is over raw UTF-8 hex bytes, so two pubkey strings differing only in case (`"AB..."` vs `"ab..."`) hash to different accents. Codebase appears consistently lowercase via `%02x`, but normalizing to lowercase inside `forSender` would make the contract bulletproof. Same case-sensitivity assumption applies to the `memberProfiles[blsPubkeyHex]` lookup at `Sources/OnymIOS/Chats/ChatThreadViewController.swift:288`.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:265` — run-grouping considers only the previous sender, never elapsed time. Two messages from Alice 6 hours apart are still one run, so the second has no header. Most chat apps re-show a header after a gap; worth a follow-up if that feels off in usage.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:309` — `nameLabel.leadingAnchor` pins to `bubble.leadingAnchor`, which for outgoing rows is the right side. Today the controller suppresses headers on outgoing (`ChatThreadViewController.swift:269-272`), so the path is dead, but it's a latent layout bug if that invariant ever loosens — a quick assert in `applyNameHeader` would lock it in.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:259-268` — when hidden, `nameLabel` has no active vertical anchor (only horizontal). UILabel still participates in layout while hidden; depending on Auto Layout's intrinsic-content-size resolution this may emit ambiguity warnings. Consider giving it a zero-height constraint or a low-priority top anchor when not in use.
- 6-color palette → ~17% pairwise collision odds; in a 5-person group, a clash is more likely than not. PR doesn't claim uniqueness, just "cheap fingerprint," so this is acceptable, but worth surfacing for designers.
- PR description says the fallback "mirrors `ChatMembersView`," but `ChatMembersView` uses `"(unnamed)"` for empty aliases + a separate 12-char prefix; the chat header uses `"BLS " + 8-char prefix`. Description nit, not a code issue.

PR #164, opened 2026-05-23
feat(groups): pick or AI-generate a group avatar (local) [1/3]

lead-onym:
First of a 3-PR stack adding group avatars (pick from gallery or generate on-device with AI), small enough to ship over NOSTR.

## This PR — local only
- **`GroupAvatarImage`**: one funnel for every source — centre-crop → 256×256 px → JPEG under a **16 KB** budget (base64 ≈ 22 KB, NOSTR-safe).
- Wired the previously-decorative create-flow camera button to a real picker: **gallery** (`PhotosPicker`) or **on-device Image Playground** (iOS 18.1+, capability-gated; option hidden on unsupported devices). No backend, no keys — prompt never leaves the device.
- `ChatGroup.avatarJPEG` + AES-GCM `encryptedAvatar` column on `PersistedGroup` (optional → lightweight migration, no wipe).
- `OnymGroupAvatar` renders the photo clipped to a circle when set, else the existing brand mark; chats list + create success hero use it.

Propagation over the wire is in the next two PRs.

## Tests
`GroupAvatarImageTests` (square-crop + byte budget incl. high-entropy worst case), `SwiftDataGroupStoreTests` (avatar round-trip / clear-on-update).

### DIFF
--- a/Sources/OnymIOS/Chats/ChatsView.swift
+++ b/Sources/OnymIOS/Chats/ChatsView.swift
@@ -152,16 +152,15 @@ private struct ChatsRow: View {
 
     var body: some View {
         HStack(spacing: 12) {
-            // Reuse the broken-ring brand mark as the per-chat avatar
-            // — same identity the user saw on the Create Group hero.
-            // Once group avatars / uploads ship this becomes the
-            // image-or-mark fallback.
+            // Group photo when set, else the broken-ring brand mark —
+            // same identity the user saw on the Create Group hero.
             OnymGroupAvatar(
                 size: 44,
                 accent: OnymAccent.blue.color,
                 ringPulse: false,
                 spinning: false,
-                brand: false
+                brand: false,
+                imageData: group.avatarJPEG
             )
 
             VStack(alignment: .leading, spacing: 2) {
--- a/Sources/OnymIOS/Group/ChatGroup.swift
+++ b/Sources/OnymIOS/Group/ChatGroup.swift
@@ -67,6 +67,12 @@ struct ChatGroup: Identifiable, Equatable, Sendable {
     /// `accepted = true`. Persisted-but-not-anchored groups can be
     /// retried.
     var isPublishedOnChain: Bool
+    /// Square JPEG (256×256, ≤16 KB — see `GroupAvatarImage`) the user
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
**APPROVE** — Clean, well-bounded PR with good tests. A few performance / nit items worth addressing before the wire-format PRs land.

- `Sources/OnymIOS/Group/GroupAvatarImage.swift:32-34` — Loop condition `quality > 0.2` stops *before* trying 0.2 (sequence: 0.7, 0.6, 0.5, 0.4, 0.3, then exit). The comment says "floor (0.2)" but actual floor is 0.3. Use `>= 0.2` or drop the comment.
- `Sources/OnymIOS/Group/GroupAvatarImage.swift:25-35` — If even q=0.3 overshoots, returns oversize JPEG silently. The doc claim "bytes…are always small enough to ship" is then false. Either hard-fail (return `nil`) or shrink `dimension` further as a last step; otherwise the next PR's wire layer must enforce.
- `Sources/OnymIOS/Group/GroupAvatarPickerButton.swift:80-86` — `setFromURL` runs `Data(contentsOf:)` + JPEG re-encode synchronously on the main thread when ImagePlayground returns. Wrap in `Task.detached` like `loadFromGallery` does.
- `Sources/OnymIOS/Group/OnymBrand.swift:378-384` — `UIImage(data: imageData)` re-decodes the JPEG on every SwiftUI body eval, including every `ChatsRow`. For a list this can churn. Cache via `@State` + `onChange(of: imageData)` (or a small `NSCache` keyed on bytes hash).
- `Sources/OnymIOS/Group/GroupAvatarPickerButton.swift:63-65` — After processing, reset `photosItem = nil`; `PhotosPicker` doesn't always re-emit when the user picks the same asset twice.
- `Sources/OnymIOS/Group/PersistedGroup.swift:52-55` — Defaulted optional + defaulted initializer arg is the right call for SwiftData lightweight migration; existing call sites stay green.
- Tests (`GroupAvatarImageTests`, `SwiftDataGroupStoreTests`) pin the right invariants — square output, byte cap under high-entropy worst case, garbage-input nil, avatar round-trip + clear-on-update.

PR #165, opened 2026-05-23
feat(groups): carry avatar in the group invitation snapshot [2/3]

lead-onym:
Second of the group-avatar stack. **Base: #164.**

Threads the create-time avatar from the flow through `CreateGroupInteractor` onto `ChatGroup` and into `GroupInvitationPayload` as an optional, `decodeIfPresent` `avatar` field — absent on the wire when nil, so photo-less invites stay small. The receiver's `materializeGroup` adopts it.

Covers every snapshot path:
- **1-on-1 / anarchy** — send the invitation directly at create time.
- **Tyranny** — invitees get the photo when the admin ships the full snapshot at join-approval (`JoinRequestApprover`).
- **State refresh** — re-sends it (`GroupStateVerifier`).

## Wire shape
New optional key `avatar` (base64 `Data`) on `GroupInvitationPayload`. Back-compat verified: pre-avatar senders (no key) decode to `nil`.

## Tests
`GroupInvitationPayloadTests`: round-trip with/without avatar, legacy-without-avatar decode, and "absent from wire when nil".

### DIFF
--- a/Sources/OnymIOS/Chats/ChatsView.swift
+++ b/Sources/OnymIOS/Chats/ChatsView.swift
@@ -152,16 +152,15 @@ private struct ChatsRow: View {
 
     var body: some View {
         HStack(spacing: 12) {
-            // Reuse the broken-ring brand mark as the per-chat avatar
-            // — same identity the user saw on the Create Group hero.
-            // Once group avatars / uploads ship this becomes the
-            // image-or-mark fallback.
+            // Group photo when set, else the broken-ring brand mark —
+            // same identity the user saw on the Create Group hero.
             OnymGroupAvatar(
                 size: 44,
                 accent: OnymAccent.blue.color,
                 ringPulse: false,
                 spinning: false,
-                brand: false
+                brand: false,
+                imageData: group.avatarJPEG
             )
 
             VStack(alignment: .leading, spacing: 2) {
--- a/Sources/OnymIOS/Group/ChatGroup.swift
+++ b/Sources/OnymIOS/Group/ChatGroup.swift
@@ -67,6 +67,12 @@ struct ChatGroup: Identifiable, Equatable, Sendable {
     /// `accepted = true`. Persisted-but-not-anchored groups can be
     /// retried.
     var isPublishedOnChain: Bool
+    /// Square JPEG (256×256, ≤16 KB — see `GroupAvatarImage`) the user
+    /// or admin set as the group photo. `nil` falls back to the Onym
+    /// brand mark in `OnymGroupAvatar`. Travels in the invitation
+    /// snapshot at create time and via `GroupAvatarPayload` afterwards.
+    /// Defaulted so existing construction sites need no change.
+    var avatarJPEG: Data? = nil
 
     /// Group ID as the raw 32-byte payload (parsed back from `id`).
     /// Used directly when building chain payloads + invitations.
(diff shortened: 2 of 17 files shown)


ux-onym (commented):
APPROVE — clean back-compat, tests cover the wire shape well, and `GroupRepository.insert` is actually `insertOrUpdate` so a `GroupStateVerifier` re-send will land the avatar on already-materialized peers too.

- `IncomingMessageDispatcher.swift:320` — receiver accepts `invitation.avatar` of any size; the 16 KB cap in `GroupAvatarImage.maxBytes` is sender-side only. A misbehaving/older sender could ship a larger blob and it'd be persisted as-is. Consider a `guard data.count <= GroupAvatarImage.maxBytes` (or a generous multiple to allow encoder drift) on the wire side.
- `JoinRequestApprover.swift:298` — Tyranny invitees only receive the photo via `JoinRequestApprover`, but `JoinRequestApprover` snapshots `anchored.avatarJPEG` at approval time. If the admin sets an avatar *after* a join request is in flight but before approval, the joiner will see the new one — correct, just worth noting since the PR description says "the admin ships the full snapshot at join-approval" (matches behavior).
- `GroupInvitationPayload.swift:117-136` — the custom `init(from:)` re-declares `decodeIfPresent` for `avatar`, which the synthesized decoder would also do for an `Optional`. Not a bug (since the rest of the fields use the custom path), but a future refactor could drop the whole custom initializer and rely on synthesis since every optional field uses `decodeIfPresent`.
- `GroupInvitationPayloadTests.swift:60-95` — tests are tight and target the right invariants (round-trip, legacy-absent, encodeIfPresent behavior). Nice to also add a parity test confirming `avatar` is the exact wire key (Android contract), matching the existing `member_profiles` parity test at line 58-71.

dev-onym (commented):
APPROVE

- `Sources/OnymIOS/Group/GroupInvitationPayload.swift:135` — custom `init(from:)` correctly threads `decodeIfPresent` for `avatar`; existing fields all still decoded; round-trip equality is preserved since `Equatable` is synthesized over all stored properties including the new one.
- `Sources/OnymIOS/Group/GroupInvitationPayload.swift:82` — `avatar` (no snake_case alias) is consistent with other single-word keys (`version`, `name`, `salt`, `commitment`) — Android parity is fine.
- `Sources/OnymIOS/Group/CreateGroupInteractor.swift:96` — `avatar: Data? = nil` default keeps the public API backward-compatible for any non-flow callers; all three governance branches thread it.
- `Sources/OnymIOS/Group/CreateGroupInteractor.swift:466,653` — anarchy/1-on-1 ship `avatar:` directly at create time; tyranny path correctly omits it (no invite sent here), with `JoinRequestApprover.swift:298` covering the deferred snapshot at approval. State refresh at `Inbox/GroupStateVerifier.swift:256` re-sends — all three claimed paths covered.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:320` — receiver materializes `avatarJPEG` from `invitation.avatar`. No receive-side size guard: `GroupAvatarImage.maxBytes = 16 KB` is enforced only by the local encoder, and even there the 0.2-quality floor at `GroupAvatarImage.swift:30` "returns it anyway rather than dropping the avatar," so an over-budget JPEG can ship and a malicious peer can send arbitrarily large bytes. Minor defense-in-depth: cap `invitation.avatar?.count` at receive (e.g., reject or drop above ~64 KB) before persisting. Non-blocking — relay event-size limits provide an implicit ceiling.
- `Tests/OnymIOSTests/GroupInvitationPayloadTests.swift:56-95` — covers round-trip with/without, legacy-no-key decode, and `encodeIfPresent` omission. Pre-existing `makePayload` callers at L22/30/102 unaffected (new param defaults to nil). Solid coverage for the wire-shape claims in the PR body.

PR #166, opened 2026-05-23
feat(groups): broadcast avatar changes via a dedicated message [3/3]

lead-onym:
Last of the group-avatar stack. **Base: #165.**

Adds the dedicated "avatar" message so the photo reaches late-joiners and can be changed after creation — the cases the invitation snapshot can't cover.

## What's here
- **`GroupAvatarPayload`** — admin-signed control message carrying the JPEG (or `nil` to clear). Distinct `avatar_*` wire keys so it can't structurally decode as a `ChatMessagePayload` (which shares version/group_id/sender/timestamp).
- **`GroupAvatarBroadcaster`** — applies locally, then seals + fans out one envelope per member inbox, mirroring `JoinRequestApprover`'s member-announcement loop.
- **Dispatcher fast-path** — decodes it, verifies the envelope's Ed25519 signer against the group's stored admin key (same trust gate as member announcements; no on-chain check since the avatar isn't part of the commitment), then applies.
- **Change-photo entry point** — admin-only picker in `ChatMembersView`, wired through `AppDependencies.setGroupAvatar`; non-admins see a read-only hero.
- **i18n** — en/ru for the new picker strings, plus fills the empty catalog keys `LocalizationCatalogTests` was already flagging (unrelated in-progress feature copy that was sitting untranslated in the catalog).

## Tests
`GroupAvatarPayloadTests` (round-trip, nil=removal, anti-collision both directions vs `ChatMessagePayload`); `IncomingMessageDispatcherTests` (apply on admin match, reject on mismatch, clear on nil). Full suite: 672 pass.

## Cross-platform follow-up
Android + relayer need to learn the `avatar` field on `GroupInvitationPayload` (#165) and the `GroupAvatarPayload` kind here to interop. iOS decodes both defensively, so nothing breaks until they catch up.

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -1125,6 +1125,22 @@
         }
       }
     },
+    "Couldn’t verify — the admin is offline. The group stays hidden until it can be verified on chain." : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Couldn’t verify — the admin is offline. The group stays hidden until it can be verified on chain."
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Не удалось проверить — администратор не в сети. Группа останется скрытой, пока её не удастся проверить в сети."
+          }
+        }
+      }
+    },
     "Create an end-to-end encrypted group anchored on Stellar." : {
       "localizations" : {
         "en" : {
@@ -1286,6 +1302,22 @@
         }
       }
     },
+    "Dismiss" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Dismiss"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Закрыть"
+          }
+        }
+      }
+    },
     "Don’t have their key? " : {
       "localizations" : {
         "en" : {
@@ -1399,6 +1431,23 @@
         }
       }
     },
+    "Founder" : {
+      "extractionState" : "stale",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Founder"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Основатель"
+          }
+        }
+      }
+    },
     "Generating proof and updating the on-chain commitment. This usually takes a few seconds." : {
       "localizations" : {
         "en" : {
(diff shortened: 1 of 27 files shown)


dev-onym (commented):
COMMENT — solid implementation; logic mirrors `applyAnnouncement` and the wire-key choice for collision-avoidance is the right call. A few nits / robustness gaps below.

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:570-589` — `applyAvatar` does not validate `payload.version`. A future `version=2` shape would silently apply as v1. Cheap to add a `guard payload.version == 1 else { return }`, matching the rest of the dispatcher's defensive decoding.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:585` — no size cap on the accepted JPEG. The 16 KB / 256×256 budget is enforced only producer-side via `GroupAvatarImage` (`GroupAvatarPayload.swift:33`). An admin (or anyone who possesses the admin's Ed25519) can blow that budget and bloat local storage / break renders; a `payload.avatar.map(\.count) <= someCap` guard would harden this.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:574-589` — `sentAtMillis` is on the wire but never consulted on receive. Out-of-order redelivery (older update arriving after a newer one) will overwrite the newer photo and flicker until the next broadcast. Either store `lastAvatarUpdateMillis` on `ChatGroup` and drop older ones, or document the "next-write-wins" tradeoff in the doc-comment.
- `Sources/OnymIOS/Group/GroupAvatarBroadcaster.swift:42` — `senderBlsPubkeyHex: group.adminPubkeyHex?.lowercased() ?? ""` emits an empty hex when the group has no stored admin (Anarchy/OneOnOne). Receivers ignore the field (envelope is authoritative), but the empty value is wire noise; consider skipping the broadcast entirely if there's no admin since receive-side verification would drop it anyway.
- `Sources/OnymIOS/Group/GroupAvatarBroadcaster.swift:31-37` — `setAvatar` writes to the local store + fans out without verifying the caller is actually the admin. The UI gate in `ChatMembersView.canChangeAvatar` (`Chats/ChatMembersView.swift:117`) is the only check. Any future call site through `AppDependencies.setGroupAvatar` would update local state even though receivers would drop the broadcast — a small defense-in-depth gap.
- `Sources/OnymIOS/Chats/ChatMembersView.swift:117` — `canChangeAvatar` aliases `canShareInvite`, which requires `groupType == .tyranny`. PR text says this is intentional; worth a TODO/issue link so OneOnOne/Anarchy doesn't quietly stay read-only forever.
- `Resources/Localizable.xcstrings:1434` — "Founder" is now marked `extractionState: stale`, but the prior commit (`36640dc`) just renamed "Tyranny" → "Founder" in user-facing copy. If Xcode flags it stale, the rename may not actually be referenced in any source string. Worth double-checking before this lands so the new label isn't silently dead.
- Tests look good — round-trip + nil-removal + both directions of `ChatMessagePayload` anti-collision + admin/imposter/clear dispatcher paths. Nothing missing.

PR #169, opened 2026-05-23
fix(inbox): stop the launch-time get_commitment storm that breaks live joins

lead-onym:
## Symptom

A joiner who scans/taps an invite and is approved often **doesn't see the group live** — it only appears after an app restart.

## Root cause (relayer-side, not transport)

The invitation arrives and persists fine (hence restart shows it). The break is in the receive-side verifier's relayer usage:

- Inbox subscriptions have **no `since` filter**, so every relay (re)connect replays the **entire inbox history**.
- The dispatcher did an **uncached `get_commitment`** per replayed group-state message (`verifyTyrannyInvitation`, `verifyTyrannyAnnouncement`) — and **before any dedup** (`applyAnnouncement`'s member-key check ran *after* the chain read).
- That self-inflicted burst **throttled the relayer**. A throttled/lagging read was then treated as a **hard reject**, silently dropping the very invitation being verified.
- A later replay (after restart, once the burst subsided / chain caught up) succeeded → "works after restart".

The same hard-reject hit a legitimate **chain-behind** read: the admin anchors epoch N then immediately sends the snapshot; the joiner's relayer read still lags at N-1 → `onchain.epoch < snapshot.epoch` → reject + drop.

## Fixes

1. **Skip the chain read when already locally verified.** If a group with the exact `(commitment, epoch)` is already materialized, return `.verified` without a relayer call. The local Poseidon recompute still runs first, so a replay that swaps the roster while reusing the commitment is still rejected. This removes nearly all of the startup reads.
2. **Dedup announcements before the chain read.** Move the BLS-pubkey-hex dedup in `applyAnnouncement` *above* `verifyTyrannyAnnouncement`, so re-delivered known members don't each hit the relayer.
3. **Throttle/lag ⇒ defer, not reject.** A chain-read throw (throttled/offline/unconfigured) and a chain-behind read now return `.staleNeedsRefresh` (defer + admin-refresh retry) instead of `.reject`. **Deferral never materializes without a later exact-epoch chain match**, so forgery protection is unchanged — only the false-drop is fixed.
4. **`CachingChainStateReader`** — bounded retry + short (10s) TTL cache around the raw relayer reads, wired in `OnymIOSApp`. A stale cache entry can only ever cause a brief *deferral* (never a false reject, given #3), so caching is safe here while `SEPContractChainStateReader` stays cache-free.

## Why it's safe (security)

The only `.reject` paths that remain are genuine cryptographic failures: missing commitment, internal recompute mismatch, and commitment mismatch **at an exact epoch**. All "couldn't verify *right now*" cases defer and re-verify, and never materialize a group without an exact-epoch on-chain match. Existing reject-path tests are unchanged and still green.

## Tests

`xcodebuild test … -only-testing:OnymIOSTests/IncomingMessageDispatcherTests -only-testing:OnymIOSTests/CachingChainStateReaderTests` → **31 passed, 0 failures**, including:
- skip-chain-read-on-verified-replay (asserts `calls == 1` across two deliveries)
- announcement dedup-before-read (asserts `calls == 0` for a known member)
- throw → defer, chain-behind → defer
- cache hit / expiry / retry-recovers / retry-exhausted
- all pre-existing reject + materialize tests still pass

## Follow-ups (not in this PR)

- Cap the Nostr replay itself (a `since` / processed-event high-water-mark) to cut inbound bandwidth — this PR cuts the *relayer* RPM, which was the harmful part.

### DIFF
--- a/Sources/OnymIOS/Chain/CachingChainStateReader.swift
+++ b/Sources/OnymIOS/Chain/CachingChainStateReader.swift
@@ -1,0 +1,69 @@
+import Foundation
+
+/// Wraps a `ChainStateReading` with two defenses against the launch-time
+/// `get_commitment` request storm and the relayer throttling it triggers:
+///
+///  1. **Bounded retry** — a single transient read failure (relayer 429 /
+///     network blip) no longer surfaces as "couldn't verify". We retry a
+///     few times with linear backoff before giving up, so a momentarily
+///     throttled relayer doesn't make a fresh join silently fail.
+///  2. **Short TTL cache** — collapses a burst of reads for the *same*
+///     group (e.g. an invitation plus several member announcements
+///     replayed together on a relay reconnect) into one round-trip.
+///
+/// The TTL is deliberately short: a stale entry can only ever make a
+/// caller see an older `(commitment, epoch)`, which the dispatcher treats
+/// as "chain behind → defer + retry" (never a reject), so staleness
+/// self-heals once the entry expires. It is therefore safe to cache
+/// without risking a false rejection — the worst case is a brief deferral.
+///
+/// `SEPContractChainStateReader` itself stays cache-free (chain state is
+/// the source of truth); this decorator is the single place that trades a
+/// little staleness for far fewer relayer calls. Mirrors the posture
+/// onym-android should adopt around its own commitment reads.
+actor CachingChainStateReader: ChainStateReading {
+    private let inner: any ChainStateReading
+    private let ttl: TimeInterval
+    private let maxAttempts: Int
+    private let baseRetryDelayMillis: UInt64
+    private let now: @Sendable () -> Date
+
+    private var cache: [Data: (entry: SEPCommitmentEntry, at: Date)] = [:]
+
+    init(
+        inner: any ChainStateReading,
+        ttl: TimeInterval = 10,
+        maxAttempts: Int = 3,
+        baseRetryDelayMillis: UInt64 = 300,
+        now: @escaping @Sendable () -> Date = { Date() }
+    ) {
+        self.inner = inner
+        self.ttl = ttl
+        self.maxAttempts = max(1, maxAttempts)
+        self.baseRetryDelayMillis = baseRetryDelayMillis
+        self.now = now
+    }
+
+    func tyrannyCommitment(groupID: Data) async throws -> SEPCommitmentEntry {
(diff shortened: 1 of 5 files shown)


dev-onym (commented):
APPROVE — solid root-cause fix with strong test coverage and a clean security argument.

- `Sources/OnymIOS/Chain/CachingChainStateReader.swift:34` — cache has no bounded size or LRU eviction; entries persist until overwritten on next read. With a 10s TTL and a chat-app's group count this is fine, but worth a one-line note or a simple cap if group sets ever grow large.
- `Sources/OnymIOS/Chain/CachingChainStateReader.swift:55` — expired entries are silently rewritten but never proactively dropped; minor, same point as above.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:419` — short-circuit correctly gates on `(commitment, epoch)` AND already runs Poseidon recompute above, so a roster-swap replay still rejects. Good.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:421` — `currentGroups()` is scanned linearly per invitation; fine at current scale but a hot path if group count grows — consider an indexed lookup later. Not blocking.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:434` and `:460` — throw→defer and chain-behind→defer are the right call: the remaining `.reject` paths are all genuine crypto failures (missing/mismatched commitment at exact epoch), and deferral can't materialize without a later exact-epoch match. Security posture preserved.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:564` — moving dedup ahead of `verifyTyrannyAnnouncement` means a re-delivered known member now skips admin-sig verification entirely. Safe because we early-return without applying state, but worth keeping in mind: any future side effect added between dedup and apply must be re-evaluated.
- `Sources/OnymIOS/OnymIOSApp.swift:391` — caching reader wired only at the dispatcher; `SEPContractChainStateReader` stays the cache-free source of truth elsewhere. Correct.
- Tests: `CachingChainStateReaderTests` covers hit/expiry/retry-recovers/retry-exhausted with deterministic clock; `IncomingMessageDispatcherTests` adds skip-on-replay (asserts `calls == 1`), throw→defer, chain-behind→defer, and known-member dedup (asserts `calls == 0`). Coverage matches the claims in the PR body.
- Minor nit: the linear backoff (300ms·attempt) totals ~900ms worst-case before surfacing; acceptable for the storm scenario, but document that it blocks the dispatch loop for that long when fully throttled.

PR #173, opened 2026-05-27
feat(chat): plumb reply-to message reference end-to-end [1/3]

lead-onym:
First of three stacked PRs adding Telegram-style replies to chat. **Data layer only — no UI yet.**

## What
Carry an optional `replyToMessageID` through the full chat round-trip so a message can point at an earlier one in the same group. Only the target id travels on the wire and to disk; the quoted sender + body are resolved live from the local store at render time (PR 2).

## Changes
- **`ChatMessagePayload`** — add optional `reply_to_message_id`. Additive + optional, so it ships under `version = 1`: older senders omit it (decodes to nil), older receivers ignore it.
- **`ChatMessage` / `PersistedMessage`** — add `replyToMessageID`. Plain (unencrypted) column — it's a pointer to another row in the same table, left queryable for a future "replies to X" lookup. Adding an optional property is a SwiftData lightweight migration.
- **`SendMessageInteractor`** — `send(...)` takes the ref (default nil) → payload + optimistic row + return value; `retry(...)` reconstructs it from the stored row so retries don't drop it.
- **`IncomingMessageDispatcher`** — map the ref onto the inbound message. No new trust gate: a dangling/forged ref just renders "unavailable" since rendering only resolves local rows.

## Tests
6 new (payload round-trip + back-compat decode of a pre-replies payload, store round-trip, inbound mapping, send carrying the ref) plus every existing `ChatMessage`/`ChatMessagePayload` constructor updated. Full affected suite: **108 tests, 0 failures.**

### DIFF
--- a/Sources/OnymIOS/Chats/ChatMessage.swift
+++ b/Sources/OnymIOS/Chats/ChatMessage.swift
@@ -44,6 +44,14 @@ struct ChatMessage: Equatable, Sendable, Identifiable {
     /// always `.received` at insert and never changes.
     let status: MessageStatus
 
+    /// The message this one replies to, if any. Mirrors
+    /// `ChatMessagePayload.replyToMessageID`. Only the target ID is
+    /// stored — the UI resolves the quoted sender + body by looking
+    /// this up among the group's other messages at render time, so a
+    /// target that isn't on this device renders as "message
+    /// unavailable" instead of carrying a stale copy of its text.
+    let replyToMessageID: UUID?
+
     /// Mirrors `ChatMessageVariant.kind`. Stored alongside the body so
     /// migrating to multi-variant rendering later doesn't need to
     /// re-fetch the group to learn the flavour.
--- a/Sources/OnymIOS/Chats/ChatMessagePayload.swift
+++ b/Sources/OnymIOS/Chats/ChatMessagePayload.swift
@@ -45,6 +45,16 @@ struct ChatMessagePayload: Codable, Equatable, Sendable {
     /// `Date`) so the wire format is unambiguous and cross-platform.
     let sentAtMillis: Int64
 
+    /// The message this one replies to, if any. Optional + additive,
+    /// so it ships under `version = 1`: a sender that omits it decodes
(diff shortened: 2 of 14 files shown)


dev-onym (commented):
APPROVE

- **Backward compat is sound** — `ChatMessagePayload` (ChatMessagePayload.swift:19) uses synthesized `Codable`, so Optional decodes via `decodeIfPresent` and old receivers ignore the unknown key. Tests at ChatMessagePayloadTests.swift:75-89 cover the missing-key case.
- **Plaintext `replyToMessageIDString` is consistent with existing privacy model** — `groupID`, `sentAt`, `directionRaw` are already plain in PersistedMessage.swift:31-38; the new column is just a pointer to a sibling row, so leaving it queryable doesn't widen disk exposure.
- **No SwiftData migration risk** — adding an optional property is a lightweight migration automatically, and SwiftDataMessageStore.swift:39-48 already has a wipe-and-retry fallback on schema mismatch (pre-1.0, acceptable).
- **Dispatcher trust model is fine but the comment overclaims** — IncomingMessageDispatcher.swift:701-706 says rendering "only looks up local rows," but for safety the resolver in PR 2 must scope the lookup by `groupID`, not just any local row, since the ref is unvalidated and could point cross-group. Worth nailing down explicitly in PR 2.
- **Minor: no self-reply guard** — `replyToMessageID == message.id` isn't rejected at send (SendMessageInteractor.swift:55) or dispatch. Not exploitable, just renders "unavailable" until resolved against itself, but the sender API could cheaply reject it. Skip if not worth the churn.
- **Nit, test comment** — ChatMessagePayloadTests.swift:60-67 says "encode `null` (or omit)"; `JSONEncoder` always emits `null` for nil Optionals, never omits. The assertion still passes (the cast-to-String is nil either way), but the comment is slightly misleading.
- **Test coverage is complete for the data layer** — round-trip, snake_case key, back-compat decode, store round-trip, dispatch mapping, send/retry carry — exactly what's needed for a stack-1 data PR.

PR #174, opened 2026-05-27
feat(chat): render reply quote in bubble + tap to jump [2/3]

lead-onym:
Second of three stacked PRs for Telegram-style replies. **Read-side rendering. Stacked on #173 — review/merge that first.**

## What
A bubble whose message replies to another now shows a Telegram-style inset quote (accent bar + sender name + one-line body snippet) above its text. Tapping the quote scrolls to and briefly flashes the original.

The quote is resolved **live** from the local message list — no snapshot on the wire (per #173). A reply whose target isn't on this device renders a muted "Message unavailable" placeholder that isn't tappable.

## Changes
- **`ChatReplyQuote`** — the resolved quote handed to the cell: `name` + `snippet` + `accent`, or `.unavailable`.
- **`ChatBubbleCell`** — quote container above the body; body-top constraint toggles between the bubble top (no quote) and the quote's bottom. Quote colors track direction (on-accent over the outgoing fill, the quoted sender's accent over the incoming tint). Tap target installed only for an available target and torn down on reuse, same pattern as the failed-bubble retry tap. Adds `flashHighlight()` and a `simulateQuoteTapForTest()` seam.
- **`ChatThreadViewController`** — `replyQuote(for:)` resolves the target live (missing → `.unavailable`); `scrollAndHighlight(messageID:)` jumps to + pulses the original.

## Notes
- User-facing strings ("Message", "Message unavailable") are plain literals to match the surrounding chat UI — nothing enters `Localizable.xcstrings`, so the l10n catalog test is untouched.

## Tests
5 cell + 3 controller: quote content resolved from the target sender, unavailable placeholder, tap fires only when the target is available, dropped-on-reuse, unknown-target scroll is a no-op. Full chat UI suite: **54 tests, 0 failures.**

### DIFF
--- a/Sources/OnymIOS/Chats/ChatBubbleCell.swift
+++ b/Sources/OnymIOS/Chats/ChatBubbleCell.swift
@@ -28,6 +28,34 @@ struct ChatSenderDisplay {
     static let unknown = ChatSenderDisplay(name: "", accent: .blue, showNameHeader: false)
 }
 
+/// Resolved quote shown at the top of a bubble that replies to an
+/// earlier message — the Telegram-style inset preview. Built by
+/// `ChatThreadViewController` by looking the reply target up in the
+/// local message list; the cell just renders it.
+///
+/// The target is resolved *live* (not snapshotted into the payload),
+/// so when it isn't on this device the controller hands over
+/// `unavailable` and the cell renders a muted placeholder instead of
+/// a real sender + snippet.
+struct ChatReplyQuote {
+    /// Quoted sender's display name (alias or BLS fingerprint).
+    let name: String
+    /// Quoted message body, rendered on a single truncated line.
+    let snippet: String
+    /// Quoted sender's accent — colors the leading bar and the name,
+    /// so the quote reads as "from that person" at a glance.
+    let accent: OnymAccent
+    /// The reply target isn't in the local store (never delivered, or
+    /// deleted). Renders a muted "message unavailable" placeholder and
+    /// the quote isn't tappable.
+    let isUnavailable: Bool
+
+    /// Placeholder for a reply whose target this device doesn't have.
+    static let unavailable = ChatReplyQuote(
+        name: "", snippet: "", accent: .blue, isUnavailable: true
+    )
+}
+
 /// One message bubble row. Two visual styles toggled by
 /// `ChatMessage.direction`:
 ///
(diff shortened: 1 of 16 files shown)


ux-onym (commented):
APPROVE — clean read-side rendering; tap/teardown follow the existing retry pattern; tests cover the meaningful branches.

- `ChatBubbleCell.swift:300` (applyReplyQuote): tap recognizer torn down unconditionally before any reinstall, matching the failed-bubble pattern — reuse safety looks correct.
- `ChatBubbleCell.swift:489-509`: `quoteContainer` has no constraint pinning its bottom into the bubble, so when hidden it doesn't inflate bubble height — good. Minor: with `numberOfLines=1` labels and nil text, the hidden container still has a non-zero intrinsic frame; harmless since nothing depends on it, but you could collapse it via a height==0 constraint if you want zero layout work in the no-reply case.
- `ChatBubbleCell.swift:317-323`: when `reply.isUnavailable` and direction is `.incoming`, both `quoteBar` and `quoteNameLabel` use the same `text3` color — the accent bar visually disappears against the muted name. Consider keeping the bar at a slightly stronger tone (e.g. `text2`) so the placeholder still reads as a quote, not a stray label.
- `ChatBubbleCell.swift:308-315`: `quoteContainer.isHidden = false` set before label text is assigned in the unavailable branch — fine, but if `reply.name == ""` for an unavailable target, the name label gets the literal `"Message"` placeholder which is good; just confirming the empty-name fallback isn't reachable for `isUnavailable == false` (it isn't, since the controller always supplies a resolved name).
- `ChatThreadViewController.swift:304-313` (scrollAndHighlight): fixed 0.3s `asyncAfter` before `flashHighlight` is a fragile coupling to UITableView's scroll duration — if the user scrolls again or the row is far off-screen, `cellForRow(at:)` may return nil and the flash silently drops. Acceptable for now; consider wrapping the scroll in a `CATransaction` completion or using `scrollViewDidEndScrollingAnimation` if flakiness shows up.
- `ChatThreadViewController.swift:466-470`: quote-tap closure correctly gates on `reply?.isUnavailable == false` *and* `replyToMessageID` non-nil, captures `targetID` (not `reply`), and uses `[weak self]` — no retain cycle, no stale-quote tap.
- "Message" / "Message unavailable" string literals: PR note acknowledges this; fine for the stacked series but worth a follow-up to localize alongside any future xcstrings additions.
- Test `test_reply_droppedOnReuse` (`ChatBubbleCellTests.swift:265`) asserts container hides on reuse but doesn't verify the recognizer was removed; low risk since hidden views don't receive taps, but a `simulateQuoteTapForTest()` returning `false` after reuse would close the loop cheaply.

PR #175, opened 2026-05-27
feat(chat): swipe-to-reply gesture + composer banner [3/3]

lead-onym:
Final of three stacked PRs for Telegram-style replies. **Compose side. Stacked on #174 (which is stacked on #173) — merge those first.**

## What
Drag a bubble left to arm a reply; a banner appears over the composer showing who/what you're replying to; sending threads the reference through `SendMessageInteractor` so the new message renders its quote (from #174).

## Changes
- **`ChatBubbleCell`** — interactive drag-follow pan (the gesture fidelity called for). The bubble tracks the finger leftward (clamped to a max travel), a reply-arrow glyph grows in, and a one-shot haptic fires as it crosses the threshold; releasing past the threshold arms the reply, otherwise everything springs back. The pan is gated via UIView's `gestureRecognizerShouldBegin` override to **leftward, mostly-horizontal** drags, so vertical scrolling and the controller's rightward full-width back-pan are untouched; it recognizes simultaneously with the scroll pan. Transform resets on reuse.
- **`ChatInputPanelView`** — "Replying to {name}" banner above the composer (accent bar + one-line snippet + cancel button); the text view re-pins below it. Adds `showReplyBanner` / `clearReplyBanner` / `onCancelReply` / `focusComposer`.
- **`ChatThreadViewController`** — tracks `replyingTo`; a swipe arms it, shows the banner, and focuses the composer; send forwards `(body, replyingTo)` and then clears; cancel disarms. `onSendTapped` is now `(String, UUID?)`.
- **`ChatThreadView`** — forwards the target into `SendMessageInteractor.send(…, replyToMessageID:)`.

## Notes
- New user-facing strings ("Replying to …", "Cancel reply") stay plain literals to match the chat UI — l10n catalog test untouched.
- The drag handler itself isn't unit-tested (synthesising live pan states is brittle); the armed-reply callback, banner wiring, and send-forwarding are, via test seams.

## Tests
3 cell (swipe fires the callback, available for any direction/status, dropped on reuse) + 4 input-panel (banner show/clear/cancel/initial-hidden) + 3 controller (swipe arms banner with target sender, send forwards target + clears, cancel drops target). **Full unit target: 703 tests, 0 failures.**

### DIFF
--- a/Sources/OnymIOS/Chats/ChatBubbleCell.swift
+++ b/Sources/OnymIOS/Chats/ChatBubbleCell.swift
@@ -119,6 +119,23 @@ final class ChatBubbleCell: UITableViewCell {
     private var onQuoteTapped: (() -> Void)?
     private var quoteTapRecognizer: UITapGestureRecognizer?
 
+    /// Fired when the user drags the bubble far enough left to arm a
+    /// reply (Telegram-style). Available on every message regardless of
+    /// direction or status. Set on each `configure`; reset transform on
+    /// reuse so a recycled mid-drag cell starts clean.
+    var onSwipeToReply: (() -> Void)?
+    private var swipePan: UIPanGestureRecognizer!
+    private let replyHint = UIImageView()
+    /// Drag distance (points) past which releasing arms a reply.
+    private let swipeReplyThreshold: CGFloat = 56
+    /// How far the bubble can travel — past the threshold it resists so
+    /// the gesture has a clear "armed" ceiling.
+    private let swipeMaxTravel: CGFloat = 72
+    /// True once the current drag crossed the threshold — drives the
+    /// one-shot haptic and the release decision.
+    private var swipeArmed = false
+    private let swipeHaptic = UIImpactFeedbackGenerator(style: .medium)
+
     // Direction-dependent constraints — only one pair is active at a
     // time. Each pair contains the alignment pin (`==` to the pinned
     // edge) and the opposite-edge breathing-room gap on the *other*
@@ -160,6 +177,7 @@ final class ChatBubbleCell: UITableViewCell {
         buildStatusIndicator()
         buildNameLabel()
         layoutBubble()
+        installSwipeToReply()
     }
 
     @available(*, unavailable)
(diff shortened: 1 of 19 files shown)


dev-onym (commented):
APPROVE

- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:670-687` — `gestureRecognizerShouldBegin` override correctly gates on `=== swipePan` and uses horizontal-dominant + leftward velocity; reads well.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:688-695` — `shouldRecognizeSimultaneouslyWith` returns `true` unconditionally. Narrowing to the table's scroll pan (e.g. `other is UIPanGestureRecognizer` or specifically the scroll view's recognizer) would be safer if a long-press, edge-pan, or another row-level recognizer is added later. Minor; comment acknowledges the tradeoff.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:336-401` — drag-follow math, threshold clamp, one-shot armed haptic, and spring-back on `.cancelled`/`.failed` only firing on `.ended` look correct. `resetSwipeState()` called from `configure()` covers reuse mid-drag.
- `Sources/OnymIOS/Chats/ChatInputPanelView.swift:140-225` — banner is hidden by default but stays in the hierarchy; labels are empty so intrinsic height collapses. The top-anchor toggle (`textViewTopToPanelConstraint` ⇄ `textViewTopToBannerConstraint`) is exclusive and switched on every show/clear — no constraint conflict.
- `Sources/OnymIOS/Chats/ChatInputPanelView.swift:281-285` — `focusComposer` only calls `becomeFirstResponder()`. Worth checking on the device that arming → focus doesn't fight with an in-flight resign during scroll-to-bottom, but no logic flaw on paper.
- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:498-544` — `handleSend` → `clearReply` ordering means a send always disarms; matches the test. `armReply` no-ops if target left the snapshot — good.
- `Tests/OnymIOSTests/ChatThreadViewControllerTests.swift:518-583` — three scenarios (arm, send-forwards-and-clears, cancel-then-send) exercise the full state machine via the test seam. Reasonable given the live-pan caveat called out in the PR body.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:478-482` — `simulateSwipeToReplyForTest` short-circuits the gesture entirely (just fires `onSwipeToReply?()`). That's the documented test seam, but be aware tests can't catch regressions in the threshold/velocity gating itself — only the wiring.
- `Sources/OnymIOS/Chats/ChatBubbleCell.swift:339-352` — accessibility: the reply hint glyph has no `accessibilityLabel`; VoiceOver users can't trigger swipe-to-reply at all (no custom action either). Not a blocker for this PR, but worth a follow-up custom action ("Reply") on the cell so the gesture has a non-gestural path.

PR #187, opened 2026-07-27
fix(security): check CSPRNG status via shared SecureRandom helper (H-1)

lead-onym:
## Summary

Fixes **H-1 (High)**: four call sites invoked `SecRandomCopyBytes(...)` as `_ = SecRandomCopyBytes(...)`, discarding the `OSStatus`, then used the buffer as security material. On a CSPRNG failure the buffer stays all-zeros — a silent, catastrophic all-zero mnemonic, storage root key, or group salt/scalar. (A zero scalar is a canonical `Fr`, so `randomCanonicalFr()`'s rejection loop would have happily accepted a zero-on-failure result.)

### Fix
- New `Sources/OnymIOS/Identity/SecureRandom.swift`: `SecureRandom.bytes(_:)` / `.data(_:)` call `SecRandomCopyBytes`, check the status, and throw `SecureRandomError.csprngFailed(status:)` on any non-`errSecSuccess`. Mirrors the existing correct pattern in `OnymNostrSigner.ephemeral()`.
- Routed all four sites through it:
  - `Bip39.generateMnemonic()` — now `throws`; its only caller (`IdentityRepository.addLocked`, already `throws`) threads `try`.
  - `StorageEncryption` root secret — `loadOrCreateRootSecret()` / `saveToKeychain()` now `throw`; `storageKey` became a `get throws` property; the memoised `cachedRootSecret` is stored as a `Result` so the thread-safe once-only `static let` init is preserved while errors still propagate. `encrypt`/`decrypt` already `throws`.
  - `CreateGroupInteractor.randomBytes(_:)` and `randomCanonicalFr()` — now `throw`; all call sites are already inside `throws` contexts and thread `try`.
- Also checked the previously-ignored `SecItemAdd` status in `StorageEncryption.saveToKeychain()` — now throws `StorageEncryptionError.keychainWriteFailed(status:)` on failure.
- Removed the now-unused `import Security` from `CreateGroupInteractor`.

## Testing

Ran on iOS Simulator (iPhone 17, iOS 26), Xcode 26, unique derived data path `/tmp/onym-ddp-h1`:

- `xcodebuild build-for-testing` → **TEST BUILD SUCCEEDED**
- `xcodebuild test-without-building` restricted to the affected classes: `SecureRandomTests`, `StorageEncryptionTests`, `CanonicalFrTests`, `CreateGroupInteractorTests`, `IdentityRepositoryTests`, `IdentityRepositoryMultiIdentityTests` → **67 tests, 0 failures**.

New `Tests/OnymIOSTests/SecureRandomTests.swift` covers: helper returns the requested length (incl. 0), output is non-constant / non-zero across 100 draws, and the three former call sites (`generateMnemonic`, `randomCanonicalFr`, storage encrypt/decrypt roundtrip) still produce valid non-zero output. Existing `CanonicalFrTests` / `StorageEncryptionTests` updated to thread `try` through the newly-throwing APIs.

Forcing a real CSPRNG failure isn't feasible in a unit test, so the throw path is verified by compilation/wiring (all callers propagate) plus the success-path length/smoke assertions.

Closes #182

### DIFF
--- a/Sources/OnymIOS/Group/CreateGroupInteractor.swift
+++ b/Sources/OnymIOS/Group/CreateGroupInteractor.swift
@@ -1,6 +1,5 @@
 import CryptoKit
 import Foundation
-import Security
 
 /// Stateless orchestration for the create-group flow. Holds dependencies
 /// only — every call to `create` is independent. The view-model
@@ -147,8 +146,8 @@ struct CreateGroupInteractor: Sendable {
         // exists to close a same-`group_id_fr` collision via
         // `group_id + p (mod 2^256)` — see the contract comment at
         // sep-tyranny/src/lib.rs:290–298.
-        let groupID = Self.randomCanonicalFr()
-        let groupSecret = Self.randomBytes(32)
+        let groupID = try Self.randomCanonicalFr()
+        let groupSecret = try Self.randomBytes(32)
         let salt = GroupCommitmentBuilder.generateSalt()
         let tier: SEPTier = .small
 
@@ -326,8 +325,8 @@ struct CreateGroupInteractor: Sendable {
         // proof binding), but we use the canonical sampler anyway for
         // consistency with Tyranny + to keep all on-chain group IDs
         // valid Fr scalars in case a future contract rev adds the check.
-        let groupID = Self.randomCanonicalFr()
-        let groupSecret = Self.randomBytes(32)
+        let groupID = try Self.randomCanonicalFr()
+        let groupSecret = try Self.randomBytes(32)
         let salt = GroupCommitmentBuilder.generateSalt()
 
         // 4. Both BLS secrets must be present at create time — the SDK's
@@ -355,12 +354,12 @@ struct CreateGroupInteractor: Sendable {
         // BLS secret keys are canonical Fr by convention. The SDK reduces
         // silently if not, but any future strictness check would silently
         // break us — so generate canonical at the source.
-        var peerBlsSecret = Self.randomCanonicalFr()
+        var peerBlsSecret = try Self.randomCanonicalFr()
         // The OneOnOne SDK rejects equal secrets. Vanishingly unlikely
         // with 256 bits of entropy, but cheap to defend against —
         // resample on the off chance of a collision.
         while peerBlsSecret == creatorBlsSecret {
-            peerBlsSecret = Self.randomCanonicalFr()
+            peerBlsSecret = try Self.randomCanonicalFr()
         }
 
         let creatorMember: GovernanceMember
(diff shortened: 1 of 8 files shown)


dev-onym (commented):
**COMMENT** — the core fix is correct (all four sites now check status, callers thread `try`), but the `Result`-cached root secret plus the untouched `loadFromKeychain` error handling turn a transient Keychain failure into a permanent one.

- `Sources/OnymIOS/Persistence/StorageEncryption.swift:27` — caching a `Result` in a `static let` makes failure **sticky for the process lifetime**. A transient CSPRNG or Keychain error (e.g. `errSecInteractionNotAllowed` on a pre-first-unlock launch) permanently poisons `storageKey`, so every subsequent `encrypt`/`decrypt` fails even after the device unlocks. Previously the process at least recovered on the next launch with a usable (if unpersisted) key. Consider a lock-guarded memo that only caches success, and retries on error.
- `Sources/OnymIOS/Persistence/StorageEncryption.swift:98-101` — pre-existing but directly adjacent and made reachable by this change: `loadFromKeychain()` conflates "not found" with "read failed". Any non-`errSecItemNotFound` status returns `nil` → `loadOrCreateRootSecret` mints a fresh secret → `saveToKeychain` runs `SecItemDelete` (line 122) **before** `SecItemAdd`, destroying the real root key. Every previously-encrypted message/group column becomes permanently undecryptable. `IdentityKeychainStore.swift:101` already does this correctly (`if status == errSecItemNotFound { return nil }`, throw otherwise) — worth mirroring here, ideally in this PR since the new `keychainWriteFailed` throw sits right below it.
- `Sources/OnymIOS/Group/SwiftDataGroupStore.swift:117`, `Sources/OnymIOS/Persistence/SwiftDataInvitationStore.swift:114` — the new throw is swallowed by `try?`: a CSPRNG/Keychain failure now silently drops the commitment or returns `false` from `save`. The error surfaces in the type system but not to the user. Pre-existing pattern, but H-1's stated goal is "no silent failure on security material" — these two sites still are silent.
- `Sources/OnymIOS/Group/GroupCommitmentBuilder.swift:16` — `generateSalt()` still uses `UInt8.random(in:)` byte-by-byte rather than `SecureRandom`. Not a defect (`SystemRandomNumberGenerator` is `arc4random_buf` on Darwin), but the PR body claims group salt was among the fixed sites; it wasn't touched. Either route it through the new helper for consistency or correct the description.
- `Sources/OnymIOS/Persistence/StorageEncryption.swift:51,67` — `try AES.GCM.seal(plaintext, using: try storageKey)` has a redundant inner `try`; the outer one already covers it.
- `Sources/OnymIOS/Group/CreateGroupInteractor.swift:856` — `randomBytes(_:)` is now a one-line passthrough to `SecureRandom.data`; could be deleted and the three call sites pointed at `SecureRandom.data(32)` directly.
- Nit on coverage: the throw path is untestable because `SecureRandom` is a non-injectable `enum`. If you want a real test, a `static var generator: (Int) throws -> [UInt8]` seam (or an internal `bytes(_:using:)` overload) would let `StorageEncryptionTests` assert that a CSPRNG failure propagates rather than yielding zeros.

I could not compile-verify locally (Linux, no Xcode); the `get throws` accessor, `Result(catching:)` memo, and all `try` threading read as valid Swift 5.5+, consistent with the reported 67-test green run.

PR #189, opened 2026-07-27
fix(security): fail-closed envelope provenance (C-1) + admin-less sender auth (H-2)

lead-onym:
## Summary

Two linked receiver-side authorization vulnerabilities in the inbox pipeline. **Receiver-side only — no wire-format change.** Legitimate senders always sign the envelope (see `sealInvitation`), so all real traffic is unaffected; only forged/unsigned envelopes are now correctly rejected.

### C-1 (Critical) — envelope signature verification was skippable
`decryptSealedEnvelope` verified the Ed25519 signature over the ephemeral key only inside `if let sig, let senderPub { ... }`. An envelope that carried `senderEd25519PublicKey` but **omitted** `ephemeralKeySignature` decrypted with the signature check skipped, and `decryptInvitationWithSender` then returned the raw, unverified `envelope.senderEd25519PublicKey` unconditionally. That attacker-chosen key flowed downstream as the "verified signer" trusted for authorization (`persistChatMessage`; the admin gates on `applyName`/`applyAvatar`/`applyAnnouncement`).

**Fix:** provenance is now fail-closed. A new `decryptSealedEnvelopeWithSender` surfaces the sender pubkey **only** after a signature is present AND verifies. A present-but-invalid signature still throws (unchanged); a wholly absent signature block decrypts with a `nil` verified sender. `decryptInvitationWithSender` returns that verified value (nil when absent/invalid) instead of the raw wire field. The `InvitationEnvelopeDecrypting` protocol default is hardened identically. The plaintext-only decrypt (e.g. `JoinRequestApprover`) is unchanged.

### H-2 (High) — admin-less groups skipped sender auth
`applyAnnouncement`, `applyAvatar`, and `applyName` wrapped the signer check in `if let storedAdmin = group.adminEd25519PubkeyHex { ... }`. When no admin key was stored (anarchy / oneOnOne / pre-PR-9 rows) the check was **skipped** and the mutation applied unconditionally by anyone who knew the recipient's inbox pubkey.

**Fix:** the skip-branch is replaced with a positive check (`isAuthorizedGroupMutation`):
- A missing/unverified signer is **always** rejected.
- With a stored admin key, the verified signer must equal it (existing behavior).
- With no stored admin, the verified signer must be a **current member** — match some `memberProfiles[*].sendingPubkey`.

## Testing

Built for the iOS Simulator (iPhone 17) with a dedicated derived-data path.

The full `OnymIOSTests` suite passed locally: **833 tests, 3 skipped, 0 failures — TEST SUCCEEDED** (~22s).

Affected classes (all green), including new coverage:
- `SealedEnvelopeTests`
- `IdentityRepositoryInvitationDecryptTests` — 2 new C-1 provenance tests
- `IncomingMessageDispatcherTests` — 6 new H-2 admin-less accept/reject-non-member/reject-unsigned tests for announcement/name/avatar; 2 existing tests updated to the fail-closed semantics
- `IncomingMessageDispatcherChatMessageTests`
- `InvitationDecryptorTests`

New/updated coverage asserts: an envelope with `senderEd25519PublicKey` set but `ephemeralKeySignature` omitted decrypts yet surfaces a **nil** verified sender and is rejected by `persistChatMessage` and by `applyName`/`applyAvatar`/`applyAnnouncement`; and an admin-less group rejects a name/avatar/announcement mutation whose signer is not a current member.

Closes #180
Closes #181

### DIFF
--- a/Sources/OnymIOS/Identity/IdentityRepository.swift
+++ b/Sources/OnymIOS/Identity/IdentityRepository.swift
@@ -323,13 +323,18 @@ actor IdentityRepository: InvitationEnvelopeDecrypting, InvitationEnvelopeSealin
         } catch {
             throw InvitationDecryptError.malformedEnvelope
         }
-        let plaintext = try Self.decryptSealedEnvelope(
+        let result = try Self.decryptSealedEnvelopeWithSender(
             envelope: envelope,
             recipientX25519PrivateKey: privateKey
         )
+        // C-1: surface the sender pubkey ONLY when the envelope's
+        // Ed25519 signature was present AND verified. A key carried
+        // without a valid signature is attacker-choosable, so it must
+        // reach provenance-sensitive callers as `nil` ("no proof of
+        // sender"), never as the raw wire value.
         return DecryptedEnvelope(
-            plaintext: plaintext,
-            senderEd25519PublicKey: envelope.senderEd25519PublicKey
+            plaintext: result.plaintext,
+            senderEd25519PublicKey: result.verifiedSenderEd25519PublicKey
         )
     }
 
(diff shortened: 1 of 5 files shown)


ux-onym (commented):
Failed to authenticate: OAuth session expired and could not be refreshed

dev-onym (commented):
**APPROVE** — both fixes are correct and correctly scoped; findings below are non-blocking (couldn't build/test: Linux box, no Xcode).

**C-1**
- `IdentityRepository.swift:395-441` — logic is right: `verifiedSender` is set only inside the branch that ran `isValidSignature`, present-but-invalid still throws, absent-sig yields `nil`. `decryptSealedEnvelope` (plaintext-only, `JoinRequestApprover.swift:696`) keeps its old semantics.
- `InvitationEnvelopeDecrypting.swift:62-76` — this default is unreachable and untested: `IdentityRepository.swift:310` and `FakeInvitationEnvelopeDecrypter.swift:51` both override `decryptInvitationWithSender`, and no test conformer uses the default. It also *diverges* from the concrete impl — an envelope with a present-but-**invalid** signature returns `nil` here vs. throwing `.signatureVerificationFailed` there. Either drop it or align + cover it.
- No new test asserts a present-but-invalid signature still throws through `decryptInvitationWithSender`; existing coverage (`IdentityRepositoryInvitationDecryptTests.swift:197`) only exercises `decryptInvitation`.

**H-2**
- `IncomingMessageDispatcher.swift:585-601` — gate is sound; `sendingPubkey` really is the envelope-signing Ed25519 (`IdentitySummary.swift:21`, `CreateGroupInteractor.swift:839`), so the member comparison is well-founded, and hex compare for the admin branch is case-normalized on both sides.
- Availability narrowing worth documenting: for admin-less groups the member set comes from `memberProfiles`, and on the **creator's** device that map only ever holds the creator (`CreateGroupInteractor.swift:828`; the only other writers are `applyAnnouncement` and `JoinRequestApprover.swift:523`, both Tyranny/join-link paths). So in oneOnOne/anarchy a legitimate rename/avatar from the peer is now silently dropped on the creator side. This matches what `persistChatMessage` (`:761`) already does to inbound chat from an unknown-profile peer, so it's consistent rather than newly broken — but it is a real behavior change with no log/telemetry. Worth a line in the PR body or a debug log on rejection.
- Announcements are only emitted by `JoinRequestApprover.swift:567` (admin path), so the announcement tightening should have no legitimate-traffic impact.

**Adjacent, not introduced here**
- `applyName`/`applyAvatar` (`:679`, `:706`) still resolve the group with `groups.first(where: { $0.groupIDData == ... })` with no `ownerIdentityID` filter, unlike `persistChatMessage:759` — with two local identities in one group the mutation lands on an arbitrary row.
- `applyReceipt` (`:835`) takes no sender at all — same class as H-2 (anyone knowing the inbox pubkey can flip messages to delivered/read). Out of scope, but a natural follow-up ticket.

**Tests** — new coverage is well-targeted (member-accept / non-member-reject / unsigned-reject across announcement+name+avatar, plus the C-1 forged-envelope case); the two updated legacy tests correctly encode the new semantics rather than being weakened.

PR #192, opened 2026-07-27
fix(security): recovery-phrase biometric gate fails closed in release

lead-onym:
## Summary

The recovery-phrase biometric gate failed **open** in Release. In `LAContextAuthenticator.authenticate(reason:)`, when `LAContext.canEvaluatePolicy(.deviceOwnerAuthentication)` was false (no passcode / no enrolled biometric), the method simply `return`ed — treated as success — with no prompt. Since this shipped in Release, a device with no device security would reveal the recovery phrase with no authentication at all.

### Fix
- Fail-open is now **DEBUG-only**. In Release, an unevaluable policy fails **closed**: `LAContextAuthenticator` throws `BiometricAuthError.deviceSecurityUnavailable` ("Set a device passcode to view your recovery phrase."). The message never reveals the phrase.
- DEBUG keeps the fail-open `return` so simulator/dev flows still work on a fresh simulator without a passcode.
- The posture is a compile-time default (`failClosedByDefault`: `false` in DEBUG, `true` in Release) but injectable via a `failClosed` flag so both branches are unit-testable under the DEBUG test build. The `canEvaluate` check is likewise injectable for deterministic testing.
- The throw propagates into `RecoveryPhraseBackupFlow.authenticate()`, which routes it to `.authFailed` — the flow never reaches `.reveal` on this path.
- `AlwaysAcceptAuthenticator` was already correctly gated behind `#if DEBUG` and is left untouched.

## Testing

`xcodebuild test -scheme OnymIOS -only-testing:OnymIOSTests/RecoveryPhraseBackupFlowTests` on the iPhone 17 simulator (derived data `/tmp/onym-ddp-bio`): **18 tests, 0 failures**.

New tests:
- `test_unevaluablePolicy_failClosed_blocksReveal` — real `LAContextAuthenticator` with policy forced unevaluable + `failClosed: true` drives the flow to `.authFailed` (not `.reveal`) with the passcode message.
- `test_unevaluablePolicy_failOpen_reachesReveal` — same but `failClosed: false` reaches `.reveal` (DEBUG posture).
- `test_laContextAuthenticator_failClosed_throwsOnUnevaluablePolicy` / `..._failOpen_returnsOnUnevaluablePolicy` — direct unit tests of both branches.
- `test_laContextAuthenticator_defaultPosture_matchesBuildConfiguration` — asserts the compile-time default matches the build config.

All pre-existing tests remain green.

Closes #190

### DIFF
--- a/Sources/OnymIOS/Recovery/BiometricAuthenticator.swift
+++ b/Sources/OnymIOS/Recovery/BiometricAuthenticator.swift
@@ -7,20 +7,62 @@ import LocalAuthentication
 protocol BiometricAuthenticator: Sendable {
     /// Prompts the user. Returns on success; throws on cancel / failure.
     /// On devices/simulators where `canEvaluatePolicy` is false (no enrolled
-    /// biometric and no passcode), the call returns successfully without
-    /// prompting — matches the reference impl's "fail open in dev" behaviour.
+    /// biometric and no passcode), behaviour depends on the build: DEBUG
+    /// returns successfully ("fail open") so dev/simulator flows work, while
+    /// Release throws ("fail closed") so the recovery phrase is never revealed
+    /// without authentication.
     func authenticate(reason: String) async throws
 }
 
+/// Raised by `LAContextAuthenticator` when the device cannot evaluate the
+/// auth policy (no passcode / no enrolled biometric) and the build is
+/// configured to fail closed. The message never reveals the phrase.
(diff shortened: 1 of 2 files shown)


dev-onym (commented):
Failed to authenticate. API Error: 401 OAuth access token has expired. Re-authenticate to continue.

PR #193, opened 2026-07-29
fix(transport): self-heal relay socket so live messages don't need a relaunch

rinat-enikeev:
## Problem

Inbound messages arrive **only** over a live Nostr WebSocket subscription that is opened once, at launch — there is no push/APNs path. Nothing re-established that socket once it died, so a backgrounded/suspended app (or a Wi‑Fi↔cellular hand-off, or a relay idle-close) went permanently deaf until a **hard relaunch**, which is the only thing that rebuilds the connection and re-issues the `REQ`.

Reported symptom: messages sent from production Android (and from another iOS instance on a simulator) were received on iOS only after force-quitting and reopening the app. Reproduced on TestFlight, Debug, and simulator-to-simulator — confirming it's the socket, not push.

## Root cause — three compounding gaps

1. **Passive-only recovery.** The connection reconnected *only* if `URLSessionWebSocketTask.receive()` threw. On a half-open socket it never throws (`waitsForConnectivity = true`, `timeoutIntervalForResource = 0`). The old heartbeat was a fire-and-forget `send` with `try?` that swallowed errors and never triggered a reconnect.
2. **No foreground / connectivity trigger.** No `scenePhase` observer and no `NWPathMonitor` touched the transport, so returning from background or switching networks did nothing.
3. **Overlapping-loop hazard** once we start forcing reconnects.

## Fix

- **Active liveness monitor** in `NostrRelayConnection`: tracks last-received-frame time, sends a match-nothing `REQ` each cycle to draw an immediate `EOSE` (a pong stand-in), and force-reconnects when either the probe send fails or no frame arrives within the liveness window. Still avoids `sendPing` (documented CFNetwork crash).
- **Lifecycle triggers** in `OnymIOSApp`: a `scenePhase` observer reconnects on the real background→foreground edge, and a new `NetworkPathMonitor` (`NWPathMonitor`) stream reconnects on regained connectivity / interface change.
- **Connection-generation token** so a forced reconnect and the error-path reconnect can never leave two receive loops / monitors fighting over the same state.
- `reconnect()` added to `InboxTransport`. The per-connection subscription dict is preserved across a rebuild, so `REQ`s replay and consumer `AsyncStream`s never break.

## Tests

New `NostrRelayConnectionReconnectTests` drives a real `NostrRelayConnection` against a loopback WebSocket relay (`LocalWebSocketRelay`) and asserts a subscription keeps delivering across:
- an explicit `reconnect()` (the foreground/connectivity trigger), and
- a server-initiated drop (the passive error-path reconnect).

Full suite green: **858 tests, 0 failures** (3 pre-existing skips).

### DIFF
--- a/Sources/OnymIOS/Chain/UITestFakes.swift
+++ b/Sources/OnymIOS/Chain/UITestFakes.swift
@@ -130,6 +130,9 @@ final class UITestLoopbackInboxTransport: InboxTransport, @unchecked Sendable {
 
     func connect(to endpoints: [TransportEndpoint]) async {}
     func disconnect() async {}
+    // In-process loopback keeps its subscriber continuations across any
+    // "reconnect", so there is nothing to rebuild — subscribers stay live.
+    func reconnect() async {}
 
     @discardableResult
     func send(_ payload: Data, to inbox: TransportInboxID) async throws -> PublishReceipt {
--- a/Sources/OnymIOS/Chats/ChatThreadView.swift
+++ b/Sources/OnymIOS/Chats/ChatThreadView.swift
@@ -253,9 +253,9 @@ struct ChatThreadView: View {
             guard let owner = chatsFlow.groups
                 .first(where: { $0.id == groupID })?.ownerIdentityID
             else { return }
-            NostrRelayConnection.deliveryLog.debug("thread SUBSCRIBED group=\(groupID, privacy: .public) owner=\(owner.rawValue.uuidString, privacy: .public)")
+            NostrRelayConnection.deliveryLog.notice("thread SUBSCRIBED group=\(groupID, privacy: .public) owner=\(owner.rawValue.uuidString, privacy: .public)")
             for await snapshot in messageRepository.snapshots(groupID: groupID, owner: owner) {
-                NostrRelayConnection.deliveryLog.debug("thread SNAPSHOT count=\(snapshot.count) group=\(groupID, privacy: .public)")
+                NostrRelayConnection.deliveryLog.notice("thread SNAPSHOT count=\(snapshot.count) group=\(groupID, privacy: .public)")
                 messages = snapshot
                 // Read receipts: the thread is on-screen (this task is
                 // tied to its lifetime), so any incoming message here is
--- a/Sources/OnymIOS/Chats/ChatThreadView.swift
+++ b/Sources/OnymIOS/Chats/ChatThreadView.swift
@@ -253,7 +253,9 @@ struct ChatThreadView: View {
             guard let owner = chatsFlow.groups
                 .first(where: { $0.id == groupID })?.ownerIdentityID
             else { return }
+            NostrRelayConnection.deliveryLog.debug("thread SUBSCRIBED group=\(groupID, privacy: .public) owner=\(owner.rawValue.uuidString, privacy: .public)")
             for await snapshot in messageRepository.snapshots(groupID: groupID, owner: owner) {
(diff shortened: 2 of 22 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — generation token is sound and the direction is right, but the trigger side is too blunt and the new self-healing path is untested.

- `NetworkPathMonitor.swift:22-24` — the gate swallows the **first** `.satisfied` callback, but the `guard path.status == .satisfied` runs *before* the gate, so an unsatisfied initial path doesn't consume the latch. Launch offline → connectivity arrives → that genuine regain event is swallowed and no reconnect fires. Track `previousStatus` and emit on the unsatisfied→satisfied edge instead of a one-shot latch.
- `NetworkPathMonitor.swift:19-27` + `NostrRelayConnection.swift:113` — `pathUpdateHandler` fires on *any* satisfied update (interface/DNS/proxy churn), each one force-drops a healthy socket, and `reconnect()` resets `reconnectAttempts = 0`. On a flapping cellular link this defeats the exponential backoff entirely and hammers a dead relay at the flap rate. Needs debounce/coalescing, and `reconnect()` should not clear the backoff counter.
- `OnymIOSApp.swift:669-671`, `598-602` — reconnect is unconditional, and the REQ replay has no `since`, so every foreground and every path update triggers a full-inbox replay and re-dispatch. The comment right above at :598 records that "every relay reconnect replays the full inbox … storming the relayer into throttling fresh joins". This PR multiplies that frequency. Consider probing first and rebuilding only when the probe fails, or bounding replay with `since`.
- `NostrRelayConnection.swift:297-300` — `livenessTick` returns `false` (monitor exits, **no** reconnect) when `webSocketTask == nil` or `state != .running`. That's fail-open in the one component whose job is to catch a socket the error path missed: if `receive()` is hung and the task is non-running, nothing ever heals it. Non-running should route to `handleConnectionFailure`, not silently stop.
- `NostrRelayConnection.swift:302-305` — liveness-triggered reconnect uses `backoff: false` and `connect()` resets attempts, so a relay that never answers the probe with `EOSE` (non-conformant, or rejects `limit: 0`) loops forever at ~60s intervals with no escalation.
- `Tests/OnymIOSTests/Support/LocalWebSocketRelay.swift:105-111` — the fake relay drains and ignores client frames and `push()` sends unconditionally, so `testSubscriptionResumesAfterForcedReconnect` passes even if the REQ replay were deleted. The PR's central claim ("REQs replay") isn't actually asserted; have the relay record REQ frames and assert the replay.
- No test covers the liveness monitor at all (headline gap #1), nor `NetworkPathMonitor`, nor the scenePhase wiring — `FakeInboxTransport.reconnectCallCount` (`Support/FakeInboxTransport.swift:29`) is added "so tests can assert" but is never read. `pingInterval`/`livenessTimeout` are private statics with no injection, making the half-open path untestable; consider injecting them.
- `Tests/OnymIOSTests/Support/LocalWebSocketRelay.swift:44,57` — `Thread.sleep` inside `async` tests blocks a cooperative-pool thread for up to 6s, which can starve the actor under test; prefer an async poll.
- Nit: `NostrRelayConnection.swift:7,18` — doc comment still says "heartbeat ping" and the field is still `pingTask` after the rename to a liveness monitor.

ux-onym (commented):
API Error: 500 Internal server error. This is a server-side issue, usually temporary — try again in a moment. If it persists, check https://status.claude.com.

PR #194, opened 2026-07-29
fix(transport): connection state machine, CLOSED handling, single-REQ subscriptions (reconnect PR-2)

rinat-enikeev:
PR-2 of the reconnection design (`~/Desktop/reconnect.md`, §5.1–5.2, §9). Transport core only — app-lifecycle triggers land in PR-4, `since`-bounded replay + dedup in PR-3. Supersedes the corresponding parts of closed #193, with the gaps that sank it closed by design.

## Root cause this fixes (F8, confirmed in PR-1)

Stock strfry ships `maxSubsPerConnection = 20` (`onym-infra/strfry/strfry.conf`). The client opened **3 REQs per logical subscription** (three filter shapes) × (identities + outstanding invite links) — ≥7 logical subs blows the cap. strfry rejects the overflow with `["CLOSED", subID, …]`, which the client **silently ignored**: a random inbox (dictionary order) went permanently deaf per launch/reconnect. The relay is stock and stays unmodified — this is a client-side adaptation.

## Changes

**`NostrRelayConnection` → explicit lifecycle state machine** (Disconnected → Connecting → Live → Backoff):
- **Generation token** on every (re)connect — receive loop / liveness monitor / failure funnel no-op when superseded; overlapping reconnects can't race
- **Confirmed-live backoff** — `reconnectAttempts` resets on the first received frame, never in `connect()`; escalates across genuinely failing attempts, capped at 30 s
- **Liveness monitor** (replaces the send-only heartbeat): match-nothing REQ probe, relay's immediate EOSE is the pong (verified live against `nostr.onym.app`); no frame within 55 s ⇒ rebuild. Half-open sockets (where `receive()` hangs forever) can no longer stay deaf. Fail-closed on nil/non-running tasks.
- **`CLOSED` handled, not ignored**: first CLOSED for a live sub → re-REQ once; repeat in the same generation → rebuild via the failure funnel. Relay-side rejection is now visible and recoverable.
- **Injectable timings** for test-scale liveness/backoff.

**Single REQ per subscription** carrying all its filters (NIP-01 dedups within a sub): `NostrInboxTransport`'s 3 filter shapes ship in one REQ — 3× fewer relay subscriptions *and* 3× less duplicate decrypt/dispatch work.

**New `LocalWebSocketRelay` harness** (NWListener loopback): REQ recording w/ filter shapes, store-and-replay-on-REQ, EOSE auto-reply (disableable → silent half-open peer), CLOSED injection, hard-drop.

## Tests (8 new; full suite 864 green)

- one REQ carries all 3 filters (regression: one-REQ-per-filter)
- delivery resumes after server drop (passive error path)
- `forceReconnect()` backfills an event stored while away — **the canonical bg→fg missed-message repro**
- liveness monitor heals a silent (half-open) socket; answered probes cause no spurious rebuilds
- CLOSED → exactly one re-REQ, no rebuild; repeated CLOSED → rebuild + REQ replay; unknown-sub CLOSED ignored

## Privacy note

No activity logging. One new log line: a relay-health warning on repeated CLOSED (relay URL only, no sub id / no user data), matching the pre-existing `securityLogger` pattern (oversized frame / invalid event id). Flagging explicitly — happy to drop it if even that is out of policy.

### DIFF
--- a/Sources/OnymIOS/Transport/Nostr/NostrInboxTransport.swift
+++ b/Sources/OnymIOS/Transport/Nostr/NostrInboxTransport.swift
@@ -58,12 +58,19 @@ final class NostrInboxTransport: InboxTransport {
     }
 
     func subscribe(inbox: TransportInboxID) -> AsyncStream<InboundInbox> {
-        AsyncStream<InboundInbox> { continuation in
+        // Per-stream token: this stream's termination may fire *after* a
+        // newer subscription for the same inbox has been installed
+        // (re-subscribe on identity rebalance / pump restart). Cleanup
+        // is token-guarded so the stale termination can't tear down the
+        // fresh subscription — the same race the per-generation subIDs
+        // close at the relay-connection level.
+        let token = UUID()
+        return AsyncStream<InboundInbox> { continuation in
             Task { [state] in
-                await state.subscribe(inbox: inbox, continuation: continuation)
+                await state.subscribe(inbox: inbox, token: token, continuation: continuation)
             }
             continuation.onTermination = { @Sendable [state] _ in
-                Task { await state.unsubscribe(inbox: inbox) }
+                Task { await state.unsubscribe(inbox: inbox, ifToken: token) }
             }
         }
     }
@@ -97,7 +104,15 @@ final class NostrInboxTransport: InboxTransport {
 
     fileprivate actor State {
         private var connections: [URL: NostrRelayConnection] = [:]
-        private var activeSubscriptions: [TransportInboxID: Task<Void, Never>] = [:]
+        /// Live per-inbox subscription: the pump task plus the stream
+        /// token that owns it (see `subscribe`'s token guard).
+        private var activeSubscriptions: [TransportInboxID: (token: UUID, task: Task<Void, Never>)] = [:]
+        /// Monotonic counter for relay subscription IDs. Each subscribe
+        /// gets a unique id so an old stream's asynchronous
+        /// `onTermination` CLOSE can't tear down a freshly opened REQ
+        /// for the same inbox — the exact race `NostrMessageTransport`
+        /// already guards against with its own generation counter.
+        private var subscriptionGeneration: UInt64 = 0
 
         func connect(to endpoints: [TransportEndpoint]) async {
             for endpoint in endpoints {
@@ -110,8 +125,8 @@ final class NostrInboxTransport: InboxTransport {
         }
 
         func disconnect() async {
-            for task in activeSubscriptions.values {
-                task.cancel()
+            for entry in activeSubscriptions.values {
+                entry.task.cancel()
             }
             activeSubscriptions.removeAll()
             for conn in connections.values {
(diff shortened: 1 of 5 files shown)


dev-onym (commented):
REQUEST_CHANGES — the state machine and single-REQ change are sound, but the CLOSED escalation path can hot-loop against the exact relay condition it's meant to fix.

**Blocking**

- `NostrRelayConnection.swift:431` + `:290` — repeated-CLOSED rebuild has effectively no backoff. Each new socket receives a frame (probe EOSE / EVENT / EOSE) which resets `reconnectAttempts = 0`, so a *persistently* rejected subscription (genuine `maxSubsPerConnection` overflow — still reachable at 1 sub/inbox + 1/topic + probe once identities×topics grow) yields: connect → REQ → CLOSED → re-REQ → CLOSED → rebuild after `baseReconnectDelay` (1 s) → forever. That's a ~1 Hz reconnect storm on the relay plus battery drain, and the inbox is still deaf. Needs a CLOSED-specific escalating counter that survives the rebuild (e.g. a separate `closedRebuilds` count feeding the delay, or don't reset `reconnectAttempts` when the previous generation died via CLOSED). `testRepeatedCLOSEDRebuildsConnection` only asserts the first rebuild, so it passes either way — the loop is untested.
- `NostrRelayConnection.swift:287-297` — `task.receive()` can return *after* a newer `connect()` bumped the generation (the loop condition was checked before the suspension). The stale frame then sets `lastActivityAt` and `reconnectAttempts = 0` for the **new** generation, falsely refreshing its liveness window (delaying half-open detection by up to `livenessTimeout`) and weakening backoff. Add `guard generation == connectionGeneration else { return }` immediately after the `receive()` returns, before any state mutation.

**Should fix**

- `NostrRelayConnection.swift:310-317` — `handleConnectionFailure` doesn't bump the generation, so the liveness path and the stale receive loop's `catch` can both pass the guard for the same generation and each do `reconnectAttempts += 1`. Only one wins the `connect()` (thanks to the `webSocketTask == nil` guard), but the counter double-steps. Bump the generation on entry, or set a `rebuildInFlight` flag.
- `NostrRelayConnection.swift:244` / `NostrInboxTransport.swift:176-194` — re-subscribing an inbox reuses the same subID: the new callback is installed at `:241`, then the *old* stream's `onTermination` fires asynchronously and calls `unsubscribe(subscriptionID:)`, removing the new subscription and sending `CLOSE`. Pre-existing, but PR-4's foreground re-subscribe will hit it routinely and reintroduce silent deafness. Suggest guarding termination on identity (generation/token per stream) in this PR since it's the same failure class.
- `closedRetriedSubIDs` (`:61`) never decays within a generation — a benign CLOSED hours apart on a long-lived socket escalates straight to a rebuild. Age it out or clear on EOSE for that sub.

**Nits**

- `NostrRelayConnection.swift:266-272` — `sendREQ` swallows send failures (`try?`); a REQ lost on a wounded socket isn't noticed for up to `livenessTimeout`. Routing the error to the failure funnel would be cheap.
- `__onym_hb` (`:89`) is never `CLOSE`d and consumes a relay sub slot; worth stating the budget is `1 + subs` in the invariant comment.
- `forceReconnect()` (`:164`) has no production caller yet — fine for PR-2 scope, just flagging it ships dead in-app.
- `LocalWebSocketRelay.swift:30,155` — `autoRespondEOSE` is written by tests without the lock but read under it; TSAN race in a `@unchecked Sendable` type. Make it lock-guarded.
- `LocalWebSocketRelay.swift:134` — `guard ... isComplete else { return }` silently terminates the read loop on any non-final/errored frame, so the harness stops recording REQs mid-test. Re-arm `receive(on:)` before the guard to avoid latent flakes.

PR #195, opened 2026-07-29
feat(chat): background-arrived messages are user-visible — deferred scroll + new-messages pill

rinat-enikeev:
Stacked on #194 (reconnect PR-2). Fixes the last mile of the reconnect work: a message backfilled on foreground was **delivered but invisible** — the auto-scroll ran while the app wasn't active yet, UIKit silently dropped it, and the fresh message sat below the fold.

## Changes (`ChatThreadViewController`)

**1. Deferred auto-scroll (the reported bug).**
When new rows want the bottom but a scroll can't stick right now (backgrounded / mid-foreground-transition / view not in a window — exactly where reconnect backfill lands), the controller arms `pendingScrollToBottom` and flushes it on `viewDidAppear` + `didBecomeActiveNotification`. Result: the message received while away is on screen the moment the user is — same as after a relaunch, without the relaunch. `canScrollNow` is an injectable seam so tests drive the lifecycle without a real app transition.

**2. New-messages pill (awareness without rudeness).**
When incoming messages land while the user is scrolled up reading history, auto-scrolling would yank them away — instead a floating **"N new messages"** capsule appears above the composer:
- tap → scrolls to the latest and clears
- manually reaching the bottom clears it too
- count accumulates across updates until seen
- localized (en/ru with proper plural variations), `accessibilityIdentifier: chat.newMessagesPill`, built from `OnymTokens`

Own outgoing sends keep today's behavior (no pill; scroll only when near bottom).

## Tests (7 new; full suite 871 green)

- pill hidden initially; appears with count while scrolled up **and does not yank the scroll position**
- count accumulates across updates; clears on reaching bottom; tap lands at latest and hides
- near-bottom arrivals auto-scroll with no pill
- **the field repro**: arrival while unscrollable arms the deferred scroll; `didBecomeActive` flushes it onto screen

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -34,6 +34,59 @@
         }
       }
     },
+    "%lld new messages": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "variations": {
+            "plural": {
+              "one": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld new message"
+                }
+              },
+              "other": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld new messages"
+                }
+              }
+            }
+          }
+        },
+        "ru": {
+          "variations": {
+            "plural": {
+              "one": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld новое сообщение"
+                }
+              },
+              "few": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld новых сообщения"
+                }
+              },
+              "many": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld новых сообщений"
+                }
+              },
+              "other": {
+                "stringUnit": {
+                  "state": "translated",
+                  "value": "%lld новых сообщений"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "—": {
       "localizations": {
         "en": {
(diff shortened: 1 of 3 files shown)


ux-onym (commented):
**COMMENT**

- `Sources/OnymIOS/Chats/ChatThreadViewController.swift:197` — `flushPendingScrollToBottom()` clears `pendingScrollToBottom` and scrolls unconditionally, unlike the arm site (`:343`) which gates on `canScrollNow()`. If `didBecomeActiveNotification` (app-wide) fires while *this* VC's view isn't actually in a window (e.g. another thread pushed on top, or an inactive scene on iPad multi-window), the scroll silently no-ops but the flag is already cleared — so `viewDidAppear`'s guard bails when the user does return to the thread, and the backfilled message stays below the fold. Suggest: `guard pendingScrollToBottom, canScrollNow() else { return }`.
- `Tests/OnymIOSTests/ChatThreadViewControllerTests.swift:559` — the new field-repro test forces `canScrollNow = { true }` *before* posting the notification, so it can't catch the above gap; a variant that posts the notification while `canScrollNow` is still `false` would.

Otherwise the change is solid: pill/scroll branches are correctly mutually exclusive on `wasNearBottom`, the incoming-count diffing (`msg.direction == .incoming && messagesByID[msg.id] == nil`, computed pre-reassignment) is correct, localization keys match the `String(localized:)` interpolation, and the 7 new tests cover the pill lifecycle well.

PR #198, opened 2026-07-29
feat(inbox): dispatch lanes — invitation backlog can't starve live chat (reconnect PR-5)

rinat-enikeev:
Stacked on #197. Design doc §5.4 (F6): the fan-out's strictly serial `for await → dispatch` meant every replayed invitation — each potentially a relayer chain read — stalled live chat delivery behind it. On the invitation-spammed device this read as "messages don't arrive while the app is open."

## Changes

- **Fast lane (inline)**: chat messages + receipts — pure local work, dispatched without ever waiting on network-bound handlers.
- **Slow lane** (`SerialDispatchLane`: FIFO chained-Task actor, non-blocking enqueue): offers, refresh requests, roster announcements, avatar/name updates, group-materializing invitations, undecryptable fall-throughs. Relative order within the group-state family is preserved.
- **Lane-split race handled**: a chat message can now beat the slow-lane work that materializes its group. On group-not-found it retries exactly once, queued *behind* the slow lane's current queue — a materializing invitation in flight completes first and the message lands; a genuinely stale delivery still drops (no retry loops).
- Existing dispatcher tests migrate to `dispatchAndDrain` (dispatch + slow-lane barrier) to keep their synchronous-effects contract explicit.

## Tests (6 new; full suite green)

Lane FIFO; enqueue never blocks on a running job; `drain` covers chained jobs; **chat persists in <1 s while a 2 s group-state job occupies the slow lane** (the starvation repro); chat-before-group lands via the retry; unknown-group chat drops after one retry.

### DIFF
--- a/Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift
+++ b/Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift
@@ -78,6 +78,13 @@ struct IncomingMessageDispatcher: Sendable {
     /// raises a message to `.read` when this device also sends read
     /// receipts. Defaulted to `true` (the shipping default).
     var readReceiptsEnabled: @Sendable () -> Bool = { true }
+    /// The slow lane: FIFO, serial, non-blocking-to-enqueue. Every
+    /// invitation / group-state payload runs here so a backlog of
+    /// chain-verifying handlers can never starve live chat (F6). One
+    /// lane per dispatcher instance; the struct's copies share it
+    /// (actor reference), preserving FIFO across the app's single
+    /// dispatcher.
+    var slowLane = SerialDispatchLane()
 
     func dispatch(
         messageID: String,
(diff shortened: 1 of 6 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the lane split is well-built, but it silently breaks the durable-dedup at-least-once guarantee added in PR-3.

- **`Sources/OnymIOS/Inbox/InboxFanoutInteractor.swift:167` — `markSeen` now fires before slow-lane work runs.** The loop marks an event seen right after `dispatch` returns, and its comment ("Mark AFTER dispatch: a crash mid-processing means re-delivery … never loss") assumed dispatch had applied all effects. Every slow payload now returns from `dispatch` while only *enqueued*. Kill the app (or lose the process) with a 200-invitation backlog queued and each of those events is durably marked seen but never processed — the invitation is gone for good, no replay recovers it. Fix: mark seen from inside the lane job (pass a completion into `enqueue`, or have `dispatch` return a handle), not at the call site.
- **`Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:173` — roster removals can now land *after* chat from the removed member.** `applyAnnouncement` is on the slow lane, `persistChatMessage` is inline, so during the exact backlog this PR targets there's a multi-minute window where a member the announcement removes still passes the `memberProfiles` / insider-spoof gate at line 817-827. Pre-PR the serial loop closed that window. Worth an explicit decision + comment, or gating chat for a group behind that group's pending slow-lane work.
- **`Sources/OnymIOS/Inbox/SerialDispatchLane.swift:20` — unbounded queue, no cancellation, no backpressure.** The stream loop now drains as fast as the relayer delivers, so an invitation flood accumulates unbounded Tasks each retaining a payload `Data` — the fan-out's serial loop used to be the backpressure. Jobs are unstructured Tasks, so `ActiveSubscriptions.apply`'s `task.cancel()` (line 136) no longer stops in-flight invitation work. Consider a depth cap + `Task.isCancelled` check at job start.
- **`Sources/OnymIOS/OnymIOSApp.swift:603` — one lane per dispatcher *instance*, not per app.** The dispatcher is built inside a SwiftUI `.task`; on restart (view reappear/reconnect) a fresh `SerialDispatchLane` is created while the old lane's uncancelled jobs are still running, so two "serial" lanes mutate group state concurrently. The property comment at `IncomingMessageDispatcher.swift:87` claims "the app's single dispatcher" — either hoist the lane to a shared singleton or soften the claim.
- `Sources/OnymIOS/Inbox/SerialDispatchLane.swift:15` — `tail` is never nil'd on completion, so the last Task is retained for the dispatcher's lifetime. Cosmetic, but trivial to clear when the finished task is still `tail`.
- Nit — `Tests/OnymIOSTests/DispatchLanesTests.swift:96` asserts wall-clock `< 1.0 s` against a 2 s stall; fine locally, but it's a real-time race on a loaded CI runner. The stronger invariant (chat persisted before the stalled job completes) is already asserted at line 101, so the timing bound could be dropped or widened.
- Good: retry is correctly bounded by `allowRetry: false` (`IncomingMessageDispatcher.swift:800`), `drain()`'s count-snapshot loop correctly covers jobs chained by running jobs, and per-payload `senderKey` hoisting out of the closures avoids capturing the envelope. The `dispatchAndDrain` migration is the right way to keep the old tests' contract explicit.

Issue #111, opened 2026-05-09
[PROPOSAL] Auto-rotate invite QR code every 24 hours and after each member joins


innovesss:
# [PROPOSAL] Auto-rotate invite QR code every 24 hours and after each member joins

## Summary

Currently, the invite QR code (Invite Key) remains static once generated. For improved privacy, the QR code should automatically rotate on two triggers:

1. **Time-based:** Every **24 hours** the invite link and QR code are regenerated, invalidating the previous one.
2. **Event-based:** After **each new member joins** the group, the current invite link is invalidated and a new one is generated.

This ensures that old invite links cannot be reused or shared beyond their intended scope, reducing the risk of unauthorized access to the group.

## Motivation

A static invite link can be forwarded, screenshotted, or leaked — allowing unintended users to join the group at any time. By rotating the invite key periodically and after each join event, the exposure window is minimized:

- A leaked link becomes useless after 24 hours or after the intended recipient joins.
- Group admins don't need to manually revoke and regenerate links.
- Aligns with the app's privacy-first philosophy (end-to-end encryption, on-chain anchoring).

## Proposed behavior

| Trigger                  | Action                                                         |
|--------------------------|----------------------------------------------------------------|
| **24 hours elapsed**     | Current invite link/QR is invalidated, a new one is generated  |
| **New member joins**     | Current invite link/QR is invalidated, a new one is generated  |
| **Manual regeneration**  | Admin can force-regenerate at any time (existing or new flow)  |

- Anyone attempting to use an expired or rotated link should see a clear error message (e.g., "This invite link has expired. Ask the group admin for a new one.").
- The QR code displayed in the **Invite Key** screen should always reflect the latest active link.

## Impact

- **Priority:** Medium
- Strengthens group privacy without adding friction for admins. Rotation is automatic — no manual action required from the user.

rinat-enikeev:
@ux-onym wdyt on this user experience change?

lead-onym:
**Routed to @ux-onym.**

_

@ux-onym will respond shortly. Cross-mention me (`@lead-onym`) on this thread to override and re-route.

rinat-enikeev:
@ux-onym wdyt on this user experience change?

lead-onym:
**Routed to @ux-onym.**

_

@ux-onym will respond shortly. Cross-mention me (`@lead-onym`) on this thread to override and re-route.

rinat-enikeev:
[...]

PR #206, opened 2026-07-30
fix(transport): bind relay receive loop to its socket + honor disconnect

rinat-enikeev:
Started as one crash fix, grew into a transport-reliability series as each fix exposed the next issue on-device. One commit per problem:

## 1. EXC_BAD_ACCESS in handleMessage (task-frame corruption)
Fault address 0x10 = null array-buffer + count offset: the async task's heap frame was corrupted by cancel-during-receive churn in the reconnect path racing CFNetwork's receive completion (same UAF family as the sendPing crash documented in-file). Amplified by three reconnect bugs: backoff pinned at 1s (`connect()` reset the counter), stale receive loops cancelling a successor's socket, and `disconnect()` not surviving a backoff sleep. Fix: `receiveLoop(task:)` is bound to its socket (one loop per socket, no foreign cancels), sticky `desiredConnected`, backoff resets on successful receive only.

## 2. Device stops receiving live events (zombie sockets)
Backgrounding/lock kills the NAT path while the socket looks locally alive — `receive()` hangs forever, the old fire-and-forget heartbeat couldn't tell. Heartbeat is now probe-and-verify: track `lastFrameAt`, send a REQ that can't match anything (NIP-01 obliges an immediate EOSE), no frame within 10s ⇒ cancel the socket and recover through the normal reconnect + subscription replay.

## 3. Temporary DEBUG diagnostics
`TransportLog` (print-based — the scheme's `OS_ACTIVITY_MODE=disable` mutes os_log) covering connection lifecycle, REQ/CLOSE, every inbound frame, publish/OK outcomes, probe verdicts. ~~Remove after the dust settles.~~ **Removed** in the review round — it printed whole REQ frames including inbox pubkeys.

## 4. Relay rejected most subscriptions ("too many concurrent REQs")
3 filter shapes × ~25 inboxes = ~75 REQs vs strfry's ~20/connection cap; rejected inboxes had no live subscription (the "messages only on app start" symptom). One REQ now carries all filters per inbox (3× fewer slots) and the relay dedups overlapping matches (kills 2-3× duplicate delivery).

## 5. Receipt storm on every launch
Relay reconnects replay the full inbox; `persistChatMessage` acked every replayed message unconditionally — a receipt publish per historical message per launch, each becoming a new stored event on the other side (snowballing replays). Receipts now fire only for newly inserted messages.

## 6. Stale keychain state inflating the subscription count
The ~25 inboxes were mostly intro-key inboxes (one per invite link ever shared, no expiry) plus keychain items surviving reinstall. Fixes: 24-hour TTL on intro keys, matching Android per #111 (filtered at the single load point; the load that first sees an expired row compacts the blob); launch-time `pruneOwners` GC for intro keys whose owner identity no longer exists; fresh-install reconciliation in `IdentityRepository` (no install marker AND no persisted selection ⇒ orphaned identity items are **quarantined** into a separate keychain namespace, never destroyed — app updates are detected via the existing selection and only stamped).

## 7. Review follow-ups (#4822757300, #4822779117)
- `currentIdentities()` now throws and calls `ensureLoaded()` first, so the intro-key prune racing `bootstrap()` can no longer see an empty pre-load snapshot and wipe every pending invite-link inbox; the prune site skips on load failure.
- `ensureLoaded()` marks the repo loaded only after a fully-successful load (a thrown keychain error no longer strands an empty "loaded" repo whose retried `bootstrap()` mints a duplicate identity).
- `reconcileFreshInstall()` quarantines instead of `wipeAll()` — both of its signals are UserDefaults (weaker file protection than the identity items), so a wrong verdict must cost a hidden identity, not destroyed key material.
- Inbox subscription ids carry a monotonic generation (a resubscribe's old-stream `onTermination` no longer CLOSEs the replacement REQ) and use a hashed inbox (under NIP-01's 64-char cap).
- The `__hb` probe REQ is CLOSEd after the probe window; `probeLiveness()` polls in 0.5s slices and exits on the first frame.
- `insertOrUpdate` returns `.inserted`/`.updated`/`.failed` — a persistence failure is no longer indistinguishable from a replayed duplicate, and stays un-acked so the relay copy retries on the next replay.

## 8. Review follow-ups, round 2 (#4822938510)
- Delivered receipts regained a retry path: `PersistedMessage` carries a `deliveredAckSent` latch, `ChatReceiptSender` reports whether the transport accepted the publish, and the dispatcher acks on `.inserted` *or* an un-latched replay — a receipt lost to an unreachable relay retries on the next inbox replay, while an acked message never re-acks (receipt-storm fix intact; both directions tested).
- `reconcileFreshInstall()` is gated on an injected `ProtectedDataAvailability` seam (`UIApplication.isProtectedDataAvailable`): pre-first-unlock launches defer the verdict instead of misreading unreadable UserDefaults as "fresh install".
- `teardownDeadSocket(_:)` cancels only the socket the probe judged, guarded by `webSocketTask === task`.
- TTL compaction publishes the affected owners' streams so a live session's intro pump releases expired REQ slots; the launch prune skips an empty identity set (post-quarantine, pruning would destroy intro keys a recovery flow could still revive).
- The update branch of `insertOrUpdate` mirrors the insert branch's `do/catch` — a failed save rolls back and reports `.failed`.
- Tests in this PR for the secret-hiding paths (quarantine branches incl. deferred verdict, TTL invisibility + at-rest compaction, `pruneOwners`), and identity tests now inject `installMarker`/`protectedData` so they're order-independent.
- Not changed: the `__hb` probe's transient extra REQ slot — an over-limit REQ is answered (CLOSED/NOTICE) and any reply frame satisfies the probe, so a rejected probe still proves liveness; assumption documented at the probe.

Builds clean (`xcodebuild -scheme OnymIOS`, iphonesimulator); full `OnymIOSTests` suite passes (870 tests). Per review convention, the remaining tests land in the final PR of the stack.

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -33,6 +33,8 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<true/>
 	<key>NSCameraUsageDescription</key>
 	<string>Onym uses the camera to scan invite-key QR codes when you add someone to a group.</string>
 	<key>NSMicrophoneUsageDescription</key>
--- a/Info.plist
+++ b/Info.plist
@@ -33,15 +33,6 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
-	<!-- Export compliance. Onym uses non-exempt encryption (end-to-end
-	     message/media encryption with AES-256-GCM + X25519, beyond
-	     HTTPS), so this is `true`. The app ships only industry-standard,
-	     publicly documented algorithms — it qualifies for US License
-	     Exception ENC (ECCN 5D992.c, mass market) via a self-classification
-	     report, and needs a French encryption declaration to be sold in
-	     France. See docs/compliance/. -->
-	<key>ITSAppUsesNonExemptEncryption</key>
-	<true/>
 	<key>NSCameraUsageDescription</key>
 	<string>Onym uses the camera to scan invite-key QR codes when you add someone to a group.</string>
 	<key>NSMicrophoneUsageDescription</key>
--- a/Sources/OnymIOS/Chats/ChatReceiptSender.swift
+++ b/Sources/OnymIOS/Chats/ChatReceiptSender.swift
@@ -5,40 +5,46 @@ import Foundation
 /// Injected into the dispatcher (delivered receipts on receive) and the
 /// chat thread (read receipts on view). Best-effort: a failed seal or
 /// send is swallowed — a missing receipt only costs a check mark, never
-/// correctness.
+/// correctness. Returns whether the transport accepted the publish so
+/// the dispatcher can latch `deliveredAckSent` only on success and
+/// retry the receipt on the next inbox replay otherwise.
 protocol ChatReceiptSending: Sendable {
+    @discardableResult
     func send(
         kind: ChatReceiptPayload.Kind,
         messageIDs: [UUID],
         groupID: Data,
         to recipientInboxKey: Data
-    ) async
+    ) async -> Bool
 }
 
 /// Default no-op so the dispatcher's many test constructions don't have
 /// to thread a receipt sender they don't exercise (same posture as
-/// `pendingInvites` / `groupStateRefresher`).
+/// `pendingInvites` / `groupStateRefresher`). Reports `false` — nothing
+/// was sent, so nothing should be latched as acked.
 struct NoopChatReceiptSender: ChatReceiptSending {
+    @discardableResult
     func send(
         kind: ChatReceiptPayload.Kind,
         messageIDs: [UUID],
         groupID: Data,
         to recipientInboxKey: Data
-    ) async {}
+    ) async -> Bool { false }
 }
 
 struct ChatReceiptSender: ChatReceiptSending {
     let identity: IdentityRepository
     let inboxTransport: any InboxTransport
 
+    @discardableResult
     func send(
         kind: ChatReceiptPayload.Kind,
         messageIDs: [UUID],
         groupID: Data,
         to recipientInboxKey: Data
-    ) async {
-        guard !messageIDs.isEmpty else { return }
-        guard let active = await identity.currentIdentity() else { return }
+    ) async -> Bool {
+        guard !messageIDs.isEmpty else { return false }
+        guard let active = await identity.currentIdentity() else { return false }
         let myBlsHex = active.blsPublicKey.map { String(format: "%02x", $0) }.joined()
         let payload = ChatReceiptPayload(
             version: 1,
@@ -49,9 +55,9 @@ struct ChatReceiptSender: ChatReceiptSending {
         )
         guard let bytes = try? JSONEncoder().encode(payload),
               let sealed = try? await identity.sealInvitation(payload: bytes, to: recipientInboxKey)
-        else { return }
+        else { return false }
         let tag = TransportInboxID(rawValue: Self.inboxTag(from: recipientInboxKey))
-        _ = try? await inboxTransport.send(sealed, to: tag)
+        return (try? await inboxTransport.send(sealed, to: tag)) != nil
     }
 
     /// Same derivation as `SendMessageInteractor` / `IntroInboxPump`.
--- a/Sources/OnymIOS/Chats/MessageRepository.swift
+++ b/Sources/OnymIOS/Chats/MessageRepository.swift
@@ -47,10 +47,10 @@ actor MessageRepository {
     /// to `MessageStore.insertOrUpdate`). Receive-side replays and
     /// outgoing status flips both flow through here.
     @discardableResult
-    func insert(_ message: ChatMessage) async -> Bool {
-        let inserted = await store.insertOrUpdate(message)
+    func insert(_ message: ChatMessage) async -> MessageInsertOutcome {
+        let outcome = await store.insertOrUpdate(message)
         await refresh(ThreadKey(groupID: message.groupID, owner: message.ownerIdentityID))
-        return inserted
+        return outcome
     }
 
     /// Flip an outgoing message's status (pending → sent / failed).
--- a/Sources/OnymIOS/Chats/MessageRepository.swift
+++ b/Sources/OnymIOS/Chats/MessageRepository.swift
@@ -53,6 +53,25 @@ actor MessageRepository {
         return outcome
     }
 
+    /// Whether the delivered receipt for one incoming row still needs
+    /// to be sent — `true` until `markDeliveredAckSent` latches it.
+    /// See `MessageStore.needsDeliveredAck`.
+    func needsDeliveredAck(id: UUID, owner: IdentityID) async -> Bool {
+        await store.needsDeliveredAck(
+            id: id,
+            ownerIDString: owner.rawValue.uuidString
+        )
+    }
+
+    /// Latch one row's delivered receipt as successfully sent. No
+    /// snapshot refresh — the flag never renders.
+    func markDeliveredAckSent(id: UUID, owner: IdentityID) async {
+        await store.markDeliveredAckSent(
+            id: id,
+            ownerIDString: owner.rawValue.uuidString
+        )
+    }
+
     /// Flip an outgoing message's status (pending → sent / failed).
     /// Hot path for the send pipeline so we don't round-trip the
     /// whole row through the encryption boundary. `failureReason`
--- a/Sources/OnymIOS/Chats/MessageStore.swift
+++ b/Sources/OnymIOS/Chats/MessageStore.swift
@@ -1,5 +1,23 @@
 import Foundation
 
+/// Result of `MessageStore.insertOrUpdate`. `.updated` (the benign
+/// replay / status-flip path) and `.failed` (nothing persisted) both
+/// mean "not a new row", but callers gating side effects must tell
+/// them apart: the delivered receipt is skipped for a replayed
+/// duplicate yet must ALSO be withheld on failure so the sender's
+/// relay copy is still replayed — and re-attempted — on the next
+/// reconnect, rather than acked for a message this device never
+/// stored.
+enum MessageInsertOutcome: Sendable, Equatable {
+    /// New row persisted.
+    case inserted
+    /// A row with the same `(id, owner)` already existed and was
+    /// updated in place.
+    case updated
+    /// Nothing was persisted — encode or save failed.
+    case failed
+}
+
 /// Persistence seam for chat messages. Async surface mirrors
 /// `GroupStore` so a concrete implementation can serialise writes on
 /// its own queue without forcing callers onto a specific actor.
@@ -33,11 +51,10 @@ protocol MessageStore: Sendable {
 
     /// Idempotent on the composite `(message.id, ownerIdentityID)`.
     /// New row → insert. Same id+owner → update in place; used both
-    /// for receive-side replays (no-op result on the second insert)
+    /// for receive-side replays (`.updated` on the second insert)
     /// and outgoing status transitions (pending → sent / failed).
-    /// Returns `true` on insert, `false` on update.
     @discardableResult
-    func insertOrUpdate(_ message: ChatMessage) async -> Bool
+    func insertOrUpdate(_ message: ChatMessage) async -> MessageInsertOutcome
 
     /// Flip just the status column (and the failure-reason column that
     /// travels with it — non-nil when flipping to `.failed`, nil
--- a/Sources/OnymIOS/Chats/MessageStore.swift
+++ b/Sources/OnymIOS/Chats/MessageStore.swift
@@ -56,6 +56,19 @@ protocol MessageStore: Sendable {
     @discardableResult
     func insertOrUpdate(_ message: ChatMessage) async -> MessageInsertOutcome
 
+    /// `true` while the row exists and its delivered receipt hasn't
+    /// been successfully handed to the transport yet. The dispatcher
+    /// consults this on replayed (`.updated`) rows so a receipt that
+    /// failed to send (relay unreachable at receive time) is retried
+    /// on the next inbox replay instead of being lost forever.
+    /// `false` for unknown rows.
+    func needsDeliveredAck(id: UUID, ownerIDString: String) async -> Bool
+
+    /// Latch the delivered receipt as sent for one row. Called only
+    /// after the transport accepted the receipt publish. No-op for
+    /// unknown rows.
+    func markDeliveredAckSent(id: UUID, ownerIDString: String) async
+
     /// Flip just the status column (and the failure-reason column that
     /// travels with it — non-nil when flipping to `.failed`, nil
     /// otherwise so a retry's pending flip clears the stale reason).
--- a/Sources/OnymIOS/Chats/PersistedMessage.swift
+++ b/Sources/OnymIOS/Chats/PersistedMessage.swift
@@ -55,6 +55,15 @@ final class PersistedMessage {
     /// optional, so adding it is a SwiftData lightweight migration
     /// over the existing store — same shape as `replyToMessageIDString`.
     var failureReasonRaw: String?
+    /// `true` once the delivered receipt for this incoming row was
+    /// successfully handed to the transport. The receipt send is
+    /// fire-and-forget with no outbox, so relay-reconnect replays are
+    /// the only retry — the dispatcher re-acks a replayed row until
+    /// this flips, then stays silent (re-acking every replay snowballed
+    /// receipt storms; see `persistChatMessage`). Plain (one bit) and
+    /// optional for lightweight migration; nil reads as `false`.
+    /// Meaningless for outgoing rows.
+    var deliveredAckSent: Bool?
 
     var encryptedSenderBlsPubkeyHex: Data
     var encryptedBody: Data
@@ -89,6 +98,7 @@ final class PersistedMessage {
         groupTypeRaw: String,
         replyToMessageIDString: String?,
         failureReasonRaw: String?,
+        deliveredAckSent: Bool? = nil,
         encryptedSenderBlsPubkeyHex: Data,
         encryptedBody: Data,
         encryptedAttachmentJSON: Data? = nil,
@@ -105,6 +115,7 @@ final class PersistedMessage {
         self.groupTypeRaw = groupTypeRaw
         self.replyToMessageIDString = replyToMessageIDString
         self.failureReasonRaw = failureReasonRaw
+        self.deliveredAckSent = deliveredAckSent
         self.encryptedSenderBlsPubkeyHex = encryptedSenderBlsPubkeyHex
         self.encryptedBody = encryptedBody
         self.encryptedAttachmentJSON = encryptedAttachmentJSON
--- a/Sources/OnymIOS/Chats/SwiftDataMessageStore.swift
+++ b/Sources/OnymIOS/Chats/SwiftDataMessageStore.swift
@@ -156,7 +156,15 @@ actor SwiftDataMessageStore: MessageStore {
             existing.encryptedVideoAttachmentJSON = encoded.encryptedVideoAttachmentJSON
             existing.encryptedAlbumJSON = encoded.encryptedAlbumJSON
             existing.encryptedVoiceAttachmentJSON = encoded.encryptedVoiceAttachmentJSON
-            try? context.save()
+            do {
+                try context.save()
+            } catch {
+                // Discard the half-mutated row so the context reflects
+                // what's actually on disk; `.failed` keeps the enum's
+                // "nothing was persisted" promise on this path too.
+                context.rollback()
+                return .failed
+            }
             return .updated
         }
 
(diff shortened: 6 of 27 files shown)


dev-onym (commented):
**COMMENT** — the three named races are correctly fixed; four non-blocking issues below.

- `NostrRelayConnection.swift:214` — backoff can never reach 120s: the exponent is clamped at `6`, so the ceiling is `1 * 2^6 = 64s` and `maxReconnectDelay` (120) is unreachable dead config. The PR body's "1s→2s→…→120s" is wrong. Use `min(attempts - 1, 7)` (128 → clamped to 120) or drop the constant to 64.

- `NostrRelayConnection.swift:196` — behavior regression from moving the reset: `reconnectAttempts` now resets *only* on a delivered frame (or `disconnect()`). A relay that accepts the socket but sends nothing — the normal case right after `NostrMessageTransport.connect(to:)`, which connects before any `subscribe`, so there's no `EOSE` either — escalates backoff monotonically across *successful* connects, ending at 64s. Consider also resetting once a socket has stayed up past some threshold, or on the first successful `send`.

- `NostrRelayConnection.swift:219` — `webSocketTask == nil` tests socket state, not loop ownership, so reconnect duty isn't uniquely assigned. Loop A parks in backoff; an external `connect()` spawns loop B; B fails fast and nils the socket; A wakes, sees `nil`, and calls `connect()` — firing a reconnect while B is still serving its own (longer) backoff. B then bails at the same guard, so there's no double-socket, but the effective delay is the *shorter* of two overlapping timers and escalation isn't monotonic. A monotonic generation counter checked instead of pointer identity would make ownership exact.

- `NostrRelayConnection.swift:52` — the loop's `Task {}` is unstructured, unretained, and captures `self` strongly. `disconnect()` cancels `pingTask` but nothing cancels this, so a loop parked in `Task.sleep` keeps the actor alive for up to 64s after `NostrMessageTransport.disconnect()` has already dropped it from `connections` (:113). Bounded, not a leak, but storing the handle and cancelling it in `disconnect()` (and when the catch path hands off) would make teardown prompt and make `desiredConnected` a belt-and-braces check rather than the only stop.

Two notes on the diagnosis rather than the code:

- Nothing here touches `handleMessage`, so the link to the reported `array[1]` fault is inferential — plausible (fewer cancel-during-receive windows) but unverified. Worth saying so in the commit message rather than asserting the crash is fixed.
- Adjacent, pre-existing, and in the same continuation machinery as the crash report: `publishAndAwaitOK` (:107) overwrites `pendingOKContinuations[eventID]` if the same event id is published concurrently, leaking the first continuation (permanent hang + runtime diagnostic). Worth a look while this area is warm.

Tests deferred to the stack's final PR per stated convention — fine, but the backoff-escalation and stale-loop-bails paths are exactly what needs coverage there.

ux-onym (commented):
**REQUEST_CHANGES**

- `Sources/OnymIOS/OnymIOSApp.swift:642` — `pruneOwners` runs in a `.task` block that races the separate `.task` calling `bootstrap()` (line 489/497). SwiftUI runs multiple `.task` modifiers on one view as independent concurrent child tasks with no ordering guarantee. If prune wins the race, `currentIdentities()` returns `[]` before identities load, and `pruneOwners(keeping: [])` deletes every intro-key entry for every owner — wiping all pending invite-link inboxes, not just orphaned ones.
- `Sources/OnymIOS/Identity/IdentityRepository.swift:225` — root cause: `currentIdentities()` reads `orderedIDs` directly, skipping the `try ensureLoaded()` call every other accessor in this actor makes. Fix: add `try? ensureLoaded()` (or make it throwing/async-safe) so this can't return an empty snapshot before load.

Everything else checked out:
- `receiveLoop(task:)` binding, `desiredConnected` sticky-disconnect, and backoff-reset-on-receive-only logic (`NostrRelayConnection.swift:68-290`) are internally consistent and match the crash/hang analysis in the PR description — actor isolation makes the "stale loop" guard correctly handle the reentrancy window it targets.
- The probe heartbeat's fake event id is exactly 64 hex chars (valid NIP-01 id shape); reusing subscription id `"__hb"` on each probe replaces rather than accumulates relay-side REQ slots.
- Single-REQ-multi-filter change (`NostrInboxTransport.swift`, `NostrRelayConnection.swift:169-198`) is a straightforward, correct refactor.
- `guard inserted else { return }` in `IncomingMessageDispatcher.swift:827` relies on `MessageRepository.insert` already returning "was newly inserted" (pre-existing API, unchanged by this PR) — safe, and nothing after the guard needs to run for a duplicate.
- `IntroKeyStore` TTL filtering and `KeychainIntroKeyStore.pruneOwners` logic are correct in isolation.
- `reconcileFreshInstall()`'s UserDefaults-survives-vs-Keychain-survives heuristic is a deliberate, documented tradeoff (accepted false-positive risk on unusual backup-restore permutations) — not a bug, just worth the team's eyes-open sign-off.

dev-onym (commented):
**REQUEST_CHANGES** — transport fixes are sound, but the identity wipe is an irreversible-data-loss risk and one "stops receiving" race survives.

- `Sources/OnymIOS/Identity/IdentityRepository.swift:571` — `reconcileFreshInstall()` destroys real keys on a *soft* signal. Both inputs (install marker, selection) live in UserDefaults, which is `NSFileProtectionCompleteUntilFirstUserAuthentication`; identity items are `AfterFirstUnlockThisDeviceOnly`. Any launch where UserDefaults reads as empty but the keychain is readable (pre-first-unlock background launch, defaults DB corruption/reset, a future migration of the selection key) ⇒ `keychain.wipeAll()` and the user's identities are gone, recoverable only if they saved a mnemonic. Suggest: never destroy silently — ignore/quarantine orphans, or require an explicit user confirmation before wiping a non-empty keychain.
- `Sources/OnymIOS/Identity/IdentityRepository.swift:534` — `loaded = true` is set *before* the new throwing `reconcileFreshInstall()`. If `wipeAll()` throws (`keychainRead`/`keychainDelete`), the repo is permanently "loaded" with an empty cache and no marker; a retried `bootstrap()` (line 63) mints a brand-new identity while the real items still sit in the keychain. Set `loaded = true` only after the load succeeds.
- `Sources/OnymIOS/Transport/Nostr/NostrInboxTransport.swift:179` — subID is now the fixed `inbox-<pubkey>`, so a resubscribe races itself: `activeSubscriptions[inbox]?.cancel()` (line 178) terminates the old `AsyncStream`, whose `onTermination` (NostrRelayConnection.swift:191) calls `unsubscribe(subscriptionID:)` — which removes the *new* registration and sends `CLOSE` for it. The inbox then goes silent until the next resubscribe: the same symptom this PR is chasing. `NostrMessageTransport.swift:150` already solves this with `subscriptionGeneration` in the subID; mirror it here.
- `Sources/OnymIOS/Transport/Nostr/NostrRelayConnection.swift:325` — the `__hb` probe REQ is never CLOSEd, so it permanently holds one of the ~20 REQ slots per connection — the exact resource commit #4 is conserving (the old heartbeat sent `CLOSE`). Send `["CLOSE","__hb"]` after the probe window.
- `Sources/OnymIOS/Transport/Nostr/NostrRelayConnection.swift:331` — `probeLiveness()` always sleeps the full 10s even when frames are already arriving, and detection latency is up to `pingInterval + probeTimeout` (40s). Polling `lastFrameAt` in short slices would exit early and detect faster.
- `Sources/OnymIOS/Transport/Nostr/NostrInboxTransport.swift:179` — `"inbox-" + 64-char hex` = 70 chars, over NIP-01's 64-char subscription-id cap (strfry's default `maxSubidSize` 71 happens to tolerate it; stricter relays reject the REQ outright). Consider a truncated hash of the inbox.
- `Sources/OnymIOS/Group/KeychainIntroKeyStore.swift:173` — TTL is filtered at read only, so expired intro *private keys* stay at rest indefinitely if no write ever happens; the "compacted on the next write" comment overstates it. Cheap fix: compact when `loadAll()` sees expiry. Also, PR body says 30-day TTL; code (and d7af48e) is 24h.
- `Sources/OnymIOS/Group/IntroKeyStore.swift:69` — a defaulted protocol requirement means any real store that forgets `pruneOwners` silently skips GC with no compile error; consider dropping the default and implementing it in the test doubles.
- `Sources/OnymIOS/Chats/SwiftDataMessageStore.swift:133` — `insertOrUpdate` returns `false` for encode failure as well as for duplicates, so with the new `guard inserted` (IncomingMessageDispatcher.swift:825) a persistence failure now also suppresses the delivered receipt: message lost locally *and* sender never re-notified. Distinguishing "existed" from "failed" would be more honest.
- `Sources/OnymIOS/Transport/Nostr/NostrRelayConnection.swift:6` — `TransportLog` is correctly DEBUG-gated with `@autoclosure` (zero release cost), but it prints whole REQ frames incl. inbox pubkeys; fine for local debug, worth a tracked removal issue rather than "remove after the dust settles."
- Positive: `receiveLoop(task:)` binding, `desiredConnected` re-check after backoff, and resetting `reconnectAttempts` only on a delivered frame all look correct, and the post-sleep `webSocketTask == nil` guard correctly yields to an externally-created successor socket.

ux-onym (commented):
**APPROVE**

- Transport core (`NostrRelayConnection.swift:206-246`) — `receiveLoop(task:)` binding via `webSocketTask === task`, `desiredConnected` sticky-intent, and backoff-reset-on-receive-only are correctly implemented. Because `ensureLoaded()`/`connect()`/`disconnect()` have no `await` between the guard check and mutation, the actor gives atomic execution per call — no interleaving hazards found in the reconnect/backoff/probe state machine.
- `probeLiveness()` (`NostrRelayConnection.swift:285-306`) correctly pairs REQ+CLOSE via `defer`, exits early on any frame (not just EOSE for `__hb`), and self-neutralizes if the socket is replaced or cancelled mid-poll. Sound.
- `IdentityRepository.reconcileFreshInstall()` (`IdentityRepository.swift:592-598`) + `quarantineAll()` (`IdentityKeychainStore.swift:189-219`): logic matches the documented truth table (marker/selection → normal/update/fresh-install). `ensureLoaded()` correctly resets `loaded`/caches on retry and only flips `loaded = true` after a full success (`IdentityRepository.swift:537-568`).
- `currentIdentities() throws` fix genuinely closes the prune/bootstrap race (`IdentitiesFlow.swift:56`, `InboxFanoutInteractor.swift:47`, `OnymIOSApp.swift:646-648`) — since `ensureLoaded()` is single-shot and atomic per actor turn, whichever caller (bootstrap vs. prune task) reaches the actor first fully loads existing keychain state before returning, so the prune can no longer observe a spuriously-empty snapshot for identities that already exist on disk.
- `SwiftDataMessageStore.insertOrUpdate` failure rollback (`SwiftDataMessageStore.swift:157-172`) correctly voids the pending in-memory insert via `context.delete(encoded)` when `save()` throws, and `IncomingMessageDispatcher.swift:818-830` correctly gates the delivered-receipt only on `.inserted`.
- REQ-batching / subscription-generation fix (`NostrInboxTransport.swift:99-236`, `NostrRelayConnection.swift:141-198`) is consistent end-to-end; hashed+generation sub-ids correctly prevent a stale `onTermination` from CLOSE-ing a successor's REQ.

Non-blocking notes:
- No direct unit tests land for `NostrRelayConnection`'s reconnect/backoff/probe state machine in this PR (`NostrInboxTransportTests.swift` only covers filter-building) — this is exactly the highest-risk concurrency code in the diff. Explicitly deferred per the stated "tests land in the final PR of the stack" convention, but worth prioritizing early in that follow-up.
- `reconcileFreshInstall()`'s heuristic can quarantine a genuine identity if a pre-existing install happens to have a nil persisted selection on first post-upgrade launch (`IdentityRepository.swift:592-598`) — already acknowledged/accepted in the PR body as a "hidden, not destroyed" tradeoff.
- `IdentityKeychainStore.list()` (`IdentityKeychainStore.swift:74-95`) duplicates the `SecItemCopyMatching` logic that the new `listAllItems()` helper (added in this same PR) already encapsulates — minor missed dedup, not a defect.

dev-onym (commented):
**REQUEST_CHANGES** — the transport fixes look sound, but the receipt gating loses its only retry path and two follow-ups don't fully deliver what their comments claim.

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:830` — `guard outcome == .inserted` removes the only retry for delivered receipts. `ChatReceiptSender.send` is fire-and-forget (`_ = try? await inboxTransport.send`, `ChatReceiptSender.swift:54`) and there's no outbox for receipts, so the replay-ack *was* the retry. Receiving while the relay is unreachable (common: inbox replay fires right at launch, before the socket settles) now leaves the sender's message at "sent" permanently. Gate on `.inserted || <row has no delivered-ack yet>`, or persist an `ackSent` flag / enqueue the receipt in `ChatOutbox`.
- `Sources/OnymIOS/Identity/IdentityRepository.swift:592` — the fresh-install verdict is computed from two UserDefaults reads that are *known* to fail in exactly the ambiguous case (pre-first-unlock background launch: plist is `CompleteUntilFirstUserAuthentication`), and `bootstrap()` then mints a replacement identity, so the user sees an empty app with no recovery UI. Quarantine limits the damage but doesn't prevent it — gate the whole reconcile on protected-data availability (`UIApplication.isProtectedDataAvailable`); you lose nothing, since the identity items can't be read/renamed then anyway.
- `Sources/OnymIOS/Group/KeychainIntroKeyStore.swift:181` — TTL compaction in `loadAll()` doesn't `publish(forOwner:)`, so the header's "expired entries are invisible to … streams (so the intro pump stops subscribing their inboxes)" (line 27) only holds at launch or on the next mutation. A session running past an entry's 24h mark keeps its REQ slot and stale snapshot. Publish for the affected owners when `live.count != entries.count`.
- `Sources/OnymIOS/OnymIOSApp.swift:647` — a successful-but-empty identity load still runs `pruneOwners(keeping: [])`, which deletes every intro key. Reachable post-quarantine and in the race with `bootstrap()` (a separate `.task`, line 497). Add `guard !identities.isEmpty`.
- `Sources/OnymIOS/Chats/SwiftDataMessageStore.swift:159` — the update branch still does `try? context.save()` and returns `.updated`, so `.failed` is only ever reported on the insert path, contradicting the enum doc ("nothing was persisted"). Mirror the insert branch's `do/catch`.
- `Sources/OnymIOS/Transport/Nostr/NostrRelayConnection.swift:312` — `teardownDeadSocket()` cancels `webSocketTask`, not the socket the probe actually judged. The window is tiny, but this is the exact foreign-cancel pattern `receiveLoop(task:)` was rewritten to eliminate; pass the probed task through and check `webSocketTask === task` before cancelling.
- `Sources/OnymIOS/Transport/Nostr/NostrRelayConnection.swift:290` — the `__hb` REQ holds a subscription slot for the 10s probe window every 30s. On a connection already near strfry's ~20 cap (the problem fix #4 addresses) this can be the REQ that gets rejected; consider probing with a filter-less `CLOSE`-then-`REQ` on an id you know is already free, or keeping one inbox slot in reserve.
- Test coverage: nothing exercises `quarantineAll`/`reconcileFreshInstall`/`InstallMarker`, the 24h TTL + compaction, `KeychainIntroKeyStore.pruneOwners`, subscription generations, or `probeLiveness` — i.e. the key-material-destroying and connection-lifecycle paths. Understood that tests land at the end of the stack, but the quarantine and TTL paths destroy/hide secrets and warrant tests in this PR. Relatedly, `IdentityRepository` tests take the default `installMarker: .userDefaults`, so the reconcile branch they hit depends on the simulator's persisted defaults and test order — inject `.inMemory(initiallySet: true)`.

dev-onym (commented):
REQUEST_CHANGES — two blocking items (an unrelated compliance-key deletion and an identity-load race the PR itself widened); the transport core looks sound.

**Blocking**

- `Info.plist:36` — `ITSAppUsesNonExemptEncryption` and its whole explanatory comment are deleted. Nothing else in this PR touches export compliance; this looks like an accidental revert of `71fbb75`. `docs/compliance/README.md:16,33` and `us-self-classification-report.md:11` still assert the key is set in `Info.plist`, and without it App Store Connect re-prompts on every upload. Restore it.
- `Sources/OnymIOS/Identity/IdentityRepository.swift:541-571` — `ensureLoaded()` now has a real suspension point (`await reconcileFreshInstall()` → `MainActor.run`, line 548/604) while `loaded = true` moved to the *end* (571). Two concurrent first-callers both pass `guard !loaded`, both clear state before the await, then both run the append loop after it. `OnymIOSApp.swift:497` (`bootstrap()`) and `OnymIOSApp.swift:652` (`currentIdentities()`) are sibling `.task` modifiers on the same view, so they race on exactly the launch where the marker is absent — i.e. the first launch of this build for every existing user. Result: `orderedIDs.append(id)` (558) runs twice per identity → duplicated picker entries and a skewed `orderedIDs.count + 1` slot name in `addLocked` (634). Needs a single in-flight load (`loadTask: Task<Void, Error>?`) or a re-check of `loaded` after the await; the old sync `ensureLoaded` was interleave-proof.

**Non-blocking**

- `Chats/ChatReceiptSender.swift:60` + `Inbox/IncomingMessageDispatcher.swift:848-858` — `accepted` bottoms out in `publishAndAwaitOK`, whose 5s silence is *resumed as `true`* (`NostrRelayConnection.swift:129-133`). On the zombie socket this PR exists to fix, `task.send()` buffers locally and no OK ever arrives → `accepted == true` → `markDeliveredAckSent` latches a receipt that was never delivered, and the replay retry never fires again. The latch would be sounder gated on a real `OK`.
- `NostrRelayConnection.swift:367-370` — `CLOSED` still falls through `default: break`. Item #4 reduces the REQ count but a rejected REQ ("too many concurrent REQs") remains completely silent: the sub stays in `subscriptions`, the inbox is dead, nothing logs or retries. That was the original diagnosis path; handling `CLOSED` (drop + resubscribe, or at minimum log) would make a recurrence visible.
- `Chats/SwiftDataMessageStore.swift:187` — `deliveredAckSent` is nil for every row that predates this migration, so `needsDeliveredAck` is `true` for all of them and the upgrade launch replays one final full receipt storm before the latch takes hold. Self-limiting, but worth a backfill (mark existing incoming rows acked) or a note in the PR body.
- `Chats/SwiftDataMessageStore.swift:177` — insert-failure uses `context.delete(encoded)` while the update path (165) uses `context.rollback()`. `delete` without a save leaves both the failed insert and the delete pending, so a later `save()` from `updateStatus`/`markDeliveredAckSent` re-flushes them. Use `rollback()` on both paths.
- `OnymIOSApp.swift:652-655` — the comment justifies the `!identities.isEmpty` guard as protecting intro keys "whose quarantined owners a recovery flow could still bring back," but `currentIdentities()` excludes quarantined owners, so with ≥1 active identity (always true post-bootstrap) the prune destroys exactly those keys. Either pass `keychain.listQuarantined()` into the keep-set or drop the rationale.
- `Group/KeychainIntroKeyStore.swift:181` — TTL is `createdAtMillis > cutoff`, so a row with a future `createdAt` (clock rollback, or a peer-supplied timestamp) never expires. Clamping to `now` would close it.

dev-onym (commented):
## COMMENT — one blocking-ish item (the upgrade migration), rest are minor

**Needs a fix before merge**

- `Sources/OnymIOS/Chats/PersistedMessage.swift:65` + `Sources/OnymIOS/Chats/SwiftDataMessageStore.swift:181` — `deliveredAckSent` is a new optional with no backfill, and `needsDeliveredAck` returns `row.deliveredAckSent != true`. Every incoming row that already exists on device reads `nil` → "needs ack". On the first launch after this ships, the relay replays the inbox, every historical message lands on the `.updated` branch with an un-latched flag, and the dispatcher publishes a receipt for each — precisely bug #5 from the PR body, once. The `.inserted`-only gate it replaces did not have this hole. Fix: backfill the flag to `true` for rows that predate the migration (or treat `nil` as acked and only latch/consult the flag for rows written from now on) — the retry path is only worth having for messages received after the feature exists.

**Minor / discussion**

- `Sources/OnymIOS/Chats/ChatReceiptSender.swift:60` — "transport accepted the publish" is weaker than the doc claims: `PublishReceipt.acceptedBy` is discarded, and `NostrRelayConnection.publishAndAwaitOK` (`NostrRelayConnection.swift:131`) resolves a 5-second OK silence as `true`. A receipt the relay quietly dropped still latches and never retries. Not new behavior, but the latch now makes it permanent — worth a comment at minimum.
- `Sources/OnymIOS/Chats/SwiftDataMessageStore.swift:172` — the insert branch's failure path is `context.delete(encoded)`, the update branch's is `context.rollback()`; the comment says they mirror each other. `delete` leaves an unsaved pending change that the next `save()` flushes. Use `rollback()` in both.
- `Sources/OnymIOS/Group/KeychainIntroKeyStore.swift:190` — compaction is best-effort: `writeAll` (line 203) swallows `SecItemUpdate`/`SecItemAdd` failures, so on a write error the expired intro privkeys stay at rest while reads filter them. The doc comment promises they're "compacted out"; soften it or surface the failure.
- `Sources/OnymIOS/OnymIOSApp.swift:653` — skipping `pruneOwners` on an empty identity set means a user who deliberately removed every identity never GCs their intro keys. The tradeoff is the right one post-quarantine; just noting it's now unreachable rather than deferred.
- No tests land for the transport core of this PR (`receiveLoop(task:)` socket binding, probe timeout/early-exit, one-REQ-many-filters, `desiredConnected` surviving backoff). The stack convention covers it, but these are the crash and zombie-socket fixes — a `NostrRelayConnection` seam over `URLSessionWebSocketTask` would pin the exact races the commit messages describe.

**Verified correct**

- `project.yml:59` is the right root-cause fix — `info.path: Info.plist` confirms the plist is xcodegen output, so the key must live in the generator.
- `ensureLoaded()`'s single-flight, `loaded = true` only after a complete `performLoad()`, the `webSocketTask === task` guards in `receiveLoop`/`teardownDeadSocket`/`probeLiveness`, `desiredConnected` re-check after the backoff sleep, per-generation hashed sub ids (`inbox-<16 hex>-<gen>` stays under NIP-01's 64), and `quarantineAll()`'s prefix isolation (`identity.` never prefix-matches `identity-quarantine.`) all check out.

PR #207, opened 2026-07-31
chore: rename app to Onym Chat

rinat-enikeev:
## Summary
- Rename the app's home-screen display name from **Onym** to **Onym Chat** (`CFBundleDisplayName`)
- Change made in `project.yml` (the source of truth); `Info.plist` regenerated via `./generate-xcodeproj.sh`

No changes to bundle identifier, product name, or URL schemes.

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
-	<string>Onym</string>
+	<string>Onym Chat</string>
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
--- a/project.yml
+++ b/project.yml
@@ -38,7 +38,7 @@ targets:
     info:
       path: Info.plist
       properties:
-        CFBundleDisplayName: Onym
+        CFBundleDisplayName: Onym Chat
         # `$(...)` is Xcode's build-setting substitution, NOT xcodegen
         # env interpolation — these strings end up verbatim in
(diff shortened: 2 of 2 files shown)


dev-onym (commented):
No code or test reads `CFBundleDisplayName`, so nothing breaks.

**APPROVE** — the change is correct, minimal, and the regenerated `Info.plist` matches `project.yml` exactly with no incidental drift.

- `project.yml:41` + `Info.plist:8` — the only two places the display name lives; they agree. Bundle id, `CFBundleName` (`$(PRODUCT_NAME)`), and the `onym`/`app.onym.ios.join` URL scheme are untouched, matching the PR description.
- Regeneration looks like a real `./generate-xcodeproj.sh` run rather than a hand edit: no version-key churn from `MARKETING_VERSION`/`CURRENT_PROJECT_VERSION` resolution, formatting/tab style preserved.
- Nothing in `Sources/` or `Tests/` reads `CFBundleDisplayName`, so no code, snapshot, or UI-test selector depends on the old string.
- `.github/workflows/release.yml:488` and `.github/workflows/qa-build.yml:229` — OTA manifest `metadata.title` is still hardcoded `Onym`. This is what iOS shows in the "Install …?" prompt and under the installing icon for QA/ad-hoc builds, so testers will now see `Onym` at install time and `Onym Chat` afterward. Cosmetic and out of the stated scope, but it's the one remaining user-visible instance of the old name — worth a follow-up (or fold in here, both are one-line edits).
- `Sources/OnymIOS/Settings/AboutView.swift:209` — the About wordmark stays `Onym`. Reads intentional (brand name vs. app name, paired with the "open · anonymous · onchain" tagline); flagging only so it's a deliberate call, not an oversight.
- Home-screen truncation: "Onym Chat" is 9 characters and will render in full on standard iPhone layouts, so no ellipsis regression.

PR #208, opened 2026-08-05
Member removal from Founder groups, with groupSecret rotation

rinat-enikeev:
## Summary

iOS twin of onymchat/onym-android#192 — the founder removes a member from a Tyranny group, expressed as the existing `update_commitment` over the shrunken roster, with **groupSecret rotation** delivered per-recipient.

- **Wire**: `MemberRemovalPayload` mirrors the Android-authored format exactly (`removal_*` required keys, trial-decode-disjoint from every other inbox payload). The victim's copy omits `group_secret_new`/`salt_new` entirely — the removed member learns the fact, never the secrets. Cross-platform parity is pinned by decoding Android's frozen wire vector in `MemberRemovalPayloadTests`.
- **Admin side**: `removeMember` on `JoinRequestApprover`, serialized through the same `approvalChain` as approvals (both are read-prove-anchor-persist; interleaving would prove against a stale epoch). Ordering: anchor → persist → best-effort fanout.
- **Tombstone, never delete**: `MemberProfile.revoked` keeps past-message rendering and dedup intact while chat sender auth, mutation auth, salt-bearing refresh replies, and all fanout loops reject/skip removed members. Re-admission overwrites the tombstone.
- **Victim device**: `ChatGroup.membershipRevoked` (defaulted optional SwiftData column — lightweight migration, no store wipe): history readable, `ChatInputPanelView` swaps to a localized "You were removed from this group" banner, `SendMessageInteractor` throws `.removedFromGroup`.
- **Receive side**: converge-forward chain verification plus a bounded 12s/24s retry when the cached chain read lags the just-landed anchor.
- **UI**: admin-only destructive remove button + confirmation dialog + in-flight spinner on `ChatMembersView`; en + ru catalog entries.

## Test plan

- [x] Full `OnymIOSTests` target green on the iPhone 17 Pro Max simulator (90 suites, 0 failures) — new: `MemberRemovalPayloadTests`, `IncomingMessageDispatcherRemovalTests` (incl. chain-behind retry + idempotency counting chain reads), `JoinRequestApproverRemovalTests` (real `Tyranny.proveUpdate` on `.small`; both fanout envelopes decrypted with real X25519 keys — victim copy proven secret-free); `SwiftDataGroupStoreTests` extended
- [ ] Cross-platform 3-member session with the onym-android sibling (remove one member; victim banner + send refusal, remaining members converse on the rotated secret, restarted device converges via inbox replay)

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -34,6 +34,102 @@
         }
       }
     },
+    "Couldn't remove the member": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Couldn't remove the member"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Не удалось удалить участника"
+          }
+        }
+      }
+    },
+    "Remove": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Remove"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Удалить"
+          }
+        }
+      }
+    },
+    "Remove %@?": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Remove %@?"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Удалить %@?"
+          }
+        }
+      }
+    },
+    "Remove member": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Remove member"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Удалить участника"
+          }
+        }
+      }
+    },
+    "They will no longer receive this group's messages. The group key will be rotated for everyone else.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "They will no longer receive this group's messages. The group key will be rotated for everyone else."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Участник больше не будет получать сообщения этой группы. Ключ группы будет обновлён для остальных."
+          }
+        }
+      }
+    },
+    "You were removed from this group": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "You were removed from this group"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Вас удалили из этой группы"
+          }
+        }
+      }
+    },
     "—": {
       "localizations": {
         "en": {
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -34,6 +34,22 @@
         }
       }
     },
+    "Couldn't reach the network. Check your connection and try again.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Couldn't reach the network. Check your connection and try again."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Не удалось подключиться к сети. Проверьте соединение и попробуйте ещё раз."
+          }
+        }
+      }
+    },
     "Couldn't remove the member": {
       "localizations": {
         "en": {
@@ -46,6 +62,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Не удалось удалить участника"
+          }
+        }
+      }
+    },
+    "Only the group's founder can remove members.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Only the group's founder can remove members."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Удалять участников может только основатель группы."
           }
         }
       }
@@ -94,6 +126,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Удалить участника"
+          }
+        }
+      }
+    },
+    "Something went wrong. Try again.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Something went wrong. Try again."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Что-то пошло не так. Попробуйте ещё раз."
           }
         }
       }
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -61,4 +61,16 @@ struct AppDependencies {
     /// `GroupNamePayload` out to members. Backs the admin-only rename
     /// affordance in `ChatMembersView`.
     let setGroupName: @MainActor (_ groupIDHex: String, _ name: String) async -> Void
+    /// Admin removes a member from a Tyranny group: anchors the
+    /// shrunken roster on chain (epoch bump + salt/groupSecret
+    /// rotation), tombstones the victim locally, and fans a
+    /// `MemberRemovalPayload` out to every member (the victim's copy
+    /// omits the rotated secrets). Backs the admin-only remove
+    /// affordance in `ChatMembersView`. Routed through
+    /// `JoinRequestApprover.removeMember` so it serializes with
+    /// in-flight approvals.
+    let removeGroupMember: @MainActor (
+        _ groupIDHex: String,
+        _ victimBlsHex: String
+    ) async -> JoinRequestApprover.RemoveOutcome
 }
--- a/Sources/OnymIOS/Chats/ChatInputPanelView.swift
+++ b/Sources/OnymIOS/Chats/ChatInputPanelView.swift
@@ -39,6 +39,16 @@ final class ChatInputPanelView: UIView {
     /// Whether any media is staged in the preview strip.
     private var hasPendingMedia = false
 
+    /// Removed-from-group lockout. When `true` every input control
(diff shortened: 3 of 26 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the wire format, admin flow, and gate ordering are solid, but the relaxed announcement dedup lets an inbox replay silently undo a removal.

- **Tombstone resurrection (high)** — `IncomingMessageDispatcher.swift:689` now lets a revoked profile fall through the dedup gate, and the converge-forward check at `:586` accepts any announcement whose epoch is ≤ the on-chain epoch. The victim's *original* join announcement is still in every remaining member's inbox and is re-dispatched on each relay reconnect (the "launch-time storm" the comment at `:680` describes; `InboxFanoutInteractor.swift:136` has no messageID dedup). It passes the admin-signature gate, passes the chain gate (chain is now ahead), and `:709` rewrites the profile with `revoked` defaulting to `false` — so after the next reconnect the removed member is back in `memberProfiles`, receives fanout again (`SendMessageInteractor.swift:184`), and their sends re-authorize at `:971`. Removal is undone with no user-visible signal. Gate tombstone-clearing on `payload.epoch > group.epoch` (re-admission always anchors a strictly newer epoch — `JoinRequestApprover.swift:904` ships the post-anchor epoch), or persist a per-member removal epoch.
- **Victim device mis-branches when the identity lookup fails** — `IncomingMessageDispatcher.swift:765-768`: `currentIdentities()` throwing/empty makes `isSelf` false, so the victim's own device takes the remaining-member path: it tombstones *its own* profile and drops itself from `members`, but leaves `membershipRevoked == false`. The composer stays live and `SendMessageInteractor`'s membership guards check `!= nil`, not `revoked` (`:135`, `:246`, `:357`, `:497`, `:627`) — sends proceed and are dropped by every receiver. Prefer `guard let selfHex else { return }` (retry on the next replay) and add a `!profile.revoked` check to the self-membership guards.
- **Victim-variant applied by a remaining member desyncs salt** — `IncomingMessageDispatcher.swift:833`: epoch/commitment advance while `salt` and `groupSecret` stay old, so the persisted salt no longer matches the persisted commitment; a later `proveUpdate` from that state would prove against the wrong `saltOld`. The comment acknowledges it, but dropping the payload when `saltNew == nil && !isSelf` is strictly safer than persisting an incoherent triple.
- **Lost update across the anchor window** — `JoinRequestApprover.swift:746` writes back `anchored`, a snapshot taken before the multi-second prove + submit; an inbound rename/avatar/announcement that lands during that window is silently reverted. Same shape as `performApprove`, so pre-existing, but removal widens the window — re-read the group and re-apply the delta before `insert`.
- **Missed active-member filters** — `ChatsView.swift:375` (chat-list "N members") and `ChatThreadViewController.swift:523` (empty-state alias list) still count/list tombstoned profiles, inconsistent with the new filters in `ChatMembersView`/`ChatThreadView`.
- **Nit** — `ChatMembersView.swift:291` puts a raw `String(describing:)` enum dump (`anchorRejected("stale epoch")`) into a user-facing alert body; unlocalized and leaks internals. Map to a small set of localized reasons.

Nice work on the parts that matter most: `removal_*` key prefixing with two-directional trial-decode tests, the real-X25519 assertion that the victim's envelope carries no `group_secret_new`, auth-before-chain-read ordering (with a test counting chain reads), and the defaulted-optional SwiftData column with a legacy-JSON decode test.

dev-onym (commented):
Reviewed the diff against the surrounding code (dispatcher fast paths, approver chain, all `memberProfiles` consumers, receipt paths, persistence).

**COMMENT** — solid, well-tested design (anchor→persist→fanout ordering, local converge-forward guard *before* the chain read, fail-closed on missing admin key, victim variant proven secret-free by decrypting real envelopes); findings are gaps and inconsistencies, none blocking.

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:883` — the retry dedup key is `groupID:victim:epoch` with no `ownerIdentityID`. `performRemove` explicitly acknowledges "the same on-chain group id can belong to two local identities" (`JoinRequestApprover.swift:605`), and the removal is fanned to each identity's inbox separately. On such a device the first identity's chain-behind retry chain makes `begin(key)` return `false` for the second, so that identity's copy silently keeps trusting the removed member until the next inbox replay. Add the owner ID to the key.
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:1092` — the `.delivered` ack is not gated on `group.membershipRevoked`, while the read-receipt path was (`ChatThreadView.swift:287`). A removed device that receives an in-flight or replayed message still publishes a receipt into the group it was evicted from — same leak the read-receipt guard was added to close.
- `Sources/OnymIOS/Chats/ChatsView.swift:375` — the chat-list subtitle still counts tombstoned profiles (`memberProfiles.count`), so the list says "4 members" while `ChatMembersView.swift:328` and `ChatThreadView.swift:381` (both fixed here) say 3. `ChatThreadViewController.swift:523` has the same issue: `makeEmptyState()` lists revoked members' aliases in the empty-state roster.
- `Sources/OnymIOS/Chats/ChatInputPanelView.swift:281` — `setMembershipRevoked(true)` hides attach/text/mic/send but not `replyBanner` (line 106) or `mediaStrip` (121). A member removed while a reply is armed or media is staged keeps a banner/preview above a composer-less panel, with `onCancelReply` still the only way out; `focusComposer()` (called from the host on reply-tap) would also try to raise the keyboard on a hidden text view. Suggest clearing both in the revoked branch.
- `Sources/OnymIOS/Chats/ChatMembersView.swift:273` — `beginRemoval` is nonisolated, so the `Task {}` body resumes on the generic executor after `await removeMember(...)`; `removalInFlightBlsHex = nil` and `removalErrorText = …` are then `@State` writes off the main actor. Mark the method or the Task `@MainActor`.
- `Tests/OnymIOSTests/MemberRemovalPayloadTests.swift:117` — `test_otherInboxPayloads_doNotDecodeAsRemoval` only asserts the reverse direction for announcement + invitation. Removal is trial-decoded *before* avatar, name, chat-receipt, and chat-message (dispatcher lines 173–250), so those four are the ones that would be misrouted; they're covered only in the removal→other direction.
- Minor/pre-existing: `performRemove` snapshots `currentGroups()` before the multi-second prove and writes the whole derived row back (`JoinRequestApprover.swift:745`), so any group mutation landing during the anchor is clobbered — same pattern as `performApprove`, worth a re-read-and-merge eventually.

dev-onym (commented):
REQUEST_CHANGES

**Blocking**

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:695` — **a replayed old join announcement resurrects a removed member.** The dedup bail was deliberately weakened so a tombstone doesn't block re-admission, but relay reconnect replays the *original* `MemberAnnouncementPayload` for that member (epoch 3, admin-signed). `verifyTyrannyAnnouncement` accepts any claimed epoch ≤ chain epoch (`:592`, byte-verifies only on exact match), so the replay writes a fresh `MemberProfile` with `revoked = false`, `statusEpoch = 3` (`:715`). The ex-member is then back in every trust gate: `isAuthorizedGroupMutation` (`:651`) and the chat sender-auth check (`:1052`) accept them again, and `SendMessageInteractor` seals new messages to their inbox (`SendMessageInteractor.swift:186`) — i.e. they can read the group again despite the rotation, and this re-fires on every launch. Fix mirrors what `applyRemoval` already does per-member: only overwrite a tombstone when `payload.epoch > existing.statusEpoch` (and refuse when the announcement carries no epoch).
- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:832` — **victim's composer re-locks after re-admission.** `selfApplies` compares only against the victim's own `statusEpoch`, and after a *real* re-admission that value is `nil`: the admin filters tombstones out of the invitation snapshot (`JoinRequestApprover.swift:356`) and — since `recordJoiner` runs after the invite ships (`:394`) — the joiner's own entry isn't in it at all, so `materializeGroup` copies `statusEpoch` as nil (`:379`). A replayed epoch-2 self-removal then hits `nil ?? true` → applies (chain is ahead, so `:859`/`:860` both pass) → `membershipRevoked = true` again, every launch. Guard on the group epoch too (`payload.epoch > group.epoch` when `statusEpoch == nil`), or stamp the victim's own profile `statusEpoch` in the self branch (`:864`).
- `Tests/OnymIOSTests/IncomingMessageDispatcherRemovalTests.swift:~305` — `test_removal_replayAfterOwnReadmission_doesNotRelockComposer` seeds `statusEpoch: 3` on the victim's own profile by hand, a state the production re-admission path never produces; the test passes while the scenario it names regresses. Same for the announcement path — there's no test for "stale join announcement after removal".

**Non-blocking**

- `Sources/OnymIOS/Inbox/IncomingMessageDispatcher.swift:886` — when `advancesState` is true but the payload is the secret-free victim variant, epoch + commitment advance while `salt`/`groupSecret` stay old, leaving a group whose commitment can't be recomputed (breaks later invitation/announcement byte-verification). Cheaper to require `saltNew != nil && groupSecretNew != nil` for the state advance.
- `Sources/OnymIOS/Group/MemberRemovalPayload.swift:177` — `removedBlsHex` is length-checked but not hex-validated or lowercase-normalized at the decode boundary; a 96-char non-hex string silently becomes garbage bytes in `ChatGroup.bytes(fromHex:)` (`:877`). Also worth asserting the payload's `removal_group_id` sender is the group admin's *current* epoch lineage — currently only epoch/commitment are cross-checked.
- `Sources/OnymIOS/Chats/ChatInputPanelView.swift:292` — `setMembershipRevoked(true)` hides `recordingOverlay` but doesn't stop an in-flight voice recording; the recorder keeps running and only fails at send time.
- `Sources/OnymIOS/Group/JoinRequestApprover.swift:838` — fanout is sequential with `try?` swallowed per recipient; as the doc comment concedes, a missed envelope leaves that peer sealing to the victim's inbox indefinitely. Worth at least counting/logging failures so the known-gap is observable before the resend queue lands.

Admin-side flow (serialization through `approvalChain`, anchor → persist → fanout ordering, victim-copy secret omission) and the wire type look right; `JoinRequestApproverRemovalTests` decrypting both envelopes with real X25519 keys is a good check.

PR #209, opened 2026-08-05
fix(group): tombstone consumed intro requests so replays can't resurrect them

rinat-enikeev:
`NostrInboxTransport`'s REQ sets neither `since` nor `limit`, so every relay reconnect re-delivers an intro inbox in full. `consume` only removed the row from `pending` and `record` only deduped against `pending`, so a replayed request the user had already approved or declined was re-recorded as if it were new.

Today that stays invisible: approving also revokes the intro key, so the replayed copy fails to decrypt in `decode` and never reaches the UI. The masking disappears the moment invite links become multi-use and the key outlives the first approval — hence fixing the store first, on its own terms.

`consume` now tombstones the event id unconditionally (even when the id isn't pending, so a tombstone can be laid ahead of a slow replay) and `record` rejects a tombstoned id. Process-lifetime, same posture as `pending`: a relaunch still re-surfaces anything the user never acted on.

**Base of the multi-use-invite-links stack.** It must land before or with the burn removal, never after — landing it later leaves a window where approved requests resurrect on relaunch.

Verified: full unit suite green (`-only-testing:OnymIOSTests`), `scripts/lint-secrets.py` clean.

### DIFF
--- a/Sources/OnymIOS/Group/IntroRequestStore.swift
+++ b/Sources/OnymIOS/Group/IntroRequestStore.swift
@@ -19,7 +19,14 @@ protocol IntroRequestStore: Sendable {
     func record(_ request: IntroRequest) async -> Bool
 
     /// Drop a request after the user has acted on it (Approve or
-    /// Decline) so it stops cluttering the surface.
+    /// Decline) so it stops cluttering the surface, and remember its
+    /// id so a replay can't resurrect it.
+    ///
+    /// Conformers MUST tombstone. `NostrInboxTransport`'s REQ carries
+    /// no `since` and no `limit`, so every relay reconnect re-delivers
+    /// the intro inbox in full. Dropping the row from `pending` alone
+    /// means the next replay re-records it and the user sees a request
+    /// they already handled.
     func consume(id: String) async
 
     /// Snapshot read used by tests + bootstrap reads. UI prefers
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
COMMENT — the tombstone is correct and safe as written, but it only covers half the resurrection path it's meant to close, and ships with no test.

- `Sources/OnymIOS/Group/JoinRequestApprover.swift:646-672` + `:355` — **the collapse siblings aren't tombstoned.** `refresh` collapses N raw `IntroRequest`s (same joiner identity + group, distinct event ids from re-sends/deeplink re-opens) into one `PendingRequest` carrying only the *newest* event id, and `approve`/`decline` consume only that id. The older sibling ids stay in `pending` untombstoned. Today the revoke masks it (siblings fail `decode`) — but that's precisely the mask this PR says disappears when the intro key outlives approval under multi-use links. In that world, the handled row re-appears on the next `refresh` under a sibling id, and can be approved a second time. The PR's own framing ("base of the multi-use-invite-links stack, must land before the burn removal") is what makes this in-scope: as-is, the stated window isn't actually closed. Fix belongs in `performApprove`/`decline` — consume every raw id whose collapse key matches, not just `requestId`.
- `Tests/OnymIOSTests/` — no coverage for the new behavior. There's no `IntroRequestStoreTests.swift` at all, and nothing in `JoinRequestApproverTests` asserts `record` after `consume` returns `false`, nor the ahead-of-replay case (`consume` on a non-pending id, then `record` → rejected). The commit is 18 lines of pure state machine; both cases are three-line tests. "Full suite green" doesn't exercise either branch.
- `Sources/OnymIOS/Group/IntroRequestStore.swift:25-29` — the doc says "Conformers MUST tombstone", but nothing enforces it and `InMemoryIntroRequestStore` is the only conformer. Fine for now; just note that when the persistent store lands, the tombstone set needs to persist alongside `pending` or the relaunch contract in `:39-42` silently flips (persisted pending + non-persisted tombstones = resurrection on relaunch, the exact bug).

Non-issues I checked: failure paths in `performApprove` correctly return before `consume`, so a transport failure doesn't tombstone a still-actionable request (`:326-328`); `findIntroPub` reads `current()` before the consume, so ordering is fine (`:733`); `consumed` only grows on user action, so no unbounded-growth vector; skipping `publish()` when nothing was pending is correct.

PR #214, opened 2026-08-06
feat(group): multi-use invite links with explicit revocation (squashed)

rinat-enikeev:
**Draft — reading copy.** Same change as the #209 → #213 stack, squashed into one commit off `main` so the whole thing can be read as a single diff. Merge the stack, not this. Comments are condensed here (≤2 lines); the stack carries the long-form rationale.

Android twin: onymchat/onym-android#201.

### What changed

An invite link was killed by the first Approve *or* Decline, which was wrong three ways: it was already accidentally multi-use (two joiners racing the first approval both worked, because approve never re-validated the key); declining one stranger silently killed the link for everyone else; and after the first approval every other pending joiner's row vanished with no explanation.

Links are now deliberately multi-use, never expire, and are retired only by the inviter. Every join is still gated by an explicit tap.

- **Approve/Decline stop revoking**, and both consume the whole collapsed sibling set rather than just the tapped row.
- **A joiner already in `group.members`** skips the chain anchor and only gets the current snapshot re-shipped, so a replay or a reinstall can't push a duplicate leaf. Also makes the long-standing double-tapped-Approve race safe.
- **The share screen reuses the group's link** instead of minting one per visit; **"Generate new link"** rotates it — mint fresh, then revoke the old, in that order so a crash leaves two links rather than zero.
- **Create-time offer keys are labelled per invitee and listed with a per-row Revoke.** They had no UI before, so dropping the TTL without this would leave every group creation trailing permanent, invisible relay subscriptions.
- **Consumed request ids are tombstoned.** The relay REQ carries no `since`, so reconnects replay the whole inbox; the burn was the only thing hiding that.
- **`JoinFlow` gains an `unanswered` state** after 90s, since revocation is silent and a dead link is indistinguishable from a slow host.
- **`IntroCapability` encodes with sorted keys** — one capability could otherwise serialize to two different link strings.

### Security note

A leaked link used to be worth one join attempt, closed by the first tap. It is now an open request channel until noticed and rotated. Nothing is granted without an explicit Approve and the link carries no group secret, but the unsolicited-request surface is new and uncapped. Two follow-ups worth filing: no cap on pending requests, and `SepTier.maxMembers` is enforced nowhere in the join path.

896 tests green, secret lint clean. Nine new strings, en and ru.

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -1540,6 +1540,22 @@
         }
       }
     },
+    "Direct invites": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Direct invites"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Персональные приглашения"
+          }
+        }
+      }
+    },
     "Dismiss": {
       "localizations": {
         "en": {
@@ -1733,6 +1749,22 @@
         }
       }
     },
+    "Generate new link": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Generate new link"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Создать новую ссылку"
+          }
+        }
+      }
+    },
     "Generate with AI": {
       "extractionState": "stale",
       "localizations": {
@@ -1762,6 +1794,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Генерируется доказательство и обновляется обязательство в блокчейне. Обычно занимает несколько секунд."
+          }
+        }
+      }
+    },
+    "Generating…": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Generating…"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Создаём…"
           }
         }
       }
@@ -2455,6 +2503,22 @@
         }
       }
     },
+    "No response yet": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "No response yet"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Пока нет ответа"
+          }
+        }
+      }
+    },
     "No servers configured. Media can't be sent or received.": {
       "localizations": {
         "ru": {
@@ -2516,6 +2580,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Это не то слово. Проверьте фразу и повторите попытку."
+          }
+        }
+      }
+    },
+    "Older link": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Older link"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Старая ссылка"
           }
         }
       }
@@ -3162,6 +3242,22 @@
         }
       }
     },
+    "Revoke": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Revoke"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Отозвать"
+          }
+        }
+      }
+    },
     "Run a relayer for yourself, your team, or your org. End-to-end encryption stays intact — Onym never sees your messages.": {
       "localizations": {
         "en": {
@@ -3387,6 +3483,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Отправляем запрос…"
+          }
+        }
+      }
+    },
+    "Sent when you created the group. Each one can still be used to request to join.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Sent when you created the group. Each one can still be used to request to join."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Отправлены при создании группы. По каждому из них всё ещё можно запросить вступление."
           }
         }
       }
@@ -3701,6 +3813,38 @@
         }
       }
     },
+    "The host may be offline, or this invite may no longer work. Ask them for a new link.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "The host may be offline, or this invite may no longer work. Ask them for a new link."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Возможно, владелец группы не в сети или это приглашение больше не действует. Попросите новую ссылку."
+          }
+        }
+      }
+    },
+    "The old link stops working. Anyone still holding it won't be told.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "The old link stops working. Anyone still holding it won't be told."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Старая ссылка перестанет работать. Тех, у кого она осталась, об этом не уведомят."
+          }
+        }
+      }
+    },
     "The phrase is generated on-device and never sent off the device.": {
       "localizations": {
         "en": {
@@ -3990,6 +4134,22 @@
           "stringUnit": {
             "state": "translated",
             "value": "Ждём одобрения от хоста…"
+          }
+        }
+      }
+    },
+    "You can leave this screen open — if they approve, the chat still appears.": {
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "You can leave this screen open — if they approve, the chat still appears."
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Можно не закрывать этот экран — если заявку одобрят, чат появится сам."
           }
         }
       }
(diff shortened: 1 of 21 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the core multi-use redesign is sound and well-tested, but the replay defense and the link-identity invariant both leak.

- `Sources/OnymIOS/Group/IntroRequestStore.swift:34` — tombstones are process-lifetime and `InMemoryIntroRequestStore` is the only conformer (`OnymIOSApp.swift:280`). The burn was previously what made a handled request undecodable *across launches*; now the key stays live and the REQ has no `since`, so **every** request the admin ever approved or declined re-decodes into a pending row on each cold start, forever. Declining a stranger becomes a chore repeated at every launch, growing with the link's lifetime. The doc comment only admits re-surfacing "what the user never handled" — that's the case this PR *doesn't* regress. Needs a persisted tombstone set (keyed by event id, ideally pruned by group).
- `Sources/OnymIOS/Group/InviteIntroducer.swift:86` — `currentOrMint` reads the store across an `await`, so the actor suspends and two concurrent calls both see no shared key and both mint. SwiftUI fires `.onAppear` more than once per entry (re-entry, sheet dismissal), so two `label == nil` keys for one group is reachable: two REQ slots, and the link on screen isn't necessarily the one a later `currentOrMint` resolves. Same read-then-mutate window in `rotate` (`:118`) — a `mintFor` interleaved with a `rotate` hands the user a link the rotate then revokes.
- `Sources/OnymIOS/Group/ShareInviteFlow.swift:104` — `guard !isRotating` is checked, but `isRotating = true` happens after two `await`s (`currentGroups`, `currentSelectedID`), so a double-tap passes both guards and `.disabled(flow.isRotating)` lags by the same window. The stated intent ("rotating twice in a row would strand a live key") isn't enforced. Set the flag before the first await.
- `Sources/OnymIOS/Group/ShareInviteFlow.swift:148` — a stranded duplicate shared key from either race above is unreachable: `refreshOtherInvites` drops `label == nil`, so it appears on no list and has no revoke, and it keeps a relay subscription until the next rotate happens to sweep it.
- `Sources/OnymIOS/Group/KeychainIntroKeyStore.swift:24` — dropping `entryTTL` reverts issue #111 (leak window of a forwarded link) with no replacement, and `IntroInboxPump` opens one subscription per live entry with no cap. Slot count is now monotonic in groups × create-time invitees; an invitee who never accepts leaves a permanent subscription and an intro **private** key at rest until someone taps a hex fingerprint row. Worth stating in the PR body next to the security note, since it's the same tradeoff.
- `Sources/OnymIOS/Group/JoinFlow.swift:113` — `.unanswered` is a dead end: `send` accepts only `.ready`/`.failed`, and `JoinView`'s `.unanswered` case renders no action, so the copy "Ask them for a new link" is the only exit. Either allow re-send from `.unanswered` or add a Try again button.
- `Sources/OnymIOS/Group/JoinRequestApprover.swift:692` — `collapsedRequestIDs` is only as fresh as the last `refresh`; a sibling recorded after it isn't consumed and resurfaces as a row for an already-approved joiner. Benign now that re-join re-ships, but it undercuts the "consume the whole set" guarantee the tests assert.
- `Resources/Localizable.xcstrings` — the file was rewritten in codepoint order (non-ASCII keys pushed to the end), so ~3.5k lines of the diff are pure churn that Xcode's next extraction will flip back. Worth re-sorting to Xcode's own order so the nine real strings are reviewable.

Good calls worth keeping as-is: mint-before-revoke in `rotate`, the `alreadyInRoster` guard sitting ahead of proof generation (and the test that pins that ordering), `label` optional in `StoredIntroKey` for upgrade safety, and `.sortedKeys` on the capability encoder.

dev-onym (commented):
REQUEST_CHANGES — the design is sound and well-tested, but the new re-join fast path and the tombstone lifecycle have exploitable/lossy edges.

- **`Sources/OnymIOS/Group/JoinRequestApprover.swift:261` — `alreadyInRoster` trusts an unauthenticated, joiner-chosen BLS pubkey.** `JoinRequestPayload.joinerBlsPublicKey` carries no proof of possession (the outer envelope only signs with the joiner's *identity* key). Anyone who has ever held a `GroupInvitationPayload` knows every member's BLS pubkey. Mallory sends a request with Alice's `joiner_bls_pub` + her own inbox/sending keys and label "Alice"; admin sees one plausible row and approves → the anchor is skipped, `recordJoiner` (`:345`) overwrites Alice's `memberProfiles` entry with Mallory's inbox+sending keys, and the full `groupSecret` snapshot is sealed to Mallory. Previously this had to survive the anchor path; the new branch makes it clean, cheap and silent. Gate the fast path on the request also matching the *existing* profile (e.g. same `joinerInboxPublicKey`), or require a signature over the BLS key before treating roster membership as authorization.
- **`Sources/OnymIOS/Group/JoinRequestApprover.swift:354` — skipping `broadcastJoin` on re-join defeats the stated purpose.** A reinstalled joiner's *new* inbox/sending pubkeys are recorded only in the admin's local `memberProfiles`; other members keep the dead keys, so nobody but the admin can reach or verify them. The "receivers bail on a known BLS hex" rationale is an argument for fixing the receiver, not for withholding the update.
- **`Sources/OnymIOS/OnymIOSApp.swift:678` — tombstone prune is scoped to the active identity but `prune(keeping:)` filters globally.** Switch identities and every other identity's tombstones are dropped (`HandledIntroRequestLog.swift:56` keeps only rows whose `introPub` is in the passed set). Switch back and the whole handled inbox replays as pending. Key the prune by owner, or pass the union across identities.
- **`Sources/OnymIOS/Group/IntroRequestStore.swift:74` — `consumed = handledLog.handledIDs()` silently discards in-memory-only tombstones.** `consume` only writes durably when the id is still in `pending` (`:62`), so the "tombstone laid ahead of a replay" case that `test_consume_unknownId_stillTombstones` locks in is erased by the next intro-key snapshot. Make prune subtractive (`consumed.subtract(dropped)`) rather than a reload, or record durably unconditionally with a nil/unknown link bucket.
- **`Sources/OnymIOS/Group/ShareInviteFlow.swift:124` — `rotateLinkAsync` never calls `refreshOtherInvites`.** After a crash-stranded second shared key, rotate revokes it but the "Older link" row stays on screen; tapping Revoke is a no-op with no feedback. Add the refresh after the state assignment, as `mintFor` does at `:91`.
- Minor: `JoinFlow.swift:107` — the debounce block is dead (`!sendTask.isCancelled` && `sendTask.isCancelled == false`, empty body). Harmless before, but `.unanswered` now adds a user-tappable retry, so either delete it or make it real.
- Minor: `HandledIntroRequestLog.swift:38` — `record` decodes + re-encodes the whole ≤2000-row array and does an O(n) `contains` on every inbound request. Fine at current volume; worth a `Set` cache if the inbox gets busy.
- Agreed with the PR's own follow-ups (no cap on pending requests, `SepTier.maxMembers` unenforced) — with the TTL gone, the pending-request cap is the one I'd file first.

dev-onym (commented):
**REQUEST_CHANGES** — solid design and test coverage; two real defects (one durable-state, one UI staleness) plus a backwards eviction policy.

- `Sources/OnymIOS/OnymIOSApp.swift:678` — `pruneTombstones(keeping:)` is fed `entriesStream(forOwner: activeID)`, i.e. **only the active identity's** intro keys, but the tombstone log is global. Switching identity A→B prunes every tombstone attributed to A's links; A's keys are still in the keychain, so switching back re-subscribes those inboxes, the no-`since` REQ replays the full inbox, and every request A already approved/declined resurfaces as a pending row — exactly the bug the log exists to prevent. Prune against all identities' live keys (or scope the log per owner). No test covers the two-identity case.
- `Sources/OnymIOS/Group/HandledIntroRequestLog.swift:38` — "Oldest-first drop: a replay is likeliest to hit recent ids" is inverted. The REQ carries no `since`, so a reconnect replays the *whole* inbox; the ids evicted at `maxEntries` are precisely the ones guaranteed to come back. Past 2000 handled requests on one link, the oldest re-materialize on every cold start. Evict per-link or by relay-side retention, not by age.
- `Sources/OnymIOS/Group/ShareInviteFlow.swift:124` — `rotateLinkAsync` sets `.ready` but never calls `refreshOtherInvites`, unlike `mintFor:91` and `revokeAsync:141`. `rotate` revokes *all* unlabelled keys for the group, so a stranded superseded key that was listed stays on screen after rotate as a ghost row whose Revoke button is a no-op; the new shared link also isn't excluded from the stale list. One `await refreshOtherInvites(...)` after line 124. `test_supersededSharedKey_isListedSoItCanBeRevoked` doesn't exercise the post-rotate list.
- `Sources/OnymIOS/Group/IntroRequestStore.swift:72` — `consumed = handledLog.handledIDs()` downgrades the in-memory set to the durable one, so any tombstone laid for an id not in `pending` (the "ahead of a replay" case at :59-66, which never reaches `handledLog`) is silently dropped by the next prune — and prune fires on every keystore emission. Record durably in `consume` regardless of `pending` membership, or union rather than replace.
- `Sources/OnymIOS/Group/JoinRequestApprover.swift:673` — `consumeRequestAndSiblings` decrypts every request in the store per Approve/Decline. Combined with the uncapped `pending` array the PR body already flags, a leaked link gives an attacker unbounded memory plus O(n) seal-opens per admin tap. Worth capping `pending` in this PR rather than a follow-up, since the link is now long-lived by design.
- `Sources/OnymIOS/Group/JoinRequestApprover.swift:352` — the reinstall path is one-sided, and correctly so per `IncomingMessageDispatcher.applyAnnouncement:637` (dedup on BLS hex bails before the merge). The rejoining member gets the snapshot; every *other* member keeps the stale `inboxPublicKey`/`sendingPubkey` and can't route to them. Pre-existing receiver behavior, but the PR advertises reinstall recovery — worth a note in the type doc so it isn't read as fixed.
- `Sources/OnymIOS/Group/IntroCapability.swift:222` — `.sortedKeys` only holds if the Android twin also sorts; confirm onymchat/onym-android#201 does, otherwise the "one capability, one link string" invariant is iOS-local.
- Nits: `ShareInviteFlow.InviteRow.createdAt` is stored and never rendered; the comment at `ShareInviteView.swift:98` ("A literal key so it localizes") sits above the `Text(label)` branch it doesn't apply to — move it onto the `Text("Older link")` branch.

PR #216, opened 2026-08-08
feat(moderation): OnymModeration package — domain, authority seams, DeviceCheck gate-check rail

rinat-enikeev:
## What

First slice of the client-side moderation seat from the onym-system spec ([Moderation.md](https://github.com/onymchat/onym-system/blob/main/moderation/Moderation.md), [Moderation-DeviceCheck.md](https://github.com/onymchat/onym-system/blob/main/moderation/Moderation-DeviceCheck.md)): a new `Packages/OnymModeration` SPM package with the domain model, the swappable-authority abstraction, and the Apple DeviceCheck profile's client half. No UI, no app wiring — the package compiles standalone and is linked into the app target (unused) via `project.yml`.

## Design

**Authorities are data, not code.** A signed directory (`KnownAuthoritiesFetcher`, same shape/trust story as `KnownRelayersFetcher`) pins each authority's operator key; `AuthorityManifestFetcher` fetches the manifest as exact bytes, hashes them (SHA-256 of fetched bytes = the `manifestHash` the mandate pins), and verifies the manifest signature against the directory-pinned key (soft mode via `ModerationTrust`, mirroring `ContractsTrust`). Switching authorities = signing a fresh `ModerationMandate`; `ModerationRepository.consent(to:)` is one path for onboarding and switching, and deactivated records keep their consented manifest snapshot byte-identical forever.

**The client never computes ban state.** `EnforcementBackendClient` is the seam for the vendor backend that holds the Apple `.p8` key (the only party that can read/write the DeviceCheck bits). The client's whole authority is: generate a `DCDevice` token, present it with an identity signature in one session, render the backend's `GateCheckResult`, and degrade toward blocking — `GateCheckRepository` owns the launch + P1D cadence with a P3D offline grace window on the persisted last-known state, then `gate-check-required`. Never toward unmoderated operation.

**Verdicts are validated mechanically.** `VerdictValidator` is a pure function over the §5.6 constraints (mandate ref, class-in-mandate, marks/disposition consistency, `banExpires` unless permanent, `executeAfter` = `decidedAt`/`appealDeadline` per appeal effect, early-write refusal via `.storeUntilExecuteAfter`) so a malformed sanction is never rendered as legitimate.

**Stubs are honest.** `StubEnforcementBackendClient` countersigns with a sentinel that keeps records `countersigned: false` (a stub mandate can never pass as spec-valid), issues a stable local `deviceBinding`, and is documented as the only implementation allowed to answer `.clear` to a nil-token check (simulators must stay developable; a real backend must answer `.checkRequired`).

## Provisional (flagged in code)

- No canonical JSON exists in the draft spec — `signingBytes()` uses sorted-keys JSON and is documented as provisional; mandates signed now will need re-consent when the spec fixes canonical form.
- Interface countersignature and real bit reads are impossible client-only by design; the seams (`countersignMandate`, `gateCheck`) are shaped for the real backend to slot in.

## Stack

- **PR-1 (this): domain + seams + stubs**
- PR-2: `OnymModerationUI` (consent flow, settings, ban/gate screens)
- PR-3: app wiring (identity signing adapter, composition root, RootView gate, UITest fakes)
- PR-4: tests

## Verification

- `xcodebuild -scheme OnymModeration -destination 'generic/platform=iOS Simulator' build` — succeeds
- Full app build with the package linked — succeeds

### DIFF
--- a/Packages/OnymModeration/Package.swift
+++ b/Packages/OnymModeration/Package.swift
@@ -1,0 +1,21 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
+    name: "OnymModeration",
+    platforms: [.iOS("18.0")],
+    products: [
+        .library(name: "OnymModeration", targets: ["OnymModeration"]),
+    ],
+    dependencies: [
+        .package(path: "../OnymFoundation"),
+    ],
+    targets: [
+        .target(
+            name: "OnymModeration",
+            dependencies: [
+                "OnymFoundation",
+            ]
+        ),
+    ]
+)
--- a/Packages/OnymModeration/Sources/OnymModeration/AuthorityKey.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/AuthorityKey.swift
@@ -1,0 +1,64 @@
+import Foundation
+import CryptoKit
+
+/// The one parser for authority signing keys. Both forms feed through
+/// it — the directory listing's base64 `operatorPublicKeyBase64` and the
+/// manifest's `onym:key:<hex>` `operator` field — so the fetcher can
+/// require the manifest to declare *byte-for-byte* the key the directory
+/// pins (`AuthorityManifestFetcher`), and verdict verification can key
+/// on that same parsed key (`VerdictValidator`). One parser, one trust
+/// root.
+public enum AuthorityKey {
+    /// Prefix of the spec's key-reference form, `onym:key:<hex>`.
+    public static let referencePrefix = "onym:key:"
+
+    /// Raw Ed25519 public key bytes from an `onym:key:<hex>` reference
+    /// (32 bytes, hex-encoded — the form mandates carry). Any other
+    /// shape throws: a reference this client can't parse is not a key it
+    /// can pin consent to.
+    public static func rawBytes(fromReference reference: String) throws -> Data {
+        guard reference.hasPrefix(referencePrefix) else {
+            throw ModerationError.keyInvalid("not an \(referencePrefix) reference")
+        }
+        let hex = reference.dropFirst(referencePrefix.count)
+        guard hex.count == 64, hex.allSatisfy(\.isHexDigit) else {
+            throw ModerationError.keyInvalid("\(referencePrefix) reference is not 32 hex-encoded bytes")
+        }
+        var bytes = Data(capacity: 32)
+        var index = hex.startIndex
+        while index < hex.endIndex {
+            let next = hex.index(index, offsetBy: 2)
+            guard let byte = UInt8(hex[index..<next], radix: 16) else {
+                throw ModerationError.keyInvalid("\(referencePrefix) reference is not hex")
+            }
+            bytes.append(byte)
+            index = next
+        }
+        return bytes
+    }
+
+    /// Raw Ed25519 public key bytes from the directory listing's base64
+    /// encoding.
+    public static func rawBytes(fromBase64 base64: String) throws -> Data {
+        guard let bytes = Data(base64Encoded: base64), bytes.count == 32 else {
+            throw ModerationError.keyInvalid("operator key is not 32 base64-encoded bytes")
+        }
+        return bytes
+    }
+
+    public static func publicKey(fromReference reference: String) throws -> Curve25519.Signing.PublicKey {
+        try publicKey(rawBytes: rawBytes(fromReference: reference))
+    }
+
+    public static func publicKey(fromBase64 base64: String) throws -> Curve25519.Signing.PublicKey {
+        try publicKey(rawBytes: rawBytes(fromBase64: base64))
+    }
+
+    private static func publicKey(rawBytes: Data) throws -> Curve25519.Signing.PublicKey {
+        do {
+            return try Curve25519.Signing.PublicKey(rawRepresentation: rawBytes)
+        } catch {
+            throw ModerationError.keyInvalid("not an Ed25519 public key")
+        }
+    }
+}
--- a/Packages/OnymModeration/Sources/OnymModeration/AuthorityManifest.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/AuthorityManifest.swift
@@ -50,8 +50,10 @@ public struct AuthorityManifest: Codable, Sendable, Equatable {
     /// mandates and verdicts.
     public let componentId: String
     public let seat: String
-    /// `onym:key:<authority-identity>` — the verdict/manifest signing
-    /// key. `operator` is the spec's field name.
+    /// `onym:key:<hex>` — the verdict/manifest signing key. `operator`
+    /// is the spec's field name. Not an independent trust root: the
+    /// fetcher requires it to equal the directory-pinned key byte for
+    /// byte (`AuthorityManifestFetcher`).
     public let operatorKey: String
     public let moderationProfileId: String
     public let violationClasses: [ViolationClass]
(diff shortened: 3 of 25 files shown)


dev-onym (commented):
REQUEST_CHANGES — solid architecture and honest stubs, but the manifest signature scheme is forgeable and several consent-time constraints the comments claim are enforced aren't.

- **`AuthorityManifestFetcher.swift:91` — signature check is bypassable by injection.** `bytesExcludingSignature` strips *every* occurrence of the base64 signature string, so an attacker can insert extra copies of that same string into any JSON string value: the excised bytes are byte-identical to the original, the Ed25519 check passes, but the decoded manifest differs. Base64 chars are all JSON-string-safe, so `operator`, `appellate`, `newHolderAppeal`, or a class `definition` (the content address the user is consenting to) can be silently rewritten and still "verify". Even under `enforceManifestSignatures = true` this gives no integrity. Fix by verifying over a re-encode of the decoded manifest with `signature` omitted (or a detached `.sig` like `SignedAsset`), not string surgery. Also note the current form breaks entirely if a signature value coincidentally appears elsewhere in the doc.
- **`AuthorityManifestFetcher.swift:48` — `manifest.operatorKey` is never bound to `listing.operatorPublicKeyBase64`.** Only `componentId` is cross-checked. The manifest signature is verified against the *directory* key, but the pinned snapshot then carries a possibly different `operator` — and `VerdictValidator.swift:30` documents the verdict verifier as being keyed on "the consented manifest's operator key". So the two trust paths can diverge: a manifest naming an attacker key would have its verdicts accepted. Cross-check at fetch, and add the `onym:key:` → raw-bytes parse helper that PR-3 will otherwise have to invent.
- **`ModerationRepository.swift:122` — no consent-time manifest validation.** `validUntil` is never compared to `clock()` (an expired manifest is consented to and pinned), `moderationProfileId` is never checked against the DeviceCheck profile (`ModerationError.unsupportedProfile` is declared and never thrown anywhere), and §5.2 constraint 2 — a `permanent` class requires an external `appellate` — is unchecked despite `ISO8601Duration.swift:52` asserting it's "checked at consent". Either implement these three or move the comments to a TODO so the next reader doesn't trust them.
- **`VerdictValidator.swift:37` — missing two cheap binding checks.** Nothing asserts `consented.manifestHash == mandate.manifestHash`, so passing a non-consented manifest silently changes ban validity (permanent/appealEffect drive `validateBan`); and `verdict.accusedKeys` is never required to contain `mandate.user`. Both are one-liners in a validator whose whole value is mechanical completeness.
- **`Verdict.swift:104` — hand-mirrored `Unsigned` struct will drift.** Any field added to `Verdict` and forgotten here silently falls outside the signed bytes, with no compile error. Prefer the mandate's approach (encode `self` and drop the signature field) or add a test that asserts the two key sets match.
- **`GateCheckRepository.swift:141` — the P1D cadence won't hold on iOS.** A single `Task.sleep(86_400s)` in a suspended app fires arbitrarily late and there's no foreground re-check hook, so a warm resume after days offline serves stale state until the sleep returns. Drive the interval from `lastSuccessAt` vs. wall clock on foreground/`checkNow` instead.
- **`GateCheckRepository.swift:155` — `.notMandated` is a non-blocking escape hatch.** Losing the mandate record (plain `UserDefaults`, per `MandateStore`) turns a persisted `.banned` into `.notMandated`. Harmless once the real backend reads bits, but with the stub it's a one-key bypass; worth an explicit comment or reconciling against persisted gate state.
- Minor: `StubEnforcementBackendClient.swift:17` documents a `--moderation-scenario` launch argument that no code reads (scenario is init-only); `enrollDevice` has a benign UUID race on concurrent calls.

dev-onym (commented):
**COMMENT** — solid, well-documented seam work; nothing wired in yet, but four issues should land before PR-3 flips this on.

- `GateCheckRepository.swift:239` — the `catch` maps *every* failure to `.unreachable`, including reachable-backend errors (non-2xx, decode failure, auth rejection) and purely local ones (`signer.sign` throwing on a locked keychain). All of them enter the offline-grace path and keep serving a cached `.clear` for up to 3 days. The doc on `AttemptOutcome.unreachable:107` says it is "NOT a backend refusal", but the protocol gives the client no way to tell them apart. Either classify here (`URLError` → unreachable, everything else → `.checkRequired`) or widen `AttemptOutcome`; as written this is the one path that degrades toward operation.
- `GateCheckRepository.swift:196` — the generation guard drops results by *start* order, not restrictiveness. A slow check that returns `.banned` is discarded if a later-started check already completed `.unreachable`-within-grace on a cached `.clear`. That is precisely the reopen-the-app failure the guard was written to prevent, just inverted. Prefer coalescing to a single in-flight check, or "most restrictive wins within a round".
- `Verdict.swift:104` vs `AuthorityManifest.swift` — the manifest path pins **exact fetched bytes** (`SignedManifest.rawBytes`) and hashes those; the verdict path re-encodes the decoded object with `.sortedKeys`/`.iso8601` and verifies against that. An authority whose JSON differs by a single byte (key order, fractional seconds, `null` for an absent field) will never verify. The PROVISIONAL note covers canonical form, but this is stronger: verdicts need their own `rawBytes` snapshot the way manifests have one, otherwise `enforceVerdictSignatures = true` is unreachable in practice.
- `MandateStore.swift:81`, `GateCheckRepository.swift:67` — `save` swallows encode failure with `try?` and returns. `ModerationRepository.consent` (`:204`) then reports success with an in-memory active mandate that won't survive relaunch, after the device was already enrolled at the backend. Make `save` throwing, or at minimum log.

Smaller:

- `VerdictValidator.swift:117` — `validateDismissal` doesn't reject `banExpires`/`executeAfter`/`appealDeadline` the way `validateOpenCase` does; a dismissal carrying sanction fields validates.
- `ISO8601Duration.swift:28` — `Int("+5")` succeeds, so `"P+5D"` parses as 5 days. Non-conforming input accepted silently.
- `ModerationRepository.swift:169` — consent auto-selects every class in the manifest. Fine as a default, but PR-2 should make this an explicit user choice rather than inherit it.
- `GateCheckRepository.start()` — nothing cancels `loopTask` if the actor is released without `stop()`; the `[weak self]` task then sleeps forever. Minor leak.

Not blocking, noted for the record: manifest and directory verification are both in soft mode (`ModerationTrust`, `ContractsTrust.enforceSignatures = false`), so today a MITM can serve arbitrary violation classes as long as the `operator` field matches the directory pin — consistent with the repo-wide posture, not a regression. I could not build or run the tests here (no Xcode on this host), so the verification claims in the PR body are unverified by me.

ux-onym (commented):
**COMMENT**

- `Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift:282-293` — `VerdictValidator` is defined but has **zero call sites** anywhere in the package (confirmed via grep — no `VerdictValidator()` instantiation outside its own file/doc comments). `GateCheckRepository.status(for:)` passes `.banned(state)` straight through from the backend into `GateStatus`, and `BanState.verdict` is documented as "validated client-side (`VerdictValidator`) before display" — but nothing invokes it. The PR's own design section states "Verdicts are validated mechanically... so a malformed sanction is never rendered as legitimate," but the one piece of *wired* runtime code in this PR that touches verdicts doesn't enforce that. Since PR-3 will hang UI off `GateCheckRepository.snapshots`/`currentStatus()` directly, this gap will silently carry forward unless explicitly remembered. Worth wiring now (or at minimum tracking explicitly in the PR-3/PR-4 stack notes) rather than leaving the validator orphaned.

- No tests at all for `VerdictValidator` (only `ModerationSigningPayloadTests` is included, covering the signing-byte mirrors). Given this is the most intricate/security-relevant logic in the package (derived `executeAfter`/`banExpires`/`appealDeadline` arithmetic per appeal effect), and it's currently unreferenced besides, it'd be good to at least confirm the stack's "PR-4: tests" item explicitly covers it rather than it being forgotten because nothing calls it yet.

- `GateCheckRepository.swift:~155` (`start()`) — the cadence loop captures `[weak self]` and loops on `Task.isCancelled`, but if the repository is deallocated without an explicit `stop()`, `Task.isCancelled` never becomes true (nothing cancels the task), so the loop spins forever doing no-op `self?.checkNow()` calls. Low impact if this is a long-lived singleton (likely, given the design), but worth a doc note or defensive `guard self != nil else { return }`-and-break.

Everything else — key parsing (`AuthorityKey`, `ComponentReference`), manifest pin/signature verification, mandate/verdict signing-byte mirrors (nicely guarded by drift tests), gate-check cadence/grace arithmetic, and the stub's honesty guarantees — looks correct and carefully reasoned. No blocking correctness issues found in the wired code paths that do exist.

PR #217, opened 2026-08-08
feat(moderation): OnymModerationUI package — consent, settings, ban and gate screens

rinat-enikeev:
## What

Stacked on #216. New `Packages/OnymModerationUI` SPM package: every screen of the moderation seat's client UX, plus the (hidden until wired) Settings row.

- **`ModerationConsentView` + `ModerationConsentFlow`** — the consent surface (Moderation.md §8 obligation 1): full-screen, every violation class with all five terms (response window, decision deadline, ban term, appeal window, appeal effect), appellate + new-holder path, the manifest hash being pinned, one explicit "I agree and sign". Serves both onboarding (`.onboarding`, non-dismissable) and later switching (`.switching`).
- **`ModerationSettingsView`/`Flow`** — current authority with its pinned terms and countersignature state, open cases, read-only mandate history (old mandates keep their consented hashes), "Switch authority" → fresh consent.
- **`BannedView`** — the conforming enforcement UX: verdict reference, class, reasoning address, authority contact, expiry (or permanent + appellate note), Appeal button and a distinct "I'm this device's new owner" button. A silent brick is nonconforming; this screen is the alternative.
- **`GateCheckRequiredView`** — blocking "verification required" screen with retry, for every degraded path (offline past grace, token invalid, no attestation).
- **`OpenCaseBanner` + `CaseNoticeDetailView`** — non-blocking (the case-open mark must not degrade service, §5.5).
- **`ModerationGateFlow`** — merges repository snapshots into the single `RootGate` the app root switches on. Two documented softenings while nothing real is deployed: consent only blocks when the authority directory actually yields entries (an interface without a designated authority answers to its distribution channel, §11.12), and `.checking` renders the app rather than blocking launch on the first gate answer.

`SettingsView` gains optional moderation factories (row hidden until the app passes them — next PR), so this PR leaves app behavior unchanged.

## Stack

- PR-1: #216 — domain + seams + stubs
- **PR-2 (this): UI package**
- PR-3: app wiring
- PR-4: tests

### DIFF
--- a/Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift
+++ b/Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift
@@ -80,6 +80,8 @@ public struct SettingsContentTile<Content: View>: View {
 /// Section label above a card. All-caps, mid-grey, small letterspacing.
 public struct SettingsSectionLabel<Trailing: View>: View {
     let text: LocalizedStringKey
+    /// Runtime data rather than UI copy — see `SettingsRow.verbatimTitle`.
+    var verbatimText: String? = nil
     @ViewBuilder var trailing: () -> Trailing
 
     // The trailing-content variant has no external consumer today, so
@@ -89,9 +91,15 @@ public struct SettingsSectionLabel<Trailing: View>: View {
         self.trailing = trailing
     }
 
+    init(verbatimText: String, @ViewBuilder trailing: @escaping () -> Trailing = { EmptyView() }) {
+        self.text = ""
+        self.verbatimText = verbatimText
+        self.trailing = trailing
+    }
+
     public var body: some View {
         HStack(alignment: .bottom) {
-            Text(text)
+            (verbatimText.map { Text(verbatim: $0) } ?? Text(text))
                 .font(.system(size: 12.5, weight: .medium))
                 .foregroundStyle(OnymTokens.text2)
             Spacer()
@@ -107,15 +115,28 @@ extension SettingsSectionLabel where Trailing == EmptyView {
     public init(_ text: LocalizedStringKey) {
         self.init(text, trailing: { EmptyView() })
     }
+
+    /// Label from runtime data (e.g. an authority-supplied class id).
+    public init(verbatim text: String) {
+        self.init(verbatimText: text, trailing: { EmptyView() })
+    }
 }
 
 /// Footnote text beneath a card — italic-equivalent grey copy used as
 /// section explanations.
 public struct SettingsFootnote: View {
     let text: LocalizedStringKey
+    /// Runtime data rather than UI copy — see `SettingsRow.verbatimTitle`.
+    var verbatimText: String? = nil
     public init(_ text: LocalizedStringKey) { self.text = text }
+    /// Footnote from runtime data (e.g. an already-localized error
+    /// message, which must not be re-looked-up as a key).
+    public init(verbatim text: String) {
+        self.text = ""
+        self.verbatimText = text
+    }
     public var body: some View {
-        Text(text)
+        (verbatimText.map { Text(verbatim: $0) } ?? Text(text))
             .font(.system(size: 12.5))
             .foregroundStyle(OnymTokens.text2)
             .lineSpacing(2)
@@ -129,9 +150,16 @@ public struct SettingsFootnote: View {
 /// shows beneath the nav bar on each top-level screen.
 public struct SettingsLargeTitle: View {
     let text: LocalizedStringKey
+    /// Runtime data rather than UI copy — see `SettingsRow.verbatimTitle`.
+    var verbatimText: String? = nil
     public init(_ text: LocalizedStringKey) { self.text = text }
+    /// Title from runtime data (e.g. an authority's published name).
+    public init(verbatim text: String) {
+        self.text = ""
+        self.verbatimText = text
+    }
     public var body: some View {
-        Text(text)
+        (verbatimText.map { Text(verbatim: $0) } ?? Text(text))
             .font(.system(size: 34, weight: .bold))
             .foregroundStyle(OnymTokens.text)
             .tracking(-0.75)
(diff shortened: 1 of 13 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — solid, well-documented package; one integrity bug in the consent path plus a few leaks/dead ends.

- **The hash the user reviews is not the hash that gets signed.** `ModerationConsentFlow.selectedAuthority` fetches the manifest itself (`ModerationConsentFlow.swift:97`) and displays `manifestHash`, then `tappedAgree` calls `repository.consent(to:)` (`ModerationConsentFlow.swift:116`), which fetches the manifest *again* (`ModerationRepository.consent`) and pins whatever comes back. Between the two fetches the authority can serve different terms — the mandate then binds to terms the user never saw, which is precisely what the "Your consent binds to this hash of these exact terms" footnote (`ModerationConsentView.swift:143`) promises can't happen. Two network round-trips for one consent is also wasteful. Fix: have `consent` accept the already-fetched `SignedManifest` (or a `expectedHash:` parameter it verifies and rejects on mismatch).
- **Flows are started but never stopped.** `ModerationConsentView`/`ModerationSettingsView` call `flow.start()` in `.task` but nothing calls `stop()` (`ModerationConsentFlow.swift:83`, `ModerationSettingsFlow.swift:46`). `start()` spawns unstructured `Task`s that outlive the view, so each Settings→Moderation push or Switch-authority sheet leaks a flow plus a permanently-registered `AsyncStream` continuation in the repository actor (`onTermination` never fires while the for-await loop holds the stream). Add `.onDisappear { flow.stop() }` or a `deinit`-safe teardown.
- **New-owner button can be a no-op.** `BannedView.swift:76-80`: with `newHolderURL == nil` and the defaulted `onNewHolderClaim = {}` (`BannedView.swift:18`), tapping the most safety-critical affordance on the screen does nothing at all. Either drop the default, or hide/disable the button and fall back to showing `authorityContact` when neither path exists.
- **Stale open-case banner.** `ModerationSettingsFlow.swift:39` only assigns `openCases` inside `if case .operational`; once a case closes and the status moves to any other case (or to `.operational([])`… that one is fine), the previously captured notices persist. Handle the non-operational branches by clearing.
- **Date used as `ForEach` identity.** `ModerationSettingsView.swift:42` keys history rows on `mandate.acceptedAt`; the repository stamps `acceptedAt` and `createdAt` from an injectable clock, so two records can collide (trivially in tests with a fixed clock). Use `mandate.manifestHash` + `acceptedAt`, or make `MandateRecord: Identifiable`.
- **Runtime strings as `LocalizedStringKey`.** `ModerationConsentView.swift:97`, `ModerationSettingsView.swift:44`, plus `LocalizedStringKey(error)` and the `classId.uppercased()` section label. These are authority-controlled values being looked up as localization keys — an authority named e.g. "Settings" would render a translated string. Pass `String` through `Text(verbatim:)` / `subtitle:` instead.
- **`.done` has no exit in `.switching` mode.** `ModerationConsentView.swift:28` renders "Mandate signed" with no Done button and no toolbar; only swipe-to-dismiss closes the sheet, since `onConsented` is supplied by the app factory and can't clear `showSwitchConsent` (`ModerationSettingsView.swift:63`). Give the sheet a dismiss action.
- Minor: `OpenCaseBanner.swift:19` says "N cases are open" but only ever opens `notices.first` — no way to reach the rest. `selectedAuthority` shows no pending state, so a double-tap starts two manifest fetches (`ModerationConsentFlow.swift:97`). `GateCheckRequiredView.swift:55` offers "Try again" for `.attestationUnavailable`, where retrying can never succeed (simulator/enterprise builds) — worth suppressing the button for that reason.
- The `.checking` case comment says "keep the launch screen" (`ModerationGateFlow.swift:20`) while the `recompute` doc says `.checking` renders the app (`ModerationGateFlow.swift:104`). Reconcile before PR-3 wires the root, since the app will follow one of them.

dev-onym (commented):
**COMMENT** — solid, well-documented package; one real state race on the consent gate plus a few UX/spec gaps worth fixing before wiring.

- `ModerationConsentFlow.swift:88-94` — the manifest fetch isn't tied to the selection that started it. Tap an authority, tap **Back** while the fetch is in flight: the late task sets `reviewingManifest` and `step = .reviewingManifest` with `selectedListing == nil`, so `tappedAgree` (`:109`) hits its `guard` and returns — the consent gate's primary button silently does nothing, with no error shown. Tap A then B quickly and A's terms render under B's name (`ModerationConsentView.swift:155`); agreeing then fails the repository's componentId check as a generic "Signing failed". Take a generation token (or cancel the prior task) and drop stale completions, as `GateCheckRepository.checkNow` already does.
- `OpenCaseBanner.swift:19-46` — with N>1 notices the copy says "N moderation cases are open" but only `notices.first` is presentable; the other notices' deadlines and evidence reference are unreachable from anywhere in the app. §5.5 requires the served notice to be presented — a list (or paging in `CaseNoticeDetailView`) is needed.
- `BannedView.swift:120` vs `:74` — the expiry line always says "You can appeal below", but the appeal button only renders when `state.appealURL != nil`. `appealURL` is optional on `BanState`, so a ban served without one points at a button that isn't there; and the contact fallback footnote (`:108`) is gated on `!hasNewHolderPath`, so a ban with a new-holder URL and no appeal URL shows no appeal route at all. Gate the copy on `appealURL`, and show the contact footnote whenever the appeal path is missing.
- `BannedView.swift:61` — `verdict.reasoning` is a content address and is the one thing the banned user most needs to open; render it as a link when it's an https URL, the way `ModerationConsentView.definitionLink` (`:281`) already does.
- `OpenCaseBanner.swift:55` — `@retroactive Identifiable` on a type from the sibling `OnymModeration` package you own. Declare `id` in the domain module instead; a later conformance there silently collides with this one.
- `ModerationConsentView.swift:305` — `ISO8601Duration.display` handles only 1-vs-n; `"\(days) days"` needs a plural rule in the strings catalog for non-English locales. Also consider hosting it next to the type rather than as an internal extension in the UI package.
- `ModerationGateFlow.swift:110` — both softenings are honestly documented, but note the second-order effect: with no mandate the app renders normally until the directory fetch lands, then flips a running session into the blocking consent surface. Worth an explicit test in PR-4 alongside the `.notMandated` invariant already flagged in `GateStatus`.
- `SettingsDesign.swift:333-378` / `:449-478` — the `titleText:` initializers duplicate the primary ones verbatim; the distinct-label rationale is right, but the bodies could funnel through one private designated init.

PR #218, opened 2026-08-08
feat(moderation): wire the moderation gate into the app

rinat-enikeev:
## What

Stacked on #217. Turns the packages on:

- **`OnymIdentity`**: new `IdentityRepository.signWithStellarKey(_:)` — Ed25519 detached signature with the derived Stellar key, same posture as `blsSecretKey()`: derived per call from the Keychain snapshot, the private key never leaves the repository. Adapted into the moderation seam by `IdentityModerationSigner` (`user` = `onym:key:<stellar-pubkey-hex>`).
- **`OnymIOSApp`**: composes the moderation stack — production uses `GitHubReleasesKnownAuthoritiesFetcher` + `URLSessionAuthorityManifestFetcher` + UserDefaults stores + `DCDeviceAttestationProvider` + `StubEnforcementBackendClient` (no enforcement backend exists yet); repositories start after identity bootstrap so gate sessions carry an identity signature.
- **`RootView`**: switches on the gate — banned and gate-check-required replace the app wholesale, consent covers it (fullScreenCover, non-dismissable), an open case only banners it. Foregrounding triggers a fresh gate check (covers the P1D cadence across backgrounding).
- **UITest harness**: canned authority + fixture manifest, seeded in-memory mandate store (so existing UI tests aren't blocked by the consent gate), `--moderation-scenario clear|case-open|banned|check-required` drives the stub, `--moderation-needs-consent` starts unconsented to exercise the consent flow.

## Behavior in production today

No authorities directory is published, so the consent gate never blocks and the stub gate answers `.clear`: user-visible behavior is unchanged except the (factory-gated) Moderation row in Settings. The enforcement rail is live end-to-end the moment a real directory + backend exist.

## Verified

Simulator smoke runs: `banned` scenario shows the full ban screen (verdict ref, expiry, appeal + new-holder), `--moderation-needs-consent` shows the authority picker, default seeded run boots to the normal app.

## Stack

- PR-1: #216 — domain + seams + stubs
- PR-2: #217 — UI package
- **PR-3 (this): app wiring**
- PR-4: tests

### DIFF
--- a/Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift
+++ b/Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift
@@ -252,6 +252,25 @@ public actor IdentityRepository: InvitationEnvelopeDecrypting, InvitationEnvelop
         return snapshot.blsSecretKey
     }
 
+    /// Ed25519 detached signature over `message` with the
+    /// currently-selected identity's derived Stellar signing key. Used
+    /// by the moderation layer to sign mandates and gate-check
+    /// sessions. Same posture as `blsSecretKey()`: the private key is
+    /// derived per call from the Keychain snapshot and never leaves
+    /// this repository — only the signature does.
+    public func signWithStellarKey(_ message: Data) throws -> Data {
+        guard let currentID else {
+            throw IdentityError.identityNotLoaded
+        }
+        guard let snapshot = try keychain.read(currentID) else {
+            throw IdentityError.identityNotLoaded
+        }
+        let privateKey = try Self.stellarSigningPrivateKey(
+            fromNostrSecret: snapshot.nostrSecretKey
+        )
+        return try privateKey.signature(for: message)
+    }
+
     // MARK: - Streams
 
     public nonisolated var snapshots: AsyncStream<Identity?> {
--- a/Sources/OnymIOS/AppDependencies.swift
+++ b/Sources/OnymIOS/AppDependencies.swift
@@ -6,6 +6,7 @@ import OnymInbox
 import OnymRecovery
 import OnymChatsUI
 import OnymSettings
+import OnymModerationUI
 
 /// App-wide composition root. Constructed exactly once by `OnymIOSApp`
 /// and threaded down to views via `RootView`. Each member is a factory
@@ -68,4 +69,10 @@ struct AppDependencies {
     /// `GroupNamePayload` out to members. Backs the admin-only rename
     /// affordance in `ChatMembersView`.
     let setGroupName: @MainActor (_ groupIDHex: String, _ name: String) async -> Void
+    /// Single shared instance — the app root switches on its gate
+    /// (consent / banned / check-required / operational) and Settings
+    /// observes the same state, so a factory would split them.
+    let moderationGateFlow: ModerationGateFlow
+    let makeModerationConsentFlow: @MainActor (ModerationConsentFlow.Mode) -> ModerationConsentFlow
+    let makeModerationSettingsFlow: @MainActor () -> ModerationSettingsFlow
 }
--- a/Sources/OnymIOS/IdentityModerationSigner.swift
+++ b/Sources/OnymIOS/IdentityModerationSigner.swift
@@ -1,0 +1,23 @@
+import Foundation
+import OnymIdentity
+import OnymModeration
+
+/// Adapts `IdentityRepository` to the moderation package's signer
+/// seam. The mandate's `user` key is the identity's derived Stellar
+/// Ed25519 public key (`onym:key:<hex>`); signatures come from the
+/// matching private key, which never leaves the repository.
+struct IdentityModerationSigner: ModerationSigner {
+    let repository: IdentityRepository
+
+    func userKeyID() async throws -> String {
+        guard let identity = await repository.currentIdentity() else {
+            throw ModerationError.notImplemented("no identity bootstrapped yet")
+        }
+        let hex = identity.stellarPublicKey.map { String(format: "%02x", $0) }.joined()
+        return "onym:key:\(hex)"
+    }
+
+    func sign(_ message: Data) async throws -> Data {
+        try await repository.signWithStellarKey(message)
+    }
+}
--- a/Sources/OnymIOS/OnymIOSApp.swift
+++ b/Sources/OnymIOS/OnymIOSApp.swift
@@ -11,6 +11,8 @@ import OnymInbox
 import OnymRecovery
 import OnymChatsUI
 import OnymSettings
+import OnymModeration
+import OnymModerationUI
 
 @main
 struct OnymIOSApp: App {
@@ -32,6 +34,11 @@ struct OnymIOSApp: App {
     private let groupStateVerifier: GroupStateVerifier
     private let introKeyStore: any IntroKeyStore
     private let introRequestStore: any IntroRequestStore
+    /// Moderation seat: authority designation + mandate lifecycle,
+    /// and the DeviceCheck gate-check cadence. The enforcement
+    /// backend is a stub until one is deployed.
+    private let moderationRepository: ModerationRepository
+    private let gateCheckRepository: GateCheckRepository
     /// Factory for the per-request SEP contract transport. Normally
     /// `URLSessionSEPContractTransport`; swapped for an in-memory
     /// ledger-backed fake under `--ui-loopback` so a Tyranny group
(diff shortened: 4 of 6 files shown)


dev-onym (commented):
REQUEST_CHANGES — the wiring is clean, but the foreground gate check can run before identity bootstrap and hard-block the app.

- `Sources/OnymIOS/RootView.swift:61` — `scenePhase == .active` fires on the cold-launch `.inactive → .active` transition, i.e. potentially before `identityRepository.bootstrap()` (`OnymIOSApp.swift:664`, which is awaited *ahead* of ~8 other network starts before `gateCheckRepository.start()` at :661). With an active mandate, `signer.sign` then throws → `performAttempt` returns `.unreachable` → `derive` with no persisted state → `.gateCheckRequired(.neverChecked)` → the whole app is replaced by the blocking screen, recoverable only via Retry. That's a UI-test flake today (seeded `UITestMandateStore`) and a launch brick once real mandates exist. Suggest gating `appForegrounded()` on bootstrap completion, or better: have `GateCheckRepository.performAttempt` distinguish a local signer failure from backend unreachability — a client-side signing error shouldn't consume offline grace or degrade to blocking.
- `Sources/OnymIOS/RootView.swift:61-64` — every activation (control center pull, notification banner, incoming call) triggers a full gate check including a DeviceCheck token mint. Throttle to at most once per `policy.interval` rather than per activation; today the P1D loop plus unthrottled foregrounding are redundant.
- `Packages/OnymIdentity/.../IdentityRepository.swift:261` — `signWithStellarKey(_:)` is an unrestricted signing oracle over the identity's Ed25519 key, which is also used for invitation-envelope signatures (`:533`) and Stellar signing, with no domain separation on either side. Payloads (`GateCheckRequest.signedPayload` = token‖ISO8601, `mandate.signingBytes()` = authority-influenced JSON) are only non-confusable by accident of length. Prefix a context string (e.g. `"onym:moderation-v1"‖message`, or sign a tagged hash) before it ships against a real authority.
- `Sources/OnymIOS/OnymIOSApp.swift:660` — `moderationRepository.start()` sits behind `relayerRepository/contractsRepository/nostrRelays/blossomServers` starts in the same serial `.task`. If any of those GitHub fetches is slow, the authorities directory (and therefore the consent gate) is delayed by minutes. Worth hoisting the two moderation starts into their own `.task` that only awaits bootstrap.
- `Sources/OnymIOS/RootView.swift:59` — `showConsent` is derived only from `onChange`, so if `gate` is already `.needsConsent` at RootView's first body pass (view re-identity, or a future call site that starts the flow earlier), the cover never presents and an unconsented user gets the full app. `.fullScreenCover(item:)` over the gate, or seeding `showConsent` in the same `.task` that calls `start()`, removes the class of bug.
- `Sources/OnymIOS/IdentityModerationSigner.swift:14` — `ModerationError.notImplemented` is documented as "the implementation is a stub"; "no identity yet" is a transient runtime state and reads as a stub failure in consent-flow error copy. A dedicated case (or `IdentityError.identityNotLoaded` passthrough) is more honest.
- `Sources/OnymIOS/RootView.swift:107-108` — both factories are passed unconditionally, so the Settings Moderation row is always visible in production despite the PR describing it as factory-gated; with no published directory it lands on `noMandateCard` → an empty authority picker. Either intentional (say so) or gate on a flag.
- Nit: `OnymIOSApp.swift:113-183` — the `#else` branch duplicates the DEBUG non-UI-test branch verbatim; a single `makeModeration(...)` helper with only the UI-test substitution inside `#if DEBUG` would keep the two from drifting.

ux-onym (commented):
**APPROVE**

- `Sources/OnymIOS/OnymIOSApp.swift:659-660` — `moderationRepository.start()` / `gateCheckRepository.start()` correctly sequenced after `identityRepository.bootstrap()` (`OnymIOSApp.swift:622`) in the same `.task`, and both `start()` calls just spawn detached background tasks and return immediately, so no blocking. Ordering guarantee the PR claims (gate sessions carry an identity signature) checks out.
- `Sources/OnymIOS/IdentityModerationSigner.swift:11-17` — `userKeyID()`/`sign()` correctly delegate to the actor; `signWithStellarKey` (`IdentityRepository.swift:255-269`) mirrors `blsSecretKey()`'s posture (key derived per-call from Keychain snapshot, never cached). Verified `GateCheckRepository.checkNow()` never touches the signer when there's no active mandate, so the `notImplemented` throw path for a not-yet-bootstrapped identity is unreachable in practice.
- `Sources/OnymIOS/RootView.swift:47-49` — `showConsent` is set only inside `.onChange(of: gate)`, which doesn't fire on initial attach. This is currently safe because `moderationGateFlow.start()` is only ever triggered from this same view's `.task`, so `gate` can't already be non-`.checking` the first time the modifier attaches — but it's a latent footgun if `RootView` is ever reconstructed while the (persistent, app-lifetime) `ModerationGateFlow` already sits in `.needsConsent`. Minor; consider `.onChange(of:initial: true)` or syncing `showConsent` inside `.task` for robustness. Not blocking.
- `Sources/OnymIOS/UITestFakes.swift:328-364` (`UITestAuthorityManifestFetcher`/`UITestMandateStore`) — fixture manifest hash and seeded mandate's `manifestHash` are derived from the same `manifestJSON` bytes, so hash-pinning stays consistent for gate-check tests. No cryptographic signature check in the test fetcher, which is expected/intentional (mirrors how `URLSessionAuthorityManifestFetcher` is the only place that verifies the operator signature).
- Wiring in `AppDependencies.swift` / `RootView.swift` matches the optional-closure params already present in `OnymSettings/SettingsView.swift` (from #217) — no signature mismatches.

No blocking issues found; the stubbed backend/no-authorities-directory posture matches the "unchanged production behavior" claim in the PR body.

dev-onym (commented):
COMMENT — wiring is coherent and the DEBUG/production split is clean; the findings below are latent (they only bite once a real backend/identity switch is involved), plus a couple of small cleanups.

- `Sources/OnymIOS/IdentityModerationSigner.swift:20` — **identity switch breaks the gate-check signature.** `sign` always uses the *currently selected* identity's Stellar key, but `GateCheckRepository.performAttempt` sends `record.mandate.user` (`Packages/OnymModeration/.../GateCheckRepository.swift:222`) — the key that consented. Switch identity and the app sends `userKey = A` signed by key `B`; a conforming backend must reject that, so every multi-identity user lands in blocking `.gateCheckRequired` the moment a real backend ships. Nothing in `OnymIOSApp.swift:668` (the `currentIdentityID` listener) re-checks or re-consents on switch either. At minimum, sign as the identity the mandate names (pass an identity ID into `signWithStellarKey`), or document the mandate as bound to one identity and force re-consent on switch.
- `Sources/OnymIOS/RootView.swift:61` — `scenePhase == .active` fires on cold launch *and* on every return from inactive (Control Center, app switcher peek, system alert), so `appForegrounded()` races the launch check from `gateCheckRepository.start()` and then re-fires constantly. Each one mints a fresh `DCDevice` token plus a backend round trip. Suggest debouncing on `lastSuccessAt` (skip if a check succeeded within, say, a minute) rather than checking unconditionally.
- `Sources/OnymIOS/RootView.swift:58` — consent presentation is driven only by `onChange`, so it depends on `gate` never being `.needsConsent` before the modifier is installed. It holds today (`start()` runs in `.task`, initial value `.checking`), but `.fullScreenCover(item:)` keyed off the gate would make it structural instead of incidental.
- `Sources/OnymIOS/OnymIOSApp.swift:146-163` vs `:165-181` — the DEBUG-non-UI-testing branch and the `#else` branch are byte-identical; factor the production stack into one `makeProductionModerationStack(signer:)` helper so the two copies can't drift.
- `Sources/OnymIOS/IdentityModerationSigner.swift:14` — `ModerationError.notImplemented("no identity bootstrapped yet")` misuses a case that means "this client doesn't implement it". A dedicated case (or `.noMandate`-style sibling) would keep the diagnostic honest.
- `Sources/OnymIOS/UITestFakes.swift:310` — mandate and gate state are in-memory, but `StubEnforcementBackendClient()` still defaults to `UserDefaults.standard` for `deviceBinding`, so that one value survives across UI-test runs. Harmless today (the binding is deliberately stable), worth passing a scratch suite if you ever assert on enrollment.
- `Packages/OnymIdentity/.../IdentityRepository.swift:258` — new public "sign arbitrary bytes with the Stellar key" oracle. The other user of that key signs a bare 32-byte ephemeral pubkey (`:533`), so no collision today, but the moderation payloads are domain-separated and this API isn't — worth a note that callers must domain-separate.
- Behavior worth a product check, not a code fix: `.banned` / `.gateCheckRequired` replace the app wholesale, which also removes access to Settings → recovery-phrase backup. A banned user who never backed up can no longer reach their phrase.

PR #221, opened 2026-08-08
feat(moderation): canonical signing bytes for reports, responses and appeals

rinat-enikeev:
`Report`, `CaseResponse` and `AppealSubmission` each carry a `signature` field with no defined bytes to sign. The report path therefore can't be implemented without someone inventing a canonical form at the call site — and there is a specific wrong way to invent it that fails silently.

## The hazard

Foundation ships **two** `.sortedKeys` options that disagree:

| | sorts by |
|---|---|
| `JSONEncoder.OutputFormatting.sortedKeys` | UTF-8 byte order |
| `JSONSerialization.WritingOptions.sortedKeys` | case-insensitively |

Of the moderation boundary objects, `Report` is the only one whose keys collide under that difference: `reportId` and `reportVersion` precede `reporter` by byte order and follow it case-insensitively. `mandate`, `verdict` and `notice` sort identically under both rules — which is why the existing client and the enforcement backend have interoperated fine.

Byte order is the interoperable choice: serde_json, Go's `encoding/json` and Python's `sort_keys` all do it, so an authority written in any of them can reconstruct these bytes. A report signed the other way is rejected by all of them, and the failure surfaces as *"every report signature is invalid"* rather than as an encoding bug — which is a bad afternoon for whoever debugs it.

## What this PR does

- Adds `signingBytes()` to `Report`, `CaseResponse` and `AppealSubmission`, following the existing mirror-struct pattern from `ModerationMandate` and `Verdict`.
- Moves the encoder configuration into one `ModerationCanonicalEncoder`. There were three copies of "sortedKeys + iso8601" across the package; a canonical form with three definitions is one waiting to diverge. `ModerationMandate` and `Verdict` now route through it — no behaviour change, verified by the existing `ModerationSigningPayloadTests`.
- Adds `ModerationCanonicalOrderTests`, which pins the byte-order rule by reading key order out of the raw JSON (parsing into a dictionary would discard exactly the thing under test): the collision case, the full expected key list, a guard asserting `JSONSerialization` would order differently, and a check that mandate/verdict remain unaffected.

## Verified against the reference authority

Not just unit-tested — checked against the Rust authority in [onymchat/onym-moderation#2](https://github.com/onymchat/onym-moderation/pull/2), which reconstructs these bytes with serde_json:

| | |
|---|---|
| report signed with `Report.signingBytes()` | **accepted**, opened a case |
| same report signed over `JSONSerialization`-ordered bytes | `signature_invalid: reporter signature did not verify` |

Both directions, so the fix is load-bearing rather than cosmetic.

## Note on scope

I originally reported this as "the client's canonicalization ends in `JSONSerialization`, so report signatures will fail". That was wrong about `main` — there is no `JSONSerialization` anywhere in `OnymModeration`, and the existing signing forms were already byte-ordered and correct. The real gap was that reports had no defined signing bytes at all. This PR closes that gap and pins the rule so the wrong option can't be chosen later.

The `PROVISIONAL` caveat still stands: the draft spec fixes no canonical JSON form, so this agreement holds by construction between implementations. A spec decision naming one would let both sides drop the caveat.

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
@@ -123,6 +123,11 @@ public struct CaseStatus: Codable, Sendable, Equatable {
 /// window), and `accept-mandate` is authority-side. One implementation
 /// per authority; the repository picks the client for the mandated
 /// authority, which is what makes authorities swappable.
+///
+/// An HTTP implementation must put `CaseResponse.caseId` and
+/// `AppealSubmission.caseId` in both the request path and the signed body.
+/// The authority must reject a path/body mismatch; trusting either one alone
+/// would reopen the cross-case replay that signing `caseId` prevents.
 public protocol ModerationAuthorityClient: Sendable {
     func fileReport(_ report: Report) async throws -> ReportReceipt
     func respond(_ response: CaseResponse) async throws
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
@@ -19,11 +19,24 @@ public struct ReportReceipt: Codable, Sendable, Equatable {
 /// through the same object, so `submit-evidence` needs no separate
 /// operation client-side.
 public struct CaseResponse: Codable, Sendable, Equatable {
+    /// The case being answered. Carried in the object — and therefore
+    /// inside the signed bytes — rather than alongside it: a signed
+    /// statement that names no case can be lifted from the case it was
+    /// written for and replayed onto another, so an innocuous "that
+    /// wasn't me" would register as an answer to an accusation its
+    /// signer never saw.
+    public let caseId: String
     public let statement: String
     public let evidence: [EvidenceItem]
     public var signature: String
 
-    public init(statement: String, evidence: [EvidenceItem] = [], signature: String = "") {
+    public init(
+        caseId: String,
+        statement: String,
+        evidence: [EvidenceItem] = [],
+        signature: String = ""
+    ) {
+        self.caseId = caseId
         self.statement = statement
         self.evidence = evidence
         self.signature = signature
@@ -33,11 +46,12 @@ public struct CaseResponse: Codable, Sendable, Equatable {
     /// See `ModerationCanonicalEncoder` for why the ordering matters.
     public func signingBytes() throws -> Data {
         struct Unsigned: Encodable {
+            let caseId: String
             let statement: String
             let evidence: [EvidenceItem]
         }
         return try ModerationCanonicalEncoder.encode(
-            Unsigned(statement: statement, evidence: evidence)
+            Unsigned(caseId: caseId, statement: statement, evidence: evidence)
         )
     }
 }
@@ -51,11 +65,15 @@ public struct AppealSubmission: Codable, Sendable, Equatable {
         case newHolderClaim = "new-holder-claim"
     }
 
+    /// The case appealed. Signed, for the same replay reason as
+    /// `CaseResponse.caseId`.
+    public let caseId: String
     public let kind: Kind
     public let statement: String
     public var signature: String
 
-    public init(kind: Kind, statement: String, signature: String = "") {
+    public init(caseId: String, kind: Kind, statement: String, signature: String = "") {
+        self.caseId = caseId
         self.kind = kind
         self.statement = statement
         self.signature = signature
@@ -69,11 +87,12 @@ public struct AppealSubmission: Codable, Sendable, Equatable {
     /// the mandated user (§5.7).
     public func signingBytes() throws -> Data {
         struct Unsigned: Encodable {
+            let caseId: String
             let kind: Kind
             let statement: String
         }
         return try ModerationCanonicalEncoder.encode(
-            Unsigned(kind: kind, statement: statement)
+            Unsigned(caseId: caseId, kind: kind, statement: statement)
         )
     }
 }
@@ -106,8 +125,8 @@ public struct CaseStatus: Codable, Sendable, Equatable {
 /// authority, which is what makes authorities swappable.
 public protocol ModerationAuthorityClient: Sendable {
     func fileReport(_ report: Report) async throws -> ReportReceipt
-    func respond(caseId: String, _ response: CaseResponse) async throws
-    func appeal(caseId: String, _ submission: AppealSubmission) async throws
+    func respond(_ response: CaseResponse) async throws
+    func appeal(_ submission: AppealSubmission) async throws
     func queryStatus(caseId: String) async throws -> CaseStatus
 }
 
@@ -121,11 +140,11 @@ public struct StubModerationAuthorityClient: ModerationAuthorityClient {
         throw ModerationError.notImplemented("file-report")
     }
 
-    public func respond(caseId: String, _ response: CaseResponse) async throws {
+    public func respond(_ response: CaseResponse) async throws {
         throw ModerationError.notImplemented("respond")
     }
 
-    public func appeal(caseId: String, _ submission: AppealSubmission) async throws {
+    public func appeal(_ submission: AppealSubmission) async throws {
         throw ModerationError.notImplemented("appeal")
     }
 
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationAuthorityClient.swift
@@ -28,6 +28,18 @@ public struct CaseResponse: Codable, Sendable, Equatable {
         self.evidence = evidence
         self.signature = signature
     }
+
+    /// The bytes the accused signs: every field except `signature`.
+    /// See `ModerationCanonicalEncoder` for why the ordering matters.
+    public func signingBytes() throws -> Data {
+        struct Unsigned: Encodable {
+            let statement: String
+            let evidence: [EvidenceItem]
+        }
+        return try ModerationCanonicalEncoder.encode(
+            Unsigned(statement: statement, evidence: evidence)
+        )
+    }
 }
 
 /// An appeal filed within the declared window — or a new-holder claim,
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — key ordering is now pinned correctly, but the same class of silent interop failure survives in string escaping, and the two new signed objects aren't bound to their case.

- `ModerationCanonicalEncoder.swift:28` — **`/` escaping is the second half of this bug and isn't handled.** `JSONEncoder` escapes forward slashes as `\/` unless `.withoutEscapingSlashes` is set; serde_json, Go and Python all emit a bare `/`. `EvidenceItem.authenticityProof` is standard base64, whose alphabet includes `/` — an 86-char Ed25519 signature contains one ~74% of the time — and `disclosedContent` is arbitrary user text (URLs). The PR's Rust round-trip passed because the fixture proof `c2VuZGVyLXNpZ25hdHVyZQ==` happens to have no slash; `ModerationCanonicalOrderTests` never exercises one either. Fix: `outputFormatting = [.sortedKeys, .withoutEscapingSlashes]`, and add a fixture with `/` and non-ASCII to both the unit tests and the authority check. Cheap now, breaking once anything is signed.
- `ModerationAuthorityClient.swift:34,70` — **response and appeal signatures aren't bound to a case.** `respond(caseId:_:)` and `appeal(caseId:_:)` carry `caseId` out of band, so the signed bytes are just `{statement, evidence}` / `{kind, statement}`. A captured response replays verbatim into any other case against the same accused, and "I bought this phone" replays into any later case. `Report` avoids this via `reportId` + `filedAt`; these should include `caseId`, a `*Version` field for consistency with the other three payloads, and a timestamp for freshness.
- `ModerationCanonicalOrderTests.swift:83` — **the new mirror structs have no drift guard.** `ModerationSigningPayloadTests.swift:57-81` compares `signingBytes()` keys against the object's full encoded key set minus the signature; the new test pins a hardcoded list that mirrors `Unsigned`, not `Report`. Add a field to `Report` and forget `Unsigned` and it falls outside the signature with every test green. Extend the existing `encodedKeys(of:).subtracting(...)` pattern to `Report`, `CaseResponse`, `AppealSubmission`.
- `ModerationCanonicalOrderTests.swift:120-135` — `test_mandateAndVerdictKeysAreUnaffectedByTheOrderingDifference` tests neither mandate nor verdict. It sorts two hardcoded string literals and round-trips them through `JSONSerialization`; the docstring's claim that "a renamed field can't quietly change that" is not true — a rename leaves the literal stale and the test green. Derive the lists from `ModerationMandate.signingBytes()` / `Verdict.signingBytes()`.
- `Report.swift:75` — doc points at `SigningBytesTests`, which doesn't exist; the tests are `ModerationCanonicalOrderTests` and `ModerationSigningPayloadTests`.
- `ModerationCanonicalOrderTests.swift:26-28` (minor) — `keyOrder`'s depth counter counts `{`/`[` inside string values, so a future fixture with a brace in `disclosedContent` corrupts the extracted list rather than failing loudly.
- `Report.swift:36` (minor) — `EvidenceItem.context == nil` omits the key entirely rather than emitting `null`. Deterministic, but it's a rule the authority must match; worth one line in the `ModerationCanonicalEncoder` doc alongside the ordering rule.

The core ordering analysis holds up: `.sortedKeys` byte order is the right choice, the collision on `reportId`/`reportVersion` vs `reporter` is real, and collapsing three encoder copies into one is the right move — that consolidation is also what makes the slash fix a one-line change.

ux-onym (commented):
APPROVE

- Logic verified by hand: `Report`'s byte-order vs case-insensitive collision claim checks out (`reportId`/`reportVersion` contain uppercase `I`/`V` which sort before lowercase `e` in `reporter` under ASCII byte order, but after it case-insensitively) — `Packages/OnymModeration/Sources/OnymModeration/Report.swift:62-99`.
- `ModerationCanonicalEncoder` correctly centralizes the `.sortedKeys` + `.iso8601` config; `ModerationMandate.swift:83` and `Verdict.swift:135` route through it with no behavior change (encoder config identical to what was inlined before).
- `CaseResponse`/`AppealSubmission` gained a required `caseId` field bound into `signingBytes()`, closing a real replay hole (an unsigned-case statement could otherwise be replayed onto a different case) — `ModerationAuthorityClient.swift:21-97`.
- `ModerationAuthorityClient` protocol dropped the now-redundant `caseId:` parameter from `respond`/`appeal`; grepped the repo and confirmed no other conformers or call sites exist besides the stub, so this is a safe, non-breaking change in practice — `ModerationAuthorityClient.swift:126-149`.
- All new `Unsigned` mirror-struct field lists match their parent structs' actual stored properties (checked `Report`, `Verdict`, `ModerationMandate` field-by-field against the test's expected key lists).
- Minor nit (non-blocking): comment at `Report.swift:75` says `SigningBytesTests` pins the ordering, but the actual test class is `ModerationCanonicalOrderTests` — stale reference from an earlier draft name.
- Minor nit (non-blocking): `keyOrder(of:)` in `ModerationCanonicalOrderTests.swift:24-45` is a brace-depth scanner that doesn't account for literal `{`/`[` characters inside string values — fine for current fixture data, but fragile if evidence/statement fixtures ever contain those characters.

dev-onym (commented):
REQUEST_CHANGES — the canonicalization decision is right and well-argued; the drift guard around it is the gap.

**Blocking**

- `Report.swift:82` / `ModerationAuthorityClient.swift:47,88` — three new `Unsigned` mirrors, none covered by the repo's existing drift guard. `ModerationSigningPayloadTests.swift:57,76` pins mandate/verdict as `keys(signingBytes) == encodedKeys(object) - ["signature"]`, which catches a field added to the object but *not* to the mirror. The new tests only compare against hardcoded key lists, so that exact failure — a new `Report` field silently outside the signature — passes green. Three more cases in that file (`encodedKeys(of:).subtracting(["signature"])`) closes it; this is the one hazard that fails silently on a signing boundary, same shape as the one the PR exists to prevent.

- `ModerationCanonicalOrderTests.swift:122` — `test_mandateAndVerdictKeysAreUnaffectedByTheOrderingDifference` never touches `ModerationMandate` or `Verdict`. It sorts hardcoded string arrays, feeds them to `JSONSerialization`, and asserts the result matches a sort computed in the test — so it asserts Foundation is self-consistent, not anything about our types. The doc comment's claim ("pinned so a renamed field can't quietly change that") is not true as written: rename `mandateVersion` and this still passes. Derive the key lists from `mandate.signingBytes()` / `verdict.signingBytes()` via `keyOrder(of:)` and compare each against its own byte-order sort.

**Non-blocking**

- `Report.swift:75` — doc names `SigningBytesTests`; no such type exists. Should be `ModerationCanonicalOrderTests`.
- `ModerationCanonicalEncoder.swift:26` — documents two load-bearing settings, but a third is in force and undocumented: Swift's synthesized `encodeIfPresent` **omits** nil optionals rather than emitting `null`. `EvidenceItem.context` (`Report.swift:14`) is the only optional inside signed bytes, so an authority that normalizes absent → `null` before re-serializing rejects every report carrying evidence without context. Worth one bullet next to the other two, since this file is explicitly the single place the canonical form is defined.
- `ModerationCanonicalOrderTests.swift:19` — `keyOrder` tracks `depth` by counting `{`/`[` characters without skipping string contents, and doesn't handle `\"`. Fine for the current fixtures; a future test whose `statement` contains a brace or an escaped quote mis-scans and fails confusingly. Also, only `depth == 1` keys are collected, so `test_..._haveTheExactByteOrderedKeyList` (`:83`) doesn't in fact pin nested `EvidenceItem` order despite those bytes being signed — `disclosedContent`/`authenticityProof`/`context` happen to sort identically under both rules, so it's latent, not live.
- `ModerationAuthorityClient.swift:22,69` — adding non-optional `caseId` to two `Codable` types is a decode-breaking change for any persisted or in-flight JSON. Nothing in-tree decodes them (only `StubModerationAuthorityClient` conforms), so this is a note for whoever writes the HTTP client: `caseId` now lives in the body and will also need to be in the request path, and the authority must check the two agree — otherwise the path/body split reintroduces the replay the field was added to close.

Verified: `.iso8601` is consistent across `MandateStore.swift:89` and `GateCheckRepository.swift:66`, so signed dates round-trip to the same bytes on the wire, and no `JSONSerialization` exists in package sources.

dev-onym (commented):
**COMMENT** — the byte-order rule is right and well-pinned; the gaps are in what the newly-defined signed forms leave out, not in what they do.

- `Packages/OnymModeration/Sources/OnymModeration/ModerationCanonicalEncoder.swift:29` — the encoder is `internal`, but `signingBytes()` is `public`. The doc says `.iso8601` "matches the form every timestamp crosses the wire in" — nothing enforces that, and no `ModerationAuthorityClient` implementation exists yet to set the precedent. Every other client in this repo (`SEPContractClient.swift:30`, `KnownAuthoritiesFetcher.swift:55`) defaults to a bare `JSONEncoder`/`JSONDecoder`, which serialises `filedAt` as a numeric interval-since-2001. An out-of-package implementer who follows that pattern ships a body the authority cannot reconstruct signing bytes from — and it fails as "every report signature is invalid", the exact mode this PR exists to prevent. Consider making the encoder `public` (or adding a wire coder alongside it) and saying so in the protocol doc at `ModerationAuthorityClient.swift:127`, where the path/body note already lives.
- `ModerationAuthorityClient.swift:21`, `:62` — `CaseResponse` and `AppealSubmission` are the only signed moderation objects without a version discriminator; `Report.reportVersion`, `mandateVersion`, `verdictVersion` and `noticeVersion` all have one. This PR is the moment their signed form gets defined, so it's the cheapest moment to add `responseVersion`/`appealVersion`. Adding one later changes the bytes and invalidates anything signed before.
- `ModerationAuthorityClient.swift:71` — for `kind == .newHolderClaim` the signer is by construction *not* the mandated user, and the object carries no signer key or attestation. So a new-holder claim's `signature` is unverifiable against anything: the authority has no key to check it with, and cannot distinguish a signed claim from `signature: ""`. The doc comment names the asymmetry but the type gives the authority nothing to act on. Out of scope to fix here, but worth an issue rather than only a comment.
- `Tests/OnymIOSTests/ModerationCanonicalOrderTests.swift:162` — this test can't catch the regression its name implies. `XCTAssertEqual(mandateKeys, byteOrdered(mandateKeys))` sorts the observed keys and compares them to themselves; since mandate/verdict keys sort identically under both rules (as the PR body states), it would still pass if those two were switched to `JSONSerialization`. It's a shape check, not a rule pin — fine to keep, but the comment above it oversells it.
- The verification note in the PR body covers `Report` only. Response and appeal signing bytes — including the newly signed `caseId` — haven't been round-tripped against the Rust authority, and the path/body mismatch rejection at `ModerationAuthorityClient.swift:129` is a requirement stated in a doc comment with no counterpart in that repo yet. Worth confirming the authority side agrees before the HTTP client lands.
- Minor: `makeMandate`/`makeVerdict` are now duplicated verbatim across both test files, and the two `makeReport` fixtures differ (`ModerationCanonicalOrderTests.swift:83` sets `context`, `ModerationSigningPayloadTests.swift:56` doesn't) — the drift guard is the one that would benefit from exercising the optional-omission rule.

PR #222, opened 2026-08-09
feat(moderation): add Authority transport and mandate registration

rinat-enikeev:
## Summary

- implement the merged onym-moderation Authority HTTP operation surface and structured error mapping
- register the exact countersigned mandate before local activation, persisting it for byte-identical retry after ambiguous failures
- require an explicit HTTPS Authority API base URL designated by the published authority directory; never derive a mandate endpoint from a mirrorable manifest URL
- single-flight overlapping consent and exact-artifact registration, including launch retry racing explicit consent
- resume persisted registration at launch without allowing an older pending Authority to replace newer completed consent
- resolve pending artifacts across manifest rotations and let deterministic refusals recover without replaying doomed bytes
- surface pending registration separately from current and historical mandates with an exact-artifact retry action
- log malformed directory entries while failing closed per entry
- wire the production client factory and cover canonical requests, status credentials, receipts, registration state, proxy failures, path safety, and RFC 3339 timestamps

## Directory rollout

onym-authorities v1 was published ahead of this schema-enforcement commit and contains apiBaseURL on every entry. Missing or insecure endpoints fail closed per entry without removing other valid Authorities.

## Source of truth

This mirrors the Authority reference implementation currently merged on onymchat/onym-moderation main. Project membership remains XcodeGen-managed; no pbxproj edits are included.

## Testing

- [x] OnymModeration iOS-simulator build
- [x] Focused ModerationAuthorityClientTests and ModerationRepositoryTests: 37 tests, 0 failures
- [ ] Full OnymIOSTests was attempted, but unrelated existing simulator Keychain entitlement failures (keychainRead(-34018)) block identity/keychain tests; existing chat-dispatcher failures also remain outside this change

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
@@ -32,11 +32,35 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
         self.apiBaseURL = apiBaseURL
         self.operatorPublicKeyBase64 = operatorPublicKeyBase64
     }
-
 }
 
-struct KnownAuthoritiesDocument: Codable {
+private struct LossyAuthorityListing: Decodable {
+    let value: AuthorityListing?
+
+    init(from decoder: Decoder) throws {
+        value = try? AuthorityListing(from: decoder)
+    }
+}
+
+struct KnownAuthoritiesDocument: Decodable {
     let authorities: [AuthorityListing]
+
+    private enum CodingKeys: String, CodingKey {
+        case authorities
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let decoded = try container.decode([LossyAuthorityListing].self, forKey: .authorities)
+        authorities = decoded.compactMap(\.value)
+        guard !authorities.isEmpty else {
+            throw DecodingError.dataCorruptedError(
+                forKey: .authorities,
+                in: container,
+                debugDescription: "directory contains no valid Authority entries"
+            )
+        }
+    }
 }
 
 /// Network seam that fetches the curated list of moderation
--- a/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
@@ -33,11 +33,6 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
         self.operatorPublicKeyBase64 = operatorPublicKeyBase64
     }
 
-    /// Named at call sites to make the directory's trust designation
-    /// visible where a client is constructed.
-    public var resolvedAPIBaseURL: URL {
-        apiBaseURL
-    }
 }
 
 struct KnownAuthoritiesDocument: Codable {
--- a/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
@@ -1,5 +1,6 @@
 import Foundation
 import OnymFoundation
+import os.log
 
 /// One entry in the interface's published authority directory. The
 /// directory pins each authority's operator public key out-of-band
@@ -35,10 +36,40 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
 }
 
 private struct LossyAuthorityListing: Decodable {
+    private static let log = Logger(
+        subsystem: "app.onym.ios",
+        category: "ModerationDirectory"
+    )
+
     let value: AuthorityListing?
 
     init(from decoder: Decoder) throws {
-        value = try? AuthorityListing(from: decoder)
+        do {
+            let listing = try AuthorityListing(from: decoder)
+            guard listing.apiBaseURL.scheme?.lowercased() == "https",
+                  listing.apiBaseURL.host != nil
+            else {
+                throw DecodingError.dataCorrupted(
+                    .init(
+                        codingPath: decoder.codingPath,
+                        debugDescription: "Authority apiBaseURL must use HTTPS"
+                    )
+                )
+            }
+            value = listing
+        } catch {
+            let componentId = (try? decoder.container(keyedBy: DiagnosticKeys.self))
+                .flatMap { try? $0.decodeIfPresent(String.self, forKey: .componentId) }
+                ?? "<unknown>"
+            Self.log.error(
+                "Skipping malformed Authority \(componentId, privacy: .public): \(String(describing: error), privacy: .public)"
+            )
+            value = nil
+        }
+    }
+
+    private enum DiagnosticKeys: String, CodingKey {
+        case componentId
     }
 }
 
--- a/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
@@ -11,14 +11,32 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
     /// Human-readable name shown in the picker.
     public let name: String
     public let manifestURL: URL
+    /// Base URL of the Authority HTTP API. Optional for compatibility
+    /// with directories published before the operation surface existed;
+    /// those entries resolve the API beside `manifestURL`.
+    public let apiBaseURL: URL?
     /// The authority's Ed25519 operator public key, base64 raw bytes.
     public let operatorPublicKeyBase64: String
 
-    public init(componentId: String, name: String, manifestURL: URL, operatorPublicKeyBase64: String) {
+    public init(
+        componentId: String,
+        name: String,
+        manifestURL: URL,
+        apiBaseURL: URL? = nil,
+        operatorPublicKeyBase64: String
+    ) {
         self.componentId = componentId
         self.name = name
         self.manifestURL = manifestURL
+        self.apiBaseURL = apiBaseURL
         self.operatorPublicKeyBase64 = operatorPublicKeyBase64
+    }
+
+    /// The reference implementation serves `/manifest.json` and `/v1/*`
+    /// from one base. New directory entries should publish `apiBaseURL`
+    /// explicitly; the fallback keeps already-published entries usable.
+    public var resolvedAPIBaseURL: URL {
+        apiBaseURL ?? manifestURL.deletingLastPathComponent()
     }
 }
 
--- a/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift
@@ -11,10 +11,11 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
     /// Human-readable name shown in the picker.
     public let name: String
     public let manifestURL: URL
-    /// Base URL of the Authority HTTP API. Optional for compatibility
-    /// with directories published before the operation surface existed;
-    /// those entries resolve the API beside `manifestURL`.
-    public let apiBaseURL: URL?
+    /// Base URL of the Authority HTTP API, explicitly designated by the
+    /// Interface's directory. It is never inferred from `manifestURL`:
+    /// manifests may be safely mirrored because their bytes are signed,
+    /// while this endpoint receives the user's signed mandate.
+    public let apiBaseURL: URL
     /// The authority's Ed25519 operator public key, base64 raw bytes.
     public let operatorPublicKeyBase64: String
 
@@ -22,7 +23,7 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
         componentId: String,
         name: String,
         manifestURL: URL,
-        apiBaseURL: URL? = nil,
+        apiBaseURL: URL,
         operatorPublicKeyBase64: String
     ) {
         self.componentId = componentId
@@ -32,11 +33,10 @@ public struct AuthorityListing: Codable, Sendable, Equatable {
         self.operatorPublicKeyBase64 = operatorPublicKeyBase64
     }
 
-    /// The reference implementation serves `/manifest.json` and `/v1/*`
-    /// from one base. New directory entries should publish `apiBaseURL`
-    /// explicitly; the fallback keeps already-published entries usable.
+    /// Named at call sites to make the directory's trust designation
+    /// visible where a client is constructed.
     public var resolvedAPIBaseURL: URL {
-        apiBaseURL ?? manifestURL.deletingLastPathComponent()
+        apiBaseURL
     }
 }
 
--- a/Packages/OnymModeration/Sources/OnymModeration/MandateStore.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/MandateStore.swift
@@ -19,6 +19,12 @@ public struct MandateRecord: Codable, Sendable, Equatable {
     /// record remains persisted with this false so a retry can resend
     /// identical bytes instead of minting a second consent artifact.
     public var authorityRegistered: Bool
+    /// A real countersigned mandate awaiting acknowledgement from its
+    /// Authority. It is neither current consent nor historical consent
+    /// and must not be presented as either one.
+    public var registrationPending: Bool {
+        countersigned && !authorityRegistered
+    }
     /// Exactly one record is active per identity–device pair.
     /// Switching authorities deactivates the old record without
     /// touching anything else (mandates are immutable, spec §12).
--- a/Packages/OnymModeration/Sources/OnymModeration/MandateStore.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/MandateStore.swift
@@ -14,6 +14,11 @@ public struct MandateRecord: Codable, Sendable, Equatable {
     /// False while the interface signature is the stub sentinel — a
     /// record that can never be mistaken for a spec-valid mandate.
     public let countersigned: Bool
+    /// True only after the Authority accepted this exact countersigned
+    /// mandate and returned its matching content reference. A pending
+    /// record remains persisted with this false so a retry can resend
+    /// identical bytes instead of minting a second consent artifact.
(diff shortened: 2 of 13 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the register-before-activate design is sound, but persisted pending records leak into the history UI and the activation/rejection paths have sharp edges.

- `ModerationRepository.swift:266` — a countersigned mandate is persisted with `isActive: false` before the network hop, so a failed registration leaves a record that `ModerationSettingsFlow.previousMandates` (`ModerationSettingsView.swift:42`, filter `!isActive`) renders as a *previous mandate* the user never held. It's also `countersigned == true`, so the existing "not countersigned" warning at `ModerationSettingsView.swift:83` doesn't fire. History needs to exclude `countersigned && !authorityRegistered`, or the UI needs a pending state.
- `ModerationRepository.swift:343` — activation matches records by `mandateHash()` and sets `isActive = true` on *every* match. `mandateHash` is signature-independent and covers only (user, interface, authority, manifestHash, classes, deviceBinding, acceptedAt), so two records minted in the same second with the same binding both go active; `activeMandateRecord()` then silently returns `first`. Activate by index/identity of the record you passed in rather than by content hash.
- `ModerationAuthorityClient.swift:275` and `ModerationRepository.swift:333` — when the Authority returns `accepted: false` with a *correct* `mandateRef`, both sites throw `mandateReferenceMismatch(expected:received:)` with two identical strings, misreporting a refusal as a hash mismatch. That deserves its own case, and neither `accepted: false` branch has a test.
- `ModerationRepository.swift:186` — the pending-retry branch runs before enrollment, so a pending record pins its original `deviceBinding` permanently. If that enrollment is invalidated backend-side (reinstall, revocation), every subsequent `consent(to:)` for that manifest replays a mandate the Authority will keep rejecting, with no path to abandon the record and re-enroll. Worth at least an age/binding check or an explicit discard.
- `ModerationAuthorityClient.swift:310` — `"query-status:<caseId>:<timestamp>"` binds neither host, method, nor path, so a captured `X-Onym-Key`/`Timestamp`/`Signature` triple is replayable to any Authority for that case inside its freshness window. If the merged reference implementation specifies this exact string, please cite it in the doc comment; otherwise bind the request line.
- `ModerationAuthorityClient.swift:127-207` — this PR deletes the rationale comments it most needs to keep: the cross-case replay argument for signing `caseId`, the `submit-evidence`-folded-into-`respond` note, the `newHolderClaim` "an authority cannot require a signature here" caveat, and the requirement that an HTTP Authority **reject** a path/body `caseId` mismatch. That last one is now unstated in the very commit that first puts `caseId` in both places (`:295`, `:303`).
- `Tests/OnymIOSTests/ModerationAuthorityClientTests.swift` — no coverage for the `http_<status>` fallback on a non-JSON/empty error body, `invalidResponse`, `malformedResponse`, or the `duplicate` receipt field. The error-mapping path is the one most likely to meet a real proxy.
- Nit, `ModerationRepository.swift:96` — defaulting `authorityClients` to the stub turns a forgotten wire-up into a runtime `notImplemented` during consent rather than a compile error; the two app call sites are correct today.

dev-onym (commented):
## REQUEST_CHANGES — solid design, but the pending-record retry has no escape hatch, and the API-base fallback trusts the wrong URL

- **Permanent rejection deadlocks consent** — `ModerationRepository.swift:186-194`: the pending-retry lookup matches on authority+user+manifestHash and fires for *every* failure, but only ambiguous (network/5xx) failures are safely retryable. A deterministic refusal — `AuthorityRejection(signature_invalid|bad_request)`, `accepted:false`, or `mandateReferenceMismatch` (which means the two sides disagree on canonical bytes and *can never* agree for this artifact) — leaves a record that every later `consent(to:)` replays identically. With no active mandate the root gate sits on `.needsConsent` (`ModerationGateFlow.swift:17`) and `tappedAgree` shows "Signing failed. Try again." (`ModerationConsentFlow.swift:114`) forever. Only escape is picking a different authority. Suggest: retry only on transport-ambiguous errors, and mark the record terminally failed (or drop it) on a deterministic rejection so a fresh mandate can be minted.
- **`resolvedAPIBaseURL` promotes an untrusted host to a trusted endpoint** — `KnownAuthoritiesFetcher.swift:38-40`: manifest hosting needs no trust because `URLSessionAuthorityManifestFetcher` verifies a detached signature over the exact bytes against the directory-pinned key, so a listing may legitimately point at a CDN or mirror. Deriving the API base from it silently POSTs the signed consent artifact (user key, device binding, both signatures) to that host, with no equivalent check. Prefer failing closed when `apiBaseURL` is absent, or at minimum restrict the fallback to same-origin-with-manifest and document the trust step.
- **The double-artifact race isn't fully closed** — `ModerationRepository.swift:186` vs `:262-270`: `consent` suspends four times (`userKeyID`, `generateToken`, `enrollDevice`, `countersignMandate`) between the pending lookup and the insert. Two overlapping calls both see no pending record, both mint and register, and each `registerAndActivate` deactivates the other's record — exactly the two-consent-artifacts outcome the PR exists to prevent. An in-flight guard keyed on componentId+manifestHash would cover it.
- **Strict `.iso8601` will reject fractional-second timestamps** — `ModerationAuthorityClient.swift:406-410`: `JSONDecoder.iso8601` fails on `2023-11-14T22:13:20.123456Z`, which is what chrono/serde emits for any sub-second-precision stored value. Every fixture in `ModerationAuthorityClientTests.swift` uses whole seconds, so the suite can't catch it; the whole `queryStatus`/`fileReport` surface would degrade to `malformedResponse`. Worth one fractional-seconds fixture and, if the reference emits them, a lenient strategy.
- **Legacy countersigned records disappear from history** — `MandateStore.swift:25-27` + `ModerationSettingsFlow.swift:69`: any pre-existing record with `countersigned == true` migrates to `authorityRegistered == false`, so `registrationPending` is true and it is filtered out of `previousMandates` — silently erasing consent history rather than showing it. Benign today only because `StubEnforcementBackendClient` is wired in all three app configs (`OnymIOSApp.swift:124,147,169`), so `countersigned` is always false; consider gating the filter on `createdAt` or a schema version.
- **Mislabeled error** — `ModerationRepository.swift:354`: losing the local record *after* the Authority committed is a store-consistency failure, not `malformedResponse`, and it silently re-opens double-minting on the next consent. Deserves its own case.
- Minor: `receipt.accepted`/`mandateRef` are validated in both `URLSessionModerationAuthorityClient.registerMandate` (`:239-247`) and `registerAndActivate` (`:333-341`) — fine as defense across the protocol seam, but worth a comment saying so; and no `timeoutInterval` is set, so consent blocks on the 60s default.
- Liked: persist-before-network ordering, `isActive: !isCountersigned` keeping the sentinel path honest, identity-based (not hash-based) record lookup with a test for the colliding-content-hash case, and `rawCode` retention for unknown Authority error codes.
- Note: I reviewed by reading the tree, not by building — this is a Linux worktree, so I could not re-run the iOS-simulator tests the PR body reports.

dev-onym (commented):
**REQUEST_CHANGES** — the registration/activation ordering is sound, but the directory schema change breaks existing entries and one failure class permanently wedges consent.

- `Packages/OnymModeration/Sources/OnymModeration/KnownAuthoritiesFetcher.swift:18` — `apiBaseURL` is required and non-optional, with no fallback anywhere; the PR summary's "manifest-adjacent fallback for existing directory entries" is not implemented. Because listings decode inside `KnownAuthoritiesDocument`, a single legacy entry fails the whole document → `refresh()` throws `malformedDocument`, the picker goes empty and nobody can consent. `Tests/OnymIOSTests/ModerationAuthorityClientTests.swift:78` codifies the fail-closed behavior. Either add the described fallback, or ship the new `authorities.json` first and say so in the PR body.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:443` — only `bad_request` and `signature_invalid` are treated as terminal. A 4xx carrying `no_jurisdiction`, `class_outside_mandate`, `window_closed`, `case_state`, or `not_found` leaves the record `registrationPending`, so every later `consent(...)` hits the retry branch at `:239` and replays the identical doomed bytes forever — that authority becomes permanently unconsentable. Suggest terminal-by-default for 4xx (excluding 408/429), retriable for transport/5xx, or an attempt cap.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationSettingsFlow.swift:69` + `MandateStore.swift:75` — records written before this PR decode as `authorityRegistered: false`, so any pre-existing *real* countersigned mandate is `registrationIncomplete` and silently disappears from `previousMandates` once deactivated. Genuine past consent artifacts vanish from history. Migrating `countersigned == true` to `authorityRegistered: true`, or filtering only on `registrationPending`, avoids this.
- `MandateStore.swift:23` — the doc comment says rejected attempts are "retained for an honest local audit trail," but `registrationIncomplete` hides them from the only history surface and nothing else reads `registrationRejected`. Surface them or drop the claim; as written, repeated rejections also accumulate invisible records forever.
- `KnownAuthoritiesFetcher.swift:37` — `resolvedAPIBaseURL` is a pure passthrough to `apiBaseURL`; two public names for one value, and the doc comment ("resolves") no longer describes anything. Prefer using `apiBaseURL` directly.
- `ModerationRepository.swift:174-181` vs `:220-229` — the componentId guard and `validateForConsent` now run twice per consent, with two separate `clock()` reads. Drop the copy in `consent(...)`, keeping the one inside `performConsent`.

Worth keeping as-is: persisting the countersigned record *before* the network hop, activating only after `mandateRef` agreement, and index-based activation in `registerAndActivate` (`ModerationRepository.swift:411-420`) correctly avoid both double-minting and hash-collision cross-activation. The repository-level re-check of `accepted`/`mandateRef` on top of the client's is the right redundancy for an injected-factory seam.

I could not compile or run the tests — no Swift toolchain in this environment, so all findings are from reading the merged sources.

dev-onym (commented):
**REQUEST_CHANGES** — the transport and registration state machine are well-built, but the exact-retry invariant has a hole that can mint a second consent artifact.

- `ModerationRepository.swift:438` — `isDeterministicRegistrationRejection` treats *every* 4xx except 408/429 as terminal and deletes the persisted mandate. The ambiguous-commit case this whole feature exists for (Authority committed, response lost) most likely surfaces on retry as a duplicate/conflict rejection — 409, or `case_state`/`bad_request` from the documented vocabulary. That path discards the local record and the next consent mints a *new* mandate, producing exactly the two artifacts the design claims to prevent. Either the Authority's idempotency on identical mandate bytes must be pinned (and tested against `409`), or conflict-shaped codes must be treated as "already registered → activate locally", not as terminal.
- `ModerationRepository.swift:392` / `ModerationAuthorityClient.swift:300` — `accepted` is checked before `mandateRef == expectedRef`, so an `accepted:false` receipt bearing an unrelated `mandateRef` is thrown as `mandateNotAccepted` and (line 384/392) deletes the pending record. A refusal should be attributable to this mandate before it's allowed to destroy it: verify the reference first, then `accepted`.
- `ModerationRepository.swift:178` — the single-flight key is `(authority, user, manifestHash)`, so it only collapses taps on the *same* authority. Two concurrent consents to different authorities both persist, both register, and both are accepted; only the local `isActive` flag is exclusive. Two Authorities then each believe they hold a live mandate for this identity–device pair. Consent needs one lock, not one per key.
- `ModerationRepository.swift:230` — the retry lookup matches on `manifestHash`, but the only way to reach it is `manifestForReview` → `consent`, and `manifestForReview` refetches the manifest. If the authority rotates its terms between the failed attempt and the retry, the hash differs, the pending record is skipped, a fresh mandate is minted, and the old pending record is orphaned in `records` forever (nothing ever removes it, and it's hidden from `previousMandates`). Nothing retries pending registrations at launch either — meanwhile `activeMandate` is nil, so the gate reports `.notMandated` with no UI signal about why.
- `ModerationRepository.swift:409` — `localMandateRecordMissing` is thrown *after* the Authority accepted. That outcome leaves a registered-but-unrecorded mandate, the worst state in this design; it should at minimum re-persist the accepted record rather than drop it.
- `KnownAuthoritiesFetcher.swift:18,39` — `apiBaseURL` is non-optional and `KnownAuthoritiesDocument` decodes the array strictly, so a single future entry published without it fails the *whole* document and empties the picker for every authority. Per-entry lossy decoding fails closed on that entry only; the current shape fails closed on all of them.
- `ModerationAuthorityClient.swift:365` — `appending(path:)` does not escape `/` or `..`, and `caseId` (authority-supplied, via notices/receipts) flows straight into the path for `respond`/`appeal`/`queryStatus`. Percent-encode the component or reject ids outside a strict charset.
- `Tests/OnymIOSTests/ModerationAuthorityClientTests.swift` — request bodies are asserted against a locally re-declared `JSONEncoder.moderationFixture()` rather than `ModerationCanonicalEncoder` (reachable via `@testable`); a drift in the canonical encoder would leave these green. No test covers a 409/duplicate registration or a pending retry after a manifest rotation — both are the gaps above.

Nothing blocking in `MandateStore.swift`: the custom `init(from:)` migration and the `registrationPending` derivation are correct, and `previousMandates` filtering is the right call.

dev-onym (commented):
**REQUEST_CHANGES** — the transport and registration state machine are solid, but launch-time retry can silently switch the user's active authority.

- `ModerationRepository.swift:389` — `retryNewestPendingRegistration` picks the first *pending* record regardless of authority and calls `registerPending(..., activate: true)`. Combined with `ModerationRepository.swift:238`, where the pre-mint resolution loop filters on `$0.mandate.authority == listing.componentId`, an abandoned pending attempt for authority X survives a later successful consent to authority Y — then at next launch it registers and takes over as the active mandate, silently reverting the user's designated authority. The method name says "newest", but the array-order lookup returns the newest *pending*, which may be older than the active record. Suggest: skip retry when a newer registered record exists (compare `createdAt`), or restrict activation to the case where no registered active record post-dates the pending one.
- `ModerationRepository.swift:389` — this path bypasses the `consentTasks` single-flight entirely, so a start-time retry can run `registerPending` concurrently with a user-initiated `consent()` on the same record. If one branch discards on a deterministic refusal while the other is mid-flight, the survivor's `recordIndex` misses and the recovery branch at `ModerationRepository.swift:444` re-inserts the just-discarded record as registered and active. Worth routing the retry through the same `ConsentKey` task map.
- `ModerationSettingsFlow.swift:69` + `ModerationRepository.swift:238` — a pending record is excluded from `previousMandates` and is not `activeMandate`, so a countersigned mandate the user signed becomes completely invisible in the UI with no retry affordance, yet can auto-activate at a much later launch. Consent state that is neither shown nor cancellable is a poor fit for the "must not be presented as either one" rule; consider surfacing it as an explicit pending row.
- `KnownAuthoritiesFetcher.swift:41` — `try?` swallows every decode error with no diagnostic. A single typo'd `apiBaseURL` silently drops an authority the user already holds a mandate with; `retryNewestPendingRegistration` then can't find the listing and the pending record is stuck indefinitely with no signal. At minimum log the skipped `componentId`.
- `ModerationAuthorityClient.swift:264` — `apiBaseURL` goes straight into the client with no scheme check. Since this endpoint receives the signed mandate (unlike the mirrorable manifest), rejecting non-`https` here is cheap defense-in-depth against a bad directory entry.
- Tests cover the retry/rotation/single-flight paths well, but there's no test for a pending record belonging to a *different* authority than the active one — that gap is exactly where finding #1 hides.

dev-onym (commented):
**COMMENT** — solid, well-reasoned design; two robustness gaps around treating any 4xx as terminal, plus smaller nits.

- `ModerationRepository.swift:580` / `ModerationAuthorityClient.swift:418` — any non-2xx with a non-Authority body gets `rawCode: "http_<status>"`, and every 4xx except 408/429 is classified deterministic, which calls `discardRegistrationAttempt` and **deletes the user's countersigned mandate**. A CDN/WAF/captive-portal 403, or a temporarily misrouted path yielding 404, permanently destroys a signed artifact the Authority may already hold. Suggest gating terminality on the body actually decoding to the known vocabulary (`AuthorityRejection.code != nil`), and retaining the record otherwise.
- `ModerationRepository.swift:576-587` — 409 is also treated as terminal. If a conforming Authority answers an exact replay of an already-committed mandate with 409 (rather than a 200 idempotent receipt), the retry path discards a mandate the Authority holds, and the user re-consents into a duplicate. Nothing in the tests pins the replay contract; `RecordingAuthorityClient` always re-accepts. Worth one test asserting replay-of-committed behavior, since the whole persist-before-send design rests on it.
- `ModerationRepository.swift:451-542` — the flight entry is only removed when the **last** waiter releases, so it outlives task completion whenever `waiters > 1`. Between waiter A's release and waiter B's, a UI-initiated `retryRegistration` for identical bytes joins the already-finished task and re-throws the cached error (or re-applies a cached receipt) without any network attempt. Narrow window on the actor, but avoidable: clear the entry on completion rather than on refcount zero.
- `ModerationAuthorityClient.swift:371-375` — the allow-list rejects `.`/`..` and non-unreserved chars but accepts `""`, so an empty `caseId` builds a wrong URL instead of failing. Conversely, the unreserved-only rule will reject any Authority case id containing `:` or `+`; the id format isn't pinned anywhere in this diff, so this could fail closed on valid ids.
- `ModerationSettingsFlow.swift:98` vs `ModerationSettingsView.swift:187` — `registrationRowID` and `historyRowID` are byte-identical private extensions in two files, and the spinner works only because the view compares `state.retryingRegistration` (set from the former) to the latter. Drift silently breaks "Confirming with authority…". Also neither includes `createdAt`, which the repository uses as its disambiguator, so two pendings sharing authority+manifestHash+acceptedAt collide as `ForEach` ids.
- `KnownAuthoritiesFetcher.swift:87` — an all-invalid (or legitimately empty) directory now throws instead of returning `[]`, so `refresh()` reports "Couldn't load the authority list" and keeps the stale cached list. Reasonable fail-closed, but it's a behavior change worth calling out in the PR body alongside the per-entry skip.
- `ModerationRepository.swift:249-273` — the rotation loop is `while let pending = ...` over a mutating array; it terminates only because every branch either clears `registrationPending` or discards. That's correct today but fragile; the `recordIndex(of:) == nil` recovery branch inside `registerPending` is the one path that could reintroduce a pending entry. A comment (or a `records.count`-bounded loop) would make the termination argument explicit.
- Verified all `ModerationRepository(` call sites and the `StubURLProtocol.Handler` widening to `URLResponse` are consistent; no other conformer of `ModerationAuthorityClient` needed the new `registerMandate`. Test coverage of the happy paths, single-flight, and legacy decode is genuinely good.

PR #223, opened 2026-08-09
feat(moderation): report authenticated chat messages

rinat-enikeev:
## Summary

- sign every newly sent text message over the canonical proof preimage — the body bound to its message id, a message-salted group binding, and the send timestamp — and retain that proof encrypted on recipients
- offer Report from the long-press menu only for incoming text messages whose sender proof verifies against the sender's stored key
- let the reporter review the exact disclosure and choose a class from their active consented Authority manifest
- construct and sign the reference Authority Report shape, bind it to the active registered reporter mandate, and verify the accused proof locally before filing
- encrypt and persist the exact signed report before network delivery, reuse it after ambiguous failures, single-flight overlapping taps, correlate the returned report ID, and treat the Authority's already-on-file answer as terminal
- bound the report ledger (resolved rows prune oldest-first; rows awaiting an outcome are kept as idempotency keys) and purge it when the filing identity is removed

## Scope

This PR intentionally reports text only. Legacy messages and media do not expose Report because they do not carry the content-level sender proof required by the finalized Authority contract.

No surrounding conversation context is disclosed; the proof's group binding is keyed by the group's shared secret (never on-chain) and salted per message, so an Authority cannot resolve it to a group — even holding the public on-chain group set — nor correlate independent reports to one group.

## Limitations (deliberate)

- **Reportability is sender-opt-in by wire design.** The proof field is optional (older clients ship none), so a modified client that omits or garbles it makes its messages unreportable. Closing that would require Authority-side verification of a different commitment (e.g. envelope-level), i.e. a protocol/contract change outside this client PR. Recipients can still see that a message carries no proof (no Report entry).
- **A 409 (already filed) cannot recover the original receipt** — the Authority protocol has no report-lookup endpoint; the client records the terminal state and stops re-delivering.

## Verification

- full OnymIOSTests suite passes
- focused coverage includes payload compatibility, canonical preimage signatures, encrypted persistence, inbound evidence retention across replays, local authenticity rejection, mandate/class binding, byte-identical retry, receipt correlation, already-filed terminality, ledger bounds/purge, and long-press eligibility

### DIFF
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessage.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessage.swift
@@ -91,6 +91,12 @@ public struct ChatMessage: Equatable, Sendable, Identifiable {
     /// the image/video/album fields.
     public let voiceAttachment: ChatVoiceAttachment?
 
+    /// Sender's detached Ed25519 signature over `body`, retained solely
+    /// so an incoming text message can be disclosed as independently
+    /// verifiable moderation evidence. Nil for legacy and media-only
+    /// messages that this first reporting surface cannot file.
+    public let moderationAuthenticityProof: String?
+
     /// Canonical media list for rendering: the album when present, else
     /// the single image/video wrapped in a one-element list, else empty.
     public var media: [ChatMediaAttachment] {
@@ -131,6 +137,7 @@ public struct ChatMessage: Equatable, Sendable, Identifiable {
         replyToMessageID: UUID?,
         groupType: SEPGroupType,
         failureReason: SendFailureReason? = nil,
+        moderationAuthenticityProof: String? = nil,
         imageAttachment: ChatImageAttachment? = nil,
         videoAttachment: ChatVideoAttachment? = nil,
         albumAttachments: [ChatMediaAttachment]? = nil,
@@ -147,6 +154,7 @@ public struct ChatMessage: Equatable, Sendable, Identifiable {
         self.replyToMessageID = replyToMessageID
         self.groupType = groupType
         self.failureReason = failureReason
+        self.moderationAuthenticityProof = moderationAuthenticityProof
         self.imageAttachment = imageAttachment
         self.videoAttachment = videoAttachment
         self.albumAttachments = albumAttachments
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
@@ -90,6 +90,14 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
     /// `ChatVoiceAttachment`.
     public let voiceAttachment: ChatVoiceAttachment?
 
+    /// Detached Ed25519 signature by the sender over the exact UTF-8
+    /// bytes of `variant.body`. A recipient retains this so a voluntary
+    /// disclosure can satisfy the Authority contract's authenticity
+    /// rule without exposing an envelope or any recipient secrets.
+    /// Optional for wire compatibility with messages sent before
+    /// reporting existed.
+    public let moderationAuthenticityProof: String?
+
     public init(
         version: Int,
         messageID: UUID,
@@ -101,7 +109,8 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
         attachment: ChatImageAttachment? = nil,
         videoAttachment: ChatVideoAttachment? = nil,
         attachments: [ChatMediaAttachment]? = nil,
-        voiceAttachment: ChatVoiceAttachment? = nil
+        voiceAttachment: ChatVoiceAttachment? = nil,
+        moderationAuthenticityProof: String? = nil
     ) {
         self.version = version
         self.messageID = messageID
@@ -114,6 +123,7 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
         self.videoAttachment = videoAttachment
         self.attachments = attachments
         self.voiceAttachment = voiceAttachment
+        self.moderationAuthenticityProof = moderationAuthenticityProof
     }
 
     enum CodingKeys: String, CodingKey {
@@ -128,6 +138,7 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
         case videoAttachment = "video_attachment"
         case attachments
         case voiceAttachment = "voice_attachment"
+        case moderationAuthenticityProof = "moderation_authenticity_proof"
     }
 }
 
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
@@ -90,12 +90,15 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
     /// `ChatVoiceAttachment`.
     public let voiceAttachment: ChatVoiceAttachment?
 
-    /// Detached Ed25519 signature by the sender over the exact UTF-8
-    /// bytes of `variant.body`. A recipient retains this so a voluntary
-    /// disclosure can satisfy the Authority contract's authenticity
-    /// rule without exposing an envelope or any recipient secrets.
-    /// Optional for wire compatibility with messages sent before
-    /// reporting existed.
+    /// Detached Ed25519 signature by the sender over the canonical
+    /// preimage `ChatModerationProof.signedContent` — `variant.body`
+    /// bound to this payload's `messageID`, `groupID`, and
+    /// `sentAtMillis`, so the proof attests one concrete send rather
+    /// than a transferable (body, signature) pair. A recipient retains
+    /// this so a voluntary disclosure can satisfy the Authority
+    /// contract's authenticity rule without exposing an envelope or
+    /// any recipient secrets. Optional for wire compatibility with
+    /// messages sent before reporting existed.
     public let moderationAuthenticityProof: String?
 
     public init(
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
@@ -1,0 +1,68 @@
+import Foundation
+
+/// Canonical preimage of the moderation authenticity proof — the exact
+/// string the sender signs and a reporter later discloses to an
+/// Authority as `EvidenceItem.disclosedContent`.
+///
+/// The proof must bind the body to its message identity. A signature
+/// over the bare body would be transferable: anyone holding a
+/// (body, signature) pair — including a third party who was never a
+/// recipient — could file it as evidence for any message, in any group,
+/// at any time, and the Authority couldn't tell. Folding `messageID`,
+/// `groupID`, and `sentAtMillis` into the signed bytes pins the proof
+/// to one concrete send.
+///
+/// The preimage is a JSON object with **UTF-8 byte-ordered keys**
+/// (`JSONEncoder.OutputFormatting.sortedKeys`), matching the canonical
+/// form every other moderation signing payload uses. The Authority
+/// never reconstructs these bytes — it verifies the signature over
+/// `disclosedContent` verbatim and parses the context fields out of it —
+/// so the only implementations that must agree byte-for-byte are the
+/// signing sender and the verifying recipient, both of which are this
+/// function.
+public enum ChatModerationProof {
+    /// Version discriminator inside the preimage so a future shape
+    /// change can't be confused with this one.
+    public static let version = 1
+
+    public static func signedContent(
+        messageID: UUID,
+        groupID: String,
+        sentAtMillis: Int64,
+        body: String
+    ) throws -> String {
+        struct Preimage: Encodable {
+            let body: String
+            let groupID: String
+            let messageID: String
+            let proofVersion: Int
+            let sentAtMillis: Int64
+
+            enum CodingKeys: String, CodingKey {
+                case body
+                case groupID = "group_id"
+                case messageID = "message_id"
+                case proofVersion = "proof_version"
+                case sentAtMillis = "sent_at_millis"
+            }
+        }
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = [.sortedKeys]
+        let data = try encoder.encode(Preimage(
+            body: body,
+            groupID: groupID,
+            messageID: messageID.uuidString.lowercased(),
+            proofVersion: version,
+            sentAtMillis: sentAtMillis
+        ))
+        return String(decoding: data, as: UTF8.self)
+    }
+
+    /// Millisecond timestamp recovered from a stored `sentAt` date.
+    /// The receiver derives `sentAt` as `Date(millis / 1000)`; rounding
+    /// the reverse conversion recovers the wire value exactly (Double
+    /// carries sub-microsecond precision at epoch-seconds magnitude).
+    public static func sentAtMillis(from sentAt: Date) -> Int64 {
+        Int64((sentAt.timeIntervalSince1970 * 1000).rounded())
+    }
+}
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 
 /// Canonical preimage of the moderation authenticity proof — the exact
@@ -20,6 +21,15 @@ import Foundation
 /// so the only implementations that must agree byte-for-byte are the
 /// signing sender and the verifying recipient, both of which are this
 /// function.
+///
+/// The group is bound through `group_binding` —
+/// `SHA-256("onym-moderation-proof-v1|" + groupID + "|" + messageID)` —
+/// rather than the raw group id. The recipient recomputes it from the
+/// same locals, so the binding is just as tight, but the disclosed
+/// content carries no cross-report-linkable identifier: two reports
+/// from the same group hash to unrelated values (the message id salts
+/// them), so an Authority cannot correlate independent reports to one
+/// group or join them against on-chain data.
 public enum ChatModerationProof {
     /// Version discriminator inside the preimage so a future shape
     /// change can't be confused with this one.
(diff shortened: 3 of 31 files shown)


dev-onym (commented):
REQUEST_CHANGES — the wire-level proof and evidence retention need tightening before this ships; plumbing and idempotency logic are otherwise solid.

**Blocking**

- `SendMessageInteractor.swift:150` — the proof signs only `Data(body.utf8)`, with no binding to `messageID`, `groupID`, `sentAtMillis`, or recipient. A (body, signature) pair is therefore transferable and replayable forever: anyone who obtains it — including a third party who was never a recipient — can file it as evidence they "legitimately received," and the Authority cannot distinguish. Since this is a new wire field, changing the signed preimage later is a compat break; sign a canonical struct over `{messageID, groupID, sentAtMillis, body}` now.
- `SwiftDataMessageStore.swift:158` — `insertOrUpdate` unconditionally overwrites `encryptedModerationAuthenticityProof`, and inbound replays route through it (`IncomingMessageDispatcher.swift:811`). An abuser can re-send the same `messageID` with `moderation_authenticity_proof` omitted and silently erase the recipient's evidence, making Report vanish from the menu. Preserve a non-nil stored proof (and ideally reject body changes on replay of a known id).
- `ChatThreadViewController.swift:769` — indexing `orderedMessages[indexPath.row]` assumes the committed table rows match the model array, but `update(messages:)` sets `orderedMessages` before the animated diffable `apply`. A long-press during an in-flight insert can attach Report to the wrong message. Use `dataSource.itemIdentifier(for: indexPath)` + `messagesByID`.

**Non-blocking**

- `ModerationReportView.swift:63` — the picker shows raw `classId` strings ("csam") with no title or link to the consented `definition`; the reporter chooses an allegation class without seeing what it means. `ViolationClass` carries `definition` — surface it.
- `ReportFilingStore.swift` — `reportRecords` grows unbounded and retains full disclosed content indefinitely (encrypted, but not purged when the message, chat, or identity is deleted). Add pruning/keying by owner identity.
- `ChatThreadView.swift:434` — eligibility only checks the proof is *present*, not valid, so a sender who ships a garbage proof gets a Report entry that fails only at submit (`ModerationRepository.swift:521`). Verifying at receipt or in `makeReportableMessage` would both fix the dead-end UX and let you drop/flag bogus proofs.
- `ModerationRepository.swift:629` — `isDeterministicReportRejection` duplicates the existing `isDeterministicRejection` body verbatim; collapse to one. Also note 409 (already-filed) is treated as deterministic and discards the ledger row, so a retry mints a fresh `reportId`.
- `ModerationReportFlow.swift:67` — `rejection.message` renders authority-controlled text verbatim in the UI; consider bounding length or wrapping it.
- `SendMessageInteractor.swift:150` — signing now gates sending: any `signWithStellarKey` failure fails the whole text send, which previously had no crypto dependency beyond identity load. Acceptable if intended, but worth an explicit decision.
- `Packages/OnymChatsUI/Package.swift` — OnymChatsUI now depends on OnymModerationUI; a `(ReportableMessage) -> AnyView` factory would keep the chats layer moderation-agnostic.

Tests cover the payload/store/retry paths well, but nothing exercises `makeReportableMessage`'s eligibility rules (the controller tests stub `canReportMessage`), so the media/voice/legacy exclusions are unverified.

dev-onym (commented):
**REQUEST_CHANGES** — the evidence-retention guard is bypassable, and the class UI shows a content hash where the reporter expects prose.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift:150` — the new guard at `:157` keeps `body`+proof as a pair on a proof-less replay, but `sentAt` (`:150`), `groupID` (`:149`) and `encryptedSenderBlsPubkeyHex` (`:156`) are still overwritten unconditionally. `sentAtMillis` is part of the signed preimage (`ChatModerationProof.swift:30`), and the sender fully controls `payload.sentAtMillis`. Re-sending the same `messageID` with a shifted timestamp and no proof leaves old body + old proof + new `sentAt`, so `ReportableMessageFactory` recomputes a preimage the signature no longer covers and Report silently disappears — the exact bypass the guard was added to close. Preserve the whole signed tuple (`sentAt`, `groupID`, sender) whenever the retained proof is kept, and add a replay-with-shifted-timestamp test next to `SwiftDataMessageStoreTests.swift:431`.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationReportView.swift:76` — `ViolationClass.definition` is documented as a *content address (hash-or-url)* (`AuthorityManifest.swift:10-12`), so this renders `Definition: <sha256…>`. That defeats the stated goal ("the reporter must see what they are alleging"). Either resolve/link the definition or relabel it. Same screen, `:63`: the picker shows raw `classId` values as the only human-facing label.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:636` — 409 keeps the ledger row so the retry replays the same `reportId`, but no path ever converts that 409 into success. After an ambiguous failure plus an app restart (receipt was never persisted), every subsequent tap re-sends and surfaces the Authority's "already filed" as a plain error; the user can never reach the success state. Treat 409 as filed (persist a receipt, or re-fetch it) rather than propagating it as a rejection.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:529` — filing needs a populated `authorities` directory, and neither `ModerationReportFlow.start()` nor `fileReport` refreshes it, so a cold launch straight into Report throws `authorityUnavailable`. `ModerationReportFlow.swift:70` then maps that (and `reportingUnavailable` / `classOutsideMandate`) to "Try again to resend the exact signed report" — wrong advice for a permanent condition. Refresh on `start()`, and split permanent vs transient copy.
- `Packages/OnymModeration/Sources/OnymModeration/ReportFilingStore.swift:47` — the ledger is append-only with no cap or TTL, and successful rows retain the full disclosed body indefinitely in `UserDefaults.standard`. Nothing clears `app.onym.ios.moderation.reports` on identity deletion (`Sources/OnymIOS/OnymIOSApp.swift:697` wipes identity-scoped state only), so evidence outlives the identity that filed it. Add a bound and a wipe hook.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift:37` — `disclosedContent` embeds `group_id`, the on-chain group hex. The PR body says "no surrounding conversation context is disclosed", but this is a linkable identifier that lets an Authority correlate independent reports to one group (and cross-reference chain data). Worth stating explicitly in the sheet's disclosure copy, or binding to a per-message-derived value instead of the raw group id.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:155` — `try?` makes every signing failure silent and permanently unreportable for all recipients. Best-effort is defensible, but log it; a systematic keychain failure would otherwise be invisible.

Nits: `ModerationReportFlow.swift:40` swallows the underlying error behind one generic mandate message; `ModerationRepository.swift:571` dedupes on `classId`, so re-reporting the same message under a different class still mints a second allegation (looks intended — worth a comment).

dev-onym (commented):
**REQUEST_CHANGES** — the crypto and idempotency work is solid, but the evidence-preservation guard is bypassable and the ledger has two retention bugs.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift:158` — the anti-laundering guard only fires when the replay *omits* a proof. The sender holds the signing key, so an abuser re-sends the same `messageID` with an innocent body **and** a fresh valid proof over the new `(body, groupID, messageID, sentAt)` tuple; `preservesEvidence` is false, and body + `sentAt` + proof are all overwritten. Evidence is erased and Report disappears. Only an attacker who forgets to sign is blocked. A known-id row that already carries a proof should refuse content mutation outright (or the new tuple should be stored as a separate row), not merely require proof presence.
- `SwiftDataMessageStore.swift:173` — even on the preserve path, `encryptedAttachmentJSON` / album / voice / `replyToMessageIDString` are overwritten unconditionally, and those come straight from attacker-controlled `payload.attachment` (`IncomingMessageDispatcher.swift:830`). A proof-less replay that adds an image attachment keeps body+proof but flips `message.media.isEmpty` false, so `ReportableMessageFactory.make` returns nil — the exact silent-Report-removal the guard was written to prevent. The eligibility-gating fields need the same protection as the signed tuple.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:644` — only *resolved* rows are bounded. Unreceipted rows are retained forever by design, so every ambiguous transport failure (and every 409, which deliberately never resolves) permanently accrues a row of disclosed message content. Contradicts the stated goal that "disclosed content doesn't accumulate indefinitely"; needs a bound or an age cap on pending rows too.
- `ModerationRepository.swift:646` — pruning a resolved row destroys its idempotency key, so re-reporting a message whose row fell past 128 mints a fresh `reportId` and a second allegation, contradicting `ReportFilingStore.swift:6`. Consider retaining a small `(sourceMessageId, classId, reportId)` tombstone after dropping the evidence payload.
- `Sources/OnymIOS/OnymIOSApp.swift:717` — `(try? await identityRepository.currentIdentities()) ?? []` means a transient read failure purges the **entire** ledger for all identities, including undelivered artifacts. Skip the purge when the identity list can't be read rather than defaulting to an empty keep-set.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:153` — wire millis truncate (`Int64(x * 1000)`) while `ChatModerationProof.sentAtMillis(from:)` rounds from the stored `Date`. Recipients are unaffected (exact wire value), but the sender can never reconstruct its own preimage; `SendMessageInteractorTests` papers over this with a whole-second `now`. Persist the pending row with `Date(millis/1000)` so sender and recipient agree.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationReportView.swift:63` — the picker labels rows with the raw `classId`; `ViolationClass` carries no display name, so users choose between protocol identifiers. Worth a follow-up on the manifest shape.

Minor/positive: `group_binding` salting by `messageID` is a genuinely good unlinkability choice; local pre-flight verification in both the factory and `performFileReport` is right; delivery-gated-on-persistence and the 409-as-terminal split are correct. Note I could not build or run the suite here — no Swift toolchain in this environment — so the "full suite passes" claim is unverified from my side.

dev-onym (commented):
**REQUEST_CHANGES** — solid, well-tested design; one gap defeats the evidence-retention control this PR is built around, plus a few retention/UX issues.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift:158` — the replay guard only blocks *proof-less* replays. A sender can re-send the same `messageID` with a benign body **and a freshly valid proof over it** (they control body/messageID/sentAtMillis, so signing the new preimage is trivial); `!preservesEvidence` then overwrites `encryptedBody` + proof (line 164) and the retained evidence is gone before the recipient files. `IncomingMessageDispatcher.persistChatMessage` (line 811 → `insert`) has no already-seen guard, so this is reachable. Suggest: for `direction == .incoming` rows that already exist, never overwrite the signed tuple (body/sentAt/groupID/proof) at all — an inbound message id is immutable once received. `test_insertOrUpdate_replayWithProof_overwritesBodyAndProofTogether` currently pins the weaker behavior.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:175` — reportability is opt-in for the sender: a modified client that omits `moderation_authenticity_proof` makes its messages permanently unreportable, which is exactly the client an abuser runs. The dispatcher already verifies an envelope-level Ed25519 signature bound to `senderProfile.sendingPubkey` (`IncomingMessageDispatcher.swift:793`), and `EvidenceItem` explicitly allows "envelope commitment" — retaining that instead of (or as fallback to) a sender-supplied proof would close the opt-out. At minimum, call this limitation out in the PR scope.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:645` — the retention bound only prunes rows with a receipt. A 409 (`reportAlreadyFiled`, line 615) leaves a receipt-less row, so its disclosed content is retained indefinitely and every subsequent Submit tap re-POSTs the same report for a state that's already terminal. Consider marking 409 rows resolved (receipt-less but terminal) so they prune and short-circuit.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:661` — the ledger is purged only on identity removal. `MessageRepository.deleteGroup` wipes the local messages but the disclosed bodies survive in the ledger with no user-facing deletion. A leave/delete-group purge (or a settings-level "clear filed reports") would match the app's posture. (Storage itself is fine — `StorageEncryption` is `…ThisDeviceOnly`.)
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationReportView.swift:69` — the picker renders the raw `classId`; `ViolationClass` carries no display name, so users choose between strings like `csam`. Worth a lightweight label map or at least surfacing the definition link inline per row.
- Nit: `ModerationAuthorityClient.swift:319` and `ModerationRepository.swift:624` both correlate `receipt.reportId`; the repository check is now unreachable through `URLSessionModerationAuthorityClient` (it only fires for other client impls). Harmless, but worth a comment saying so.
- Nit: `ModerationReportFlow.swift` — after `reportAlreadyFiled` the message renders as a red error with Submit still enabled, inviting a retry that can only 409 again. Treat it as a terminal state like `receipt`.

Verified by reading: key derivations line up (`stellarPublicKey` == `MemberProfile.sendingPubkey` == `onym:key:` reporter/accused refs), group id encoding matches on both sides, and the millis→`Date`→millis round-trip is exact for both the truncating retry path and `sentAtMillis(from:)` (checked numerically over 200k values). I could not run the test suite here — it's an iOS/Xcode target and this worktree is Linux.

dev-onym (commented):
**REQUEST_CHANGES** — solid design and test coverage, but the evidence-retention guard is bypassable and one stated privacy property doesn't hold.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift:157` — the retention guard only fires when the replay carries *no* proof. An abuser re-sends the same `messageID` with a laundered body, a shifted `sentAtMillis`, and a **fresh valid signature over the new preimage**; `preservesEvidence` is false, so body/sentAt/proof/attachments are all overwritten and the row still passes `ReportableMessageFactory`. Nothing upstream blocks it — `IncomingMessageDispatcher` (`:811`) builds the row from `payload.messageID` and `MessageRepository.insert` (`:51`) is the only dedup. Since a legitimate replay is byte-identical, gate on direction instead: never mutate the signed tuple of an existing `.incoming` row.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift:27-33` — `group_binding = SHA256("…|groupID|messageID")` has no secret input: `message_id` ships in cleartext in the same disclosed JSON, and group IDs are on-chain lookup keys (`OnymChain/SEPContractClient.swift:106`). An Authority holding the chain's group set recovers the group by trial-hashing one candidate per group — exactly the "join them against on-chain data" adversary the comment claims to defeat. Either bind through a group-held secret (salt/commitment preimage) or drop the unlinkability claim from the doc comment and PR body.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:697` — `saveReportRecords()` caps only `isResolved` rows. A user who is offline or hitting a 5xx accumulates unbounded unresolved rows, each holding disclosed message content, in a `UserDefaults` plist. Worth a separate ceiling (oldest unresolved dropped with a log) rather than unbounded growth.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:625` — if `try saveReportRecords()` throws *after* the Authority accepted, the receipt is thrown away and `ModerationReportFlow` (`ModerationReportFlow.swift:88`) shows "could not be delivered. Try again to resend" for a report that succeeded; after relaunch the retry can only 409. Consider returning the receipt and downgrading the persist failure to a log.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:539` — the idempotency key includes `reporterMandate`, so re-consenting to the same authority (new mandate hash) lets the same message be filed a second time. If that's intended, say so in the comment next to the `classId` note.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationReportView.swift:99` — the picker renders raw `classId` ("csam"); `ViolationClass` carries no display name, so this is user-visible machine text. Worth a follow-up on the manifest shape.

Verified as correct: proof key matches `MemberProfile.sendingPubkey` (both derive from `stellarPublicKey`), sender/recipient `groupID` hex forms agree, the millis round-trip reconstructs the signed value, and `filedAt`'s `.iso8601` seconds precision survives the store round-trip so byte-identical retry holds across relaunch.

dev-onym (commented):
**APPROVE** — no correctness defects found; the crypto binding, ledger idempotency, and store-immutability changes hold up under inspection. Nits below.

**Verified by inspection (not just from the diff)**
- Group-ID agreement: sender signs `group.id`, receiver stores lowercase `%02x` hex (`IncomingMessageDispatcher.swift:773`) — identical strings, so the binding recomputes. `groupSecret` is non-optional and sealed into invitations, so both sides key the same hash.
- Key agreement: proof is signed by the Stellar key (`IdentityRepository.swift:261`), `memberProfiles.sendingPubkey` is that same key (`CreateGroupInteractor.swift:842`), and `accused` hex matches `IdentityModerationSigner.swift:16` — so the Authority can resolve the accused to a mandate.
- The incoming-row immutability change (`SwiftDataMessageStore.swift:146-159`) is safe: `IncomingMessageDispatcher.swift:844` is the only writer of incoming rows, and the ack/status latches (`markDeliveredAckSent`, `updateStatus`) are separate paths, so no legitimate update is lost. All `ChatThreadView`/`ChatsView` call sites are updated.

**Findings**
- `SendMessageInteractor.swift:792` — the retry path re-derives wire millis with `Int64(sentAt * 1000)` (truncating) instead of the `ChatModerationProof.sentAtMillis(from:)` helper the proof depends on. A 1 ms downshift would silently make every *retried* message unreportable for all recipients, with no diagnostic. I scanned 2M consecutive epoch-ms values through the `Double`/Core-Data round trip and found zero truncation failures, so this is latent, not live — but it should use the helper so the invariant isn't re-derived twice, two ways.
- `ChatModerationProof.swift:65-71` — the group binding is a prefix-keyed SHA-256 (`tag ‖ secret ‖ data`) rather than HMAC-SHA256. Not exploitable here (the binding is covered by the sender's signature, and length extension leaks no secret), but HMAC is the free, obviously-correct construction.
- `ModerationRepository.swift:676` — `saveReportRecords` bounds only *resolved* rows. A row that never resolves (persistent 5xx, long offline period) is kept forever, each retaining a disclosed body in UserDefaults. Worth an absolute or age-based backstop above the 128 resolved cap.
- `OnymIOSApp.swift:713-726` — the ledger is purged only on identity removal. Deleting the group, deleting the message, or "clear local message cache" leaves the disclosed body in the ledger. Defensible (it's the idempotency key), but confirm it matches the product's delete promise.
- `ModerationRepository.swift:647` — a `reportIdentifierMismatch` leaves the row unresolved permanently, and `ModerationReportFlow.submit()` maps it to generic "try again" copy, so every retry re-POSTs and then 409s into the terminal path. Cosmetic, but the copy is misleading.
- Test gap: nothing exercises send → wire → dispatcher → store → `ReportableMessageFactory` end to end. Eligibility depends on three independently-derived values agreeing (hex group id, truncated millis, shared secret); every current test builds both sides from the same literals, so a future change to any one normalization breaks reporting silently rather than failing a test.

PR #224, opened 2026-08-09
feat(moderation): accused-side case status — standing, query, snapshot

rinat-enikeev:
## Summary

First of four sequential PRs delivering the accused-side case lifecycle (status → response/appeal submissions → UI → tests). This one is the domain layer: everything a later surface needs to *discover and follow* a case, with standing resolved correctly.

- **Mandate-by-reference standing** — `mandateRecord(forRef:)` resolves the mandate a case names from mandate history by content hash. Standing in a case follows the mandate it was opened under, **not** the currently active record: switching authorities deactivates the old mandate but must not strand the user out of a case that mandate still governs (mandates are immutable, spec §12). This is the API the response/appeal PR builds on.
- **`caseStatus(caseId:mandateRef:)`** — fails closed when the referenced mandate isn't retained (`noMandate`), the active identity doesn't own it (new `caseAccessUnavailable`), or its Authority isn't in the directory (`authorityUnavailable`); single-flights per (authority, user, caseId); queries the reference Authority's signed status endpoint (`GET /v1/cases/{id}/status`, header credential `query-status:{caseId}:{timestamp}` — unchanged, already matched `authorize_case_party`); and correlates the returned `caseId` at both the client and repository layers (new `caseIdentifierMismatch`).
- **Offline snapshot** — `CaseStatusRecord` persists the last fetched status document encrypted at rest (same posture as the report ledger), upserted by `caseId`, bounded at 64 (oldest fetch pruned first), purged on identity removal alongside report records, and readable via `storedCaseStatus(caseId:)` for rendering before/without network.
- **Event log modeled** — `CaseStatus.events: [{at, kind}]?` matches what the reference actually serves (kinds like `case_opened`, `response`, `decided`, `appeal_filed`; details are never exposed to a party). The `assessment` document remains deliberately unmodeled until its UI lands.

## Reference-implementation alignment (recon findings)

Verified against onym-moderation `main` before writing code:

- The gate check already delivers `caseId` to the client: `caseOpen` carries it per notice, `banned` carries it inside `verdict.caseId` — no plumbing was needed.
- Status responses always emit every field (nullables as JSON `null`); `stage` is only `open|decided`, terminal detail lives in `disposition` (including case-level `reversed`). The existing Swift DTO's optionals decode this correctly.
- The status endpoint deliberately answers **404 for unknown case, bad signature, and non-party alike** (a case id must not be an existence oracle). The repository doc calls this out so the UI layer never presents 404 as proof of anything.
- Known pre-existing gap, out of scope here and flagged for the UI PR: `BannedView` renders `BanState.verdict` without running `VerdictValidator` (which currently has no production call sites).

## Scope

Domain only — no UI change, no respond/appeal (next PR), no gate-behavior change (a failed status fetch neither relaxes nor tightens the gate; it isn't wired into gate state at all). Per the agreed phasing, tests for this layer land in the final `test/case-lifecycle` PR.

## Verification

Full OnymIOSTests suite green locally on a signed simulator test host (no behavior of existing paths changed; the one exhaustive-switch update routes the new correlation error to the non-deterministic branch, matching `reportIdentifierMismatch`).

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/CaseStatusStore.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/CaseStatusStore.swift
@@ -1,0 +1,95 @@
+import Foundation
+import OnymFoundation
+
+/// The last authenticated case-status document fetched for one case,
+/// retained so the case surface can render offline and across
+/// relaunches. A snapshot is a cache of the Authority's answer — the
+/// Authority remains the source of truth, and a stale snapshot is
+/// replaced wholesale by the next successful fetch.
+public struct CaseStatusRecord: Codable, Sendable, Equatable {
+    public let caseId: String
+    /// Content reference of the mandate this case proceeds under.
+    /// Standing follows this mandate — not the currently active one —
+    /// so the record keeps the linkage the next fetch must re-resolve.
+    public let mandateRef: String
+    /// `onym:key:` reference of the mandated user, retained so
+    /// identity removal can purge the snapshots that identity owned.
+    public let user: String
+    public var status: CaseStatus
+    public var fetchedAt: Date
+
+    public init(
+        caseId: String,
+        mandateRef: String,
+        user: String,
+        status: CaseStatus,
+        fetchedAt: Date
+    ) {
+        self.caseId = caseId
+        self.mandateRef = mandateRef
+        self.user = user
+        self.status = status
+        self.fetchedAt = fetchedAt
+    }
+}
+
+public protocol CaseStatusStore: Sendable {
+    func load() -> [CaseStatusRecord]
+    func save(_ records: [CaseStatusRecord]) throws
+}
+
+/// Encrypted-at-rest store, same posture as the report ledger: the
+/// status document names the violation class and deadlines of an
+/// accusation against this device's user, so it never lands in
+/// UserDefaults as plaintext.
+public struct UserDefaultsCaseStatusStore: CaseStatusStore, @unchecked Sendable {
+    private static let recordsKey = "app.onym.ios.moderation.caseStatus"
+    private let defaults: UserDefaults
+
+    public init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    public func load() -> [CaseStatusRecord] {
+        guard let encrypted = defaults.data(forKey: Self.recordsKey),
+              let data = try? StorageEncryption.decrypt(encrypted),
+              let records = try? Self.decoder().decode([CaseStatusRecord].self, from: data)
+        else { return [] }
+        return records
+    }
+
+    public func save(_ records: [CaseStatusRecord]) throws {
+        let data = try Self.encoder().encode(records)
+        let encrypted = try StorageEncryption.encrypt(data)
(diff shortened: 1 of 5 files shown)


dev-onym (commented):
**APPROVE** — clean, consistent with the existing report-ledger patterns; one ownership gap to close before the UI PR consumes it.

- `ModerationRepository.swift:551` — `storedCaseStatus(caseId:)` returns a snapshot with **no standing check**. The fetch path enforces ownership via `caseStanding`, and `CaseStatusRecord.user` exists precisely to identify the owner, but the offline read path ignores both. On a multi-identity device the accused-side surface can render identity A's case document (class ID, deadlines, disposition) while B is active. No call sites yet, so it's latent — but the next PR is the consumer. Gate on `signer.userKeyID()`, or take `mandateRef` and route through `caseStanding`.
- `ModerationRepository.swift:539` — `mandateRecord(forRef:)` is public and deliberately unchecked (correct for the immutability rationale), but it's also billed as the API respond/appeal builds on. Make sure that PR calls `caseStanding`, not this, or the same ownership hole reappears on the write path.
- `ModerationRepository.swift:565-569` — the single-flight key is (authority, user, caseId) and omits `mandateRef`. Two concurrent calls for the same case under different refs collapse into one flight, and the persisted snapshot's `mandateRef` is whichever call won — the linkage "the next fetch must re-resolve" (CaseStatusStore.swift:12) is then wrong for the loser. Add `mandateRef` to the key or document why it can't diverge.
- `ModerationRepository.swift:578` — standing requires a populated directory, so on cold launch before `refresh()` lands, a valid retained mandate still fails `authorityUnavailable`. Consistent with `fileReport`, but this is the path a user hits immediately after a ban notice. Worth awaiting `startTask` or documenting `storedCaseStatus` as the intended fallback in the doc comment.
- `ModerationRepository.swift:539` — `try? $0.mandate.mandateHash()` collapses an encode failure into "not retained", surfacing as `.noMandate` and sending the user to re-consent for what is actually a serialization bug. Cheap to distinguish.
- `ModerationRepository.swift:612-615` — prune sorts `caseRecords` by `fetchedAt` descending, silently reordering the array that `insert(at: 0)` otherwise maintains. Correct today (oldest falls off first); brittle if anything later depends on insertion order.
- `OnymIOSApp.swift:714-720` — the "skip the purge if the list can't be read" comment above still says "every identity's ledger"; it now guards snapshots too. One-word fix.
- Tests: deferring to the final PR is the agreed phasing, but `maxCaseStatusRecords` pruning, upsert-by-caseId, `purgeCaseStatusRecords`, and both `caseIdentifierMismatch` guards are all unit-testable today against `InMemoryCaseStatusStore` — three PRs will stack on untested persistence. Consider pulling those four forward.
- Verified as claimed: the credential message matches `authorize_case_party` unchanged (`ModerationAuthorityClient.swift:364-375`), the custom RFC 3339 decoder handles the new `CaseEvent.at` in both fractional and whole-second forms, the existing `testStatusQueryUsesSignedHeaders` fixture already returns `caseId: "case-1"` and `"events":[]` so the new guard and field don't break it, and the exhaustive-switch update at `:1039` routes correlation failures to non-deterministic exactly like `reportIdentifierMismatch`.

PR #225, opened 2026-08-09
feat(moderation): accused-side case submissions — response and appeal

rinat-enikeev:
## Summary

Second of four sequential PRs for the accused-side case lifecycle (status ✅ #224 → **submissions** → UI → tests). This adds the write half of the domain: filing a signed statement on a case, and filing an appeal or new-holder claim.

- **`respond(caseId:mandateRef:statement:)`** — standing resolved by the case's mandate reference (#224), statement validated client-side against the reference's 16384-byte cap, signed over the existing canonical bytes (`caseId` inside the signature — cross-case replay stays closed), persisted before delivery, single-flighted per (authority, user, caseId, kind, statementHash).
- **`appeal(caseId:mandateRef:kind:statement:)`** — `.appeal` requires mandate ownership; `.newHolderClaim` deliberately does not (its premise is that the device's holder is *not* the mandated user — the reference accepts it unsigned for the same reason; we sign anyway since it costs nothing).
- **Exact-retry ledger** (`CaseSubmissionStore`, encrypted at rest): delivery never starts unless the signed artifact is durably persisted; ambiguous outcomes retain it so retry re-delivers byte-identical content; deterministic refusals discard it; an identical statement after acknowledgment returns the stored receipt. Resolved rows cap at 128, purged on identity removal alongside the other ledgers.
- **Receipts** — client `respond`/`appeal` now decode and correlate the reference's success bodies (`{caseId, recorded, late}` / `{caseId, filed, kind, note}`) instead of discarding them. `late` matters: the reference enforces **no response deadline** (late filings accepted and flagged), so client-side deadline gating stays advisory, per the plan.

## Reference-implementation alignment

- **The reference does not deduplicate responses** — every delivery files a new row (bounded at 32 per case, then 409 `case_state`). The client-side ledger is therefore the *only* double-file protection; that's why identical-statement resubmission short-circuits to the stored receipt. A duplicate caused by an ambiguous outcome where the server actually committed remains possible and is byte-identical by construction — the API offers no idempotency token to do better.
- Appeal gating (`disposition == ban`, 410 `window_closed`, supplements while `pending`, decided-review 409) is enforced server-side; the repository surfaces those as-is. UI-level advisory gating comes in the next PR.
- New-holder claims answer 200 unconditionally by design (non-oracular): `filed: true` never proves recording — only the authenticated status endpoint's `newHolderState` can. The receipt doc-comment says exactly this so the UI PR can't oversell it.
- Anti-oracle 404 (unknown case / bad signature / non-party are indistinguishable) is treated as a deterministic refusal: exact replay cannot change it, so the artifact is discarded rather than deadlocking the retry key.

## Scope

Domain only — no UI (next PR), no gate-behavior change, no media evidence on responses (statement + empty evidence array; disclosure UX for counter-evidence is its own decision). Per the agreed phasing, tests for this layer land in the final `test/case-lifecycle` PR.

## Verification

Full OnymIOSTests suite green locally on a signed simulator test host. One existing client test updated: its canned respond/appeal success body was `{"recorded":true}`, which the client used to discard; it now returns the reference's actual success shapes, which the client decodes and correlates — the test additionally asserts both receipts.

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/CaseSubmissionStore.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/CaseSubmissionStore.swift
@@ -18,8 +18,13 @@ public struct CaseSubmissionRecord: Codable, Sendable, Equatable {
 
     public let caseId: String
     public let mandateRef: String
-    /// `onym:key:` reference of the submitting user, for
-    /// identity-removal purge (same contract as the other ledgers).
+    /// `onym:key:` reference of the LOCAL identity that filed this
+    /// submission, for identity-removal purge (same contract as the
+    /// other ledgers). For a response or ordinary appeal this equals
+    /// the case mandate's user; for a new-holder claim it is the
+    /// device's current identity — the mandated user is by premise not
+    /// local, and keying the row to them would let the purge silently
+    /// drop a pending claim's retry artifact.
     public let user: String
     public let authorityName: String
     public let kind: Kind
--- a/Packages/OnymModeration/Sources/OnymModeration/CaseSubmissionStore.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/CaseSubmissionStore.swift
@@ -1,0 +1,125 @@
+import Foundation
+import OnymFoundation
+
+/// Exact signed case submission (response or appeal), retained across
+/// ambiguous delivery outcomes so a retry re-delivers byte-identical
+/// content — never differently signed content. Records with a receipt
+/// remain as a local idempotency ledger: the reference Authority does
+/// NOT deduplicate responses (each delivery files a new row, bounded
+/// at 32 per case), so client-side "same statement → same artifact →
+/// return the stored receipt" is the only thing keeping a double-tap
+/// or a re-entered identical statement from filing twice.
+public struct CaseSubmissionRecord: Codable, Sendable, Equatable {
+    public enum Kind: String, Codable, Sendable {
+        case response
+        case appeal
+        case newHolderClaim
+    }
+
+    public let caseId: String
+    public let mandateRef: String
+    /// `onym:key:` reference of the submitting user, for
+    /// identity-removal purge (same contract as the other ledgers).
+    public let user: String
+    public let authorityName: String
+    public let kind: Kind
+    /// Exactly one of these is set, matching `kind`. Stored whole —
+    /// including the signature — so retry re-encodes the identical
+    /// canonical bytes.
+    public let response: CaseResponse?
+    public let appeal: AppealSubmission?
+    public var responseReceipt: CaseResponseReceipt?
+    public var appealReceipt: AppealReceipt?
+    public let createdAt: Date
+
+    public init(
+        caseId: String,
+        mandateRef: String,
+        user: String,
+        authorityName: String,
+        kind: Kind,
+        response: CaseResponse? = nil,
+        appeal: AppealSubmission? = nil,
+        responseReceipt: CaseResponseReceipt? = nil,
+        appealReceipt: AppealReceipt? = nil,
+        createdAt: Date
+    ) {
+        self.caseId = caseId
+        self.mandateRef = mandateRef
+        self.user = user
+        self.authorityName = authorityName
+        self.kind = kind
+        self.response = response
+        self.appeal = appeal
+        self.responseReceipt = responseReceipt
+        self.appealReceipt = appealReceipt
+        self.createdAt = createdAt
+    }
+
+    /// Delivery reached a terminal acknowledged outcome; the row is
+    /// retained only as the idempotency ledger and prunes under the
+    /// resolved retention bound.
+    public var isResolved: Bool {
+        responseReceipt != nil || appealReceipt != nil
+    }
+}
+
+public protocol CaseSubmissionStore: Sendable {
+    func load() -> [CaseSubmissionRecord]
+    func save(_ records: [CaseSubmissionRecord]) throws
+}
+
+/// Encrypted-at-rest, same posture as the report ledger — a statement
+/// is the user's own words about an accusation and never lands in
+/// UserDefaults as plaintext.
+public struct UserDefaultsCaseSubmissionStore: CaseSubmissionStore, @unchecked Sendable {
+    private static let recordsKey = "app.onym.ios.moderation.caseSubmissions"
+    private let defaults: UserDefaults
+
+    public init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    public func load() -> [CaseSubmissionRecord] {
+        guard let encrypted = defaults.data(forKey: Self.recordsKey),
+              let data = try? StorageEncryption.decrypt(encrypted),
+              let records = try? Self.decoder().decode([CaseSubmissionRecord].self, from: data)
+        else { return [] }
+        return records
+    }
+
+    public func save(_ records: [CaseSubmissionRecord]) throws {
+        let data = try Self.encoder().encode(records)
+        let encrypted = try StorageEncryption.encrypt(data)
+        defaults.set(encrypted, forKey: Self.recordsKey)
+    }
+
+    private static func encoder() -> JSONEncoder {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .iso8601
+        return encoder
+    }
+
+    private static func decoder() -> JSONDecoder {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .iso8601
+        return decoder
+    }
+}
+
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — solid persist-before-deliver discipline and receipt correlation; the new-holder-claim path is inconsistent with the ledger's ownership contract.

- `ModerationRepository.swift:844` — a `.newHolderClaim` row stores `user: record.mandate.user`, i.e. the *mandated* (previous) holder. `CaseSubmissionStore.swift:24` documents that field as "the submitting user, for identity-removal purge", and `purgeCaseSubmissionRecords(keepingUsers:)` (`:773`) drops every row whose `user` isn't a current local identity. By the claim's own premise the mandated user is *not* a local identity, so the next identity-removal sweep silently deletes a pending new-holder claim's retry artifact — the one row the persist-before-deliver invariant exists to protect. Store the signing identity's key (`try await signer.userKeyID()`) as `user` and keep the mandated user separate, or exclude `.newHolderClaim` from the purge.
- `ModerationRepository.swift:600,728` — `requireOwnership: false` only relaxes half the precondition: `caseStanding` still requires `mandateRecord(forRef:)` to resolve, i.e. the *former* holder's mandate must still be retained on this device. A genuine new holder (fresh install, or after the old identity was removed and its mandate purged) throws `.noMandate` before reaching the relaxed check, so the path is reachable only in the same-device/identity-still-retained case. If that narrower scenario is the intent, say so in the doc comment; otherwise the standing resolution needs a mandateRef source that doesn't depend on local retention (the `CaseNotice` carries one).
- `ModerationAuthorityClient.swift:411` — `appeal` correlates `receipt.caseId` but not `receipt.kind`, even though the reference echoes it and the client already knows the submitted value. A response mixing up `appeal` vs `new-holder-claim` would be accepted and surfaced to the UI as-is. Cheap symmetric guard.
- `ModerationRepository.swift:941` — `saveCaseSubmissions()` applies the prune to `caseSubmissions` before the store write; if `caseSubmissionStore.save` throws, the pruned resolved rows are already gone from memory while still on disk, and `persistNewSubmission`'s rollback (`:861`) restores only the new row. Prune into a local, assign after a successful save.
- `ModerationRepository.swift:875,900` — `statementInvalid` is used for two internal invariant violations ("submission holds no response"/"no appeal") that no user input can cause; it'll render as a validation message in the UI PR. These are unreachable-by-construction — prefer a distinct error (or make the record enum carry its payload so the state is unrepresentable).
- Ledger growth is bounded only for *resolved* rows (`:945`); rows stuck in ambiguous state accumulate without limit. Same posture as the report ledger, so consistent — just noting it's now two unbounded pending sets in encrypted UserDefaults.
- Reference-alignment reasoning (no server-side response dedup → client ledger is the only double-file guard; 404 anti-oracle treated as deterministic; `late`/`filed` deliberately non-probative) matches the code and is documented at the right places. No repository-level test exercises the dedup/discard/prune paths in this PR — acceptable under the stated phasing, but those three are the parts most likely to regress silently.

dev-onym (commented):
**REQUEST_CHANGES** — the ledger discipline is sound, but the appeal-kind correlation is unverified for the path it most affects, and two failure modes leave the ledger able to double-file.

- `ModerationAuthorityClient.swift:418` — the kind check compares the receipt against `AppealSubmission.Kind.rawValue`, which is `"new-holder-claim"` (hyphens, line 142), while the spec identifier quoted in this file's own doc comment is `new_holder_claim` (line 138). If the reference echoes anything but the hyphenated form, *every* new-holder filing throws `malformedResponse` after a successful 200. Nothing in the repo pins the reference's echoed string, and the only test (`ModerationAuthorityClientTests.swift:449`) covers `"appeal"`. Please confirm against `authority/src/api.rs` and add a `.newHolderClaim` case to that test.
- `ModerationAuthorityClient.swift:409,418` / `ModerationRepository.swift:903` — a post-200 correlation failure (caseId or kind mismatch) isn't `AuthorityClientError.rejected`, so `isDeterministicReportRejection` returns false and the artifact is retained as "ambiguous". It isn't ambiguous: the server answered 200 and, per your own note, does not deduplicate — the retry files a second row. A 200 that fails correlation should be terminal (discard the artifact, surface the error).
- `ModerationRepository.swift:943` — if `purgeCaseSubmissionRecords` runs while a delivery is in flight, `finishSubmission` doesn't find the index and re-inserts the row, re-persisting a removed identity's statement text and defeating the purge contract asserted at line 780. The receipt already returns to the caller from `deliverResponse`; when the row is gone, prefer leaving it gone over resurrecting it (or re-check membership before insert). Same shape exists in the report path (line 1666), but this PR is where the guarantee is newly documented.
- `ModerationRepository.swift:686` vs `792` — the single-flight key omits `mandateRef` while the ledger dedupe includes it. Two concurrent calls differing only by `mandateRef` collapse into one flight, and the loser receives a receipt for a row it didn't create. Add `mandateRef` to `CaseSubmissionKey` (or drop it from the dedupe predicate) so the two notions of identity agree.
- `ModerationRepository.swift:698` — cancelling the awaiting caller runs the `defer` and frees the flight key while the unstructured `Task` keeps delivering. A subsequent call finds the persisted pending row and re-delivers it concurrently — a duplicate against a server with no idempotency token. Worth a `pending`-in-flight marker on the record, or holding the key until the inner task completes.
- `ModerationRepository.swift:962` — only resolved rows prune. Each edited-statement retry while the Authority is unreachable adds a permanent row of up to 16 KB with no bound; consider a pending cap too.
- `ModerationRepository.swift:970` — `submissionIndex(of:)` identifies a row by (caseId, kind, user, statements, createdAt). With an injected fixed test `clock()`, two distinct rows alias; a stored row id would be sturdier than tuple matching.

Non-blocking: the receipt doc comments and the new-holder ownership rationale are unusually clear, and prune-after-save is the right ordering fix.

dev-onym (commented):
**COMMENT** — solid, carefully-reasoned domain layer; one edge case inverts the PR's own double-file invariant, and the repository half ships untested.

- `ModerationRepository.swift:929-932`, `1045-1052` — post-200 correlation/decode failure discards the ledger row. The server *accepted* and doesn't dedupe, so the user's retry of the identical statement finds no row, signs a fresh artifact and files a **second** response — exactly the outcome the ledger exists to prevent. Discard is right for a deterministic 4xx (nothing was recorded); for post-acceptance it should mark the row terminal (e.g. `receipt == nil` + a `resolvedUnverified` flag consumed by `isResolved`) so a re-file short-circuits instead of duplicating.
- `ModerationAuthorityClient.swift:184-193`, `199-212` — `recorded`, `late`, `filed`, `kind` are all non-optional. Any Authority that omits `late` (or `recorded`) on success turns a *successful* filing into `malformedResponse` → discard → the duplicate above. Given `note` is already optional, decoding `late`/`recorded` as `Bool?` defaulted false/true costs nothing and removes a whole class of that failure. The `caseId`/`kind` correlation guards are worth keeping strict.
- `ModerationRepository.swift:991-1013` — the pending prune can drop a row whose delivery is still in flight; `finishSubmission` then no-ops (`:974`) and the acknowledgment is never persisted, so a later identical statement re-files. Excluding rows with a live flight key, or pruning only on the pending-insert path, closes it.
- `ModerationRepository.swift:713-723`, `790-796` — the janitor releases the key only after `task.value` settles, so a caller arriving in that window is served the *failed* flight's error rather than starting a retry. Harmless for deterministic refusals, mildly annoying for a transient timeout; worth clearing the key eagerly when the flight threw a non-deterministic error.
- `ModerationRepository.swift:800-802` — `caseSubmissions(caseId:)` isn't scoped by identity, so on a multi-identity device the UI can be handed another local identity's statement text. Add the `user` filter here rather than trusting every call site.
- `ModerationRepository.swift:1238-1244` — dropping the recovered-row re-insert in the *report* path is a behavior change to a ledger this PR otherwise doesn't touch. The reasoning (don't resurrect purged rows) is right, but it belongs with a test asserting the receipt still reaches the caller; today nothing covers it.
- Tests — `Tests/OnymIOSTests/ModerationRepositoryTests.swift` only gets signature fixups. The client-level receipt tests are good, but the ledger *is* the double-file protection and every claim in the PR body about it (persist-before-deliver, exact retry, discard-vs-retain, prune, purge) is currently unverified. If the phasing plan holds them for the test PR, at minimum land the retain/discard branch tests here — they're the ones a later refactor will silently break.
- Nits: `AppealSubmission.Kind` wire-value doc-comment at `ModerationAuthorityClient.swift:136-141` is the right call and well-evidenced by the new `testNewHolderClaimKindCorrelates…` test; `maxPendingCaseSubmissions` and `maxResolvedCaseSubmissions` being equal constants (`:96-105`) reads as accidental — a comment saying the equality is coincidental would help.

PR #226, opened 2026-08-09
feat(moderation): case UI — actionable open-case and ban surfaces

rinat-enikeev:
## Summary

Third of four sequential PRs for the accused-side case lifecycle (status ✅ #224 → submissions ✅ #225 → **UI** → tests). The open-case notice and the ban screen become actionable.

- **`ModerationCaseFlow` + `ModerationCaseView`** — authenticated case status with the persisted snapshot as a labeled offline fallback ("shown as of …"), the reference's event log (known kinds translated, unknown kinds shown verbatim rather than invented), a response composer while the case is open, and appeal / new-holder sections for a ban. Full state coverage: loading, offline snapshot, retry, submitted (late-flagged when the Authority says so), rejected with bounded authority-controlled text, deadline-closed annotations.
- **Advisory gating, as planned** — the composer availability follows returned state, but nothing hard-blocks near a boundary: the reference accepts late responses (flagging them) and answers its own 410 for a closed appeal window; local clocks only annotate.
- **Anti-oracle 404 honored in copy** — the Authority deliberately conflates unknown case / bad signature / non-party into one 404; the screen presents that as "neither confirms nor denies", never as proof.
- **Entry points**: `CaseNoticeDetailView`'s "responding isn't available yet" stub → "Review and respond" into the case screen; `BannedView` → in-app "Review case and appeal" (in-app outranks link-out; URL fallback retained) with the new-holder affordance routed to the same sheet's dedicated section (`banContext` shows appeal/new-holder before any status fetch, since the caller holds the verdict); Settings' open-case banner wired identically.
- **Copy stays factual**: filing an appeal "does not suspend the ban unless the consented terms say so"; a new-holder claim is "submitted for expedited human review" — `filed: true` is never presented as recorded, matching the reference's non-oracular contract.

## Enforcement invariants

Case-screen network activity never touches gate state: a failed status fetch annotates the screen and keeps the stale snapshot; it neither relaxes nor tightens enforcement. `BannedView` remains full-screen and non-dismissable; the case sheet rides on top of it.

Known pre-existing gap, unchanged and re-flagged: `BannedView` renders `BanState.verdict` without running `VerdictValidator` (no production call sites). Wiring validation into gate handling is enforcement-side work, deliberately not smuggled into a UI PR.

## Scope

UI + composition wiring only — no domain changes. `OnymSettings` gains an `OnymModeration` dependency (it now names `CaseNotice` in a factory signature). No moderator/admin features, no media evidence, no push infrastructure. Per the agreed phasing, flow/UI tests land in the final `test/case-lifecycle` PR.

## Verification

Full OnymIOSTests suite green locally on a signed simulator test host (all new SettingsView/ModerationSettingsView parameters are additive; existing call sites compile unchanged).

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
@@ -579,8 +579,18 @@ public actor ModerationRepository {
 
     /// The last fetched snapshot for a case, if any — for rendering
     /// offline or before the first fetch of a session completes.
-    public func storedCaseStatus(caseId: String) -> CaseStatusRecord? {
-        caseRecords.first { $0.caseId == caseId }
+    /// Owner-checked: a snapshot persisted for another local
+    /// identity's case never leaves the repository, so no caller can
+    /// render foreign case content while a network fetch is in flight.
+    public func storedCaseStatus(caseId: String) async -> CaseStatusRecord? {
+        guard let record = caseRecords.first(where: { $0.caseId == caseId }) else {
+            return nil
+        }
+        guard let userKey = try? await signer.userKeyID(),
+              record.user == userKey else {
+            return nil
+        }
+        return record
     }
 
     /// Drop every case snapshot not owned by one of `users`
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
@@ -23,8 +23,13 @@ public struct BannedView: View {
     /// the contact fallback remain for verdicts without a case path.
     let makeCaseFlow: (@MainActor (_ caseId: String, _ mandateRef: String) -> ModerationCaseFlow)?
 
+    private enum CaseSheet: String, Identifiable {
+        case appeal, newHolder
+        var id: String { rawValue }
+    }
+
     @Environment(\.openURL) private var openURL
-    @State private var showsCase = false
+    @State private var caseSheet: CaseSheet?
 
     public init(
         state: BanState,
@@ -91,13 +96,16 @@ public struct BannedView: View {
                     // In-app appeal outranks a link-out: the flow
                     // retains the exact signed filing and its receipt.
                     if let builder = caseFlowBuilder {
-                        SettingsPrimaryButton(action: { showsCase = true }) {
+                        SettingsPrimaryButton(action: { caseSheet = .appeal }) {
                             Text("Review case and appeal")
                         }
                         .accessibilityIdentifier("moderation.banned.appeal")
-                        .sheet(isPresented: $showsCase) {
+                        .sheet(item: $caseSheet) { sheet in
                             NavigationStack {
-                                ModerationCaseView(flow: builder())
+                                ModerationCaseView(
+                                    flow: builder(),
+                                    focusNewHolder: sheet == .newHolder
+                                )
                             }
                         }
                     } else if let appealURL = state.appealURL {
@@ -118,8 +126,9 @@ public struct BannedView: View {
                                 openURL(url)
                             } else if caseFlowBuilder != nil {
                                 // The case sheet carries the
-                                // new-holder section (banContext).
-                                showsCase = true
+                                // new-holder section (banContext);
+                                // open it scrolled to that section.
+                                caseSheet = .newHolder
                             } else {
                                 onNewHolderClaim?()
                             }
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
@@ -17,17 +17,34 @@ public struct BannedView: View {
     /// in-app path pass `nil`, and the screen falls back to the
     /// authority's contact instead of offering a button that lies.
     let onNewHolderClaim: (() -> Void)?
+    /// Builds the in-app case flow for the banning verdict's case.
+    /// When present and the verdict is attached, appeal and new-holder
+    /// filings happen in-app; URLs (if the authority serves them) and
+    /// the contact fallback remain for verdicts without a case path.
+    let makeCaseFlow: (@MainActor (_ caseId: String, _ mandateRef: String) -> ModerationCaseFlow)?
 
     @Environment(\.openURL) private var openURL
+    @State private var showsCase = false
 
-    public init(state: BanState, onNewHolderClaim: (() -> Void)? = nil) {
+    public init(
+        state: BanState,
+        onNewHolderClaim: (() -> Void)? = nil,
+        makeCaseFlow: (@MainActor (_ caseId: String, _ mandateRef: String) -> ModerationCaseFlow)? = nil
+    ) {
         self.state = state
         self.onNewHolderClaim = onNewHolderClaim
+        self.makeCaseFlow = makeCaseFlow
+    }
+
+    /// In-app path into the banning case, when the verdict is attached.
+    private var caseFlowBuilder: (@MainActor () -> ModerationCaseFlow)? {
+        guard let makeCaseFlow, let verdict = state.verdict else { return nil }
+        return { makeCaseFlow(verdict.caseId, verdict.mandateRef) }
     }
 
     /// Whether the new-holder path can actually be reached from here.
     private var hasNewHolderPath: Bool {
-        state.newHolderURL != nil || onNewHolderClaim != nil
+        state.newHolderURL != nil || onNewHolderClaim != nil || caseFlowBuilder != nil
     }
 
     public var body: some View {
@@ -71,7 +88,19 @@ public struct BannedView: View {
                 SettingsFootnote("The ban covers this device on this app only. The Onym protocol itself remains open.")
 
                 VStack(spacing: 10) {
-                    if let appealURL = state.appealURL {
+                    // In-app appeal outranks a link-out: the flow
+                    // retains the exact signed filing and its receipt.
+                    if let builder = caseFlowBuilder {
+                        SettingsPrimaryButton(action: { showsCase = true }) {
+                            Text("Review case and appeal")
+                        }
+                        .accessibilityIdentifier("moderation.banned.appeal")
+                        .sheet(isPresented: $showsCase) {
+                            NavigationStack {
+                                ModerationCaseView(flow: builder())
+                            }
+                        }
+                    } else if let appealURL = state.appealURL {
                         SettingsPrimaryButton(action: { openURL(appealURL) }) {
                             Text("Appeal this ban")
                         }
@@ -87,6 +116,10 @@ public struct BannedView: View {
                         Button {
                             if let url = state.newHolderURL {
                                 openURL(url)
+                            } else if caseFlowBuilder != nil {
+                                // The case sheet carries the
+                                // new-holder section (banContext).
+                                showsCase = true
                             } else {
                                 onNewHolderClaim?()
                             }
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift
@@ -93,8 +93,14 @@ public struct BannedView: View {
                 SettingsFootnote("The ban covers this device on this app only. The Onym protocol itself remains open.")
 
                 VStack(spacing: 10) {
-                    // In-app appeal outranks a link-out: the flow
-                    // retains the exact signed filing and its receipt.
+                    // In-app appeal is primary — the flow retains the
+                    // exact signed filing and its receipt. The external
+                    // appellate link SUPPLEMENTS it rather than being
+                    // replaced: the case screen can be unable to help
+                    // (mandate no longer retained, authority missing
+                    // from the directory, anti-oracle 404), and a
+                    // permanent ban's header explicitly promises the
+                    // external appellate remains available.
                     if let builder = caseFlowBuilder {
                         SettingsPrimaryButton(action: { caseSheet = .appeal }) {
                             Text("Review case and appeal")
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
REQUEST_CHANGES — solid state modeling and honest copy, but the offline snapshot bypasses the repository's ownership check, and the shared appeal state swallows new-holder failures.

- `ModerationCaseFlow.swift:55` — `storedCaseStatus(caseId:)` is rendered unconditionally, without comparing the record's `user`/`mandateRef` against the active identity or this flow's `mandateRef`. `caseStatus()` fails closed with `.caseAccessUnavailable` for a non-owned mandate (`ModerationRepository.swift:604`), and snapshots are purged only on identity *removal*, not on switch — so after switching identities the screen shows the other identity's violation class, deadlines and event log, with the ownership error rendered merely as an annotation beside it. Drop (or refuse to render) the snapshot when `mandateRef` doesn't match, and clear `state.status` when the fetch returns `.caseAccessUnavailable`.
- `ModerationCaseView.swift:271`,`237` — `appealErrorMessage` is only rendered inside `appealSection`. A failed new-holder claim shows nothing in `newHolderSection`, and when `showsAppealSection` is false (non-ban disposition, or `appealState` denied/granted) the error is dropped entirely — the user taps "File new-holder claim" and gets silence. Render the error in `newHolderSection` too, or split the error/receipt state per kind.
- `ModerationCaseFlow.swift:134-142` — `showsAppealSection` consults `state.status` first, so a *stale snapshot* predating the ban (`stage: "open"`, `disposition: nil`) hides the appeal section on the ban surface — "Review case and appeal" opens a screen with no appeal affordance, exactly in the offline case the snapshot exists for. Prefer `banContext ||` when the last fetch failed (`statusErrorMessage != nil` / `snapshotFetchedAt != nil`).
- `ModerationCaseView.swift:271` + `ModerationCaseFlow.swift:26` — one `appealReceipt` slot serves both kinds: filing a new-holder claim erases the appeal confirmation, and a later appeal makes the new-holder editor reappear as if nothing was filed. `newHolderText` also isn't cleared on success (unlike `responseText`/`appealText`), inviting a duplicate filing.
- `ModerationCaseFlow.swift:126` — `canRespond` requires a fetched `stage == "open"`, so an offline user with no snapshot gets no composer at all. That's a hard block on the response path, which the file's own doc comment says this flow never does; the notice already asserts the case is open. Consider allowing the composer when the status is unavailable and letting the Authority arbitrate.
- `SettingsView.swift:171` — the moderation section now requires all three factories; a caller passing settings+consent but not `makeModerationCaseFlow` silently loses the entire MODERATION entry rather than degrading. Not "additive" for optional-arg callers — worth gating case entry separately from the section.
- Minor: `ModerationCaseView.swift:41` `.task { await flow.start() }` re-runs on every re-appearance, momentarily relabeling a fresh status as "Shown as of …"; and `BannedView.swift:119` routes the new-holder tap into the same sheet with no scroll anchor, landing the user on STATUS/appeal rather than the section they asked for.

dev-onym (commented):
**REQUEST_CHANGES** — solid, honest surfaces; two filing-receipt/routing defects worth fixing before merge.

- `ModerationCaseView.swift:41-46` — the response receipt lives *inside* `responseSection`, which is gated on `flow.canRespond`. `submitResponse` clears the editor and then `refresh()`es; if the fresh status comes back `stage != "open"` (decided in the interim, or an authority that closes on first response), the whole section — receipt included — vanishes. The user's statement is gone from the box with zero confirmation it landed, on the one screen whose contract is "every filing surfaces the Authority's actual answer." Render `state.responseReceipt` outside the `canRespond` gate.
- `ModerationCaseView.swift:47-49` — same shape for the appeal: `appealReceipt` is inside `appealSection`, and `showsAppealSection` goes false as soon as a fresh status reports `appealState` outside `nil/none/pending` (or `disposition == "reversed"`) — i.e. exactly when the appeal succeeded. The receipt disappears at the moment it matters most.
- `BannedView.swift:125-131` — new-holder precedence is inverted relative to the appeal button and to the PR body: `newHolderURL` wins over the in-app flow, so for any authority that serves that URL the in-app new-holder section is never reached from its own button (only sideways via "Review case and appeal"). Also makes an explicitly-passed `onNewHolderClaim` dead whenever `caseFlowBuilder != nil`. Either flip to match the appeal path or amend the PR body — the current code isn't what it describes.
- `ModerationCaseFlow.swift:88` — `try? await repository.refresh()` refetches the whole directory on *every* status refresh, including the post-submit refresh and pull-to-refresh. It also flips repository-wide `fetchStatus` to `.fetching`/`.failed` and `publish()`es, which `ModerationConsentFlow` observes (`ModerationConsentFlow.swift:64`) — a case-screen pull can repaint an unrelated consent surface with "Couldn't load the authority list." Gate it on the mandate's authority actually being absent from the directory.
- `ModerationCaseView.swift:200-217` — once `newHolderReceipt` is set the editor is removed permanently. `filed: true` is explicitly non-oracular (`ModerationAuthorityClient.swift:196-200`), and the repository dedupes by statement hash, so a corrected claim *is* a legitimate new filing — but this screen can't file it.
- `ModerationCaseFlow.swift` has no tests despite being pure, injectable logic (`now:`, repository) — `canRespond`, `showsAppealSection`, and especially the snapshot `mandateRef` guard at `:69-75` are exactly what a later UI-test PR won't cover well. Deferring flow tests here is a bigger gap than deferring view tests.
- Nit: `OpenCaseBanner.swift:105-116` — the `if let makeCaseFlow {` branch leaves `SettingsCard` and its body un-indented one level, unlike the `else`.
- Fine as-is: 404 anti-oracle copy, bounded authority text (`prefix(280)`), snapshot-standing check, per-kind receipts/errors, and keeping `BannedView` gate state untouched by case-screen network activity.

dev-onym (commented):
**REQUEST_CHANGES** — solid, careful UI work; two correctness issues on the ban screen and the snapshot standing claim, plus some smaller ones.

- `BannedView.swift:111` — the in-app path *replaces* rather than supplements `state.appealURL` (`else if`). The PR body says "URL fallback retained", but once `makeCaseFlow` + `verdict` are present the external appeal link is unreachable — including when the case screen can't help: `noMandate` (mandate not retained, e.g. identity removed after the ban), the anti-oracle 404, or an authority missing from the directory. On a permanent ban the header copy explicitly promises "appealable at the authority's external appellate". Render both (primary in-app, secondary link), or fall back to the URL when the flow reports it can't resolve the case.
- `BannedView.swift:125` — the new-holder button does the opposite: `newHolderURL` link-out outranks in-app, contradicting the appeal rule two blocks up. Pick one ordering; in-app is the stated principle.
- `ModerationCaseFlow.swift:67` — the snapshot standing guard doesn't achieve what its comment claims. `mandateRef` is supplied by the same caller that supplied `caseId`, so `stored.mandateRef == mandateRef` is tautological for every real call site; the field that identifies the owner (`CaseStatusRecord.user`) is never checked. Combined with `refresh()` doing a network directory fetch (line 82) *before* `caseStatus` runs the local ownership check, another identity's snapshot — violation class, deadlines, event log — renders for the whole request timeout. Compare `stored.user` against the active identity (or defer the snapshot render until standing resolves).
- `ModerationCaseFlow.swift:82` — `repository.refresh()` isn't side-effect-free: it sets `fetchStatus = .fetching` / `.failed` and `publish()`es. A failed case-screen refresh (or pull-to-refresh, or the refresh after each submission) will flip "Couldn't load the authority list." onto the moderation settings surface. Prefer a non-publishing directory-ensure, or only refresh when `authorityUnavailable` actually surfaced.
- `ModerationCaseFlow.swift:122` — `isSubmittingAppeal` is shared across both kinds, which undercuts the per-kind independence the receipts/errors go to lengths to preserve: an in-flight new-holder claim disables the appeal button and makes it read "Filing…", and vice versa (`ModerationCaseView.swift:196,304`). Split the flag.
- `ModerationCaseFlow.swift:238` — 410 → "The appeal window has closed." is in `submissionMessage`, shared with `submitResponse`; a 410 on a response filing would be mislabeled. Key off the kind.
- `ModerationCaseView.swift:281` — after a new-holder receipt the composer is replaced permanently for the session, but the reference answers `filed: true` unconditionally, so that 200 proves nothing was recorded. Keep the refile path available.
- Nit: `OpenCaseBanner.swift:105` — `if let makeCaseFlow {` leaves `SettingsCard` un-indented inside the branch.
- No tests for `showsAppealSection` / `canRespond` / error mapping. Deferring to `test/case-lifecycle` is the agreed phasing, so this isn't a blocker — just noting the branch logic is where the risk sits.

dev-onym (commented):
**REQUEST_CHANGES** — solid, well-reasoned UI layer, but the in-app new-holder path can dead-end the one user this screen exists to protect.

- `BannedView.swift:148` — **new-holder regression.** When `caseFlowBuilder != nil`, `state.newHolderURL` and `onNewHolderClaim` become unreachable. The canonical new-holder case is a resold device with a fresh install: DeviceCheck bits survive, the mandate store does not, so `repository.appeal(kind: .newHolderClaim)` → `caseStanding` → `mandateRecord(forRef:)` is nil → `ModerationError.noMandate` (`ModerationRepository.swift:618`), and the sheet can only say "the mandate … is no longer on this device." A working authority URL is suppressed in favor of a form that always fails. This is exactly the "button that lies" the property's own doc comment at `BannedView.swift:13` forbids. The appeal path got this right (external link kept as a supplement, `BannedView.swift:131`); make new-holder symmetric — keep the `newHolderURL`/contact affordance alongside the sheet, or fall back to it when the flow reports `noMandate`.
- `BannedView.swift:46` — `caseFlowBuilder` routes signed filings using `verdict.caseId` / `verdict.mandateRef` from a `BanState.verdict` that is still never run through `VerdictValidator`. The PR re-flags the gap, which is fair, but it was previously display-only; it now selects a case identifier *and* suppresses the URL fallback. Worth at least gating `caseFlowBuilder` on a validated verdict.
- `ModerationCaseView.swift:60` — `proxy.scrollTo("newHolder")` in `onAppear` fires before the scroll content is laid out and is commonly a no-op in practice; combined with the section rendering only when `banContext`, the "I'm this device's new owner" tap can silently land on the status card. Defer it (`Task { @MainActor in … }` / after first layout), and consider `anchor: .center` for a short section.
- `ModerationCaseFlow.swift:79` + `:134`/`:168` — `refresh()` returns early when `isLoading`, so the post-submission refresh is silently dropped if a pull-to-refresh (`ModerationCaseView.swift:68`) is in flight; the user files a response and the status card keeps saying "Response on file: No". Queue a follow-up refresh instead of no-oping.
- `ModerationCaseFlow.swift:112` — if the directory-recovery `repository.refresh()` itself throws, the error escapes the `authorityUnavailable` mapping and the user gets the generic "Couldn't reach your authority" instead of the actionable connectivity copy.
- `ModerationCaseView.swift:105`/`:113`/`:118` — rejection messages are bounded to 280 chars, but `classId`, `appealState`, `newHolderState`, unknown `stage` (`stageLine` default) and unknown event `kind` are rendered verbatim and unbounded. Same authority-controlled-text principle applies; bound them.

Correct and good: the owner check in `storedCaseStatus` (`ModerationRepository.swift:585`), clearing status on `caseAccessUnavailable` while preserving the snapshot otherwise, per-kind receipts/errors, receipts rendered outside the composer gates, the 404 anti-oracle copy, and additive optional factories so non-wiring call sites degrade rather than vanish. Deferring flow/UI tests to the final PR is fine as agreed, but the new-holder routing above is the case that most needs one.

dev-onym (commented):
COMMENT — solid, well-reasoned UI PR; the advisory-gating and receipt-survival design holds up. A few real gaps, none blocking.

- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationCaseFlow.swift:211` — `canRespond` contradicts the one hard-standing failure. `performRefresh` (`:105-112`) deliberately clears `status` on `caseAccessUnavailable` ("nothing about the case may stay on screen"), but `canRespond` then falls through to `return !banContext` → the notice path renders the composer, and the submit is guaranteed to fail with the same error. Track "refused standing" explicitly and hide the composer for it.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:757` — comment now stale: "A fresh holder with nothing retained files from the ban surface … that flow lands with the case UI." This is that PR, and the in-app new-holder path still routes through `caseStanding`, which throws `.noMandate` without a retained mandate. The external-URL supplement (`BannedView.swift:176`) is the right mitigation, but the repository comment promises a path that doesn't exist — correct it or drop it.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/BannedView.swift:45` — the known unvalidated-`BanState.verdict` gap grows in scope here: `verdict.caseId`/`verdict.mandateRef` no longer just render, they now select what the client fetches and what it signs a statement under. Bounded (the repo resolves `mandateRef` against locally retained mandates and requires ownership for `.appeal`), but a forged verdict can steer a signed appeal at an arbitrary `caseId`. Worth naming in the PR body's re-flag, since "display-only" no longer describes it.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationCaseView.swift:54` — `.id("newHolder")` on a `Group` propagates to *each* child (label + card), so two views share the anchor ID and `scrollTo` targets an ambiguous one. Put the `.id` on a single wrapping container.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationCaseView.swift:134` — first render has `status == nil && !isLoading` (the `.task` hasn't suspended into `start()` yet), so "No case status is available yet." + **Try again** flash before the fetch begins. Seed `isLoading = true` in `State`, or gate on a `hasAttempted` flag.
- `ModerationCaseView.swift:77` + `ModerationCaseFlow.swift:84` — `.refreshable` returns the moment `refresh()` queues a follow-up, so the pull-to-refresh spinner dismisses while the fetch is still in flight. Cosmetic, but the queueing fix makes it more likely.
- `ModerationCaseFlow.swift:243,252` — `appealDeadlinePassed`/`responseDeadlinePassed` call `now()` inside non-observed computed properties; if a deadline crosses while the screen is open, the advisory annotation never appears until some other state change. Fine to leave, worth a note.
- `Packages/OnymModerationUI/Sources/OnymModerationUI/OpenCaseBanner.swift:133` — "Responding from this surface isn't wired up" is developer-facing phrasing in user copy; the old "isn't available yet" read better for the fallback.
- No tests in this PR — consistent with the stated phasing, but `storedCaseStatus`'s new owner check (`ModerationRepository.swift:585`) is domain behavior, not UI; a direct repository test for it belongs here rather than in the UI-test PR.

PR #227, opened 2026-08-09
feat(moderation): live enforcement backend — wire the deployed interface service

rinat-enikeev:
## Summary

Removes the last gap between the shipped moderation machinery and the deployed services: the app talked to `StubEnforcementBackendClient` in every build because no backend existed. It does now — `moderation.onym.app` (interface/enforcement), `authority.onym.app` (authority), and the `onym-authorities` v1 directory release are all live and were verified by hand before writing this.

- **`URLSessionEnforcementBackendClient`** — `POST /v1/enroll`, `/v1/mandates/countersign`, `/v1/gate-check`, shapes pinned against the reference `apple` service: camelCase keys, base64 `Data`, RFC 3339 second-precision timestamps (Foundation `.iso8601` on both sides), `GateCheckResult` decoded from the server's deliberately Swift-enum-shaped JSON, `{error, message}` envelope surfaced as `AuthorityClientError.rejected` with the apple vocabulary kept verbatim. The countersign body is the canonical mandate encoding — the server recomputes signing bytes from the raw body, verifies the user's signature is the *only* one present, and returns only its own signature, so no consented field can change behind the user's back.
- **Wiring** — DEBUG (non-UI-test) and release construct the real client; UI tests keep the `--moderation-scenario` stub. Nothing else about gate cadence, grace, or enforcement behavior changes.
- **Trust switches** — `enforceVerdictSignatures` ON (the deployed authority signs verdicts with the directory-pinned operator key; no UI-test path routes fixtures through `VerdictValidator`). `enforceManifestSignatures` deliberately **stays OFF**, with the reason documented in place: the client verifies a detached signature at `<manifest-url>.sig`, and the deployed authority currently answers **404** for `https://authority.onym.app/manifest.json.sig` (verified 2026-08-09). Flipping it before that asset exists would hard-fail every consent. **Deployment follow-up: publish the `.sig` (operator signature over the exact manifest bytes), then flip the switch.**

## What this unlocks

The full loop end-to-end on a device build: consent → enroll (DeviceCheck) → countersign → mandate registration with the authority → report → moderator verdict (`POST /v1/cases/{id}/decide`, bearer token) → gate check delivers case-open/ban → case screen (status/respond/appeal, #226) → reversal picked up by the next gate check.

Practical notes for dogfooding: DeviceCheck attests only on real hardware — with `deviceCheck: true` server-side, simulators land in `gate-check-required` by design; session signatures are single-use with a freshness window, which the existing signed-payload builders already satisfy.

## Scope

One new file + composition wiring + the trust-switch defaults. No changes to gate logic, repositories, or UI. `--enforcement-base-url <url>` (DEBUG builds) points the client at a local deployment; plain http is accepted for loopback only, and only in DEBUG binaries.

## Verification

Full OnymIOSTests suite green locally on a signed simulator test host. Live-deployment checks performed while writing: `/health` on both services, directory release asset, manifest fetch, and the `.sig` 404 that kept manifest enforcement off.

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/AuthorityManifestFetcher.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/AuthorityManifestFetcher.swift
@@ -24,13 +24,10 @@ public struct URLSessionAuthorityManifestFetcher: AuthorityManifestFetcher {
         enforceSignature: Bool = ModerationTrust.enforceManifestSignatures
     ) {
         self.session = session
-        if let decoder {
-            self.decoder = decoder
-        } else {
-            let d = JSONDecoder()
-            d.dateDecodingStrategy = .iso8601
-            self.decoder = d
-        }
+        // The shared tolerant decoder: `consentedManifest()` is
+        // load-bearing for verdict display, and a manifest that
+        // decodes here must keep decoding from the pinned snapshot.
+        self.decoder = decoder ?? ModerationJSON.decoder()
         self.enforceSignature = enforceSignature
     }
 
--- a/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
@@ -167,6 +167,11 @@ public enum CheckRequiredReason: String, Codable, Sendable, Equatable {
     case offlineGraceExpired
     /// Local policy: no successful check has ever completed.
     case neverChecked
+    /// Client-derived (never sent by the backend): the backend was
+    /// reachable and deterministically refused the session — a
+    /// replayed or invalid signature, or no enrollment/mandate on
+    /// record. Retrying with a fresh session usually clears it.
+    case backendRefused
     /// Local policy: the device clock now reads earlier than the last
     /// successful check, so elapsed time can't be trusted to bound
     /// staleness — the grace window is refused rather than extended.
--- a/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
@@ -169,9 +169,17 @@ public enum CheckRequiredReason: String, Codable, Sendable, Equatable {
     case neverChecked
     /// Client-derived (never sent by the backend): the backend was
     /// reachable and deterministically refused the session — a
-    /// replayed or invalid signature, or no enrollment/mandate on
-    /// record. Retrying with a fresh session usually clears it.
+    /// replayed or invalid signature (401/403). Retrying with a fresh
+    /// session usually clears it; a persistent refusal points at the
+    /// device clock being outside the server's freshness window.
     case backendRefused
+    /// Client-derived: the backend answered `no_mandate` — its record
+    /// of this device's enrollment/mandate is gone (state loss,
+    /// redeploy, or a mandate never countersigned). Not user-transient:
+    /// the recovery is consenting again, which re-runs enrollment,
+    /// countersignature, and registration — the gate flow routes this
+    /// state back to consent rather than to a retry that cannot succeed.
+    case enrollmentLost
     /// Local policy: the device clock now reads earlier than the last
     /// successful check, so elapsed time can't be trusted to bound
     /// staleness — the grace window is refused rather than extended.
--- a/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
@@ -226,10 +226,10 @@ public enum GateCheckResult: Codable, Sendable, Equatable {
 
 /// The interface vendor's enforcement backend — the only party
 /// holding the Apple DeviceCheck key, therefore the only possible
-/// write/read path for device marks. Not deployed yet: the app ships
-/// against `StubEnforcementBackendClient`, and a future
-/// `URLSessionEnforcementBackendClient` (SEPContractTransport-style
-/// wire client) slots in behind this protocol untouched.
+/// write/read path for device marks. Production builds ship
+/// `URLSessionEnforcementBackendClient` against the deployed service;
+/// UI-test builds keep `StubEnforcementBackendClient` for
+/// scenario-driven gates.
 ///
 /// One protocol for enrollment, countersignature, and gate check
 /// because all three are the same service sharing auth and transport.
--- a/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift
@@ -179,6 +179,12 @@ public enum CheckRequiredReason: String, Codable, Sendable, Equatable {
     /// the recovery is consenting again, which re-runs enrollment,
     /// countersignature, and registration — the gate flow routes this
     /// state back to consent rather than to a retry that cannot succeed.
+    ///
+    /// Never accepted from a 200 body: the reason is plain `Codable`,
+    /// so a backend *could* serialize it into `checkRequired`, but
+    /// `GateCheckRepository` normalizes that onto the same refusal
+    /// path as the `no_mandate` envelope instead of treating it as a
+    /// successful check.
     case enrollmentLost
     /// Local policy: the device clock now reads earlier than the last
     /// successful check, so elapsed time can't be trusted to bound
--- a/Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift
@@ -233,6 +233,17 @@ public actor GateCheckRepository {
     /// self-inflict a replay refusal (consentCompleted + the cadence
     /// loop, or a double-tapped retry). Bumping into the next second
     /// stays comfortably inside the server's freshness window.
+    ///
+    /// Known limits, accepted deliberately:
+    /// - Each same-second bump sends a timestamp one second further
+    ///   into the future, so an N-call burst drifts N seconds ahead —
+    ///   fine against a symmetric freshness window (bursts are small:
+    ///   retry taps + foreground + cadence), and the drift resets the
+    ///   moment a second passes between calls.
+    /// - Not persisted: a relaunch inside the same second could still
+    ///   replay. The relaunch path takes far longer than a second in
+    ///   practice, and the cost of a collision is one refused check
+    ///   that the next retry clears.
     private var lastSessionTimestamp = Date.distantPast
 
     private func performAttempt(for record: MandateRecord) async -> AttemptOutcome {
@@ -264,7 +275,18 @@ public actor GateCheckRepository {
                 timestamp: timestamp,
                 signature: signature
             )
-            return .success(await sanitize(try await backend.gateCheck(request)))
+            let result = try await backend.gateCheck(request)
+            // `.enrollmentLost` is client-derived from the backend's
+            // `no_mandate` error envelope; a conforming backend never
+            // puts it in a 200 body. If one does anyway (the reason is
+            // plain `Codable`, so it decodes), normalize it onto the
+            // same refusal path as the envelope: it must not persist
+            // as a successful check, and downstream must see exactly
+            // one `.enrollmentLost` route.
+            if case .checkRequired(.enrollmentLost) = result {
+                return .refused(.enrollmentLost)
+            }
+            return .success(await sanitize(result))
         } catch {
             // Only a recognizable SESSION refusal blocks: 401/403, or
             // the backend's own session codes. A broader any-4xx rule
--- a/Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift
@@ -107,10 +107,12 @@ public actor GateCheckRepository {
         /// keeps serving the last known state within the window.
         case unreachable
         /// The backend answered and deterministically refused the
-        /// session (4xx: replayed/invalid signature, no mandate).
-        /// NOT unreachable: cached `.clear` must not keep serving for
-        /// the grace window on the say-so of a refusal.
-        case refused
+        /// session. NOT unreachable: cached `.clear` must not keep
+        /// serving for the grace window on the say-so of a refusal.
+        /// Carries the reason so `no_mandate` (recoverable only by
+        /// re-consenting) routes differently from a signature refusal
+        /// (recoverable by retrying / fixing the clock).
+        case refused(CheckRequiredReason)
     }
 
     private static let logger = Logger(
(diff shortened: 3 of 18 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the wire client itself looks faithful, but two of the PR's trust claims don't hold in the shipped code paths, and the new client has no tests.

- `ModerationTrust.swift:26` — flipping `enforceVerdictSignatures` to `true` is inert: `VerdictValidator` has **zero production call sites** (grep: only `VerdictValidatorTests`/`ConsentedTermsEnforcementTests`, and both helpers pass `enforceSignature: false` explicitly). `BanState.verdict` decoded from the backend goes straight to `BannedView` at `Sources/OnymIOS/RootView.swift:41` unvalidated — the "verdict signature enforced" property in the PR body isn't actually obtained. Either wire the validator into the gate/ban path or drop the claim to what the flag does.
- `Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift:237` — `performAttempt` funnels *every* thrown error into `.unreachable`. With the stub that was unreachable; with a live backend a 401 replayed-signature, `no_mandate`, or `signature_invalid` is now indistinguishable from being offline, so the grace arithmetic keeps serving the last cached `.clear` for the full window instead of degrading to `.gateCheckRequired`. Deterministic `AuthorityClientError.rejected` (4xx) should map to `.gateCheckRequired`, mirroring `isDeterministicRegistrationRejection` (`ModerationRepository.swift:1465`).
- No tests for `URLSessionEnforcementBackendClient.swift` while the sibling `ModerationAuthorityClient` has a full `URLProtocol` suite (`Tests/OnymIOSTests/ModerationAuthorityClientTests.swift`). The riskiest assumption — that the server emits `GateCheckResult` in Swift's synthesized shape (`{"caseOpen":{"_0":[…]}}`, and `{"clear":{}}` for the payload-less case) — is asserted only in a comment. A mismatch fails silently into `.unreachable` (see above), not loudly. Round-trip tests for all four cases + the `{error,message}` envelope + base64/RFC-3339 body shapes are cheap here.
- `Sources/OnymIOS/OnymIOSApp.swift:147` — the DEBUG non-UI-test branch now points at production `moderation.onym.app`, so every dev/simulator run enrolls and gate-checks against the live service (real DeviceCheck bits on hardware) with no base-URL override. The PR defers the dev harness; at minimum a launch-arg/env override for `baseURL` belongs in this change, since the constructor already takes one.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:396` — `isCountersigned` is now "anything that isn't the stub sentinel", so `{"signature":""}` or a garbage string from the backend records the mandate as countersigned and active. Nothing verifies the interface signature client-side; a non-empty/base64-decodable check is the minimum.
- `AuthorityManifestFetcher.swift:24` — worth stating in the `ModerationTrust` comment that with `enforceManifestSignatures` off, the operator key that *would* verify verdicts comes from an unauthenticated manifest, so the verdict switch can't be stronger than the manifest one. The documented `.sig` follow-up is the right fix; consider tracking it as an issue rather than only a code comment.

Nits: the date-decoding closure (`URLSessionEnforcementBackendClient.swift:110-130`) duplicates `ModerationAuthorityClient.swift:531+` and allocates two `ISO8601DateFormatter`s per date — worth a shared `ModerationJSON.decoder()`. `URLSession.shared` (line 39) shares cookie/cache storage app-wide; an ephemeral session with `httpShouldHandleCookies = false` fits this codebase's linkability posture better (same nit applies to the existing clients).

dev-onym (commented):
**REQUEST_CHANGES** — solid wire client and tests, but the new `sanitize()` silently guts the UI-test ban screen and `.refused` hard-blocks on any 4xx.

- `GateCheckRepository.swift:269` — `sanitize()` runs in **every** build, including the UI-test composition, contradicting the PR body's "no UI-test path routes fixtures through `VerdictValidator`". The stub ban fixture (`StubEnforcementBackendClient.swift:113-140`) declares `authority: "onym:component:stub-authority"` / `deviceBinding: "stub-enrollment"` while the UI-test mandate is `onym:component:uitest-authority` / `uitest-enrollment` (`Sources/OnymIOS/UITestFakes.swift:388,391`), so validation throws on the authority check *before* signatures even matter — and the sentinel `"stub:authority-unsigned"` fails too now that `enforceVerdictSignatures` is on. Result: `--moderation-scenario banned` renders `BannedView` without violation class, decided date, reasoning, or appeal deadline (`BannedView.swift:58-62`). No UI test asserts those rows, so nothing catches it. Fix the fixture to derive from the mandate, or pass `enforceSignature: false` on the stub path.
- `GateCheckRepository.swift:253-257` — treating *all* 4xx except 408/429 as `.refused` converts any deploy/route/encoding regression (404 on a renamed path, 400 from a body-shape drift) into an immediate hard block for every user, with grace discarded and a message telling them to re-consent (`GateCheckRequiredView.swift:61`). Suggest narrowing to 401/403 plus the known rawCodes (`no_mandate`, invalid/replayed signature) and letting other 4xx fall through to `.unreachable`.
- `URLSessionEnforcementBackendClient.swift:104` — `JSONEncoder` omits nil optionals, so a nil `deviceToken`/`mandateRef` ships as an **absent key**, not `null`, while `signedPayload` hashes them as empty. That's exactly the simulator/no-attestation path the PR calls out for dogfooding; `testGateCheck*` never asserts the request body for that shape. Worth pinning a nil-token body test against the apple service's parser.
- Test coverage: no tests for `derive(.refused) → .backendRefused` (a one-liner in `GateCheckDeriveTests`), for `sanitize()` stripping an unverifiable verdict while keeping the ban, or for `ModerationError.countersignatureInvalid` (`ModerationRepository.swift:404-411`). All three are new load-bearing branches.
- `URLSessionEnforcementBackendClient.swift:88-92` — the loopback-http exception is compiled into release, unlike `URLSessionModerationAuthorityClient`, which is https-only (`ModerationAuthorityClient.swift:466`). Only the DEBUG launch arg can reach it today; `#if DEBUG` around the exception would make that structural.
- `EnforcementBackendClient.swift:170-174` — `backendRefused` is documented as client-derived but is wire-decodable, so a backend can send it. Consider excluding it from the decoded reason set.
- Stale doc: `VerdictValidator.swift:33-35` still says "no real authority signs verdicts yet; see `ModerationTrust`" — that's now the opposite of the shipped default.
- The `enforceVerdictSignatures` ON / `enforceManifestSignatures` OFF asymmetry is honestly documented (`ModerationTrust.swift:22-30`); no objection, but the `.sig` publication is a real prerequisite, not a nicety — a tampered manifest at consent time supplies the operator key that verdicts then verify against.

dev-onym (commented):
**REQUEST_CHANGES** — the wire client and validation seam look right; the new `.refused` → hard-block path can brick users with no in-app recovery.

- `GateCheckRepository.swift:277,346` — `no_mandate` / 401 / 403 immediately yields `.gateCheckRequired(.backendRefused)` with grace discarded. The causes are not all user-transient: interface-backend state loss or a redeploy that drops enrollments, a mandate registered with the authority but never countersigned by the backend, or a device clock outside the server's freshness window (401) all pin *every* affected user in this state permanently. That is the same class of outage the comment above `isSessionRefusal` says it narrowed the rule to avoid — it just moved the blast radius from route/body drift to session codes.
- `ModerationGateFlow.swift:121`, `GateCheckRequiredView.swift:61` — the only affordance is "Try again", which replays the identical failing check; the copy tells the user to "re-consent to your moderation authority" but no route to consent or settings exists from this gate. Either map `no_mandate` to a re-consent route (`recompute` can fall to `.needsConsent`, or reuse `.reidentificationRequired`) or give the screen an escape hatch before shipping the live backend.
- `EnforcementBackendClient.swift:120` — the signed payload's timestamp is second-precision and, with a nil device token (attestation unsupported / `generateToken()` throws), two `checkNow()` calls in the same second produce byte-identical signatures. Against a single-use-signature server that is a self-inflicted 401 → `.refused` → block. `consentCompleted()` + the cadence loop, or a double-tapped retry, both hit it. A nonce in the payload, or serializing checks in the actor, closes it.
- `GateCheckRepository.swift:298` — the validator's `Outcome` is discarded, so a ban whose `executeAfter` is still in the future (`.storeUntilExecuteAfter`) renders the full ban UX with narrative. The marks are the enforcement, but the validator computes the early-execution answer and the caller drops it.
- `GateCheckRepository.swift:270` — `isSessionRefusal` is the function that decides hard-block vs. grace and has no test; `GateCheckDeriveTests.swift:17` only exercises `derive(attempt: .refused)`. Add a case that pushes `AuthorityClientError.rejected(401)` and a `400 no_mandate` through `performAttempt`, plus one confirming a 404/500 falls through to `.unreachable`.
- `ModerationJSON.swift:19` vs `AuthorityManifestFetcher.swift:28` and `MandateStore.swift:83` — the new "one decode configuration for every moderation wire response" isn't used by either manifest decode path, which still accept whole seconds only. Consistent today, so no bug, but `consentedManifest()` is now load-bearing for verdict display: if those two ever diverge, verdicts get silently stripped with only an `os_log` line.
- `ModerationTrust.swift:26` — the honesty note understates the guarantee. `AuthorityManifestFetcher.directoryPinnedOperatorKey` already requires the manifest's declared operator to equal the directory-pinned key byte-for-byte, so the verdict-signing key is directory-pinned regardless of `enforceManifestSignatures`. What the missing `.sig` leaves unverified is the rest of the manifest (ban terms, appeal windows) that `VerdictValidator` derives dates from — worth saying precisely, since that's what the follow-up actually closes.
- PR body says "No launch-arg base-URL override yet"; `OnymIOSApp.swift:590` adds `--enforcement-base-url`. Update the body.
- Not verified locally: no Swift toolchain in this environment, so the "full suite green" claim is taken as reported.

dev-onym (commented):
**REQUEST_CHANGES** — the wiring, transport shapes, and refusal/grace split are solid; one new fail-open path in the `enrollmentLost` route needs fixing.

- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationGateFlow.swift:126` — **fail-open**. `.enrollmentLost` falls back to `.operational(openCases: [])` when `authoritiesAvailable` is false, but `ModerationRepository.authorities` is in-memory only (`ModerationRepository.swift:167`, set solely in `refresh()` at :234) — it is empty on every cold launch until the GitHub directory fetch returns, and stays empty if that fetch fails (rate limit, missing release asset, partial connectivity). So: device has an active mandate, backend is reachable and answers `no_mandate`, directory unavailable → the app operates unmoderated indefinitely. Before this PR that same answer produced a blocking `.gateCheckRequired`. The `!hasMandate` softening at :110 ("an interface with no designated authority answers to its distribution channel") doesn't transfer here — `hasMandate == true` proves an authority was designated. Suggest `gate = authoritiesAvailable ? .needsConsent : .gateCheckRequired(.enrollmentLost)`; `GateCheckRequiredView.swift:64` already has the copy for it, and the "normally unreachable" note there would then be accurate.
- `Packages/OnymModeration/Sources/OnymModeration/EnforcementBackendClient.swift:176` — `.enrollmentLost` is a plain `Codable` `CheckRequiredReason`, so a backend answering `{"checkRequired":{"_0":"enrollmentLost"}}` decodes and takes the same route, contradicting the "never sent by the backend" invariant in the doc comment. Harmless while that route blocks; combined with the finding above it becomes a backend-triggerable path to operational. Either enforce the invariant in `status(for:)`/decode or drop the claim.
- `Info.plist` — no `NSAppTransportSecurity`/`NSAllowsLocalNetworking` key, so the DEBUG `--enforcement-base-url http://localhost:8080` affordance (`OnymIOSApp.swift:594`, allowance at `URLSessionEnforcementBackendClient.swift:80`) will likely be refused by ATS before the request leaves the app. Worth verifying on a simulator run, or adding the DEBUG-only exception.
- `Packages/OnymModeration/Sources/OnymModeration/GateCheckRepository.swift:249` — `max(clock(), last + 1)` correctly guarantees a distinct second (`floor(x+1) == floor(x)+1`), but each same-second call pushes the sent timestamp one second into the future; a burst (retry double-tap + `appForegrounded` + cadence loop) lands several seconds ahead. Fine against a symmetric freshness window, self-inflicted refusal against a server that rejects future timestamps. Also not persisted, so a relaunch inside the same second can still replay — low risk, worth a line in the comment.
- `Packages/OnymModeration/Sources/OnymModeration/ModerationTrust.swift:35` — the "verdict-signing key is directory-pinned regardless" argument checks out against `AuthorityManifestFetcher.swift:75`. Flipping verdicts on while manifests stay soft is defensible; just note `VerdictValidator` derives `appealDeadline`/`banExpires` bounds from that unverified manifest (`VerdictValidator.swift:144`, :179), so the residual gap is exactly as documented — good that onym-infra#4 is named.
- `Tests/OnymIOSTests/ModerationRepositoryTests.swift:9` — nit: the file-scope `kInterfaceCountersignature` was inserted between the class doc comment and the class, so the comment now documents the constant; make it `private` and move it inside the class.

dev-onym (commented):
**COMMENT** — solid, well-reasoned wiring that fails closed in the right places; the concerns are coverage and wire-contract fragility rather than defects I can prove.

- `Packages/OnymModeration/Sources/OnymModeration/ModerationTrust.swift:34` flips `enforceVerdictSignatures` to `true`, but nothing exercises the *positive* path end-to-end. `GateHardeningTests.swift:317` and `:353` only cover strip-on-failure and validation-disabled; every existing `VerdictValidatorTests` call passes `enforceSignature: false` (`VerdictValidatorTests.swift:157`). If the authority's `signingBytes()` serialization differs from Swift's (`ModerationMandate.swift:64` explicitly calls the canonical form PROVISIONAL and non-cross-implementation), *every* real ban verdict is stripped at `GateCheckRepository.swift:358` with only an `os_log` — users see a ban screen with no reason and no telemetry. Please add a test where a verdict signed with the consented manifest's operator key survives `sanitize`, and verify one live signed verdict before merge.
- `URLSessionEnforcementBackendClient.swift:17-19` pins the wire format to Swift's *synthesized* enum encoding (`{"caseOpen":{"_0":[…]}}`). `_0` is a compiler implementation detail, not a stable contract, and the producer is a different implementation deliberately mimicking it. `EnforcementBackendClientTests.swift` pins it well, but a Swift upgrade would break production and the test simultaneously. Recommend explicit `CodingKeys`/custom `Codable` on `GateCheckResult` so the shape is owned by this file.
- `EnforcementBackendClient.swift:154` — `CheckRequiredReason` is a closed `String` enum, so a server adding any new reason code fails the whole `GateCheckResult` decode → `malformedResponse` → `.unreachable` → grace → block. Fails closed, but silently and confusingly. An `unknown` fallback mapping to a blocking reason would be cheaper to debug.
- `OnymIOSApp.swift:157` — DEBUG non-UI-test builds now hit `moderation.onym.app`. Per the PR's own note, simulators land in `gate-check-required`, so every local dev run is blocked at the gate unless someone stands up an https tunnel. Consider keeping `--moderation-scenario` (or a `--stub-enforcement` flag) usable in DEBUG without `--ui-testing`.
- `GateCheckRepository.swift:297` — the first 401/403 is an immediate hard block with grace discarded. The doc at `EnforcementBackendClient.swift:173` says "retrying with a fresh session usually clears it", but the code never does that retry; a single automatic re-attempt with the bumped timestamp (`GateCheckRepository.swift:260`) before surfacing `.backendRefused` would avoid flashing a block screen on a transient collision, and a server-side signature-verification regression currently blocks the entire user base instantly.
- `GateCheckRepository.swift:427` — `.caseOpen` notices carry a `signature` (`CaseNotice.swift:20`) and reach `OpenCaseBanner` with no validation at all, while ban verdicts now get the full `VerdictValidator` treatment. Same channel, same trust; worth closing the asymmetry or documenting why case notices are exempt.
- Nit: `ModerationRepository.swift:407` throws `countersignatureInvalid` after `enrollDevice` already succeeded, so a garbage countersignature leaves an orphaned server-side enrollment on every retry. Harmless today, but worth a comment noting the backend must tolerate repeat enrollments for the same user key.

PR #230, opened 2026-08-09
Device recovery behind reidentificationRequired — moderator-decided

rinat-enikeev:
Client half of onymchat/onym-moderation#27–#29: the way back for a marked device whose enrolled identity did not survive a reinstall or a change of hands. **No self-serve unban** — the holder files a claim with a real contact and their account of holding the device; a human moderator decides; only the moderator-signed grant moves anything.

## Flow
1. `gateCheckRequired(.reidentificationRequired)` now offers **Ask a moderator to recover this device** (the case-ID appeal path stays for retained identities).
2. Claim form → `POST /v1/recovery-claims` (signed by the new identity over canonical bytes). Claim id persisted (`UserDefaultsRecoveryClaimStore`) so the screen resumes across relaunches.
3. Poll `GET /v1/recovery-claims/:id` (signed headers; answered only to the claim's key). Refusal shows the moderator's reasons and allows a fresh claim.
4. On grant: `GateCheckRepository.redeemRecoveryGrant` presents the grant's **exact signed bytes** to `POST /v1/recover` in an `onym-moderation-recover-v1` session bound to the grant reference (`RecoveryGrant.reference()` mirrors the server's canonical-bytes hash), folds the backend's reconciled gate answer into published state — success dismisses the blocking screen with no extra round trip. `markInForce` keeps the grant and shows the authority's routes.

## Notes
- `recover` / `fileRecoveryClaim` / `recoveryClaimStatus` are protocol **requirements** with refusing extension defaults — existential dispatch reaches the real clients; stubs and test fakes stay source-compatible.
- Grant bytes travel verbatim end to end; the client never re-encodes what the operator key signed.
- Tests: stacked PR follows. Unit suite: 1038 passed.

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/DeviceRecovery.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/DeviceRecovery.swift
@@ -1,0 +1,236 @@
+import CryptoKit
+import Foundation
+
+// Device recovery (REFERENCE-AUTHORITY-POLICY §6): the way back for a
+// marked device whose enrolled identity did not survive a reinstall or
+// a change of hands. There is deliberately no self-serve path — the
+// holder files a claim carrying a real contact and their own account
+// of how they hold the device, a human moderator at the authority
+// decides it, and only the grant that decision signs can move the
+// case's record at the enforcement backend.
+
+// MARK: - The grant
+
+/// A moderator-issued recovery grant, held as the **exact bytes** the
+/// authority signed. They travel verbatim: the operator signature is
+/// over the canonical form of these bytes, so re-encoding them would
+/// orphan it.
+public struct RecoveryGrant: Sendable, Equatable {
+    public let raw: Data
+    public let caseId: String
+    /// The identity key the grant was issued to. The enforcement
+    /// backend refuses a session by any other key, so a stolen grant
+    /// is inert.
+    public let grantee: String
+    public let authority: String
+    public let issuedAt: String
+
+    public init(raw: Data) throws {
+        struct Wire: Decodable {
+            let caseId, grantee, authority, issuedAt, signature: String
+        }
+        let wire: Wire
+        do {
+            wire = try JSONDecoder().decode(Wire.self, from: raw)
+        } catch {
+            throw ModerationError.grantInvalid("recovery grant: \(error)")
+        }
+        self.raw = raw
+        self.caseId = wire.caseId
+        self.grantee = wire.grantee
+        self.authority = wire.authority
+        self.issuedAt = wire.issuedAt
+    }
+
+    /// The grant's reference: SHA-256 over the canonical signing bytes
+    /// (every field except `signature`, sorted keys). The enforcement
+    /// backend derives the same value from the raw bytes to make the
+    /// grant single-use, and the session signature binds it — so the
+    /// two derivations must agree. Reconstructed through a typed
+    /// mirror, like `Verdict.signingBytes()`: JSONEncoder's
+    /// `.sortedKeys` is UTF-8 byte order, matching serde's map order.
+    /// A grant carrying fields this mirror doesn't know would derive a
+    /// different reference and fail the session — acceptable, since
+    /// grants are consumed by the interface whose vocabulary this is.
+    public func reference() throws -> String {
+        struct Unsigned: Encodable {
+            let grantVersion: Int
+            let caseId, grantee, authority, issuedAt: String
+        }
+        let bytes = try ModerationCanonicalEncoder.encode(
+            Unsigned(
+                grantVersion: 1,
+                caseId: caseId,
+                grantee: grantee,
+                authority: authority,
+                issuedAt: issuedAt
+            )
+        )
+        return SHA256.hash(data: bytes).map { String(format: "%02x", $0) }.joined()
+    }
+}
+
+// MARK: - Redemption at the enforcement backend
+
+/// `POST /v1/recover` as the client presents it: the grant's exact
+/// bytes plus the same session envelope as a gate check, signed by the
+/// grantee identity with the grant's reference bound into the payload.
+public struct RecoveryRequest: Codable, Sendable, Equatable {
+    public let deviceToken: Data?
+    public let userKey: String
+    /// The grant's exact signed bytes; travels as base64.
+    public let grant: Data
+    public let timestamp: Date
+    public let signature: Data
+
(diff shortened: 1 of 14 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the design and crypto plumbing are sound, but the flow has two dead ends and a re-entrancy hole that can undo a successful recovery.

- `DeviceRecoveryView.swift:180` — the "Check again" button in `markInForce` is dead: `pollableClaimId` (`DeviceRecoveryFlow.swift:127`) returns nil for `.markInForce`, so `checkClaim()` returns immediately. The holder must close and reopen the sheet to get anywhere. Either include `.markInForce` in the pollable set (the claim id is still in the store) or drop the button.
- `DeviceRecoveryFlow.swift:89` — `.checking` is itself pollable, so nothing guards re-entry. `.task { await flow.checkClaim() }` (`DeviceRecoveryView.swift:38`) firing while a tapped check is in flight polls twice and redeems the same **single-use** grant twice; the loser's `.failed` overwrites `.recovered` with `.awaitingReview` + an error the user can't act on. Guard on `.awaitingReview` only.
- `DeviceRecoveryFlow.swift:119` — `startOver()` is reachable only from `.refused` (`DeviceRecoveryView.swift:144`). A persisted claim id that the authority no longer knows (re-consent, pruned claim) leaves the holder pinned in `.awaitingReview` with a recurring error and no way to file again. Surface "file a new claim" whenever `errorMessage` is set in that phase.
- `DeviceRecovery.swift:62` — `reference()` hardcodes `grantVersion: 1` while `Wire` (line 20) never decodes it. If the authority ever issues v2 grants, every redemption fails with an opaque session rejection and the client can't tell why. Decode the version and reject unknown ones explicitly instead of silently assuming 1.
- `ModerationAuthorityClient.swift:533` — the `recovery-claim-status:{id}:{ts}` credential message, the claim's canonical field set, and the `{status, gate, …}` recover response shape are unverifiable in this repo. The existing analogue pins itself to the server by name ("the exact credential message verified by … `authority/src/api.rs`, `authorize_case_party`", line 441); please add the same citation for onym-moderation#27–#29 — a silent mismatch here is a 100% failure rate on a path only banned users hit.
- `DeviceRecovery.swift:141` — `ModerationError.grantInvalid` for an unknown status is unreachable through the real client: `URLSessionEnforcementBackendClient.swift:133` wraps every decode failure into `AuthorityClientError.malformedResponse`. Harmless, but the case's stated purpose isn't met.
- `GateCheckRequiredView.swift:95,103` — two `.sheet` modifiers on the same view, both live under `.reidentificationRequired`. This is a long-standing SwiftUI pitfall; verify both actually present on device, or fold them into one `.sheet(item:)` over an enum.
- `GateCheckRepository.swift:342` — `generation &+= 1` then `await sanitize(gate)`: a `checkNow` starting inside that await takes a newer generation and can still overwrite the recovered status. Re-check the generation after the await, as `checkNow` does. Also nothing verifies `grant.grantee == userKey` locally — a mismatched grant surfaces only as an opaque backend rejection.
- `ModerationRepository.swift:502` — `recoveryAuthorityClient()` duplicates the guard/lookup in `recoverableCaseIDs()` (476–483); route the older one through it.
- `DeviceRecovery.swift:203` — the claim key isn't scoped to the identity, so a claim filed under one grantee key is resumed and polled under the next one.
- No tests land with this PR (stacked). At minimum `reference()` against a server-produced fixture and the flow's phase transitions should not wait — they're the two places a wire mismatch would show up cheaply.

dev-onym (commented):
**COMMENT** — sound design and the wire/canonicalization work is careful; three issues worth fixing before merge.

- `DeviceRecoveryFlow.swift:155` — `.recovered` discards the associated `GateStatus`. `RecoveryRedemption.recovered` can carry `.banned`, `.gateCheckRequired`, or `.notMandated` (`GateCheckRepository.swift:76-95`), yet the UI unconditionally renders "Device recovered — The record was cleared" (`DeviceRecoveryView.swift:62-67`) and clears the claim store. A reconciliation that lands on a still-blocking status shows a success screen behind a blocking gate, with the claim gone. Switch on the status: only `.operational` is "recovered".
- `GateCheckRepository.swift:345` — the recovery path skips the `.checkRequired(.enrollmentLost)` normalization that `performAttempt` applies at line 286. A `recovered` body carrying `enrollmentLost` gets persisted through `derive(attempt: .success(...))` as a *successful* check — exactly the state that path exists to prevent. Route it through the same `.refused(.enrollmentLost)` normalization (or share a helper).
- `GateCheckRepository.swift:352-355` — on a stale generation it returns `.recovered(cached)`, i.e. reports success for an answer it deliberately discarded; the flow then clears the claim (`DeviceRecoveryFlow.swift:158`). A distinct "superseded" outcome that keeps the claim would be safer.
- `DeviceRecoveryFlow.swift:174` — `describe` handles only `AuthorityClientError.rejected` and `.noMandate`. A `ModerationError.grantInvalid` (e.g. the deliberate v2 refusal, `DeviceRecovery.swift:50-55`, thrown out of `recoveryClaimStatus`) surfaces as "The authority could not be reached. Check the connection and try again." — so the by-name version refusal never reaches the holder, and every re-poll returns the same granted+v2 claim: a permanent loop with a wrong explanation. Same for `.notImplemented` against a stub. Add a `grantInvalid` branch.
- `ModerationAuthorityClient.swift:495-510` — `fileRecoveryClaim` is the first signing form over free user text, and `ModerationCanonicalEncoder` does not set `.withoutEscapingSlashes`, so `/` is signed as `\/`. If the authority recomputes signing bytes by re-serializing the parsed body (serde_json does not escape `/`), any contact/statement containing a slash (a URL, a Telegram handle) fails verification. `appeal` has the same exposure, so this may already be settled — worth confirming against `authority/src` rather than assuming.
- `GateCheckRepository.swift:314-316` — the grantee-vs-`userKeyID()` mismatch is checked after `lastSessionTimestamp` is bumped; harmless, but moving the guard above the timestamp/signing block keeps the failure free of side effects.
- `GateCheckRequiredView.swift:88-92` — the button has no in-flight guard; a double tap runs `makeDeviceRecoveryFlow()` twice and the second assignment replaces `activeSheet` with a different flow identity mid-presentation.
- PR body says "Tests: stacked PR follows," but `Tests/OnymIOSTests/DeviceRecoveryTests.swift` is in this diff — fine, just reconcile the description. Note the grant fixture (`:16-18`) pins client canonicalization to itself for the absent-`grantVersion` case; that path is only truly verified if the server also injects the serde default before hashing.

PR #231, opened 2026-08-09
Tests: device recovery client — grant bytes, wire, redemption, claim flow

rinat-enikeev:
Stacked on #230 (base: `feat/device-recovery`). Rebased on the base's review-fix commit (6185585) and merged with the tests that landed there — 27 tests; full unit suite 1065 pass.

- **Grant**: parses and keeps the exact signed bytes; `reference()` asserted two ways — against a hand-written canonical JSON hash pinning the `grantVersion` < `grantee` byte-order case where a case-insensitive sort would diverge from serde, and byte-for-byte against a fixture produced by the reference Authority's `issue_grant` (fixed key + clock). Unknown versions refused by name; an absent version derives the authority's serde-default bytes.
- **Session payload**: `onym-moderation-recover-v1` is domain-separated from gate/enroll and binds one grant ref.
- **Wire**: `/v1/recover` posts base64 of the exact grant bytes; `RecoveryResult` decodes `recovered`+gate and `markInForce` with/without routes, refuses unknown statuses as a `DecodingError`; claim intake signature verified to cover the transmitted body minus `signature` in canonical order; status poll sends the claim-bound signed headers.
- **Redemption**: recovered → adopted as the newest gate check (published + persisted, no second round trip); `markInForce` → moves nothing, persists nothing; rejection → message surfaced; a grant issued to a different grantee → refused locally with an actionable message before anything is signed.
- **Claim flow**: resume-from-persisted only for the grantee that filed it; contact/statement required; refusal clears the claim and shows reasons; grant redeems immediately and retires the claim; `markInForce` keeps it and "check again" re-polls; a check in flight cannot be re-entered to double-redeem the single-use grant; a recurring poll error stays actionable via start-over.

### DIFF
--- a/Tests/OnymIOSTests/DeviceRecoveryTests.swift
+++ b/Tests/OnymIOSTests/DeviceRecoveryTests.swift
@@ -1,13 +1,60 @@
+import CryptoKit
 import XCTest
-@testable import OnymModeration
+@testable import OnymIOS
+import OnymModeration
 import OnymModerationUI
 
-/// Device recovery: the grant's reference derivation pinned to a
-/// server-produced fixture, and the claim flow's phase transitions —
-/// the two places a wire mismatch or a UI dead end would show first.
+/// Device recovery: the grant object, its wire crossings, the gate
+/// repository's redemption, and the claim flow's state machine. The
+/// invariant under test everywhere: nothing in this path unbans a
+/// device on the client's own say-so — the moderator's signed grant is
+/// the only object that moves anything, and the client treats it as
+/// opaque signed bytes.
 final class DeviceRecoveryTests: XCTestCase {
+    private var session: URLSession!
+    private let baseURL = URL(string: "https://moderation.test")!
 
-    // MARK: - The grant against the reference Authority's bytes
+    override func setUp() {
+        super.setUp()
+        session = StubURLProtocol.makeSession()
+    }
+
+    override func tearDown() {
+        StubURLProtocol.reset()
+        session = nil
+        super.tearDown()
+    }
+
+    // MARK: - The grant object
+
+    /// Grantee matches `CapturingSigner.userKeyID()` — the repository
+    /// refuses to present a grant issued to any other key.
+    private static let grantJSON = #"{"grantVersion":1,"caseId":"case-1","grantee":"onym:key:test","authority":"onym:component:authority","issuedAt":"2026-08-09T00:00:00Z","signature":"c2ln"}"#
+
+    func testGrantParsesAndKeepsTheExactBytes() throws {
+        let raw = Data(Self.grantJSON.utf8)
+        let grant = try RecoveryGrant(raw: raw)
+        XCTAssertEqual(grant.version, 1)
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
**COMMENT** — solid, well-targeted suite; the findings below are weak assertions and coverage holes, not defects.

- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:266` — "no second round trip" isn't actually pinned: `currentStatus()` just returns `cached` (`GateCheckRepository.swift:499`), and `RecoveryAnsweringBackend.gateCheck` returns `.clear` too (line 468), so a spurious re-check would yield identical assertions. Make `gateCheck` `XCTFail(...)` (or return `.blocked`) to make the claim testable.
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:301` — `store` starts empty, so `XCTAssertNil(store.load())` only rules out a *new* write. "moves nothing" needs a pre-seeded `PersistedGateState` asserted unchanged, plus `currentStatus()` unchanged after the `markInForce`.
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:32,46` — `grantVersion` is the one field `RecoveryGrant.init` never decodes while `reference()` hardcodes `1` (`DeviceRecovery.swift:37,57`). A grant with `grantVersion: 2` parses fine and silently derives a wrong reference — rejected only later by the backend. The reference test's whole premise is this field; worth a case asserting the intended behavior (local refusal, or explicitly-documented divergence).
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:61` — comment and PR body say "domain-separated from gate/enroll" but only gate is compared. `EnrollmentRequest.signedPayload` (`EnforcementBackendClient.swift:52`) is one more `XCTAssertNotEqual`.
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:344-395` — two live branches of `checkClaim` are untested: `granted` with `grant == nil` (stays `awaitingReview` + message) and `redeem → .failed` (returns to `awaitingReview`, keeps the claim, surfaces the message). The latter is the only path where a failed redemption must *not* drop the claim.
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:110` — unknown `status` is covered, but not `{"status":"recovered"}` with no `gate` or `markInForce` with no `authorityContact`; both are `decode` (not `decodeIfPresent`) and are the likelier server-drift shapes.
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:229` — the suite domain is cleared in `setUp` but never in teardown, so the test leaves a plist behind. Add `removePersistentDomain` after the assertions (or `defer`).
- `Tests/OnymIOSTests/DeviceRecoveryTests.swift:186` — expected JSON is rebuilt with raw interpolation and `as! String`; fine for these fixtures, but it will crash rather than fail if a non-string field is ever added to the claim body. `try XCTUnwrap(... as? String)` is the cheap fix.

Nothing here blocks merge; the first two are the ones I'd fix, since both tests currently pass for reasons weaker than their names claim.

PR #233, opened 2026-08-09
Report screen UX: violation class as a visible, deliberate choice

rinat-enikeev:
## What was wrong

- The class picker rendered **raw manifest slugs** (`csam`, `credible-violence`, `unsolicited-pornography`) verbatim in a `.menu` picker.
- `start()` **preselected the first class** — the live manifest orders gravest-first, so opening the sheet and tapping Submit filed a **CSAM report with no deliberate choice ever made**.
- The menu hid the options; three high-stakes choices should be comparable at a glance.
- Jargon copy: "VIOLATION CLASS", "Consented definition (content address): …".

## What it is now

- **No preselection** — Submit stays disabled until the reporter explicitly chooses (the disabled state existed; the default made it dead).
- **Radio-style list card** showing every class at once, checkmark on the selected row, matching the message card above it.
- **Names are typography only** — hyphens→spaces, sentence case, known acronyms uppercased ("CSAM", "Credible violence"). Deliberately not a renaming: the class id is consented vocabulary, so the words stay the manifest's words.
- **The selected row expands with what choosing it means**: a "What counts as this?" link to the consented definition (or the pinned content address for a bare hash), and a statutory-reporting notice for classes declaring `lawfulReporting` (e.g. CSAM) — the reporter learns the report may go beyond the authority *before* filing.
- Accessibility ids preserved (`moderation.report.class` group + per-row `moderation.report.class.<id>`); rows carry `.isSelected` and disable while submitting.

Full app build passes. Known limitation: the acronym set (`csam`, `ncii`) is a hardcoded allowlist — a new acronym class an authority coins would render capitalized-first until added.

### DIFF
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift
@@ -1,0 +1,230 @@
+import SwiftUI
+
+/// One block of a parsed markdown document. Foundation's markdown
+/// parser (`AttributedString(markdown:)` with `.full` syntax) styles
+/// inlines but only *annotates* block structure via
+/// `PresentationIntent` — `Text` renders it all as one flow. This
+/// splits the parse into blocks a view can lay out, using nothing but
+/// the system parser.
+public struct MarkdownBlock: Identifiable, Equatable {
+    public enum Kind: Equatable {
+        case paragraph
+        case heading(level: Int)
+        /// `ordinal` is set for ordered lists, nil for bullets.
+        case listItem(ordinal: Int?, depth: Int)
+        case codeBlock
+        case quote
+        case divider
+    }
+
+    public let id: Int
+    public let kind: Kind
+    public let text: AttributedString
+
+    public static func blocks(from markdown: String) -> [MarkdownBlock] {
+        let options = AttributedString.MarkdownParsingOptions(
+            allowsExtendedAttributes: false,
+            interpretedSyntax: .full,
+            failurePolicy: .returnPartiallyParsedIfPossible
+        )
+        guard let parsed = try? AttributedString(markdown: markdown, options: options) else {
+            // Unparseable input is still readable input.
+            return [MarkdownBlock(id: 0, kind: .paragraph, text: AttributedString(markdown))]
+        }
+        var blocks: [MarkdownBlock] = []
+        for (intent, range) in parsed.runs[\.presentationIntent] {
+            let slice = AttributedString(parsed[range])
+            var kind = Kind.paragraph
+            var ordinal: Int?
+            var ordered = false
+            var listDepth = 0
+            var isDivider = false
+            for component in intent?.components ?? [] {
+                switch component.kind {
+                case .header(let level):
+                    kind = .heading(level: level)
+                case .codeBlock:
+                    kind = .codeBlock
+                case .blockQuote:
+                    kind = .quote
+                case .listItem(let value):
+                    ordinal = value
+                case .orderedList:
+                    ordered = true
+                    listDepth += 1
+                case .unorderedList:
+                    listDepth += 1
+                case .thematicBreak:
+                    isDivider = true
+                default:
+                    break
+                }
+            }
+            if isDivider {
+                blocks.append(MarkdownBlock(id: blocks.count, kind: .divider, text: AttributedString()))
+                continue
+            }
+            if ordinal != nil || listDepth > 0 {
+                kind = .listItem(ordinal: ordered ? ordinal : nil, depth: max(listDepth, 1))
+            }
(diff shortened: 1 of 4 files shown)


ux-onym (commented):
**REQUEST_CHANGES**

- `ModerationReportView.swift:137-179` — `definitionLine(for:)` renders a `Link` (for the `https`/`http` case) nested inside the outer row's `Button` label (the `Button` starting at `ModerationReportView.swift:136`). SwiftUI does not support nested interactive controls: the outer `Button`'s `.contentShape(Rectangle())` and tap-gesture capture the touch before it reaches the inner `Link`, so tapping "What counts as this?" on an already-selected row re-triggers `flow.selectClass(item.classId)` (a no-op) instead of opening the definition URL. This defeats the PR's stated goal of letting the reporter read the consented definition before filing. Move the `Link` out of the row's `Button` label (e.g. render it as a separate row/element below the selectable row, or make the whole row a plain tappable area with the link as an adjacent, non-nested control).
- `ModerationReportView.swift:80-83` — `item.classId != flow.state.classes.last?.classId` recomputes `.last` for every row (O(n²)); harmless at manifest sizes but simpler to zip with index or use `.enumerated()` if touched again.
- Everything else checks out: no-preselection logic (`ModerationReportFlow.swift:39-45`) correctly leaves `selectedClassId` nil so Submit (`ModerationReportView.swift:118-122`) stays disabled until an explicit `selectClass` call; accessibility ids (`moderation.report.class`, `.class.<id>`, `.definition`, `.lawful`) are preserved/extended as described; `displayName(for:)` acronym/capitalization logic is correct for the disclosed hardcoded-allowlist limitation. No existing tests reference the removed `Picker` or old copy, so nothing else breaks.

ux-onym (commented):
APPROVE

- **No functional bugs found.** Removing the `state.selectedClassId = classes.first?.classId` line (ModerationReportFlow.swift:41-46) correctly leaves selection nil; Submit's `disabled` check at ModerationReportView.swift:122-126 already gates on `selectedClassId == nil`, so the disabled state now actually engages as the PR claims.
- ModerationReportView.swift:139-142 correctly moves the definition `Link`/statutory notice outside the `Button` label — nested interactive controls inside a SwiftUI `Button` label are inert, so this avoids a real tap-swallowing bug.
- `displayName(for:)` (ModerationReportView.swift:213-222) handles the sampled cases correctly (`csam`→"CSAM", `credible-violence`→"Credible violence"); acronym-allowlist limitation is called out honestly in the PR body.
- Accessibility identifiers (`moderation.report.definition`, `moderation.report.lawful`) are reused across rows but only one row is ever expanded at a time, so no duplicate-id collision at runtime.
- No other call site references `selectedClassId` or `displayName`, so this is a self-contained, low-blast-radius change.

**Minor, non-blocking:**
- `OnymModerationUI` has no test target (find turned up nothing under `*OnymModerationUI/Tests*`), so `displayName`'s acronym/capitalization logic and the no-preselection behavior ship untested — pre-existing gap, not introduced by this PR, but worth a follow-up given the PR touches exactly the logic (acronym allowlist) most likely to silently regress.
- `displayName` is a `static func` on the view struct rather than a free function/extension on `ViolationClass` — stylistic only, not a defect.

ux-onym (commented):
APPROVE

- `ModerationReportFlow.swift:41-46` — preselection removal is correct and matches `submit()`'s existing `guard let classId = state.selectedClassId` gate; `ModerationReportView.swift:133-137` still disables Submit on `selectedClassId == nil`, so the fix is load-bearing, not cosmetic.
- `ModerationReportView.swift:148-196` — button/expanded-content split correctly avoids the nested-control-inertness pitfall it calls out in the comment (the `Link`/`Button` for the definition lives outside the row's `Button` label).
- `ModerationReportView.swift:204-226` — scheme allowlist (`https`/`http`) is checked before constructing `DefinitionDocument`, so `MarkdownDocumentView` only ever fetches from vetted schemes.
- `ModerationReportView.swift:233-241` / `displayName` — typography-only transform (hyphens→spaces, first-word capitalization, hardcoded acronym allowlist) matches the PR's stated limitation; behaves correctly for both leading and non-leading acronym positions.
- `MarkdownDocumentView.swift` — block splitter's pattern matches (`case .codeBlock:`, `case .listItem(let value):`) correctly ignore/bind `PresentationIntent` associated values per the real API; HTML-vs-document sniffing (content-type + regex fallback) and the browser hand-off are reasonable; no fetched-content size cap, but low risk for authority-manifest definition docs.
- Accessibility ids (`moderation.report.class`, `moderation.report.class.<id>`, `.isSelected` trait) are preserved per the PR description; no existing tests reference the removed `.menu` picker.
- Could not compile/run the Swift package in this Linux sandbox (no `swift` toolchain, SwiftUI-only code) — review is static; recommend confirming `swift test`/Xcode build green before merge as stated in the PR body.

PR #234, opened 2026-08-09
Markdown viewer: relative links resolve, tables render, links stay in-app

rinat-enikeev:
Follow-up to #233, found by reading the live `authority.onym.app/policy/csam` document in the viewer:

- **Relative links were dead** — the policy's `[What these windows mean](./evidence-rules#what-the-windows-mean)` parsed as a base-less URL: rendered as a link, tapping did nothing. `MarkdownBlock.blocks(from:baseURL:)` now resolves relative links against the document's own address; absolute links pass through untouched.
- **Tables rendered as disjointed paragraphs** — the policy's "Terms" table (response window / decision deadline / ban term / appeal) arrives as `PresentationIntent` table cells; the splitter now reassembles them into rows rendered as a `Grid`, dropping the empty header row of the `| | |` key-value idiom.
- **Followed links stay in-app** — the policy docs cross-reference each other, and Safari would show raw markdown for those hops anyway. Web links now push the destination document onto the viewer's navigation stack (`mailto:`/`tel:` stay with the system; the HTML hand-off's "Open in browser" deliberately bypasses the in-app handler).

Still zero rendering dependencies — Foundation's parser only.

Tests: relative-link resolution + absolute passthrough, table extraction with empty-header dropping. Full unit suite: 1071 pass.

### DIFF
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift
@@ -162,6 +162,10 @@ public struct MarkdownBlock: Identifiable, Equatable {
             let rows = byRow.keys.sorted().map { rowIndex -> [AttributedString] in
                 byRow[rowIndex]!.sorted { $0.column < $1.column }.map(\.text)
             }
+            // Tradeoff, accepted: this also drops a legitimate row
+            // whose cells are ALL blank, not just the `| | |` header.
+            // A fully blank data row carries nothing a policy reader
+            // loses; rows with any content survive intact.
             return rows.filter { row in
                 !row.allSatisfy { $0.characters.allSatisfy(\.isWhitespace) }
             }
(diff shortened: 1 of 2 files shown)


ux-onym (commented):
**REQUEST_CHANGES**

- `MarkdownDocumentView.swift:205-213` — the in-app `openURL` override follows *any* `https`/`http` link found in a document with no same-origin/same-authority check, then `load()` (`MarkdownDocumentView.swift:358`) fetches it via bare `URLSession.shared.data(from:)`. A link to an unrelated domain (injected into an otherwise-trusted authority document, or reached transitively via a hop-of-a-hop) gets silently fetched and rendered inside the trusted viewer chrome, with only a slug-derived title (`impliedTitle`, line 219) and no host/address shown anywhere — unlike bouncing to Safari, the moderator has no way to see they've left the original authority's domain. The PR's stated intent is same-document cross-references (`./evidence-rules`); the implementation doesn't restrict to that. Worth gating in-app follow to same-host-as-`baseURL` (or same-authority) and falling back to `.systemAction`/browser hand-off otherwise.
- `MarkdownDocumentView.swift:191-213` — pushed destinations (`navigationDestination(for: URL.self)`) don't inherit the root's `.toolbar` Done button (`line 199-203`); once a moderator follows a link, closing the sheet requires stepping back through each pushed document rather than one tap. Minor if this is always sheet-presented (swipe-to-dismiss covers it) — worth confirming presentation context.
- `MarkdownDocumentView.swift:165-167` — the empty-cells filter that drops the `| | |` header row also silently drops any legitimate row whose cells are all blank; low risk given the current table content but worth a comment noting the tradeoff.

Everything else checks out: relative-link resolution (`MarkdownDocumentView.swift:43-53`) correctly captures ranges before mutating attributes (safe against invalidation), the table accumulator's row/column reconstruction and flush-on-table-switch logic (`lines 125-169`) is sound, and `UIApplication.shared.open` deliberately bypassing the custom `openURL` for the HTML hand-off (line 283) is the right call. Tests cover the two documented behaviors adequately.

ux-onym (commented):
**APPROVE**

- `MarkdownDocumentView.swift:229` — `followsInApp` accepts `http` as well as `https` for same-host destinations, so a plain-`http` link on the authority's own host renders inside the trusted, address-less viewer chrome unencrypted. Minor; consider restricting in-app follow to `https` only.
- `MarkdownDocumentView.swift:45` — protocol-relative links (`//evil.example/x`) have `scheme == nil` too, so they go through the "relative" resolution path. Verified this is benign: the resolved absolute URL keeps `evil.example` as host, so `followsInApp` (line 228) still routes it to the system browser at tap time — but worth a one-line comment noting this is intentionally not a bypass, since a future reader might assume "relative" means "same-document."
- `MarkdownDocumentView.swift:169` — the empty-row filter drops *any* all-blank row, not just the `| | |` empty header; already called out honestly in the adjacent comment as an accepted tradeoff, no action needed.
- Table rows render with uniform (non-bold) styling even when a table has a genuinely non-empty header row (only the `| | |` idiom is special-cased) — cosmetic gap, out of scope for the policy doc this targets.
- Logic, tests (relative/absolute link resolution, table reassembly, same-host gate including subdomain rejection and case-insensitivity), and the nav-stack restructuring for in-app link following all look correct. No blocking issues found.

PR #236, opened 2026-08-09
Support case-free unban grants

rinat-enikeev:
## Summary
- decode the new `onym-unban-grant-v1` format
- preserve exact signed bytes and derive its claim-bound reference
- keep legacy case-bound recovery grants compatible

Depends on [onym-moderation#37](https://github.com/onymchat/onym-moderation/pull/37).

## Validation
- Swift package test invocation is blocked here by pre-existing macOS availability errors in `OnymFoundation`

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/DeviceRecovery.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/DeviceRecovery.swift
@@ -83,46 +83,71 @@ public struct RecoveryGrant: Sendable, Equatable {
     /// (every field except `signature`, sorted keys). The enforcement
     /// backend derives the same value from the raw bytes to make the
     /// grant single-use, and the session signature binds it — so the
-    /// two derivations must agree. Reconstructed through a typed
-    /// mirror, like `Verdict.signingBytes()`: JSONEncoder's
-    /// `.sortedKeys` is UTF-8 byte order, matching serde's map order.
-    /// A grant carrying fields this mirror doesn't know would derive a
-    /// different reference and fail the session — acceptable, since
-    /// grants are consumed by the interface whose vocabulary this is.
+    /// two derivations must agree. Canonicalize the raw JSON itself
+    /// rather than reconstructing a typed subset: fields such as
+    /// `grantType` are part of the signature and must not be silently
+    /// omitted when the wire schema grows.
     public func reference() throws -> String {
-        struct CaseUnsigned: Encodable {
-            let grantVersion: Int
-            let caseId, grantee, authority, issuedAt: String
+        guard var object = try JSONSerialization.jsonObject(with: raw) as? [String: Any] else {
+            throw ModerationError.grantInvalid("recovery grant is not a JSON object")
         }
-        struct UnbanUnsigned: Encodable {
-            let grantVersion: Int
-            let claimId, grantee, authority, issuedAt: String
-        }
-        let bytes: Data
-        if grantType == "onym-recovery-grant-v1", !caseId.isEmpty {
-            bytes = try ModerationCanonicalEncoder.encode(
-                CaseUnsigned(
-                    grantVersion: version,
-                    caseId: caseId,
-                    grantee: grantee,
-                    authority: authority,
-                    issuedAt: issuedAt
-                )
-            )
-        } else if grantType == "onym-unban-grant-v1", let claimId {
-            bytes = try ModerationCanonicalEncoder.encode(
-                UnbanUnsigned(
-                    grantVersion: version,
-                    claimId: claimId,
-                    grantee: grantee,
-                    authority: authority,
-                    issuedAt: issuedAt
-                )
-            )
-        } else {
-            throw ModerationError.grantInvalid("grant has neither caseId nor claimId")
-        }
+        object.removeValue(forKey: "signature")
+        let bytes = try ModerationCanonicalEncoder.encode(RecoveryCanonicalJSON(object))
         return SHA256.hash(data: bytes).map { String(format: "%02x", $0) }.joined()
+    }
+}
+
+/// An Encodable bridge for the raw JSON object in a grant. Using a
+/// dynamically keyed value keeps every signed field, while
+/// `ModerationCanonicalEncoder` still supplies the same sorted-key and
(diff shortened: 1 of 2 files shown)


rinat-enikeev (commented):
## Overview

This PR teaches the iOS `RecoveryGrant` type to decode the new `onym-unban-grant-v1` shape introduced in `onym-moderation#37`: a `grantType`/`claimId`-bearing variant alongside the existing case-bound `caseId` variant, with `reference()` branching to build the right canonical unsigned payload for each. Legacy case-bound grants keep decoding as before. Small, well-scoped diff, and the accompanying tests (`testCaseFreeUnbanGrantParsesWithoutACaseId`) correctly check that decode preserves exact bytes and derives the right `claimId`.

## Blocking: `reference()`'s canonical bytes omit `grantType`, so signature verification will fail against the real interface

`reference()` derives its canonical bytes by hand-listing fields into a local `Encodable` struct (`CaseUnsigned` / new `UnbanUnsigned`) rather than deriving them structurally from `raw`. Neither struct includes `grantType`.

But on the Rust side (`apple/src/canonical.rs`), the signing/reference bytes are computed the other way: parse the **actual raw JSON**, drop only `"signature"`, re-serialize with sorted keys — everything else present in the document, including `grantType`, stays. And `grant_type` is an always-serialized (non-optional) field on both `authority/src/recovery.rs::RecoveryGrant` (case-bound, already on `main`) and the new `UnbanGrant` (`onym-moderation#37`). So every real grant byte-stream the device receives already contains `"grantType"`.

This isn't cosmetic: the enforcement interface's `/v1/recover` handler (`apple/src/api.rs`) does **not** trust whatever `grantRef` the client asserts. It recomputes `grant_ref` itself from the raw bytes it received (`canonical::grant_signing_bytes(&grant_raw)`), rebuilds the expected session payload from that, and verifies the device's signature against it. If the client's locally-computed `reference()` differs from the server's recomputation, the device's signature — which is over a payload built with the *client's* value — will never match what the server reconstructs, and `verify_user_signature` refuses every redemption.

I checked this isn't just an unbanGrant issue — it's the same shape as the pre-existing `CaseUnsigned` (renamed from `Unsigned`, unchanged in this PR), which has always omitted `grantType` too. Concretely (field-set difference alone, independent of Python/Swift JSON serialization quirks):

```
RUST canonical:  {"authority":...,"claimId":"claim-1","grantType":"onym-unban-grant-v1","grantVersion":1,"grantee":...,"issuedAt":...}
SWIFT canonical: {"authority":...,"claimId":"claim-1","grantVersion":1,"grantee":...,"issuedAt":...}   # no grantType
```
Different byte strings → different SHA-256 → session signature mismatch → refused redemption.

Worth noting: the existing cross-implementation fixture test, `testGrantReferenceMatchesServerProducedFixture` (comment: "Produced by the reference Authority's `issue_grant`... onym-moderation#28"), also has **no `grantType` in its fixture JSON** — meaning that fixture predates `grant_type` being added to the wire schema and no longer represents what the current authority actually emits. It's passing, but it isn't validating current interop; it's giving false confidence. All the other `reference()` tests in this file (and the new one added here) only check Swift against Swift's own hand-written expected literal, never against real Rust output.

**Fix suggestion:** make `reference()` derive canonical bytes structurally — parse `raw` as JSON, drop `"signature"`, re-serialize with `.sortedKeys` — the same way `canonical::grant_signing_bytes` does on both Rust sides, instead of hand-listing a field subset in a local struct. That's the only way this stays correct as the wire schema gains fields in the future (it already silently broke once, for `grantType`, and this PR's `UnbanUnsigned` repeats the pattern for `claimId`/`grantType` again). Also worth regenerating `serverGrantJSON`/`serverGrantRef` from the *current* authority build (and adding an equivalent fixture for `onym-unban-grant-v1` from `issue_unban_grant`) so the fixture test actually pins current byte-for-byte interop rather than a stale schema.

## Minor

- `caseId` becomes `""` rather than `nil` for unban grants — a reasonable, minimal-diff choice given `Equatable`/existing call sites, but it does mean any future code checking "is this case-bound" must remember to test `!caseId.isEmpty` rather than `caseId != nil` (currently only `reference()` does this, correctly).
- Removed trailing blank line at EOF — harmless.

## Summary

Directionally the right shape for decode/legacy-compat, and the new decode-side test is solid. But `reference()`'s canonical-bytes construction has a field-omission bug (`grantType`) that appears to already affect the existing case-bound path and is now replicated in the new unban path — and given how `/v1/recover` verifies the session signature, this looks like it would make **every** case-free unban grant redemption fail against the real interface. Given this PR's entire purpose is to make that redemption work, I'd want this fixed (and the stale fixture regenerated) before merging.

rinat-enikeev (commented):
## Re-review: blocking issue resolved ✅

This directly fixes the `grantType`-omission bug flagged in the previous pass. `reference()` now derives canonical bytes structurally — `JSONSerialization.jsonObject(with: raw)`, drop `"signature"`, re-encode via the new `RecoveryCanonicalJSON` bridge through `ModerationCanonicalEncoder` — instead of hand-listing a field subset in a typed struct. That mirrors `canonical_bytes(raw, &["signature"])` on the Rust side exactly, so it no longer silently drops fields the wire schema grows (which is exactly how `grantType` got lost before).

I independently recomputed both regenerated fixtures from scratch (parse JSON → drop `signature` → sort keys → SHA-256) and both matched:
- `serverGrantRef` (case-bound) ✅
- `serverUnbanGrantRef` (new unban path) ✅

These are now genuine cross-implementation fixtures rather than the stale pre-`grantType` one from before — good catch adding `testUnbanGrantReferenceMatchesServerProducedFixture` as the unban-side equivalent.

Also verified: `RecoveryCanonicalJSON`'s NSNumber handling correctly distinguishes JSON booleans from numbers via `CFGetTypeID`/`CFBooleanGetTypeID` (the classic `JSONSerialization` gotcha), and uses `int64Value` rather than a value derived from `doubleValue` when encoding whole numbers, so no precision is lost. The absent-`grantVersion` test was updated correctly too — canonicalization now follows the literal raw bytes (no synthesized `grantVersion` key when it's genuinely absent), which is the subtle case a less careful fix could have gotten wrong.

No remaining concerns. **Approving.**

PR #238, opened 2026-08-11
Sign attachment bytes at send time (proof v2)

rinat-enikeev:
A recipient can report a text message but not a photo. The reason is not a missing screen — **nothing in the system is a sender commitment to attachment bytes**, so a reported photo could never be authenticated against the accused.

## What is missing today

- `ChatModerationProof` v1 signs `{body, group_binding, message_id, proof_version, sent_at_millis}`. No attachment digest, type, or length.
- Only the plain-text `send(...)` path produces a proof at all. `sendImage`, `sendVideo`, `sendAlbum` and `sendVoice` construct `ChatMessagePayload` without one, so it defaults to `nil`.
- `SealedEnvelope.ephemeralKeySignature` covers the ephemeral key, not the ciphertext. It authenticates who sent an envelope; it yields nothing transferable a third party could verify against the content.

The commitment can only be made **at send time**, which is why this ships alone and first: every day it is not merged adds attachments that can never be authenticated. That cost is not recoverable — a sender can only be *asked* to re-sign later, and a bad actor never will.

## What this adds

A v2 preimage — v1 plus a `media` array:

```json
{
  "body": "<caption, may be empty>",
  "group_binding": "<unchanged v1 derivation>",
  "media": [{
    "blob_sha256": "<ciphertext digest — the blob address>",
    "height": 1600,
    "mime_type": "image/jpeg",
    "plaintext_byte_length": 421337,
    "plaintext_sha256": "<digest of the decrypted bytes>",
    "width": 1200
  }],
  "message_id": "<uuid lowercase>",
  "proof_version": 2,
  "sent_at_millis": 1785580800123
}
```

Two hashes because they answer different questions. `plaintext_sha256` is what an Authority re-derives from disclosed evidence **without ever needing the attachment key**. `blob_sha256` keeps the commitment tied to the stored object a recipient actually fetched.

All four media paths sign — including video, album and voice, whose evidence no Authority accepts yet. The commitment is only creatable at send time, so omitting them now would strand today's sends and force a `proof_version: 3` later.

## Compatibility

A message with no media produces **v1 bytes, byte for byte**. This is pinned by a golden-bytes test whose group binding was computed independently of the implementation. Every proof already signed by a shipped client keeps verifying.

The version — not the presence of the field — is what a verifier dispatches on. Honouring a `media` array inside a v1 preimage would let a sender choose which rules their own signed bytes were read under.

## Two deliberate choices worth reviewing

**Slashes stay escaped** (`image\/jpeg`). The preimage encoder does not set `.withoutEscapingSlashes`. Adding it would be a silent breaking change: a v1 body containing `/` would stop reproducing the bytes its sender already signed, retroactively invalidating the proof on every such message already delivered.

**"Plaintext" means the encoded JPEG that travelled** — `ChatImageEncoder`'s downscaled, re-encoded output — not the camera original. The evidence is what was received, which is the only thing a recipient can honestly disclose.

## Incidental fix

The media paths persisted `sentAt` from the raw clock rather than the truncated wire millis, which the text path deliberately derives so a stored row can reconstruct its own preimage. Harmless before; a latent trap once media is signed.

## Known limitation

Attachments sent before this cannot be authenticated, ever. That is documented on the type rather than worked around.

## Testing

`xcodebuild test -only-testing:OnymIOSTests` — 1084 tests, 3 failures, all in `DeviceRecoveryTests` (grant canonical bytes / server fixture). Confirmed identical on a clean tree with these changes stashed: pre-existing and unrelated.

New coverage: golden v1 and v2 preimage bytes, `media` key position, per-field digest sensitivity, album ordering inside the signed bytes, dimension keys absent for audio, and send-path tests asserting an image proof does **not** verify against the caption-only preimage.

### DIFF
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessagePayload.swift
@@ -99,6 +99,11 @@ public struct ChatMessagePayload: Codable, Equatable, Sendable {
     /// contract's authenticity rule without exposing an envelope or
     /// any recipient secrets. Optional for wire compatibility with
     /// messages sent before reporting existed.
+    ///
+    /// When the payload carries media the preimage is v2, which also
+    /// commits to each attached blob's plaintext digest, MIME type and
+    /// byte length — the only sender commitment to attachment bytes
+    /// that exists anywhere in the system.
     public let moderationAuthenticityProof: String?
 
     public init(
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift
@@ -13,6 +13,16 @@ import Foundation
 /// `groupID`, and `sentAtMillis` into the signed bytes pins the proof
 /// to one concrete send.
 ///
+/// A message carrying media signs a v2 preimage, which adds a `media`
+/// array committing to each attached blob's exact bytes. Without it an
+/// attachment is unreportable in principle, not merely unimplemented:
+/// nothing else in the system is a sender commitment to those bytes.
+/// The transport envelope is not one — `SealedEnvelope` signs only the
+/// ephemeral key it carries, so it authenticates who sent an envelope
+/// but produces nothing a third party could later verify against the
+/// content. Media sent before v2 therefore cannot be authenticated,
+/// ever, and recipients simply cannot report it.
+///
 /// The preimage is a JSON object with **UTF-8 byte-ordered keys**
 /// (`JSONEncoder.OutputFormatting.sortedKeys`), matching the canonical
 /// form every other moderation signing payload uses. The Authority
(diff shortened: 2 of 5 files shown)


dev-onym (commented):
**COMMENT** — design is sound and the v1-compatibility story checks out (I independently recomputed the golden `group_binding` — `18413f67…27ce6a` matches); findings are test-coverage and doc nits, nothing blocking.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:618` — the album's lockstep `commitments` accumulation is the trickiest new logic (skip-on-failure `continue`, poster+video pair emitting two entries, collapse-to-single at `:691`) and has **no send-path test**. `test_mediaOrderIsSigned` only exercises the pure encoder. A test that sends `[.image, .video, .image]` and verifies the proof against a hand-built 4-entry commitment list would pin the ordering contract where it can actually drift. Same gap for `sendVideo` (`:496`).
- `Tests/OnymIOSTests/SendMessageInteractorTests.swift:584` — `test_sendVoice_signsTheAudioBytes` only asserts the empty-body preimage does *not* verify. A proof over arbitrary garbage passes that assertion too, so the test doesn't establish that the audio bytes were committed. It needs the positive `isValidSignature` against the `audio/mp4` commitment, as the image test at `:512` does.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift:166` — doc references `MediaPreimageKeyOrderTests`, which doesn't exist anywhere in the tree; the assertion lives in `ChatModerationProofTests.test_mediaKeySortsBetweenGroupBindingAndMessageID`. Stale pointer in exactly the comment a future editor would trust before changing key order.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift:80` — no shared derivation from attachment descriptors to `[MediaCommitment]`. The recipient/report side must independently re-derive the poster-before-video ordering and the `mimeType` literals, and nothing but a comment records the convention. A `MediaCommitment.from(_ attachment:, plaintext:)` factory (plus an array builder over `ChatMessage.media` + `voiceAttachment`) would make the follow-up verifier reuse these bytes instead of reimplementing them — the one place drift silently invalidates every media proof.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatModerationProof.swift:60` — `mediaVersion`'s stated rule ("a `media` array riding along in a v1 preimage must be refused") isn't enforceable here since `signedContent` can't produce that shape. Fine as guidance for the Rust verifier, but worth saying so explicitly, or the comment reads like an in-tree invariant that isn't one.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:346` — the keychain signature is now awaited *before* the optimistic insert on the media paths, so it sits in front of the bubble the method's own doc (`:280`) promises appears immediately. Ed25519 signing is fast enough that this is almost certainly imperceptible; noting it only because moving the `await` after `messageRepository.insert(pending)` would cost nothing (the proof is only needed for the payload and the row update).
- Verified no regression on the report affordance: `ReportableMessageFactory:20-22` still gates on `media.isEmpty && voiceAttachment == nil`, so media messages don't get a Report entry that would dead-end on a v1 reconstruction. Confirms the PR's "ships alone and first" framing — but it also means v2 bytes are unexercised end-to-end until the verifier lands, and the golden tests are the only thing holding the format.

PR #239, opened 2026-08-11
Report a received photo

rinat-enikeev:
Stacked on #238 (base branch is `feat/media-authenticity-proof`) and pairs with onymchat/onym-moderation#38.

With a sender commitment in place, a recipient can disclose a received photo on the same terms as text.

## Verification happens where the bytes are

`ReportableMessageFactory` accepts a single-image message, and the check is the same one in a second place: recompute the digest of the bytes **this device actually holds** and require the sender to have signed *that*. A photo whose bytes don't match its commitment produces no Report entry — it isn't evidence of anything and would only dead-end at the Authority.

## Eligibility is split from disclosure

A photo's verification needs decrypted bytes, which come from an actor, and a context menu cannot `await`. So `isReportable` checks a photo as far as it can without bytes, and the byte-level check happens in `make`, before anything leaves the device.

The asymmetry is safe in exactly one direction: the menu may offer Report for a photo that later fails verification, and the flow then presents nothing. It must never refuse something disclosure would have accepted. There's a test asserting that direction.

## `plaintext(for:)` vs `image(for:)`

`ChatImageLoader.image(for:)` returns a decoded `UIImage`. Re-encoding one produces **different bytes with a different digest** — precisely the value that authenticates the photo. So `plaintext(for:)` returns the exact decrypted bytes, and the report screen renders from the same bytes it uploads. What the reporter agrees to disclose and what leaves the device cannot diverge.

## Upload, then file

The other order would leave a filed report pointing at an image the Authority does not hold — a case opened, and a mark set, on evidence nobody can look at.

Uploading is idempotent because the digest is the address, so a retry after a partial upload re-sends the same bytes rather than creating anything new, and an upload that never gets referenced expires on the Authority's own sweep. Two tests cover this: one asserts ordering (the fake throws if a report arrives before its upload), one asserts a failed upload files no report.

## Cross-repo detail worth noting

While wiring this I hit a status-code collision: the Authority's `media_missing` was mapped to **409**, which this client already reads as "these exact bytes are already on file" — a terminal, benign state, and the opposite of "upload and re-file". Changed to **424 Failed Dependency** in onym-moderation#38.

## No PhotosPicker

Deliberately absent from the flow. An arbitrary camera-roll photo has no signed commitment and could never authenticate.

## Known limitation

Photos sent before proof v2 carry a v1 signature over the caption alone. They stay unreportable, and there's a test naming that as the reason rather than leaving it to be rediscovered.

## Testing

`xcodebuild test -only-testing:OnymIOSTests` — 1090 tests, 3 failures, the same pre-existing `DeviceRecoveryTests` ones confirmed unrelated in #238.

New coverage: a received photo is reportable and its disclosed content is the v2 preimage; one flipped byte makes it unreportable; a photo without its bytes is unreportable; a pre-v2 photo is unreportable; eligibility never offers what disclosure would refuse; upload ordering; and a failed upload filing nothing.

### DIFF
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
@@ -94,7 +94,13 @@ public actor ChatImageLoader {
         cacheDir.appendingPathComponent(key).appendingPathExtension("img")
     }
 
+    /// Atomic, because a half-written cache file is no longer only a
+    /// broken thumbnail. `plaintext(for:)` hands these bytes back
+    /// verbatim as evidence, so a truncated file hashes to the wrong
+    /// digest, fails its commitment check, and makes the photo
+    /// permanently unreportable — with the picture still rendering fine
+    /// from whatever decoded.
     private func writeDisk(_ key: String, _ data: Data) {
-        try? data.write(to: fileURL(key), options: .completeFileProtection)
+        try? data.write(to: fileURL(key), options: [.atomic, .completeFileProtection])
     }
 }
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
@@ -46,6 +46,26 @@ public actor ChatImageLoader {
         return image
     }
 
+    /// The **exact** decrypted bytes of `attachment`.
+    ///
+    /// Distinct from `image(for:)` on purpose. A `UIImage` is a decoded
+    /// bitmap, and re-encoding one produces different bytes with a
+    /// different digest — which is precisely the value that
+    /// authenticates a reported photo against the sender's signature.
+    /// Anything disclosing evidence must take the bytes from here and
+    /// never round-trip them through an image.
+    ///
+    /// The disk cache already holds the plaintext verbatim, so a photo
+    /// the user has looked at costs nothing to re-read.
+    public func plaintext(for attachment: ChatImageAttachment) async throws -> Data {
+        let key = attachment.sha256
+        if let cached = try? Data(contentsOf: fileURL(key)) { return cached }
+        let plaintext = try await downloadAndDecrypt(attachment)
+        writeDisk(key, plaintext)
+        if let image = UIImage(data: plaintext) { memory[key] = image }
+        return plaintext
+    }
+
     /// Sender-side warm cache: after uploading, prime the decrypted
     /// image so the sender renders instantly without re-downloading.
     func prime(sha256: String, plaintext: Data) {
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadView.swift
@@ -130,6 +130,10 @@ public struct ChatThreadView: View {
     /// hands the controller fresh data via `updateUIViewController`.
     @State private var messages: [ChatMessage] = []
     @State private var reportableMessage: ReportableMessage?
+    /// The message whose disclosure is being prepared, if any. Doubles
+    /// as the in-flight indicator and the repeat-tap guard.
+    @State private var preparingReportFor: UUID?
+    @State private var reportUnavailable = false
 
     public var body: some View {
         ChatThreadControllerBridge(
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the photo path itself is sound (byte-level re-verification, no re-encode, upload-before-file), but the upload's error handling sits outside every safeguard the filing path has.

- `ModerationRepository.swift:1266` — the `for image in images { try await client.uploadEvidenceImage(...) }` loop is *outside* the `do/catch` that wraps `fileReport`. Consequences:
  - A **409 from the blob PUT** propagates raw as `AuthorityClientError.rejected(409)`, never reaching the `statusCode == 409` → `.reportAlreadyFiled` mapping, and the report is never filed. A duplicate PUT is the *expected* case on any retry (`performFileReport` line 1216 re-enters `submitReport` and re-uploads). The PR body argues the upload is idempotent — that only holds if the Authority returns 2xx for an existing blob. Worth pinning down against `put_evidence_blob` in onym-moderation#38, and tolerating 409 here regardless, since the client already reads 409 as "already on file" everywhere else.
  - A **deterministic upload rejection** (400/413/415/403) skips `isDeterministicReportRejection`/`discardReport`. The record was already inserted and `saveReportRecords()`-persisted at line 1247 *before* `submitReport`, so a disclosure row that can never be filed is retained on disk indefinitely — exactly the outcome the text path's discard exists to prevent — and every retry re-fails identically.
- `ChatThreadView.swift:180` — `Task { reportableMessage = await makeReportableMessage(...) }` has no in-flight or failure feedback. For an uncached photo this is a full Blossom download + decrypt; the user taps Report and gets nothing for seconds, or nothing at all if the download fails or the bytes don't verify. Given the menu predicate is deliberately optimistic, "presents nothing" is a reachable normal path, not an edge case — it needs at least a spinner and a failure alert. Nothing debounces repeat taps either, and the sheet can appear after the user has moved on.
- `ModerationReportView.swift:81` — `if let rendered = UIImage(data: image.bytes)` silently renders nothing on decode failure. With an empty `displayBody` (line 92) the user then sees a "PHOTO TO DISCLOSE" header, the disclosure footnote, and no photo — and Submit still uploads the bytes. That inverts the invariant the PR is built on; prefer refusing to present over disclosing something unseen.
- `ReportableMessageFactory.swift:51,85` — `message.media.count <= 1` is dead: `ChatMessage.media` (ChatMessage.swift:102) derives from `albumAttachments ?? imageAttachment ?? videoAttachment`, all three already guarded, so it is always exactly 1 here. Harmless, but it reads as a real bound.
- `ChatImageLoader.swift:98` — `writeDisk` isn't atomic. A truncated cache file now means more than a broken render: `plaintext(for:)` returns it verbatim, the digest mismatches, and the photo is permanently unreportable with no signal. `options: [.atomic, .completeFileProtection]` is a cheap fix.
- `ModerationReportView.swift:81` — `ForEach(Array(...enumerated()), id: \.element.sha256)` discards the index; `ForEach(flow.message.images, id: \.sha256)` is equivalent.

Test coverage on the new logic is genuinely good — the tamper, missing-bytes, pre-v2, ordering, and failed-upload cases are the right ones, and the cross-repo fixture is the correct answer to two encoders in two languages. Missing: nothing covers an upload that fails with a rejection status, which is where finding #1 lives.

dev-onym (commented):
COMMENT — the core design (byte-level commitment check, upload-before-file, exact-bytes rendering) is sound and well tested; findings below are layering, perf, and doc issues, none blocking.

- `ModerationRepository.swift:1281` — the repository re-verifies the accused's signature over `disclosedContent` precisely because it doesn't trust its caller, but `images` ride along unverified: nothing checks that each `image.sha256` actually appears in the verified `disclosedContent`. A caller (or a future factory change) can make the client PUT arbitrary photo bytes under the reporter's credential attached to a report that never references them — the tests demonstrate exactly that shape (`disclosedContent: "prohibited content"` + `sha256: "digest-of-the-photo"`). A one-line `guard message.images.allSatisfy { message.disclosedContent.contains($0.sha256) }` in `performFileReport` restores the boundary the rest of the function argues for.
- `ModerationAuthorityClient.swift:411` — `uploadEvidenceImage(sha256:bytes:)` never checks that `sha256` is the digest of `bytes`; the address it PUTs to can disagree with the content it sends. The Authority catches it, but recomputing locally (or asserting) turns a remote rejection into an impossible state.
- `ModerationReportView.swift:66,115` — both `hasUndisplayableImage` and the `ForEach` decode the full-size JPEG inside `body`, so every state change (selecting a violation class, an error appearing, submit toggling) re-decodes multi-megabyte evidence twice on the main thread. Decode once — e.g. `let decoded = flow.message.images.map { UIImage(data: $0.bytes) }` computed in `init`/`.task` and stored — then gate and render from that.
- `ModerationReportView.swift:110` — the comment says "Force-unwrapped deliberately… a silent fallback here is what would let an unseen photo be disclosed," but line 115 *is* a silent fallback (`?? UIImage()`, a blank image). The `hasUndisplayableImage` gate makes it unreachable today, so this is only a comment/code contradiction — but it's contradicting itself about the one invariant the screen exists for. Render from the pre-decoded value and the branch disappears.
- `ChatImageLoader.swift:62` — `plaintext(for:)` returns disk-cached bytes with no integrity check, and nothing invalidates the cache when the commitment check later fails. A file corrupted by a pre-atomic-write partial (the case the new `.atomic` flag acknowledges) or a disk-full truncation makes that photo unreportable *permanently*, with the alert blaming an incomplete download. Consider deleting the cache entry and re-downloading once when `make` rejects the bytes, or having `plaintext` verify before returning.
- `Tests/OnymIOSTests/Fixtures/media-commitment-v2.json:2` — the note says the identical copy lives at `onym-ios/Tests/OnymIOSTests/Fixtures/media-commitment-v2.json`, which is this file's own path. It should name the onym-moderation copy (`authority/fixtures/…`, as `MediaCommitmentFixtureTests.swift:16` correctly does) — the cross-repo pointer is the fixture's entire reason for existing.
- `ChatThreadView.swift:314` — the preparing overlay is a full-screen tap-blocking dim with no cancel; an uncached photo on a bad network leaves the user stuck until the Blossom download times out. A Cancel button (clearing `preparingReportFor`, which the post-await identity guard already tolerates) would cover it.

PR #241, opened 2026-08-11
Move join requests into the chat thread

rinat-enikeev:
## Why

A founder only learned about a join request by noticing a badged toolbar button on the chat list and opening a separate **Join requests** modal. New users never found it — nothing in the conversation they were actually looking at said someone was waiting to get in.

The request now renders as a row **inside the group's own thread** with Accept / Decline, and the separate screen and its toolbar badge are gone. Admitting someone writes an "X joined" notice into every member's thread; the joiner's newly materialized thread opens with "You joined \<group\>".

## Design notes

**Founder-only comes for free.** A join request arrives sealed to a per-invite intro key that only the inviting device holds, so a non-founder's pending list is empty by construction — no new filtering was needed. The thread additionally checks the active identity's BLS pubkey against the group's stored admin pubkey as defence in depth (same derivation as `ChatMembersView.canShareInvite`).

**System notices are local-only.** `ChatMessagePayload` has no `system_event` key and must not grow one — a wire field would let any peer forge "Alice joined" history. Each device mints its own copy from an event it independently verified:

| Device | Trigger |
|---|---|
| admin | its own successful `approve` |
| existing members | signature- **and** chain-checked `MemberAnnouncementPayload` |
| joiner | its verified `GroupInvitationPayload` |

Row ids are derived from `(group, owner, event)`, so a relay replaying the inbox on reconnect collapses onto the existing row instead of stacking a duplicate notice on every launch. Insertion also sits behind each caller's existing dedup guard.

**Two different lifetimes, two mechanisms.** "X joined" is permanent history → a persisted `ChatMessage`. The request prompt is transient and consumed on action → a synthetic row backed by the (now persisted) `IntroRequest`, so crypto material isn't duplicated into the message store.

## Behaviour change beyond the UI

Intro requests are now **persisted**. They were deliberately process-lifetime, which was fine when approval was a modal to act on there and then. As a row in the thread, vanishing on force-quit reads as a message the app lost — and the joiner has no way to know, they're sitting on "Waiting for the host to approve…".

## Carve-outs

System rows are excluded from unread counts (a join would otherwise light up a badge on a chat nobody spoke in), and skip sender headers, run-grouping, and the Report menu. Approve errors are now keyed to the request that failed, since the thread shows one row per request rather than one banner.

## Bug found while wiring this up

The message table sits at `alpha 0` until the cold-open reveal, which is gated on the first *non-empty message* snapshot. A brand-new group with no messages — exactly when the first join request lands — would have rendered the row invisible. Fixed in `update(joinRequests:)`.

## Verification

- Build succeeds; full `OnymIOSTests` suite passes, including 14 new tests: system-event persistence round-trip, unread-count carve-out, replay idempotence, per-owner rows, persisted intro-request store, and error attribution.
- **Not run:** the two-simulator end-to-end pass. It needs live relays and a chain relayer. `MultiIdentityChatUITests` was rewritten to accept in-thread and assert the "Bob joined" notice, but was not executed — worth running before merge.


dev-onym (commented):
REQUEST_CHANGES — solid architecture (locally-minted, derived-id system rows are the right call), but two behavioural gaps.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadView.swift:374` — `sendReadReceipts` filters on `direction == .incoming` only, and system rows are minted `.incoming` with a **real member's** BLS hex. So opening a thread ships a `.read` receipt to that member for a UUID no other device has ever seen; for `.youJoined` the sender hex is your own key (it's in `memberProfiles`, dispatcher line 363), so you send a read receipt to yourself. This contradicts `ChatSystemEventRecorder`'s stated invariant ("`.incoming` + `.received` keeps the row off every outgoing path (retry, receipts, …)") and receipts are missing from the PR's carve-out list. Worst case: opening a brand-new group with zero real messages now emits relay traffic to the joiner. Fix: add `&& !message.isSystem` to the loop at :374.
- `ChatThreadView.swift:511,532` — `canAccept: request.groupName != nil` and the "Decline stays live so the row can still be cleared" comment are dead code. The row renders *only* inside the thread of a group found in `chatsFlow.groups` (:511-513), and `groupName == nil` means precisely that no such local group exists. So a request for an unknown group can no longer be declined at all — and since it's now persisted, it sits on disk forever, invisible. The deleted modal was the only surface that could clear it. Needs either a fallback surface or server-side/TTL pruning.
- `Packages/OnymGroup/Sources/OnymGroup/SwiftDataIntroRequestStore.swift:91-93` — no TTL/pruning anywhere, and `record` returns `false` for both "duplicate" and "encrypt/save failed", so a failure is silently read as a dup by callers. Unlike `SwiftDataMessageStore.insertOrUpdate`, a failed `save()` doesn't `context.delete(row)`, leaving a half-inserted object a later `consume()` `save()` can commit without publishing. Approve's early `return .transportFailed(...)` paths also skip `consume`, so those rows now survive relaunch (previously cleared by process death).
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatSystemNoticeCell.swift:64-70` vs `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatMessage.swift:217` — the same two sentences are hardcoded twice. The `ChatSystemEvent` doc claims the sentence is assembled in the UI, but Core assembles it too; change one and the chat-list subtitle silently diverges from the thread pill. Have the cell call `event.previewText`.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadViewController.swift:336` — `Dictionary(uniqueKeysWithValues:)` traps on a duplicate request id rather than degrading; matches the existing idiom at :284, so low risk, but a crash rather than a dropped row.
- `Tests/OnymIOSTests/ChatSystemEventTests.swift` `test_recordMemberJoined_isPerOwner` asserts a per-owner guarantee the delivery path can't honour: `IncomingMessageDispatcher.applyAnnouncement` resolves `groups.first(where: groupIDData == ...)`, so with two local identities in one group only one gets the row. Pre-existing dispatcher limitation, but the new test reads as coverage for it.
- `Resources/Localizable.xcstrings` — the file was re-sorted end-to-end (~800 lines of churn for ~6 real key additions) and lost its trailing newline. Worth reverting to the original key order to keep the diff and future merges readable.

Non-blocking: `isGroupAdmin` (`ChatThreadView.swift:495`) omits the `groupType == .tyranny` check that `canShareInvite` has — currently harmless since `adminPubkeyHex` is nil for admin-less groups. I verified `insertOrUpdate` is genuinely idempotent for incoming rows, so the replay-collapse claim holds; no dangling references to the deleted `ApproveRequestsView`/`lastSuccessMessage` remain. The two-simulator E2E pass is still unrun, as the PR discloses — the rewritten `JoinRequestRow` page object is the only check on the new accessibility ids, so please run it before merge.

dev-onym (commented):
**REQUEST_CHANGES** — sound architecture (local-only notices, derived ids, receipt/unread carve-outs all check out); the blockers are a discoverability hole the PR reopens and a UI test that will fail as written.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatSystemNoticeCell.swift:66` — `configure` sets `isAccessibilityElement = true` on the **cell**, collapsing the subtree, so the inner `UILabel` is no longer exposed. `ChatBubbleCell` deliberately doesn't do this (`ChatFlowScreens.swift:223` finds bubbles via `app.staticTexts[text]`). The new `waitForJoinedNotice` (`ChatFlowScreens.swift:191`) queries `app.staticTexts["Bob joined"]` and won't match — the cell surfaces as `.cell`, not `.staticText`. Since the E2E pass wasn't run, this ships red. Either drop `isAccessibilityElement` or query `app.cells["chat.system_notice"]`/match on label.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift:99` — with `ApproveRequestsToolbarButton` gone there is now **no ambient signal anywhere** that a request is waiting: no chat-list badge, no `chatListPreview` change (a request isn't a `ChatMessage`), and system rows are explicitly excluded from `unreadCount` (`SwiftDataMessageStore.swift:110`). A founder who doesn't open that specific thread learns nothing, and after 7 days `SwiftDataIntroRequestStore.retention` sweeps it silently while the joiner still sits on "Waiting for the host to approve…". The PR trades one undiscoverable surface for another; consider a chat-list badge keyed on `pending` per group.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:381` and `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:731` — `joinerDisplayLabel` / `newMember.alias` go into `.memberJoined` unsanitized. `ChatThreadView.swift:592` trims and falls back to `(unnamed)` for the request row; the notice doesn't, so an empty or whitespace alias renders as " joined" permanently in history. Apply the same trim+fallback in `ChatSystemEventRecorder`.
- `Packages/OnymGroup/Sources/OnymGroup/ApproveRequestsFlow.swift:88,105` — `lastErrorRequestID` was added precisely so one row's failure doesn't paint every row, but `approve` success and `decline` both null it unconditionally: approving request A erases the visible error under request B. Only clear when `lastErrorRequestID == id`. Same shape means two concurrent failures can only ever show one.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatJoinRequestCell.swift:180` — when `canAccept` is false the button is disabled with **no** explanation; the deleted view said "This request is for a group that isn't on this device. Decline to clear it." If the state is truly unreachable (as the comment argues, and `GroupRepository.currentGroups()` lazy-loads so `groupName` does resolve on cold launch), delete `canAccept`; otherwise restore the copy.
- `Packages/OnymGroup/Sources/OnymGroup/SwiftDataIntroRequestStore.swift:112,183` — `record()` calls `pruneExpired()`, then `publish()` → `loadAll()` → `pruneExpired()` again: two fetch+delete passes per insert. Drop the explicit call in `record`.
- Doc rot from the deleted modal: `ApproveRequestsFlow.swift:6-8` ("the toolbar badge on `ChatsView`… the modal `ApproveRequestsView`"), `IntroRequestStore.swift:3-9` (still says process-lifetime, contradicting the new persisted impl), `Sources/OnymIOS/AppDependencies.swift:39`, `Packages/OnymInbox/Sources/OnymInbox/PendingInvitesView.swift:5`. `ChatsView.swift:99-105` also leaves an orphaned comment paragraph attached to the *invitations* toolbar item.
- Minor: `ChatThreadViewController.swift:379` reveals the table without latching `hasAppliedFirstSnapshot`, so the first real message (usually the "X joined" notice right after Accept) still takes the `isFirstApply` branch and hard-jumps an already-visible table instead of animating the append.

Verified as correct and not worth changing: derived-UUID idempotence against `insertOrUpdate`'s immutable-incoming rule, the `isSystem` skips in `rebuildSenderDisplays`/context menu/`ChatReadReceiptTargets`, `isGroupAdmin` (anarchy/1:1 have `adminPubkeyHex == nil` and can't share invites anyway), and all six new xcstrings keys matching their `String(localized:)` interpolation forms.

dev-onym (commented):
**REQUEST_CHANGES** — solid design (local-only notices, derived ids, keyed errors all check out); the chat-list signal is under-gated and the one genuinely new UIKit path has no tests.

- `ChatsView.swift:312` — `pendingJoinRequestCounts` isn't gated on admin/owner, but the thread row is (`ChatThreadView.swift:556`). `approveRequestsFlow.pending` is device-wide: `IntroKeyStore.find(introPublicKey:)` is not identity-scoped, so decode succeeds under any active identity. With two local identities in one group (explicitly supported — see the `MessageRepository` header), the *non-admin* identity's row shows "Someone wants to join" + badge, and opening the thread offers nothing to act on. Reuse the `isGroupAdmin` derivation (or filter `pending` by `group.adminPubkeyHex` == active identity's BLS hex) when building the counts.
- `ChatsView.swift:448` — the join-request badge replaces the unread badge, and `:416` replaces the subtitle. A founder with unread messages *and* a pending request loses the unread count and the last-message preview entirely. Both signals fit; suggest keeping the unread badge and prepending the join hint, or at minimum not dropping `chats.row.unread.<id>`/`chats.row.subtitle.<id>` from the tree (UI-test surface too).
- `ChatThreadViewController.swift:334` — `update(joinRequests:)` is the trickiest new code (duplicate-id drop, append order, empty-state toggle, cold-open reveal) and has zero tests, while `ChatThreadViewControllerTests` already drives `update(messages:)` in-process. The 14 new tests cover Core and the flow well; this path is the gap.
- `ChatThreadViewController.swift:379` — the reveal sets `alpha = 1` synchronously in the completion, whereas `jumpToBottomForColdOpen` defers it a runloop precisely to mask the height-settle (and, opened-from-search, does a second `scrollToRow`). When a message snapshot and a request snapshot both land while `alpha == 0` (cold open of a founder thread with a pending request), this reveals before the settle pass — a frame at the wrong offset. Gate on `hasAppliedFirstSnapshot`/`orderedMessages.isEmpty` instead of `alpha == 0`.
- `ChatSystemEventRecorder.swift:81,104` — `(Unnamed)`/`(unnamed)` are localized at *record* time and frozen into permanent history, which contradicts the "structured data, not prose" rationale on `ChatSystemEvent` and won't follow a later language change. Store the raw/empty value and fall back in `localizedText`.
- `SwiftDataMessageStore.swift:353` — an undecodable `systemEvent` (unknown `kind` after a downgrade) yields `systemEvent: nil`, so the row renders as an ordinary *empty bubble* attributed to the joiner rather than being hidden. Consider dropping the row on decode failure.
- `SwiftDataMessageStore.swift:183` — the outgoing→outgoing overwrite branch doesn't copy `encryptedSystemEventJSON`. Unreachable today (notices are `.incoming`), but it's a silent field drop if that ever changes; add the assignment for uniformity.
- `PageObjects/ChatFlowScreens.swift` — `waitForJoinedNotice` matches the literal `"\(alias) joined"`, so the E2E pass is locale-dependent now that the string comes from the catalog. Worth noting alongside the already-flagged "not run" caveat.

Verified as claimed: replay idempotence holds (`insertOrUpdate` treats incoming rows as immutable, so `sentAt` doesn't shift), the unread predicate stays a `fetchCount`, no delivered-ack or local-notification path enumerates stored rows, swipe-to-reply is `ChatBubbleCell`-only, all new catalog keys resolve, and the removed modal/toolbar has no dangling references.

dev-onym (commented):
**REQUEST_CHANGES** — solid design (local-only system rows, derived ids, per-request error keying all check out); one real regression in the cold-open path.

- `ChatThreadViewController.swift:388` — the reveal latch fires on the *wrong* first render. `messages` is `@State` starting empty (`ChatThreadView.swift:139`), while `approveRequestsFlow.pending` is already populated from the app-lifetime flow. So for a founder opening a group that **has** history with a request pending, the bridge's first pass is `update(messages: [])` (no latch, `sorted.isEmpty`) then `update(joinRequests: [r])` → `orderedMessages.isEmpty` is true, so it sets `alpha = 1` and latches `hasAppliedFirstSnapshot = true`. The real history then lands via `update(messages:)` with `isFirstApply == false` → `animatingDifferences: true` + `scrollToBottom(animated:)`: every existing message flies in and scrolls, which is exactly what the cold-open branch exists to prevent (`ChatThreadViewController.swift:307-315`).
- `ChatThreadViewController.swift:436` — same root cause, worse consequence: `openAtMessageID` is consumed *only* inside `jumpToBottomForColdOpen()`, which now never runs in that scenario. Opening a Search hit (`RootView.swift:152`) in a group with a pending request lands at the bottom instead of the target message, with no flash — silently, since nothing else clears `openAtMessageID`.
- `ChatThreadViewControllerTests.swift` (`test_joinRequestAfterMessages_leavesTheRevealToTheColdOpen`) only covers messages-first. The failing order is request-first-then-messages, which is the actual production order. Suggested fix: don't latch on the request path — have the host pass a "message snapshot has arrived" flag so an empty thread can be distinguished from an unloaded one, and let `update(messages:)` keep owning the cold open (it can skip the redundant re-reveal when alpha is already 1).

Non-blocking:

- `ChatsView.swift:318` — `pendingJoinRequestCounts` is a computed property read once per row inside `List(flow.items)`, so it rebuilds the whole dict (and re-runs `isAdmin` per request) for every row on every render. Hoist it into a `let` in `groupList`.
- `ApproveRequestsFlow.swift:122` — `dismissError()` now has no caller outside a test; with the modal gone there's no dismiss affordance, so a failed approve's red text sits under the row until that same request is retried or declined. Fine if intentional, but the API is dead as written.
- `ChatSystemNoticeCellTests.swift:32` — asserts `label?.text == "Bob joined"`, hardcoding English for a string that now resolves through the catalog. That's the same brittleness the UI-test page object was just rewritten to avoid; `event.localizedText` (as the third test does) is the durable form.

Verified as claimed: the founder gate matches `ChatMembersView.canShareInvite:118` and is consistent for Anarchy/1-on-1 (nil `adminPubkeyHex`, and `anchorTyrannyJoin` rejects those anyway); no positional `indexPath` indexing survives the appended request rows; delivered-acks are dispatcher-driven so notices stay off that path; search snippets match on `body`, which is empty for notices; all new catalog keys are present with `ru`.

dev-onym (commented):
**REQUEST_CHANGES** — the in-thread request row renders on a blank, never-revealed table in the exact case the PR set out to fix; everything else is minor.

- **Blocking — the request-only cold-open reveal is unreachable in production.** `ChatThreadViewController.swift:353` early-returns when the request list is unchanged, but the bridge's *first* call always arrives with `messagesLoaded: false`: `hasLoadedMessages` starts `false` (`ChatThreadView.swift:149`) and only flips at `ChatThreadView.swift:408`, after the first render. So for a founder opening an empty group with an **already-pending** request (the common path — they tapped the row because of the new "Someone wants to join" hint): call 1 stores the row, hides the empty state (`ChatThreadViewController.swift:349`), and skips the reveal because `messagesLoaded` is false; call 2 has `messagesLoaded: true` but an identical list, so `:353` returns before reaching `:403`. Table stays at `alpha 0` with the empty state suppressed → an entirely blank thread until some real message lands and the cold open fires. Fix: evaluate the reveal above the unchanged-guard, or latch "reveal still owed" and re-check it on every call.
- **Test gap for the above.** `test_emptyLoadedThreadWithARequest_isStillRevealed` calls `update(joinRequests:messagesLoaded: true)` as the first request push, and `test_requestBeforeHistory_*` only cover the case where real history follows. Nothing covers `messagesLoaded: false` → same list with `messagesLoaded: true`, which is the production sequence.
- **Retrying a failed approve shows spinner + stale error together.** `ApproveRequestsFlow.swift:95` clears the keyed error only on `.sent`, so a second `approve(id)` renders "Anchoring on chain…" underneath the previous failure. Clear `lastError`/`lastErrorRequestID` for `id` at the start of `approve`/`decline` instead.
- **`pruneExpired` doesn't notify subscribers** (`SwiftDataIntroRequestStore.swift:178`). A sweep that happens on a `current()` read deletes the row but leaves it on screen until the next `record`/`consume` publishes. Note `publish()` → `loadAll()` → `pruneExpired()`, so calling `publish()` from the sweep needs a recursion guard.
- Nit: the green on-chain seal is suppressed whenever a request is pending (`ChatsView.swift:497`), which contradicts the "both badges when both apply" comment directly above it.
- Nit: `JoinRequestBadge` (`ChatsView.swift:545`) renders only an icon, never the number, unlike `UnreadBadge` — 2 vs 9 pending requests are indistinguishable outside VoiceOver and the subtitle.
- Nit: duplicate sibling requests collapsed by `JoinRequestApprover.refresh` are never consumed (only the representative id is, at `JoinRequestApprover.swift:397`/`:667`). Harmless now that the intro key is revoked (`decode` drops them silently), but they occupy disk for the full 7-day retention where process death used to clear them.

Design side, the parts I checked hold up: no `system_event` wire field, derived row ids are genuinely replay-idempotent against `insertOrUpdate`'s incoming-immutable branch, the read-receipt carve-out is real, and the delivered-ack path is untouched by notices (it only runs off a wire payload). Agreed the two-simulator E2E should run before merge — `waitForJoinedNotice` is unexercised.

dev-onym (commented):
**REQUEST_CHANGES** — solid design and unusually good test coverage; two concrete defects in the new persistence/E2E paths.

**Blocking**

- `Packages/OnymGroup/Sources/OnymGroup/SwiftDataIntroRequestStore.swift:167` — retention is measured against `IntroRequest.receivedAt`, which is **sender-asserted**, not local receipt time: `IntroInboxPump.swift:109` passes `message.receivedAt`, sourced from the Nostr event's `ms` tag (`NostrInboxTransport.swift:207` → `NostrEvent.displayMilliseconds:69`, a tag the *joiner* writes). Two consequences, both new with persistence: a founder offline >7 days gets the replayed request pruned on the first `loadAll()` and never sees it (joiner still on "Waiting for the host to approve…" — the exact failure this PR exists to fix); and a joiner who stamps `ms` far in the future gets a row that never expires and permanently sorts first. Stamp `Date()` at `record()` into a separate `firstSeenAt` column and prune on that; keep `receivedAt` for display/ordering.
- `Tests/OnymIOSUITests/PageObjects/ChatFlowScreens.swift` (`waitForJoinedNotice`) — `app.staticTexts.matching(identifier:).containing(NSPredicate(format: "label CONTAINS…"))`. `containing` filters by *descendants*; a `UILabel`-backed static text has none, so this query matches nothing and the (unrun) E2E will fail at the assert. Every other query in this suite uses `.matching(NSPredicate…)` — e.g. `IdentityManagementUITests.swift:54`. Fold both conditions into one predicate: `identifier == "chat.system_notice" AND label CONTAINS[c] %@`.

**Non-blocking**

- `Packages/OnymGroup/Sources/OnymGroup/ApproveRequestsFlow.swift:27` — `lastError`/`lastErrorRequestID` is a single slot, but the UI is one row per request. Two failing requests → the first row silently drops its explanation while its request is still pending. `[String: String]` keyed by request id matches the stated design; the current shape only *attributes* one error.
- `ApproveRequestsFlow.swift:135`+ (`failureReason`) — returns hardcoded English. These now render in `ChatJoinRequestCell.errorLabel`, alongside copy that all goes through the catalog; `ChatSystemEvent.localizedText` makes a point of this. Worth routing through `String(localized:)` in the same pass.
- `ChatThreadView.swift:568` `isGroupAdmin` and `ChatsView.swift:335` `isAdmin(ofGroupID:)` are the second and third copies of `ChatMembersView.canShareInvite:117`, and both drop its `groupType == .tyranny` clause. Extract one helper rather than letting the derivations drift.
- `Resources/Localizable.xcstrings` — `"%@ присоединился"` is masculine-only for a self-asserted alias of unknown gender; a gender-neutral phrasing (e.g. "%@ — теперь участник") avoids misgendering. `"%lld join requests"` also has no Russian plural variants, though the rephrased value sidesteps it.
- `ChatJoinRequestCell.swift:190` — when `!canAccept && isInFlight`, the hint shows "isn't on this device" rather than the anchoring copy. Unreachable in practice, but the in-flight branch should win.

Verified good: `insertOrUpdate` refuses incoming-row overwrites, so derived ids genuinely collapse replays (`SwiftDataMessageStore.swift:180`); delivered-acks are per-wire-message only (`IncomingMessageDispatcher.swift:914`), so no sibling leak to the read-receipt fix; the `revealOwed`-before-guard ordering and its four tests are correct.

dev-onym (commented):
COMMENT — solid, well-reasoned change; the design claims in the PR body hold up against the code. One real functional hole and a coverage gap worth closing before merge.

**Verified as claimed** — no wire field for system events; the member-side notice sits strictly after the dedup guard, admin-signature check and on-chain commitment check (`IncomingMessageDispatcher.swift:723`); `recordYouJoined`'s guard is at function end so nothing is skipped; `insertOrUpdate` really is idempotent on `(id, owner)` for incoming rows (`SwiftDataMessageStore.swift:176`); all new UI strings exist in `Localizable.xcstrings` with `ru` values (including `Decline`, `Anchoring on chain…`, both hints); no dangling references to the deleted `ApproveRequestsView`/`ApproveRequestsToolbarButton`; both `ChatThreadView(` call sites updated; every table-delegate path is identifier-based, so appending request rows can't misindex `orderedMessages`; system rows stay off the delivered-ack path (its only caller is the wire path at `IncomingMessageDispatcher.swift:914`).

**Findings**

- `ChatThreadView.swift:612` — requests for a group that isn't on this device are now **unreachable**, not just un-acceptable. The old modal listed them cross-group with "Decline to clear it", and `decline` is one of only two callers of `introKeyStore.revoke` (`JoinRequestApprover.swift:665`). `GroupRepository.delete` doesn't revoke intro keys, so: founder deletes a group locally while its invite link is live → every request is stored, rendered by nothing, swept after 7 days (`SwiftDataIntroRequestStore.swift:182`) **without revoking the key** → the link keeps minting requests indefinitely and each joiner sits on "Waiting for the host to approve…" forever. The `canAccept` comment acknowledges the row is unreachable but not that the revoke path went with it. Either revoke on the retention sweep, or revoke a group's intro keys on local delete.
- `ChatsView.swift:320` — `pendingJoinRequestCounts` + `isAdmin(ofGroupID:)` and `ChatGroup.isAdmin` (`ChatGroup.swift:168`) have **zero tests**, and this is the signal the PR's whole discoverability thesis now rests on (the two-identity gating case the doc-comment argues for is exactly what a test would pin). The E2E that would have covered it end-to-end wasn't run. Worth a small unit test on `isAdmin` + the counts dictionary.
- `ChatSystemEventRecorder.swift:104` vs `ChatMessage.swift:233` — `displayAlias` trims the alias before storing while `ChatSystemEvent.display` trims again at render, and `recordYouJoined` (`:79`) stores `groupName` untrimmed. The stated invariant is "the stored row keeps whatever arrived"; alias is the one field that doesn't. `displayAlias` is redundant — drop it and let `display` do the trimming for both.
- `ApproveRequestsFlow` — `errors` entries are only removed by acting on that same request id, so a request swept by retention while showing a failure leaves an orphan entry for the app's lifetime. Harmless (rows key off `pending`), but a `pending`-change prune would close it.
- E2E: `JoinRequestRow.declineFirst` and `waitForRequest`'s public use are dead in the rewritten flow, and the pass still doesn't assert the new chat-list signal — the surface that replaced the badge it used to assert on.

dev-onym (commented):
**REQUEST_CHANGES** — the trust model and idempotence hold up; the blocker is unintended localization-catalog churn, plus some dead/contradictory carry-over.

**Blocking**

- `Resources/Localizable.xcstrings:1670` — this branch adds `"extractionState": "stale"` to **164** entries (254/276 now stale), including strings *this PR starts using*: `"Decline"`, `"Anchoring on chain…"`, `"Generating proof and updating the on-chain commitment…"`, `"This request is for a group that isn’t on this device…"`. Stale marks live keys as removal candidates, so the next catalog clean drops their `ru` translations. This reads as a partial-target build artifact, not a curated edit — revert the noise and keep only the genuinely dead entries (`Join requests`, `Approved on chain`, `No pending requests`, `wants to join “%@”`) plus the new `manual` ones.

**Dead code the PR's own comments admit is unreachable**

- `ChatThreadView.swift:612` + `ChatJoinRequestCell.swift:204` — `canAccept` is always `true` at the sole construction site: the row is only built when `chatsFlow.groups.first(where: { $0.id == groupID })` resolves, and `PendingRequest.groupName` is nil exactly when no local group exists. So the cell's disabled-Accept branch, its hint copy, and that catalog string cover a state nothing can reach. Drop the flag, or keep it and stop shipping the string.

**Orphaned doc comments (each doc landed on the wrong declaration)**

- `ChatsView.swift:532` — `JoinRequestBadge` is documented "Red pill showing the unread-message count", immediately above its own "Deliberately not red" comment; `UnreadBadge` (`:567`) lost its doc entirely.
- `ChatsView.swift:510` — `subtitle`'s doc now sits on `joinRequestSubtitle`, and the appended line "Replaces the subtitle while requests are outstanding" contradicts the `HStack` above it, which deliberately renders both.
- `ChatMessage.swift:233` — `localizedText`'s doc landed on `private static func display`; `localizedText` (`:238`) has none.

**Non-blocking**

- `ChatsView.swift:320` / `ChatThreadView.swift:568` — both surfaces gate on the *active* identity being admin, but `IntroKeyStore.find` decodes device-wide. On a two-identity device, a request arriving while the non-admin identity is selected is invisible everywhere (no badge, no row) until the user happens to switch — the old toolbar badge was identity-agnostic. It then ages out silently at 7 days. Worth a follow-up.
- `ApproveRequestsFlow.swift:29,137` — `errors` entries are only dropped by a later approve/decline of that same id, so a request removed by the retention sweep leaves its entry behind for the process lifetime. Pruning against `pending` on each stream update would close it.

**Verified as claimed:** replay idempotence (derived ids + `insertOrUpdate` treating incoming rows as immutable, `SwiftDataMessageStore.swift:186`), the `applyAnnouncement` notice sitting past dedup + signature + on-chain checks (`IncomingMessageDispatcher.swift:697`), retention measured on `firstSeenAt` with a working `isPruning` recursion guard, and the reveal ordering (`orderedMessages.isEmpty` is what saves the history case, since `hasAppliedFirstSnapshot` latches asynchronously). The unrun two-simulator E2E is disclosed and I agree it should run before merge — `waitForJoinedNotice` is now the only thing pinning the accept→notice round trip.

dev-onym (commented):
**COMMENT** — solid, well-tested change; one plausible stuck-state and one security-posture shift worth confirming before merge.

- `Packages/OnymInbox/Sources/OnymInbox/PendingInvitesView.swift:145` — `.chainSettling` renders a spinner with **no Retry and nothing that re-drives it**. `deferLocally` sets no timeout and no target (`GroupStateVerifier.swift:183`), and the auto-retry in `deferVerification` only fires for `.unreachable` (`GroupStateVerifier.swift:150`). The only path back is the relay re-delivering the invitation and `materializeGroup` re-reading the chain. If the relay connection stays up, a joiner sits on "usually takes a few seconds" indefinitely with no affordance. Either offer Retry here too (the `retry` → `reverify` path already works for it) or schedule a delayed re-verify.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:648` — accepting an archived-epoch match changes the trust posture, not just latency: a snapshot up to 64 epochs stale now materializes without the admin confirming current state. There's no member-removal or secret-rotation path in the tree today, so nothing is exploitable, but the PR body frames this purely as "the admin no longer has to be awake". Worth stating explicitly, since a future revocation feature would silently inherit a 64-epoch acceptance window.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadView.swift:580` — a pending request whose group isn't on this device, or that decoded under a non-admin local identity, now has **no surface at all**: the thread gates on `isGroupAdmin`, the chat-list signal gates on the same, and the cross-group modal is gone. It can't be declined and ages out after 7 days while the joiner watches "Waiting for the host to approve…". The comment argues the un-acceptable state is unreachable, which is true for the *row*, but the request itself is still reachable state with no exit.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:607` — `_ = readError` is a dead binding, and mapping *all* `ChainReadError` to `.chainNotConfigured` mislabels the common transient case: `SEPContractChainStateReader.client()` throws `.noActiveRelayer` whenever `selectURL()` is momentarily empty, so a plain connectivity gap reads as "Still setting up this device's connection to the chain."

- `Packages/OnymChain/Sources/OnymChain/RelayerRepository.swift:79` — releasing `startTask` means `start()` is no longer one-shot; every call refetches. Only caller is the launch `.task` (`OnymIOSApp.swift:762`) so it's fine today, but the doc comment at `RelayerRepository.swift:44` still describes it as a one-time bootstrap.

- Test gap: `ChatThreadViewControllerTests` covers row ordering, dedup, empty state and the reveal, but nothing asserts that tapping Accept actually reaches `onJoinRequestAccepted` — the one wire between the new cell and `ApproveRequestsFlow.approve`. Cheap to pin given the E2E pass is admittedly unrun.

Verified locally: no remaining references to `ApproveRequestsView`/`ApproveRequestsToolbarButton`/`approve_requests.*` in sources, tests, or docs; both `ChatThreadView` call sites updated; `PendingVerificationStore` is in-memory like `localDeferrals`, so Retry can't outlive its backing state.

PR #242, opened 2026-08-11
Force fresh consent when the authority's terms change

rinat-enikeev:
## Why

A mandate binds to the exact manifest bytes it signed, and the reference Authority keeps judging it under those bytes forever (`consented_manifest`). That is the right design — republishing a manifest with a longer ban term must not silently re-term everyone who consented earlier — but it left a gap on this side: republishing was **invisible** to every install that had already consented.

Nothing in the app ever re-read the published manifest. `manifestForReview` is the only fetch, and it only runs once the user is already inside the consent flow; the stored `MandateRecord.manifestHash` was never compared against anything. So an authority could update its terms and existing iPhone/iPad installs would go on operating under terms it no longer stands behind, with no signal at all.

## What

Check the published manifest against the active mandate's hash **on app open and on every return to the foreground**, and block on the consent surface when they disagree.

**Detection — `ModerationRepository`**

New `TermsCurrency` on `ModerationState` (`unknown` / `current` / `superseded(publishedHash:)` / `authorityDelisted`), established by `refreshActiveTerms()`: resolve the active mandate's authority in the directory, fetch its manifest through the existing verified fetcher, compare against the pinned hash.

Deliberately unable to lock anyone out on a bad answer — an unreachable authority, a directory that hasn't loaded, or a manifest that fails signature verification all leave the last known currency in place. Only an authenticated manifest that hashes differently can supersede consent, so a forged or corrupted document cannot push a user into re-signing. Overlapping calls (launch racing the first foreground notification) share one check, and a mandate that changes mid-fetch is re-read before the verdict lands.

**Gate — `ModerationGateFlow`**

New blocking `RootGate.needsReconsent(ReconsentReason)`, reached only from within the `.operational` branch. It therefore ranks below every enforcement answer: a banned user re-signing terms is still banned, an unanswered gate check is still the more urgent block, and it never pre-empts `.checking` — a re-consent screen thrown up before the gate has spoken would be the first thing a cold launch shows on a flaky network. `authorityDelisted` falls back to operational when the directory lists nobody, matching the existing "don't brick an install on an empty directory" softening.

**UI**

`ModerationConsentFlow.Mode.reconsent(reason)` — mechanically identical to switching, since consent is always a fresh mandate and never an edit (spec §12), differing only in being undismissable and in naming what changed. The picker marks the current authority's row so re-signing reads as re-signing rather than switching. `RootView` moves from a `Bool` cover to `fullScreenCover(item:)` keyed on the mode, so both consent gates share one surface and it drops as soon as the gate says so.

The re-consent copy leads with what is *not* happening — the mandate is untouched, anything already under way is still judged by the agreed terms — before asking for a signature. Worth a read; it is the part most likely to want wording changes.

## Two reasons to re-consent

| | user can re-sign same authority | remedy |
|---|---|---|
| `termsChanged` | yes, and it's the expected answer | review new terms, sign |
| `authorityDelisted` | no — it can't publish terms any more | pick another authority |

## Testing

Implementation only, per the usual split — tests land in the follow-up PR. Verified: app and `OnymModerationUI` build clean for iOS; existing `GateHardeningTests` pass unchanged.

## Not in this PR

The `signature invalid: reporter signature did not verify` failure on report filing is unrelated and still open. `ModerationCanonicalEncoder` sets `.sortedKeys` but not `.withoutEscapingSlashes`, so Swift signs `"...\/..."` while the authority's `canonical_bytes` re-serializes to `"...​/..."` — any signed string containing a slash breaks, which for a base64 `authenticityProof` is roughly three reports in four. `authority/src/canonical.rs` already documents the assumption the encoder never implements. Separate branch off main.

### DIFF
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
@@ -339,12 +339,14 @@ public actor ModerationRepository {
 
     private func performTermsCheck() async {
         guard let record = activeMandateRecord() else { return }
-        if fetchStatus != .success {
-            // The directory names the operator key each manifest is
-            // pinned against, so there is no authenticated fetch
-            // without it.
-            try? await refresh()
-        }
+        // Always re-read the directory, not just when it has never
+        // loaded. It is what says the authority is still designated,
+        // and it carries the operator key the manifest below is
+        // authenticated against — a cached copy makes delisting a
+        // cold-start-only discovery and pins the check to a key that
+        // may since have rotated. Callers share one fetch, so a
+        // foreground that races the launch check costs nothing.
+        try? await refresh()
         guard fetchStatus == .success else { return }
 
         guard let listing = authorities.first(where: {
@@ -355,6 +357,13 @@ public actor ModerationRepository {
             noteAuthorityDelisted(record.mandate.authority)
             return
         }
+        // Presence in the directory is the same quality of answer as
+        // absence from it, so it retires the delisting here rather than
+        // hostage to the manifest fetch below. Otherwise a relisted
+        // authority whose manifest momentarily fails to fetch keeps
+        // showing "Authority Unavailable", whose only offered remedy is
+        // switching away from an authority that is in fact listed.
+        noteAuthorityListed(listing.componentId)
 
         let published: SignedManifest
         do {
@@ -404,6 +413,13 @@ public actor ModerationRepository {
         let before = termsCurrency
         publishedManifest = (authority: authority, hash: hash)
         if delistedAuthority == authority { delistedAuthority = nil }
+        if termsCurrency != before { publish() }
+    }
+
+    private func noteAuthorityListed(_ authority: String) {
+        guard delistedAuthority == authority else { return }
+        let before = termsCurrency
+        delistedAuthority = nil
         if termsCurrency != before { publish() }
     }
 
@@ -493,12 +509,26 @@ public actor ModerationRepository {
 
         // These bytes were fetched, authenticated, and validated moments
         // ago, so they are the freshest observation this repository has
-        // of what the authority publishes. Recording it here — before
+        // of what this authority publishes. Recording it here — before
         // any of the paths below activate a record — is what lets
         // currency stay derived: whichever record ends up active is
         // compared against this, and consenting to current terms comes
         // out `.current` without anyone having to say so.
-        notePublishedManifest(signedManifest.manifestHash, for: listing.componentId)
+        //
+        // Only for the authority the active mandate already names. The
+        // other remedy the re-consent copy offers is moving to a
+        // different authority, and an observation about *that* one
+        // makes the derived currency `.unknown` — dropping the
+        // undismissable gate while this method is still enrolling and
+        // signing. A failure after that point would return the user to
+        // the app under the stale mandate, ungated, without ever
+        // showing the signing error. The gate must stay up until a new
+        // record is actually active, and with the observation withheld
+        // it does: currency keeps reading the old record against the
+        // old authority's published hash.
+        if listing.componentId == activeMandateRecord()?.mandate.authority {
+            notePublishedManifest(signedManifest.manifestHash, for: listing.componentId)
+        }
 
         // Resolve every older ambiguous registration for this Authority
         // before minting under newly reviewed terms. A manifest rotation
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
@@ -209,9 +209,13 @@ public actor ModerationRepository {
     private var caseAppealTasks: [CaseSubmissionKey: Task<AppealReceipt, Error>] = [:]
     private var refreshTask: Task<Void, Error>?
     private var termsCheckTask: Task<Void, Never>?
-    /// Last authenticated, consentable manifest seen for an authority.
+    /// Last authenticated, consentable manifest seen for an authority,
+    /// and the authority the directory has stopped listing. Each holds
+    /// one authority, not a table: they are observations about the one
+    /// the active mandate names, and are only ever read through
+    /// `termsCurrency`, which discards them when they describe anyone
+    /// else. Reading either directly would need that check restated.
     private var publishedManifest: (authority: String, hash: String)?
-    /// Authority the directory has stopped listing, if any.
     private var delistedAuthority: String?
 
     public init(
@@ -298,7 +302,25 @@ public actor ModerationRepository {
             throw error
         }
         refreshTask = nil
+        // Every successful directory read is evidence about the active
+        // mandate's authority, wherever it was asked for. The consent
+        // flow refreshes on appear and on its retry button without ever
+        // going through the terms check, so leaving reconciliation
+        // there let a stale `publishedManifest` outlive the listing it
+        // came from — an undismissable "these terms have changed" over
+        // a picker reading "No authorities are published yet", with no
+        // remedy on screen.
+        reconcileDirectoryWithActiveMandate()
         publish()
+    }
+
+    private func reconcileDirectoryWithActiveMandate() {
+        guard let record = activeMandateRecord() else { return }
+        if authorities.contains(where: { $0.componentId == record.mandate.authority }) {
+            noteAuthorityListed(record.mandate.authority)
+        } else {
+            noteAuthorityDelisted(record.mandate.authority)
+        }
     }
 
     // MARK: - Terms currency
@@ -319,6 +341,12 @@ public actor ModerationRepository {
     /// first foreground notification) share one check.
     public func refreshActiveTerms() async {
         if let task = termsCheckTask {
+            // A foreground arriving mid-check takes that check's answer
+            // even though its fetches may predate the backgrounding. A
+            // deliberate gap: the alternative is queueing a second full
+            // round of fetches behind every one already running, and
+            // the verdict this returns is at worst one foreground
+            // stale, which the next one corrects.
             await task.value
             return
         }
@@ -349,21 +377,17 @@ public actor ModerationRepository {
         try? await refresh()
         guard fetchStatus == .success else { return }
 
+        // Listing and delisting were both settled by the refresh above,
+        // which reconciles them for every caller. Presence is the same
+        // quality of answer as absence and is recorded the moment the
+        // directory lands — never held hostage to the manifest fetch
+        // below, which would leave a relisted authority whose manifest
+        // momentarily fails to fetch showing "Authority Unavailable",
+        // offering only the remedy of switching away from an authority
+        // the directory does list.
         guard let listing = authorities.first(where: {
             $0.componentId == record.mandate.authority
-        }) else {
-            // A published directory that no longer lists this authority
-            // is a definitive answer, not a fetch failure.
-            noteAuthorityDelisted(record.mandate.authority)
-            return
-        }
-        // Presence in the directory is the same quality of answer as
-        // absence from it, so it retires the delisting here rather than
-        // hostage to the manifest fetch below. Otherwise a relisted
-        // authority whose manifest momentarily fails to fetch keeps
-        // showing "Authority Unavailable", whose only offered remedy is
-        // switching away from an authority that is in fact listed.
-        noteAuthorityListed(listing.componentId)
+        }) else { return }
 
         let published: SignedManifest
         do {
--- a/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
+++ b/Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift
@@ -11,6 +11,31 @@ public enum AuthorityFetchStatus: Equatable, Sendable {
     case failed(message: String)
 }
 
+/// Whether the active mandate still pins the manifest its authority
+/// publishes today.
+///
+/// A mandate stays valid under the bytes it signed — the reference
+/// Authority judges every case under the manifest that mandate pinned,
+/// so superseded terms are not an outage and nothing in flight breaks.
+/// What they are is stale *consent*: the user is operating under terms
+/// the authority no longer stands behind, and only fresh consent can
+/// move them onto the new ones.
+public enum TermsCurrency: Sendable, Equatable {
+    /// Not established yet — no active mandate, the directory hasn't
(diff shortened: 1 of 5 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the design and layering are right; two paths can leave a user stuck on an undismissable gate.

- `Packages/OnymModeration/Sources/OnymModeration/ModerationRepository.swift:444` — the pending-dedup path in `performConsent` returns without `setTermsCurrency(.current)`, unlike both branches at :546/:566. Scenario: an active mandate is `.superseded`, a prior re-consent attempt for the same authority/hash is still `registrationPending`; the user signs, that record activates pinning the *published* hash, but currency stays `.superseded`, so `needsReconsent` keeps a cover with no dismiss button up. Re-tapping agree re-enters the same branch; only a background→foreground round trip clears it. Set currency inside `registerPending(…, activate: true)` (covers all three call sites), or add it here.
- `ModerationRepository.swift:344` — the published manifest is authenticated but never passed through `manifestValidator.validateForConsent`. An authority that republishes an expired/unsupported manifest supersedes every install, and then `manifestForReview` rejects those very bytes (`ModerationConsentFlow.selectedAuthority` → "Couldn't load this authority's terms"), so the `.termsChanged` remedy the copy promises ("read the new terms and sign them") is unavailable — same-authority re-consent is impossible. Terms you cannot consent to shouldn't supersede consent: `guard (try? manifestValidator.validateForConsent(published, now: clock())) != nil else { return }`.
- `ModerationRepository.swift:313` — at cold launch `start()`'s `refresh()` has already set `fetchStatus = .fetching`, so the terms check fires a *second* concurrent directory fetch (a third from `ModerationConsentFlow.start()`). Whichever loses can overwrite `.success` with `.failed`, surfacing a spurious "Couldn't load the authority list." + retry on the onboarding picker. Guard on `.fetching` (or share one refresh task, as the consent/report paths already do).
- `ModerationRepository.swift:297-304` — the dedup window is one suspension too narrow: a caller arriving after `task` completes but before line 304 clears it awaits the finished task and returns a stale verdict without checking. A foreground event can be silently swallowed. Clearing inside the task body (`defer`) closes it.
- `ModerationGateFlow.swift:71` — the launch terms check is an untracked `Task`, unlike `moderationTask`/`gateTask`; `stop()` won't cancel it. Minor, but `start()` is documented idempotent and this call isn't guarded by the `moderationTask == nil` check above it, so repeated `start()` calls re-fire it.
- `Sources/OnymIOS/RootView.swift:44` — `case .switching` in `ConsentPresentation.id` is unreachable (`presentation(for:)` never yields it; Settings presents its own cover). Harmless, but it reads as if the root can host the switching flow.
- `TermsCurrency.superseded(publishedHash:)`'s payload is never consumed — the gate collapses it to `.termsChanged` and no UI shows it. It does usefully re-trigger `publish()` on a second republish, so keep it, but consider surfacing the hash on the re-consent screen next to the pinned one (`ModerationSettingsView` already shows `mandate.manifestHash`).

Re-consent copy reads well; `.termsChanged` correctly leads with what hasn't changed. One wording note: "so it can't publish terms or answer for you any more" (`ModerationConsentView.swift:171`) overstates — a delisted authority is still bound by mandates it registered; only *this interface* stops honouring it.

dev-onym (commented):
REQUEST_CHANGES — the design is sound, but recording the observation before consent commits can drop the blocking gate mid-signing, and the delisted flag never clears on a transient answer.

- `ModerationRepository.swift:501` — `notePublishedManifest` fires before enrollment, signing, countersign, and registration, and it keys the observation to the *selected* authority, not the mandate's. Superseded-terms path where the user takes the second remedy the copy offers ("or move to another authority"): active record is A, user picks B, taps agree → `publishedManifest = (B, hash)` → derived currency flips to `.unknown` (observed.authority ≠ record.authority) → `ModerationGateFlow.swift:155` recomputes `.operational` → `RootView.swift:88` sets `consentPresentation = nil` and tears the undismissable cover down while `tappedAgree`'s task is still signing. If that task then throws (enrollment/countersign/registration failure), the user is back in the app under the stale mandate with no gate and never sees "Signing failed. Try again." Record the observation only when `listing.componentId == activeMandateRecord()?.mandate.authority`, or after the new record is activated — the derived-currency comment's own logic ("whichever record ends up active") holds either way.
- `ModerationRepository.swift:355` vs `:361` — `delistedAuthority` is set when the directory omits the authority, but the only thing that clears it is a successful *and* consentable manifest fetch in `notePublishedManifest`. Once the authority is listed again, a manifest fetch that throws (or a manifest that fails `validateForConsent`) returns early and leaves the latch set, so the app keeps showing an undismissable "Authority Unavailable" whose only remedy is switching, for an authority the directory currently lists. Clear `delistedAuthority` as soon as the listing resolves at `:353`, before the fetch — presence in the directory is the same "definitive answer" quality the delisting branch claims.
- `ModerationRepository.swift:342` — the directory is only re-fetched when `fetchStatus != .success`, so after a successful launch fetch every foreground check reuses the cached list. `.authorityDelisted` is therefore effectively cold-start-only; an authority delisted during a long-lived session is never noticed (and the operator key used to authenticate the manifest is equally stale). Worth either re-fetching the directory on the terms check or scoping the PR's "on every return to the foreground" claim to `termsChanged`.
- `RootView.swift:88` — `fullScreenCover(item:)` does not reliably re-present when the item's identity changes while a cover is up, so `termsChanged` → `authorityDelisted` (authority republishes, then leaves the directory) can leave the old reason and old flow on screen despite the differing `id`. The comment asserts the rebuild happens; worth verifying on device or forcing a nil tick between presentations.
- Minor: `ModerationGateFlow.swift:88` `stop()` cancels `termsTask`, but that only cancels the wrapper — the actor's inner `termsCheckTask` keeps running; and `appForegrounded()` spawns an untracked task, which is inconsistent with the tracking rationale in the `start()` comment.
- Copy reads well. `authorityDelisted` blurb says reports "can't reach it from this app" — confirm that matches what the enforcement client actually does with a delisted authority, since it's the one factual claim in there a user could test.

dev-onym (commented):
REQUEST_CHANGES — the detection/derivation design is sound; two paths can strand a user on an undismissable screen.

- `Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationGateFlow.swift:160` — `.superseded` is the one blocking branch without the empty-directory softening that `:144`, `:166` and `:190` all apply. `publishedManifest` survives a later `refresh()` that returns an *empty* list (e.g. `ModerationConsentFlow.start()`/`tappedRetry()` at `ModerationConsentFlow.swift:106,167`, which don't go through `performTermsCheck`, so nothing calls `noteAuthorityDelisted`). Result: undismissable "These terms have changed" over `ModerationConsentView.swift:119`'s "No authorities are published yet." — zero remedy, no dismiss, no way out until the directory refills. Gate `.superseded` on `authoritiesAvailable` too, or clear `publishedManifest` whenever a successful directory fetch stops listing its authority.
- `Sources/OnymIOS/RootView.swift:100` — the deferred re-present captures `next` and applies it unconditionally on a later main-actor turn. If the gate moves again before that turn (delisted → operational, or a swap back), the task overwrites the newer state with a stale one: a `needsReconsent` cover comes back up over an operational app, undismissable (`ModerationConsentFlow.isDismissable` is `.switching`-only), until the gate happens to change again. Re-derive inside the task instead of capturing: `consentPresentation = Self.presentation(for: dependencies.moderationGateFlow.gate)`.
- `ModerationRepository.swift:320` — `refreshActiveTerms()` joins an in-flight check unconditionally, so a foreground arriving mid-check (the check's directory/manifest fetch having started *before* backgrounding) gets that pre-background verdict and never runs one of its own. The completed-task window is handled carefully at `:335`; this one isn't. Probably fine given the next foreground retries, but it undercuts "on every return to the foreground" — worth a comment at minimum.
- `ModerationGateFlow.swift:117` — `appForegrounded()` overwrites `termsTask` without cancelling the prior one, and cancelling it never reaches `runTermsCheck` (unstructured `Task` inside the actor doesn't inherit cancellation), so `stop()` doesn't stop the check. The comment concedes the second half; the dropped handle is still a silently orphaned task.
- `ModerationRepository.swift:285` — `performRefresh` nils `refreshTask` on the failure path *before* `publish()`/`throw`, so on a cold offline start `start()`, the consent flow, and the terms check each fire their own directory fetch back-to-back. Intentional per the comment, but it does mean the coalescing only helps the success path.
- Minor, `ModerationConsentView.swift:211` — `SettingsRow.subtitleLineLimit` defaults to 1, so "Your current authority · onym:component:…" will truncate away the component id it's appended to. Consider `subtitleLineLimit: 2` on that row, or putting the marker after the id.
- `notePublishedManifest`/`noteAuthorityDelisted` (`:412`, `:426`) hold a single authority each; a user who switches away leaves `delistedAuthority` pinned to the old id forever. Harmless today because currency is derived against the active record — just noting it as a latent trap if anything ever reads those directly.
- Given the tests-in-follow-up split: the two cases above (`superseded` + empty directory, gate swap while a cover is up) are the ones worth landing tests for first, alongside "forged manifest leaves currency untouched" and "mandate switches mid-fetch".

dev-onym (commented):
**REQUEST_CHANGES** — the design is sound and the fail-open reasoning holds up (the directory *is* signature-verified via `SignedAsset.verify`, so `authorityDelisted` is an authenticated signal too), but the withheld-observation trick leaves one path that raises a bogus undismissable gate.

- **`ModerationRepository.swift:553` — stale observation survives a switch away and then fires a false `.superseded` on switching back.** `publishedManifest` is only overwritten for the currently-active authority, but it is never *invalidated* when the active record moves elsewhere. Sequence: on A, a terms check records `(A, H1)`; user switches to B (note withheld, `(A, H1)` retained, currency `.unknown`, gate drops); A republishes `H2`; in the same session the user switches back to A and signs `H2`. The note is withheld again (active is B), then the new A/`H2` record activates and the derived currency compares `H2` against the retained `H1` → `.superseded(publishedHash: H1)`. The user lands on an undismissable "Terms Updated" gate immediately after signing, with a stale hash presented as "Published now". Reachable through both activation paths (`:667` and `:681`). Fix keeps the property you wanted: record the observation *at* the activation points rather than gating it on the old active authority — the gate still stays up until a record is actually active, and no observation outlives the record it describes.
- **`Sources/OnymIOS/RootView.swift:96` — the nil-then-reassign is unlikely to fix what the comment says it fixes.** One `Task { @MainActor }` hop lands well inside `fullScreenCover`'s dismiss animation, which is the same window that swallows re-presentation; the likely outcome for a reason swap is *no* cover, leaving the app ungated — strictly worse than the stale reason. Simpler and dependable: keep the item non-nil and put `.id(presentation.id)` on `ModerationConsentView` so the flow is rebuilt in place, no dismiss/represent at all.
- **`ModerationRepository.swift:378` — `guard fetchStatus == .success` reads shared state across an await.** Since `refresh()` now coalesces, a second caller can start a fresh `performRefresh` (setting `.fetching`) between `refreshTask = nil` and this awaiter resuming, silently skipping the whole check for that foreground. Have `performRefresh` return the outcome (or the reconciled listing) and branch on that value instead of on actor state.
- **`ModerationRepository.swift:369` — a launch whose active mandate arrives via `retryNewestPendingRegistration` gets no terms check at all.** `performTermsCheck` returns immediately on `activeMandateRecord() == nil`, and nothing re-triggers when a pending record later activates; those installs wait until the next foreground. A cheap fix is re-running the check when a registration resolves into an active record.
- **`ModerationGateFlow.swift:112` — every foreground now forces a directory fetch, which publishes `fetchStatus = .fetching`.** Harmless for the gate, but any consent/settings surface that is open at the time will show its loading/failure affordance flicker on each resume; worth confirming that's acceptable rather than gating the refresh on staleness.
- Drive-by (pre-existing, but you're rewriting this block): the cover is driven purely by `onChange(of: gate)`, so a `RootView` that appears when the gate is *already* `.needsConsent`/`.needsReconsent` never presents — seeding `consentPresentation` from `presentation(for:)` in `.task` closes it.
- Copy reads well and the reason/remedy split is right; `WHAT CHANGED` showing both hashes is a good call. `interactiveDismissDisabled(!isDismissable)` and the `.operational`-only ranking both look correct.

dev-onym (commented):
**COMMENT** — the derivation and fail-soft discipline on the manifest path are solid; the delisting path is the one place that doesn't match it.

- `ModerationGateFlow.swift:170` / `ModerationRepository.swift:358` — **delisting is fail-hard while everything else fails soft.** One successful directory read that omits the active authority immediately raises an undismissable gate whose only exit is signing a mandate with someone else. The manifest path demands authenticated bytes and returns on any doubt; delisting needs only a 200. And the directory is soft-verified (`KnownAuthoritiesFetcher.swift:136`) and lossy-decoded — `LossyAuthorityListing` silently `compactMap`s away entries with one bad field (`:86`), so a malformed listing in a release asset delists every install of that authority. The empty-directory softening can't catch this: decoding throws on an empty list (`:87`), so `authoritiesAvailable` is always true on a successful fetch. Suggest requiring N consecutive delisted observations (or a first-seen-delisted timestamp plus grace) before `.authorityDelisted` is allowed to gate.
- `ModerationRepository.swift:453` — **an in-flight check clobbers a fresher observation.** `performTermsCheck` binds `record` at entry, then awaits two network hops; `notePublishedManifest` at the end writes unconditionally. If the user consents to authority B meanwhile, the launch check for A overwrites the `(B, hashB)` that `activate()` just recorded, and derived currency drops to `.unknown` until the next foreground. The verdict is re-derived against the current record as the PR describes, but the *observation* isn't guarded. Cheap fix: re-read `activeMandateRecord()` before recording and drop the result if the authority no longer matches.
- `ModerationRepository.swift:418` — `refreshedAuthorities()` may hand back a fetch that started long before this check. On a slow network a foreground check joins the launch fetch and authenticates the manifest against that older directory's operator key — the exact staleness the "always re-read the directory" comment at `:411` argues against. Same acceptable-gap class as the coalesced terms check, but the comment currently claims a stronger property than the code gives.
- `ModerationGateFlow.swift:112` — only launch and foreground trigger a check, so a session that stays foregrounded for days never re-reads. Fine if deliberate; worth a line in the doc comment saying so.
- `ModerationConsentFlow.swift:64` — `isDismissable` is `mode == .switching`, so `.done` in reconsent mode has no exit but the gate. That holds today (`activate()` → `.current` → `.operational`), but it's an invariant with no test; make sure the follow-up PR pins it, along with the `.checking`/`banned` ranking in `recompute`.
- Nit `ModerationConsentView.swift:220` — `String(localized:) + " · " + componentId` bakes the separator into the layout; a single format string (`"%@ · %@"`) localizes better.

PR #244, opened 2026-08-13
Add OnymDiscovery package: models, trust, store, repository

rinat-enikeev:
## Summary

PR 1 of the discovery-driven Settings sequence (contract: onymchat/onym-system#28, reference impl: [onym-discovery](https://github.com/onymchat/onym-discovery)). A new local package `Packages/OnymDiscovery` — models, verification, persistence, and repository for the static-snapshot/Ed25519 Discovery profile. **No UI, no app wiring yet**; the app builds unchanged.

- Strict top-level decoding of `DiscoveryProviderManifest` / `CatalogSnapshot`, lossy per-entry decoding (malformed entries skipped and counted, never defaulted).
- `DiscoveryCanonical`: structural signature removal + `.sortedKeys`/`.withoutEscapingSlashes` — byte-compatible with the Rust reference, proven by the `canonical-input.json → canonical-bytes.bin` fixture test.
- `DiscoveryTrust` (hard-enforced, no soft mode): manifest self-signature with TOFU key return for pinning, snapshot chains (sequence +1, `previousDigest` over exact bytes, rollback/gap/fork rejection), ≤90-day expiry, size/entry caps, §7 URI rules, destination digest binding. Errors mirror the profile's §9 codes. Ed25519 via OnymFoundation's existing `Ed25519DetachedSignatureVerifier` with injected per-source keys — the legacy soft `ContractsTrust` path is untouched.
- `DiscoverySourcesConfiguration` with `removedDefaultProviderIds` (a removed default source never silently returns) and RelayerConfiguration-style backward-compat decoding.
- `actor DiscoveryRepository` in the established repository shape (AsyncStream snapshots, per-source failure tolerance, TOFU add→confirm, durable remove).

## Testing

27/27 unit tests via `xcodebuild test -scheme OnymDiscovery` (iPhone 17 Pro simulator), zero warnings. Fixtures are byte-copied from onym-discovery's conformance suite — drift on either side fails a build. Notes: standalone `swift test` is impossible (OnymFoundation is iOS-only); redirect tightening (≤3, HTTPS→HTTPS) deferred to the wiring PR with a code note.

Next in sequence: PR 2 (generic consent + offers primitives in OnymFoundation), PR 3 (composite fetchers + boot wiring), PR 4–6 (Settings IA, consent flow, offers UI), PR 7 (fakes/UI tests).


dev-onym (commented):
REQUEST_CHANGES — the snapshot chain rule rejects both the steady state (unchanged snapshot) and the normal onboarding case (provider past sequence 1), so hard-enforced trust would brick discovery in production.

- **`Sources/OnymDiscovery/DiscoveryTrust.swift:259` — an unchanged snapshot is rejected on every refresh.** Once snapshot N is retained, the next refresh re-fetches the same bytes and `snapshot.sequence == previous.sequence + 1` fails with "rollback or gap". Any refresh where the provider hasn't published since the last one permanently surfaces a trust error and `.failed` status. Needs a digest-equality short circuit (identical retained bytes → accept as a no-op) before the +1 rule. `testChainAdvancesAcrossRefreshesAndRollbackIsSurfaced` only ever serves 1→2, so it misses this.
- **`Sources/OnymDiscovery/DiscoveryTrust.swift:274` — first observed snapshot is forced to `sequence == 1`.** `CatalogDescriptor` carries a single `snapshot` URL (the current one) with no archive mechanism, so a client adding a provider whose catalog is at sequence 42 can never accept anything. TOFU should adopt whatever sequence is first observed and enforce +1 from there. Please confirm against the reference impl before landing — if the profile really mandates sequence 1, the default source becomes unusable after its second publication.
- **`Sources/OnymDiscovery/DiscoveryRepository.swift:242,252` — partial catalog failure loses accepted records.** `withLastAccepted` accumulates into a local `updated`, but `replaceSource(updated)` only runs after the whole catalog loop; a throw on catalog 2 discards catalog 1's record even though its bytes were already persisted at :237. Those entries then never appear in `rebuildAggregateFromRetained` (it iterates `lastAccepted`), and the orphaned bytes still act as `previousRaw` on the next refresh. Persist per catalog inside the loop, or catch per catalog.
- **`Sources/OnymDiscovery/DiscoveryRepository.swift:306` — `removeSource` deletes only `lastAccepted` catalogs.** Any orphan from the above (or a catalog dropped from the manifest) leaks in UserDefaults and, on a re-add, is loaded as `previousRaw`, breaking the chain for a source the user believes is fresh. Prefer removing by key prefix.
- **`Sources/OnymDiscovery/DiscoveryFetcher.swift:47` — the size cap is enforced after the body is fully buffered.** `session.data(from:)` reads unbounded bytes into memory before line 53 checks the cap, so a hostile host defeats the §7 bound. Use a byte-counting delegate / `bytes(from:)` with early abort.
- **`Sources/OnymDiscovery/DiscoveryStore.swift:61` — up to 1 MiB per (provider, catalog) in UserDefaults.** The sibling stores keep small config there; multi-MB snapshot blobs bloat the plist and get faulted in at launch. A file in Application Support fits better.
- **`Tests/OnymDiscoveryTests/FixtureTestSupport.swift:16` — missing fixtures `XCTSkip` instead of failing.** If `.copy("Fixtures")` ever stops landing in the bundle, the entire cross-language conformance suite passes vacuously — the opposite of the "drift fails a build" claim. Use `XCTFail`/throw.
- **Not wired into the Xcode project or CI.** `OnymDiscovery` appears in no `.pbxproj` or workflow, so the 27 tests only ran locally and will never gate a PR. Either add the scheme now or note the wiring PR explicitly.
- `Sources/OnymDiscovery/DiscoveryCanonical.swift:19` — the caveat understates the divergence: `.sortedKeys` also sorts `.numeric`/`.widthInsensitive`, not just case-insensitively. No schema key is attacker-controlled today, so it's a comment fix, but the "byte-compatible with Rust" claim rests on the schema, not on the serializer.
- Minor: duplicate-`componentId` detection (`DiscoveryTrust.swift:230`-ish) runs over surviving entries only, so a duplicate whose twin was lossily skipped passes; `policyUnavailable`/`queryUnsupported`/`sourceConflict`/`rateLimited` are declared but never thrown.

dev-onym (commented):
**REQUEST_CHANGES** — solid, well-tested trust core, but the snapshot fetch isn't bound to the catalog it was fetched for, and refresh has a re-entrancy hole in chain state.

- `DiscoveryTrust.swift:205` / `DiscoveryRepository.swift:271` — `verifySnapshot` resolves the descriptor from the *document's* `catalogId`, so a snapshot for catalog B served at catalog A's `snapshot` URL verifies fully (B's policy digest matches B's descriptor) and is then persisted and attributed under A (`DiscoveryRepository.swift:280`, `:288`). Effects: B's entries render as A's, and A's retained sequence/digest are clobbered — A's next genuine snapshot then fails as a rollback, wedging the catalog permanently. Fix: pass the expected `catalogId` into `verifySnapshot` (or assert `accepted.snapshot.catalogId == catalog.catalogId` before saving).
- `DiscoveryRepository.swift:255-308` — `updated` is a copy taken before any `await`; `replaceSource(updated)` writes the whole source back. `refresh()` is public and re-entrant (and `start()` fires one concurrently), so two overlapping refreshes let the loser's stale `lastAccepted` overwrite the winner's — a regressed sequence makes a subsequent rollback acceptable. Guard `refresh()` with an in-flight task like `start()` does, or merge per-catalog records into the current `configuration` instead of replacing a stale snapshot.
- `DiscoveryFetcher.swift:73` — `for try await byte in bytes` iterates the 1 MiB snapshot cap one byte per async iteration; this is orders of magnitude slower than `bytes.lines`-style chunking. Accumulate over `AsyncBytes`' underlying chunks (e.g. wrap in a buffered read) while keeping the mid-stream cap.
- `DiscoveryTrust.swift:131` — manifest `validUntil` is compared to `now` with no skew allowance, while snapshot expiry gets the symmetric 10-minute `clockSkewSeconds`. A device 30 s fast rejects a just-expiring manifest and takes the whole source down; apply `clockSkewSeconds` here too (or state why the asymmetry is intended).
- `DiscoveryModels.swift:365` — entry `componentId` and `operator` are decoded as free strings even though `DiscoveryFormat.isComponentId` / `operatorKeyHex` exist and every sibling field is format-checked. An entry with a junk `componentId` survives, becomes the dedup key and the `AttributedCatalogEntry.id`.
- `DiscoveryCanonical.swift:31` vs `DiscoveryTrust.swift:189` — signing bytes come from `JSONSerialization`, the decoded values from `JSONDecoder`. Two parsers over the same bytes means duplicate-key (and numeric) resolution differences are a signature-substitution surface. Worth either canonicalizing *from* the decoded tree or rejecting documents whose top-level key count differs from the deserialized dictionary's.
- `DiscoveryStore.swift:110` — the type doc claims `providerId`/`catalogId` are "validated to `[a-z0-9-:]` / `[a-z0-9-]` shapes", but no validation exists here; `providerId` is interpolated straight into a path component and `removeRetainedSnapshots` does a recursive `removeItem`. Values happen to be trust-validated upstream, but `removeSource(providerId:)` is public — add the assertion the comment promises.
- `DiscoveryRepository.swift:217` — when every source is unpinned (the seeded default's normal state), the loop skips all of them and status still lands on `.success` with an empty aggregate and no per-source note. Users get a silent empty list; surface a "needs confirmation" state.
- `DiscoveryRepository.swift:307` — `lastNotes` is only assigned after the catalog loop, so a manifest-stage failure leaves the previous refresh's notes displayed next to the new error. Clear notes at the top of `refreshSource`.
- `.github/workflows/release.yml:195` — the new step writes `/tmp/onym-discovery.xcresult` but nothing uploads it, so `report-to-issue` can't parse discovery failures. Add a matching upload step.

Fixtures, canonical-serializer vectors, and the chain/expiry tests are genuinely good — the byte-pinned `canonical-*` vectors and the born-expired/skew-boundary cases are the right things to pin.

dev-onym (commented):
**REQUEST_CHANGES** — excellent trust/canonicalization work; two real defects in `DiscoveryRepository` plus a §7 gap on the entry URL.

**Must fix**

- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryRepository.swift:402` — `confirmAddSource` mutates configuration but never calls `rebuildAggregateFromRetained()` / `publish()`. Every other mutator (`removeSource:441`, `setEnabled:450`, `start:159`) does. Subscribers on `snapshots` never see the newly added source until an unrelated refresh lands — the add-source flow in PR 4–6 will silently show nothing.
- `DiscoveryRepository.swift:198` — refresh coalescing is correct against `start()`, but wrong for the add-then-refresh sequence: `performRefresh` snapshots `configuration.sources` at entry, so `confirmAddSource(...)` followed by `refresh()` while a pass is in flight coalesces onto a pass that never saw the new source, then returns `.success`. Either re-check for unrefreshed sources after `await task.value`, or queue a follow-up pass rather than joining the in-flight one.
- `DiscoveryRepository.swift:387` — `addSource(manifestURL:)` hands the URL straight to the fetcher with no `DiscoveryFormat.isValidURI` check. Catalog/entry URIs are §7-validated at decode, but the user-supplied provider manifest URL is not: `http://…`, `https://192.168.1.1/manifest.json`, or a `?`/userinfo URL is accepted. `CappedFetchHandler` only applies the IP-literal/HTTPS rule to *redirect* targets (`DiscoveryFetcher.swift:109`), so nothing catches the initial hop. Same applies to `manifestURL` rehydrated from persisted config.

**Worth addressing**

- `DiscoveryFetcher.swift:80` — the comment claims `body` is "touched only on the session's serial delegate queue", but `reserveCapacity` at :137 runs in the *async* `didReceive response` variant, which is not the delegate queue. `redirects += 1` at :109 is likewise unsynchronized. URLSession serializes callback *ordering*, so this is unlikely to misbehave in practice, but the comment is wrong and the class is `@unchecked Sendable` on that premise.
- `DiscoveryModels.swift:366` — `seatType` gets no format check (unlike `componentId`/`operatorKey`), and no stage cross-checks an entry's `seatType` against the declaring descriptor's `seatTypes` (§4.1). A catalog declaring `["notary"]` can list `transport.message` entries and they flow into the aggregate.
- `DiscoveryRepository.swift:301` — a catalog whose `snapshotURL` fails to parse is `continue`d with no note and no error; unreachable today given decode-time URI validation, but it's the one silent-skip path in an otherwise consistently surfaced design.

**Non-blocking**

- `.github/workflows/release.yml:195` — the package test step has no `if: always()` guard around ordering; a unit-test failure short-circuits it, so the discovery suite is never gated on red unit runs. Consider `if: always()` (uploads already have it).
- `confirmAddSource` on a re-add resets `lastAccepted` to `[:]` without purging retained blobs, leaving orphans until the next `removeSource`. Chain state reset on explicit re-add is documented and intended; the orphaned bytes aren't.

Everything else checks out: the hand-rolled §3 serializer (escape-case ordering, UTF-8 byte sort, float rejection), the single-parse duplicate-key hardening, the `expectedCatalogId` substitution binding, symmetric skew on both `validUntil` and `expiresAt`, and the fixture-date arithmetic in the tests are all correct as written.

rinat-enikeev:
Round 3 addressed in 77852ca + febf61e (merged up through discovery-wiring 9cfa973 and discovery-settings-ui 4d02e5d). One process note first, then point by point.

**Backport note — why two of your findings were "already fixed" somewhere you couldn't see.** Later review rounds on PR #247 fixed repository behavior *at that branch* while this base branch stayed behind — notably `confirmAddSource` carrying `lastAccepted`/`userLabel` forward on in-place re-confirm, and a targeted `refreshSource(providerId:)`. You were reading `discovery-package`, which had neither. 77852ca backports the package-level pieces (`DiscoveryRepository` in-place confirm + targeted refresh, `DiscoveryUserMessage`, the integrity flag on `DiscoverySourceStatus`, and their tests) so this PR now stands alone; those hunks merged up-stack as no-ops.

**Must fix**

1. **`confirmAddSource` never publishes — real defect, nice catch.** This one was broken on *every* branch, including up-stack: subscribers only saw the new source after an unrelated refresh. It now ends with `rebuildAggregateFromRetained()` + `publish()` like every other mutator. `testConfirmAddSourcePublishesToSnapshotSubscribers` verifies through the `snapshots` stream: replay, confirm, next yielded state carries the source.

2. **Add-then-refresh coalescing — the sharpest catch of the round.** You're right that the up-stack `refreshSource(providerId:)` (now backported) only covers the UI path; `refresh()` itself still lied. Fixed at the repository level: `performRefresh` records the providerIds it attempted (`lastPassAttemptedIds`), and `refresh()` loops after joining a pass — any enabled+pinned source the finished pass never attempted triggers a follow-up full pass. `testConfirmThenRefreshDuringInFlightPassRunsFollowUpPass` runs exactly your scenario: gated fetch holds a pass, confirm mid-pass, plain `refresh()` → the new source is fetched and `.success` is only reported once true.

3. **§7 on the manifest URL.** `addSource` now validates with `DiscoveryFormat.isValidURI` *before any network touch*, and the rehydrated configuration is validated too: a persisted source with a rule-violating URL surfaces its error state at `start()` and is excluded from fetching (`performRefresh` / `refreshSource`) and from the offline aggregate — an error card, never silent use. Tests: `testAddSourceRejectsProfileURIRuleViolations` (http / IP literal / query / fragment / userinfo / port, plus "fetcher never called") and `testHostilePersistedManifestURLIsSurfacedNeverUsed` (hostile persisted config: error surfaced, nothing fetched, retained blobs kept out of the aggregate).

**Worth addressing**

4. **`CappedFetchHandler` comment vs. reality.** Made the premise true rather than the comment weaker: `body` and `redirects` are now guarded by the existing lock (the async delegate variants resume on the concurrency executor, not the delegate queue), and the comment now says exactly that. The `@unchecked Sendable` claim rests on the lock, not on queue folklore.

5. **`seatType`.** Two layers, as suggested: entry decode format-checks `[a-z0-9.-]{1,64}` (and rejects a literal `"*"` — wildcard is descriptor-only), and `verifySnapshot` cross-checks every entry against the declaring descriptor's `seatTypes` honoring `"*"` — violations skipped + counted into `skippedEntryCount`, which feeds the existing §9 note. One subtlety your example exposed: the offline aggregate is rebuilt from the retained *bytes*, which still contain the violating entries — so the acceptance record now retains the descriptor's `seatTypes` and `rebuildAggregateFromRetained` applies the same binding. Tests: `testEntrySeatTypeFormatIsCheckedAtDecode`, `testEntrySeatTypeMustBeAdmittedByDeclaringDescriptor` (violating / wildcard / exact match), and `testSeatTypeViolationsStayOutOfAggregateAndOfflineRebuild`.

6. **Silent `snapshotURL` skip.** Now surfaces a per-source note ("its snapshot URL couldn't be parsed, so the catalog was skipped"). Agreed it's unreachable today given decode-time URI validation — which is also why there's no test: a §7-valid string that `URL(string:)` refuses doesn't exist to construct. Defense in depth, consistently surfaced.

**Non-blocking**

7. **`release.yml`** — the OnymDiscovery test step now has `if: always()`, matching its xcresult uploads.

8. **Re-add orphans.** Post-backport, re-confirm carries chain state forward, so "reset" only happens when it should: a fresh add, or a re-confirm under a *rotated operator key* — which now also resets `lastAccepted` instead of chaining the new operator onto the old identity's sequence (rotation is "re-add as new" per the pinning rules). Both paths purge the providerId's retained blobs. Tests: `testRotationReconfirmResetsChainAndPurgesRetainedBlobs`, `testFreshConfirmPurgesOrphanedRetainedBlobs`.

**Verification.** `discovery-package`: OnymDiscovery 84/84, app builds. `discovery-wiring` (merge 9cfa973): app builds, `DiscoverySeatAdaptersTests` green. `discovery-settings-ui` (merge 4d02e5d): OnymDiscovery 111/111, app builds. The backported hunks merged up-stack cleanly; the only real conflict was `confirmAddSource`'s `lastAccepted` line, resolved to the new key-aware carry-forward (a strict refinement of the settings-ui behavior).

dev-onym (commented):
COMMENT — solid, well-reasoned package; five things worth addressing, one before PR 3 wires this in.

- **Crash-on-launch from persisted state** — `DiscoveryStore.swift:118`: `assertPathSafe` uses `precondition`, justified by "ids reach this store only from documents `DiscoveryTrust` already format-checked". That isn't true of rehydrated config: `DiscoverySource.init(from:)` (`DiscoverySource.swift:125`) decodes `providerId` with no format check, and `start()` → `rebuildAggregateFromRetained` → `store.loadRetainedSnapshot(...)` (`DiscoveryRepository.swift:674`) passes it straight through. A corrupted/hostile defaults blob with `providerId: "../x"` plus one `lastAccepted` entry is an unrecoverable launch crash loop. This is exactly the hostile-persistence vector `surfaceInvalidPersistedSources` already defends for URLs — either validate `providerId`/`catalogId` on decode (drop the source, surface it), or have the store skip-and-log rather than trap.
- **§7 redirect policy is entirely untested** — `DiscoveryFetcher.swift:112`: no test covers the ≤3 cap, HTTPS→HTTPS, or IP-literal refusal, and the doc-comment's claimed outcome ("completes with the 3xx status and fails as `badStatus`") depends on whether `didReceive response:` fires for a refused redirect — which nothing here verifies. `StubURLProtocol` can serve a 302 chain cheaply; please pin the actual outcome rather than the assumed one.
- **Signature verified last in `verifySnapshot`** — `DiscoveryTrust.swift:368`: chain, policy, and expiry all evaluate before the signature. An off-path attacker replaying old bytes gets `snapshotInvalid(reason: "rollback")`, which `DiscoveryUserMessage.isIntegrityFailure` badges as tampering *by the provider*. Verifying the signature first makes unauthenticated junk a generic rejection and keeps the integrity badge meaning what it says.
- **Empty-by-policy inconsistency** — `DiscoveryTrust.swift:121`: a manifest publishing a literal `"catalogs": []` (nothing skipped, nothing malformed) is fatal, while the audience-skip path deliberately treats a zero-public-catalog manifest as valid. Same end state, opposite verdicts; worth aligning or noting why.
- **CI: new suite is gated but invisible in the report** — `discover_tests.py:108-109` only globs `Tests/OnymIOSTests` / `Tests/OnymIOSUITests`, so `parse_xcresult.py --target OnymDiscoveryTests` will emit results with no matching checkbox in the issue body. The `xcodebuild test` step does fail the job, so coverage is real — but `report-to-issue` will silently drop these classes.
- Nit: `DiscoveryRepository.swift:747` `subscriberCount` is documented as a test seam for the continuation-leak check, but no test reads it — the leak it guards is unverified.

PR #245, opened 2026-08-13
Add generic service-manifest consent and offers primitives to OnymFoundation

rinat-enikeev:
## Summary

PR 2 of the discovery-driven Settings sequence (independent of #244 — branched off main; later wiring PRs consume both). Generic primitives that let any seat module (courier, notary, authority) get the moderation seat's review-then-pin consent treatment:

- `SignedServiceManifest` — tolerant common-spine parse of any operator manifest (componentId / seat / operator key / lossy offers / validUntil), exact raw bytes + `sha256:` digest retained; hard-enforced Ed25519 self-signature over canonical bytes (structural signature removal + sortedKeys, byte-compatible with the moderation and discovery canonicalizers, no cross-package import).
- `ReviewedServiceManifest` — no-public-init token mintable only by `ServiceManifestReviewer.review(raw:expectedDigest:now:)`; what was reviewed is what gets pinned, enforced at the type level.
- `PinnedConsentRecord` + `PinnedConsentStore` — local pinned acceptance {componentId, seatType, manifestHash, manifestBytes, sourceLabel, offerId, acceptedAt, isActive}; re-consent deactivates the previous record, history kept (MandateRecord shape). This is the "local pinned acceptance" consent model for Transport/Notary; moderation's full mandate flow stays layered on top, untouched.
- `ServiceOffer` + `EntitlementProviding`/`FreeTierEntitlementProvider` — offers surfaced from manifests; free tiers grantable now, the provider protocol is the seam StoreKit will implement later.

## Testing

First test target for OnymFoundation (26/26 green on iPhone 17 Pro simulator; manifests generated and signed in-test with CryptoKit). Downstream verified by building OnymModeration against the modified package. Library target layout untouched — new files only.


dev-onym (commented):
**REQUEST_CHANGES** — the canonicalizer is built on the one API this codebase documents twice as unusable for signing bytes.

- `ServiceManifest.swift:139` — signing bytes are produced with `JSONSerialization`'s `.sortedKeys`, which sorts *case-insensitively and numerically*, not by UTF-8 byte order. `ModerationCanonicalEncoder.swift:8-13` says outright "signing bytes must never be produced through `JSONSerialization`", and `Report.swift:63-73` explains why. So the PR's "byte-compatible with the moderation and discovery canonicalizers" claim does not hold: a manifest with keys `seatID`/`seatable` (byte order: `seatID` first; case-insensitive: `seatable` first) or `relay2`/`relay10` (numeric-aware ordering flips them) canonicalizes differently here than under serde_json / Go / Python `sort_keys`. Every such manifest gets rejected with `signatureInvalid` against a correctly-signed document.
- `ServiceManifest.swift:139` — the repo already has the correct pattern for exactly this shape: `DeviceRecovery.swift:91-97` parses raw JSON, strips `signature` structurally, then re-encodes through `ModerationCanonicalEncoder` via the `RecoveryCanonicalJSON` dynamic-key `Encodable` bridge (`DeviceRecovery.swift:104-152`). Port that bridge into OnymFoundation and route `signingBytes` through `JSONEncoder`. It also fixes the secondary hazard: `JSONSerialization` re-serializes parsed `NSNumber`s with no control over int-vs-double form, so `1.0` or an int above 2^53 can round-trip to different bytes than the operator signed.
- `Tests/.../ServiceManifestTestSupport.swift:55` — the factory signs via the same `ServiceManifestCanonical.signingBytes`, so 26/26 green proves self-consistency only, never cross-language agreement. Add a fixture signed by the reference implementation, or at minimum a manifest with `seatID`/`seatable`-style keys asserting the bytes match `JSONEncoder`+`.sortedKeys` output — the moderation side pins this in `ModerationCanonicalOrderTests`.
- `PinnedConsentStore.swift:20` — `accept` is a non-atomic load/modify/save on a `Sendable` protocol with no serialization. Two concurrent accepts both read the same array and the second `save` drops the first record, or leaves two records active for one componentId; `activeRecord` (`:44`) then silently picks `last`. Moderation avoids this by keeping the whole flow inside `ModerationRepository`'s actor with a `consentTasks` dedup (`ModerationRepository.swift:520-545`). Since this is the generic primitive every seat will call, it needs an actor or a documented "callers must serialize" contract.
- `ServiceOffer.swift:51` — same `JSONSerialization` re-encode for `serviceData`, described as a "faithful carrier". It isn't byte-faithful (key order, number form). Not signature-relevant today, but a seat adapter that ever hashes it inherits the trap; worth saying "re-encoded, never hash this" in the doc comment.
- `ReviewedServiceManifest.swift:58` — a malformed `expectedDigest` throws `digestMismatch(expected:actual:)`, conflating "caller passed garbage" with "bytes were swapped"; an uppercase-hex digest from a catalog is also rejected rather than normalized. A distinct case (or lowercasing before compare) makes the failure diagnosable. `testMalformedExpectedDigestIsRejected` only asserts *that* it throws, so it wouldn't catch a change here.
- `PinnedConsentStore.swift:36` — re-consent appends full `manifestBytes` forever with no retention bound, all in UserDefaults. Moderation caps its analogous ledger (`maxResolvedReportRecords = 128`).
- `Entitlement.swift:47` — the consent-store-backed provider does a full defaults read plus JSON decode of every record on each entitlement check. Fine at current scale; worth a cached lookup once a UI polls it.
- Nit: `ServiceManifestFormat.decodeBase64AcceptingUnpadded` has no test covering the unpadded branch.

The type-level discipline (`ReviewedServiceManifest` with no public init, exact-bytes retention, lossy offers, hard-enforced expiry) is right and mirrors the moderation precedent faithfully — the canonicalizer is the one thing blocking.

dev-onym (commented):
REQUEST_CHANGES — solid design and unusually well-tested primitives, but the tests never run in CI and the "tolerant" parser hard-rejects a plausible class of real manifests.

- `Package.swift:12` — CI only runs `-only-testing:OnymIOSTests` against the Xcode project (`.github/workflows/release.yml:187`); nothing invokes `swift test` on `Packages/OnymFoundation`. All 26 new tests are green locally and dead in CI. Either add a `swift test` step or wire the target into the OnymIOS scheme, otherwise this suite silently rots.
- `ServiceManifest.swift:190` — any non-integer number *anywhere* in the document fails the whole manifest as `manifestInvalid`. The type's contract is "seats own their schemas, unknown fields preserved untouched" — but a courier publishing `{"offers":[{"service":{"pricePerGB":0.5}}]}` becomes unreviewable, and the error says nothing about which field. Offers are explicitly the commercial-terms carrier, so decimals there are likely, not exotic. Same for `ServiceManifest.swift:203` on any value JSONSerialization returns that isn't in the switch. If §3 really pins integers-only, this needs to be stated in `ServiceOffer`'s docs and ideally caught with a distinct error case.
- `ServiceManifestCanonicalTests.swift:118` — the cross-language fixture is all-lowercase keys, no control chars, no non-ASCII, no mixed-case. It proves nothing about the exact vectors the hand-rolled encoder exists for; those are only checked against hand-written Swift string literals (`:33`, `:44`, `:88`), i.e. self-consistency. If the Rust reference is reachable, add fixtures with `seatID`/`seatable`, a `\u0001`, and a non-ASCII string; otherwise the "Change one, change both" invariant has no enforcement.
- `ServiceManifest.swift:131` — the doc anchors correctness on `DiscoveryCanonical` in `OnymDiscovery`, which doesn't exist in this repo (no such package, no such symbol). Reads as a present-tense fact about a twin that isn't there yet; reword to point at the spec until #244 lands.
- `PinnedConsentStore.swift:110` — a decode failure returns `[]`, and the next `accept` saves a fresh array over the key, permanently destroying every prior consent record and its history. Silent, unrecoverable, and exactly the case the immutable-artifact design is trying to protect. Distinguish "absent" from "corrupt" and refuse to overwrite on corrupt.
- `PinnedConsentStore.swift:116` — `save` swallows encode failure and returns, while `accept` still returns a record with `isActive: true`. The caller has no way to learn consent wasn't persisted.
- `PinnedConsentStore.swift:20` — the doc claims atomicity "even through different store instances over the same defaults". True in-process only; an `NSLock` gives nothing across the app + share extension over a shared suite. Worth narrowing the claim.
- `ReviewedServiceManifest.swift:90` — `validUntil` is enforced at review and never again. A pinned record whose manifest expired stays `activeRecord` forever and `FreeTierEntitlementProvider` keeps granting; `ServiceEntitlement.expiresAt` is always `nil`. Intentional for pinned consent, but there's no API for a caller to ask "is this pin still fresh" — consider surfacing it on the record.
- `ReviewedServiceManifest.swift:53` — signature is self-referential (manifest's own `operator` key), and `expectedDigest` defaults to `nil`, so the default call is pure TOFU. Docs say key-binding is the caller's protocol, but nothing in the API makes that visible at the call site. An optional `expectedOperatorKey:` param would make the unbound case an explicit choice.
- `ServiceOffer.swift:32` — `model == "free"` is case- and whitespace-sensitive; `"Free"` gates as paid with no diagnostic. Cheap to normalize.
- `ServiceManifestCanonicalTests.swift:130` — `expectedDigest: ServiceManifestFormat.sha256Digest(of: raw)` computes the pin from the same bytes, so the digest check is vacuous there; hardcode the expected digest.
- Nits: `ServiceManifest.swift` `date(fromRFC3339:)` allocates two `ISO8601DateFormatter`s per call, on the parse path for every manifest; `digestMismatch(expected:)` reports the lowercased pin rather than what the caller passed, so a log won't show the original.

rinat-enikeev:
All addressed in 3b4351c (branch rebased onto main now that #244 landed). Point by point:

**CI never ran the suite — the sharpest catch of the round.** You're right that all of these tests were green locally and dead in CI: `release.yml` only ran `-only-testing:OnymIOSTests` against the Xcode project, and a SwiftPM package target never rides along. There's now an `xcodebuild test (OnymFoundation package)` step in the unit-tests job mirroring the OnymDiscovery one — `always()`-guarded so a red unit step can't short-circuit it, xcresult uploaded as `foundation-tests-xcresult`, and `report-to-issue` parses it as target `OnymFoundationTests` alongside the other bundles. The suite (51 tests after this round) now gates the release pipeline.

**Non-integer numbers — keeping the rejection, on spec grounds, but making it diagnosable.** I looked hard at loosening this for offers, and it would be wrong: the §3 canonical form exists so signing bytes re-serialize byte-stably, and a float genuinely breaks that — there is no pinned decimal form, so `0.5` parsed and re-emitted is whatever the serializer picks, and cross-language byte agreement (the entire point of the canonicalizer) is gone. serde_json, the reference, would face the same ambiguity; §3 pins integers precisely to close it. So v1 signed manifests are integer-only *everywhere in the signed document*, offers included — `pricePerGB: 0.5` must be published as a string (`"0.5"`) or a scaled integer. What was wrong was the diagnostics: it's now a distinct `nonIntegerNumber(path:)` naming the offending field (`offers[0].service.pricePerGB`, pinned by test), the unknown-value branch is likewise a distinct `unsupportedValueType(path:)`, and the integer-only rule is documented on both `SignedServiceManifest` and `ServiceOffer` so an operator hits the doc before the error.

**Cross-language vectors are now Rust-emitted, not self-written.** onymchat/onym-discovery#5 adds a `foundation-vectors` fixture pair to the reference implementation's byte-pinned conformance set: one input combining `seatID`/`seatable`, `relay10`/`relay2`, a `` control character, and a non-ASCII string — plus the canonical bytes serde_json actually emitted, regenerated via the existing `DISCOVERY_REGEN_FIXTURES=1` flow. The Swift side carries the pair verbatim and asserts byte equality, so "change one, change both" is enforced by real Rust bytes on both ends, on exactly the divergence classes the hand-rolled encoder exists for. (The vacuous digest in the Rust-signed-fixture test is also fixed — the pin is hardcoded now.)

**Doc anchor:** reworded to point at spec §3 and the onym-discovery reference (whose emitted bytes are the pinned fixture) rather than asserting a twin file as present fact.

**Corrupt store overwrite — the other catch that mattered.** Absent and corrupt are now different states: a blob that exists but doesn't decode throws `PinnedConsentStoreError.corruptStore`, gets parked under a `.corrupt` sibling key for forensics, and the live key is left untouched — `accept` propagates the throw before writing, so the next consent can no longer silently destroy every prior record. Recovery is an explicit `clearCorruptRecords()` (nothing calls it automatically), with `hasCorruptRecords` to surface the state. `save` failures now throw `persistenceFailed` and `accept` propagates that too, so a returned record always means persisted consent. The protocol is throwing accordingly; tests pin never-overwrite, parking, refusal-until-cleared, and the failing-save path.

**Atomicity claim** narrowed to in-process — an `NSLock` gives nothing across an app + extension over a shared suite, and the doc now says so.

**Pin freshness:** `PinnedConsentRecord.validUntil` / `isExpired(now:)`, derived from the pinned manifest bytes. Deliberately not enforced anywhere — an expired pin stays the active consent, entitlement behavior unchanged — but callers can now ask, to badge a stale pin or prompt a re-fetch; documented as such.

**Operator-key binding:** `review(raw:expectedDigest:expectedOperatorKey:now:)` — when supplied, the manifest's `operator` must equal it or review fails with a distinct `operatorKeyMismatch(expected:actual:)`; when nil, TOFU is an explicit call-site choice rather than a hidden property of the default. `digestMismatch` now reports the caller's original pin string, not the lowercased form, so logs line up with what was passed; `digestPinMalformed` stays distinct.

**Nits:** `isFree` trims and lowercases (`"Free"` no longer gates as paid; `"freemium"` still does), the two `ISO8601DateFormatter`s are cached statically off the per-manifest parse path, and the fixture digest is hardcoded as noted above.

OnymFoundation 51/51 on the iPhone 17 Pro simulator, on the branch as pushed. Per the restacking plan, this round is on `service-consent-primitives` only — #246/#247 will absorb it when they're recut. Thanks especially for the CI-dead-suite and corrupt-overwrite finds; the first would have let every future regression through silently, and the second was a designed-in data-loss path hiding behind a `try?`.

dev-onym (commented):
**COMMENT** — solid, well-tested, well-documented; no blocking correctness defects, but one structural duplication and one integrity gap worth addressing.

- `Packages/OnymFoundation/Sources/OnymFoundation/ServiceManifest.swift:184` — `serialize` and `appendEscaped` (:239) are a near-verbatim copy of `DiscoveryCanonical.serialize`/`appendEscaped` (`Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryCanonical.swift:71,124`), ~70 lines, differing only in error type and the added `path`. The "no cross-package import" rationale is backwards: `OnymDiscovery` already depends on `OnymFoundation` (`Packages/OnymDiscovery/Package.swift:11`), so the dedupe direction is available today — `DiscoveryCanonical.signingBytes` can delegate here and keep its error mapping. Two byte-identical canonicalizers that must stay in sync are exactly the drift risk the fixture tests exist to catch; better not to have the drift surface at all.
- `Packages/OnymFoundation/Sources/OnymFoundation/PinnedConsentRecord.swift:9` — the type is `public` + `Codable`, so the synthesized `init(from:)` is public. That is a second, ungated construction path around `ReviewedServiceManifest`, which the doc comment (:37) claims holds "at the type level". Combined with `PinnedConsentStore.load()` never re-checking `manifestHash == sha256(manifestBytes)`, a decoded record can pin a hash that doesn't match its bytes and nothing notices. Cheap fix: custom `init(from:)` that recomputes the digest and throws on mismatch (or make the record's decode internal to the module).
- `Packages/OnymFoundation/Sources/OnymFoundation/PinnedConsentRecord.swift:61,73` — `validUntil` and `isExpired` each call `consentedManifest()`, which is a full `JSONSerialization` parse *plus* a SHA-256 over the whole manifest (`SignedServiceManifest.init` computes `manifestHash` eagerly). These are documented as the "badge a stale pin" callers, i.e. SwiftUI row bodies. A stored `validUntil` on the record, or a lazily-cached decode, avoids hashing every manifest per render.
- `Packages/OnymFoundation/Sources/OnymFoundation/ReviewedServiceManifest.swift:86` — `expectedDigest` is case-normalized (:74) with a good rationale, but `expectedOperatorKey` is compared raw. The same copy-paste-from-a-catalog path applies to `onym:key:<hex>`; an uppercased key yields `operatorKeyMismatch`, which reads as an attack rather than a caller typo.
- `Packages/OnymFoundation/Sources/OnymFoundation/PinnedConsentStore.swift:159` — `save` only throws `persistenceFailed` on *encode* failure; `defaults.set` returns nothing, so the "or written" half of the doc contract (:14) and the "a returned record always means persisted consent" claim (:88) are stronger than the implementation. Either narrow the comment or add a read-back check.
- `Packages/OnymFoundation/Sources/OnymFoundation/PinnedConsentStore.swift:42` — one process-global lock serializes accepts across *all* store instances, including ones over different `UserDefaults` suites (notably the per-test suites). Correct, just broader than needed; keying by suite name would avoid the false contention.
- `Packages/OnymFoundation/Sources/OnymFoundation/Entitlement.swift:54` — `hasCorruptRecords`/`clearCorruptRecords()` live on the concrete store, so a consumer holding `any PinnedConsentStore` (this initializer) can observe the fail-closed `nil` but has no way to surface or recover the corruption. Consider hoisting both onto the protocol.
- `.github/workflows/release.yml:783` — results land in `results.json` as `OnymFoundationTests.*`, but `discover_tests.py:107` only enumerates `Tests/OnymIOSTests` and `Tests/OnymIOSUITests`, so those FQNs match no checkbox and `apply_results.py` silently leaves them out. Pre-existing (`OnymDiscoveryTests` has the same gap), so not a regression — but the gate is weaker than it looks.

PR #246, opened 2026-08-13
Wire discovery-backed seat fetchers with legacy fallback

rinat-enikeev:
## Summary

PR 3 of the discovery sequence. **Stacked on #245** (base branch set accordingly; #244 is already merged to main). Restacked 2026-08-15 as a **single commit** on top of `service-consent-primitives` — the diff now shows only this branch's own wiring work, with all eleven review rounds' fixes preserved in the squashed tree.

- `Sources/OnymIOS/DiscoverySeatAdapters.swift`: discovery-backed implementations of all four seat fetcher protocols (relayers, nostr relays, blossom servers, authorities). Each maps verified catalog entries → destination manifests reviewed via `ServiceManifestReviewer` (digest pin + Ed25519 + expiry, plus catalog↔manifest operator-key equality; the authority key is converted from the verified catalog entry, never taken from the fetched manifest). Any failure, or an empty catalog, falls back to the wrapped legacy GitHub fetcher — discovery can only add sources, never make the app worse than today.
- `SeatSelectionStore` in OnymDiscovery: append-only provenance records `{seatType, componentId, providerId, manifestHash, offerId, selectedAt}`; nothing writes it yet — the Settings UI PR does.
- Boot wiring in `OnymIOSApp.swift`: `DiscoveryRepository` constructed (nil under UI-testing flags, matching the existing pattern), started in the repository cluster, legacy fetchers wrapped at their construction sites, UITest slots for the discovery-backed fetchers. `AppDependencies` untouched.
- `project.yml`: app target picks up the OnymDiscovery package product.

## Restack note

The branch previously carried many merge commits tracking #244/#245 review rounds; it is now exactly one commit (2 original wiring commits + the merge-round adaptations, squashed). It sits on #245's tip **including its round-2 throwing `PinnedConsentStore` API** — the wiring layer itself needed no call-site changes for that.

**Merge order: #245 → #246 → #247, squash-merge recommended for each.**

## Testing (re-run after restack)

OnymDiscovery package 91/91, OnymFoundation package 51/51, app-target `DiscoverySeatAdaptersTests` 13/13, full app build for iPhone 17 Pro simulator.

### DIFF
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryModels.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryModels.swift
@@ -688,11 +688,13 @@ struct AnyCodingKey: CodingKey {
 // MARK: - Format rules
 
 /// Identifier / digest / URI syntax from §2 and §7 of the profile.
-enum DiscoveryFormat {
+public enum DiscoveryFormat {
     static let implementationProfileId = "onym:discovery-implementation:static-ed25519-v1"
 
-    /// `onym:key:<64-lowercase-hex>` → the hex, or nil.
-    static func operatorKeyHex(_ value: String) -> String? {
+    /// `onym:key:<64-lowercase-hex>` → the hex, or nil. Public for the
+    /// same reason as `isValidURI`: the app target's seat adapters
+    /// must parse operator keys with THIS parser, not a second one.
+    public static func operatorKeyHex(_ value: String) -> String? {
         guard value.hasPrefix("onym:key:") else { return nil }
         let hex = String(value.dropFirst("onym:key:".count))
         guard hex.count == 64, isLowercaseHex(hex) else { return nil }
@@ -739,15 +741,21 @@ enum DiscoveryFormat {
         }
     }
 
-    /// §7 URI rules: https only, DNS host (no IP literal, including
-    /// integer/hex single-label forms), no userinfo, no query, no
-    /// fragment, no port component in the **raw string** (URL
-    /// libraries normalize a redundant `:443` away before any
-    /// parsed-port check is reachable, so the raw string is checked
-    /// directly).
-    static func isValidURI(_ value: String) -> Bool {
+    /// §7 URI rules: a single allowed scheme (https for every URI the
+    /// discovery documents themselves carry; seat adapters pass "wss"
+    /// for WebSocket endpoints where the seat expects it), DNS host
+    /// (no IP literal, including integer/hex single-label forms), no
+    /// userinfo, no query, no fragment, no port component in the
+    /// **raw string** (URL libraries normalize a redundant default
+    /// port away before any parsed-port check is reachable, so the
+    /// raw string is checked directly).
+    public static func isValidURI(_ value: String, scheme allowedScheme: String = "https") -> Bool {
         guard let components = URLComponents(string: value) else { return false }
-        guard components.scheme == "https" else { return false }
+        // Schemes are case-insensitive (RFC 3986 §3.1) and
+        // `URLComponents` preserves the published case, so compare
+        // lowercased — `WSS://relay.example` must validate, not be
+        // silently dropped.
+        guard components.scheme?.lowercased() == allowedScheme else { return false }
         guard components.user == nil, components.password == nil else { return false }
         guard components.query == nil, components.fragment == nil else { return false }
         guard components.port == nil else { return false }
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryTrust.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryTrust.swift
@@ -447,7 +447,10 @@ public enum DiscoveryTrust {
 
 extension Data {
     /// Strict lowercase-hex decode (the profile's key encoding).
-    init?(lowercaseHex: String) {
+    /// Public for the same reason as `DiscoveryFormat.operatorKeyHex`:
+    /// the app target's seat adapters decode key hex with THIS
+    /// decoder, not a second one.
+    public init?(lowercaseHex: String) {
         guard lowercaseHex.count % 2 == 0,
               DiscoveryFormat.isLowercaseHex(lowercaseHex)
         else { return nil }
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/SeatSelectionStore.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/SeatSelectionStore.swift
@@ -1,0 +1,137 @@
+import Foundation
+
+/// Provenance record for one seat selection made through discovery:
+/// which component the user picked, which provider listed it, and the
+/// exact manifest hash the pick was bound to.
+///
+/// **Provenance only** — the legacy per-seat configurations
+/// (`RelayerConfiguration`, the Nostr fanout list, the Blossom server
+/// list, the moderation mandate) remain the operational source of
+/// truth. A discovery-driven pick writes BOTH this record and the
+/// derived endpoint into the legacy configuration (the dual-write
+/// lands with the consent UI); endpoints present in a legacy
+/// configuration with no matching `ModuleSelection` are "Manual /
+/// legacy" entries.
+public struct ModuleSelection: Codable, Equatable, Hashable, Sendable {
+    /// The seat this selection fills (e.g. "notary",
+    /// "transport.message", "blob.storage", "moderation").
+    public let seatType: String
+    /// `onym:component:<id>` of the selected component.
+    public let componentId: String
+    /// `onym:component:<id>` of the discovery provider whose catalog
+    /// the entry came from (the selection's source).
+    public let providerId: String
+    /// `sha256:<hex>` of the exact manifest bytes the user consented
+    /// to — ties the selection to `PinnedConsentRecord` /
+    /// `CatalogEntry.manifest.digest`.
+    public let manifestHash: String
+    /// The offer accepted at selection time, when the manifest carried
+    /// offers.
+    public let offerId: String?
+    public let selectedAt: Date
+
+    public init(
+        seatType: String,
+        componentId: String,
+        providerId: String,
+        manifestHash: String,
+        offerId: String? = nil,
+        selectedAt: Date
+    ) {
+        self.seatType = seatType
+        self.componentId = componentId
+        self.providerId = providerId
+        self.manifestHash = manifestHash
+        self.offerId = offerId
+        self.selectedAt = selectedAt
+    }
+}
+
+/// Persistence seam for selection provenance. One **active** selection
+/// per seat type (the most recently recorded); prior records for that
+/// seat are kept as bounded history so "what did I run in March, via
+/// which provider?" stays answerable without growing without limit.
+public protocol SeatSelectionStore: Sendable {
+    /// The most recently recorded selection for `seatType`, or nil if
+    /// no discovery-driven selection was ever made for that seat.
+    func activeSelection(seatType: String) -> ModuleSelection?
+    /// The retained selections for `seatType`, oldest first (the last
+    /// element is the active selection).
+    func history(seatType: String) -> [ModuleSelection]
+    /// Record a selection. It becomes the active selection for its
+    /// `seatType`; previously recorded selections stay in (bounded)
+    /// history.
+    func record(_ selection: ModuleSelection)
+}
+
+/// Serializes every `record` in the process — `record` is a
+/// load-modify-save, and without mutual exclusion two concurrent
+/// records read the same array and one selection is silently dropped.
+/// A process-wide lock rather than an actor for the same reason as
+/// `PinnedConsentStore`'s serialization: the flows that record a
+/// selection are synchronous, and an actor hop would force that whole
+/// path async for the same guarantee.
+private enum SeatSelectionStoreSerialization {
+    static let lock = NSLock()
+}
+
+/// Production `SeatSelectionStore`. Single UserDefaults blob under
+/// `app.onym.ios.discovery.selections` — an append-only-shaped array
+/// of records; "active" is a projection (last record per seat), not
+/// separate state that could drift from the history.
+///
+/// Retention mirrors `PinnedConsentStore`'s treatment: per seat type,
+/// the active selection plus the last `maxHistoryRecordsPerSeat`
+/// prior records; the oldest beyond that are evicted on `record`.
+///
+/// `@unchecked Sendable` for the same reason as the sibling stores:
+/// `UserDefaults` is documented thread-safe but not formally `Sendable`.
+public struct UserDefaultsSeatSelectionStore: SeatSelectionStore, @unchecked Sendable {
+    private static let key = "app.onym.ios.discovery.selections"
+
+    /// Prior (non-active) records kept per seat type. Same cap as the
+    /// consent store's per-component history — these records are tiny
+    /// (no manifest bytes), but unbounded re-selection history in a
+    /// UserDefaults blob is the same failure mode.
+    static let maxHistoryRecordsPerSeat = 8
+
+    private let defaults: UserDefaults
+
+    public init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    public func activeSelection(seatType: String) -> ModuleSelection? {
+        history(seatType: seatType).last
+    }
+
+    public func history(seatType: String) -> [ModuleSelection] {
+        loadAll().filter { $0.seatType == seatType }
+    }
+
+    public func record(_ selection: ModuleSelection) {
+        SeatSelectionStoreSerialization.lock.lock()
+        defer { SeatSelectionStoreSerialization.lock.unlock() }
+
+        var all = loadAll()
+        all.append(selection)
+        // Retention: evict the oldest records of this seat type beyond
+        // active + cap. Records of other seats are untouched.
+        let seatIndices = all.indices.filter { all[$0].seatType == selection.seatType }
+        let excess = seatIndices.count - (Self.maxHistoryRecordsPerSeat + 1)
+        if excess > 0 {
+            for index in seatIndices.prefix(excess).reversed() {
+                all.remove(at: index)
+            }
+        }
+        guard let data = try? JSONEncoder().encode(all) else { return }
+        defaults.set(data, forKey: Self.key)
+    }
+
+    private func loadAll() -> [ModuleSelection] {
+        guard let data = defaults.data(forKey: Self.key),
+              let all = try? JSONDecoder().decode([ModuleSelection].self, from: data)
+        else { return [] }
+        return all
+    }
+}
--- a/Packages/OnymDiscovery/Tests/OnymDiscoveryTests/DiscoveryModelsTests.swift
+++ b/Packages/OnymDiscovery/Tests/OnymDiscoveryTests/DiscoveryModelsTests.swift
@@ -231,6 +231,9 @@ final class DiscoveryModelsTests: XCTestCase {
 
     func testURIRules() {
         XCTAssertTrue(DiscoveryFormat.isValidURI("https://discovery.onym.app/manifest.json"))
+        // Schemes are case-insensitive (RFC 3986 §3.1): an uppercase
+        // scheme validates instead of being silently dropped.
+        XCTAssertTrue(DiscoveryFormat.isValidURI("HTTPS://discovery.onym.app/manifest.json"))
         // http scheme
         XCTAssertFalse(DiscoveryFormat.isValidURI("http://discovery.onym.app/manifest.json"))
         // IP literals, v4 and v6
@@ -249,6 +252,23 @@ final class DiscoveryModelsTests: XCTestCase {
         XCTAssertFalse(DiscoveryFormat.isValidURI("https://3232235777/m.json"))
         XCTAssertFalse(DiscoveryFormat.isValidURI("https://0xc0a80101/m.json"))
         XCTAssertFalse(DiscoveryFormat.isValidURI("https://192.168.1.0x1/m.json"))
+    }
+
+    func testURIRulesWithAllowedScheme() {
+        // Seat adapters validate WebSocket endpoints under the same §7
+        // rules with the scheme swapped for the one the seat expects.
+        XCTAssertTrue(DiscoveryFormat.isValidURI("wss://relay.example", scheme: "wss"))
+        XCTAssertTrue(DiscoveryFormat.isValidURI("wss://relay.example/inbox", scheme: "wss"))
+        // Case-insensitive: `WSS://…` must validate, not silently drop.
+        XCTAssertTrue(DiscoveryFormat.isValidURI("WSS://relay.example", scheme: "wss"))
+        // The default stays https-only; wss passes only when asked for.
+        XCTAssertFalse(DiscoveryFormat.isValidURI("wss://relay.example"))
+        XCTAssertFalse(DiscoveryFormat.isValidURI("https://relay.example", scheme: "wss"))
+        // The rest of the rules still apply under an allowed scheme.
+        XCTAssertFalse(DiscoveryFormat.isValidURI("wss://relay.example:8443", scheme: "wss"))
+        XCTAssertFalse(DiscoveryFormat.isValidURI("wss://user@relay.example", scheme: "wss"))
+        XCTAssertFalse(DiscoveryFormat.isValidURI("wss://relay.example/x?a=1", scheme: "wss"))
+        XCTAssertFalse(DiscoveryFormat.isValidURI("wss://192.168.1.10/x", scheme: "wss"))
     }
 
     func testIdentifierAndDigestFormats() {
--- a/Packages/OnymDiscovery/Tests/OnymDiscoveryTests/SeatSelectionStoreTests.swift
+++ b/Packages/OnymDiscovery/Tests/OnymDiscoveryTests/SeatSelectionStoreTests.swift
@@ -1,0 +1,133 @@
+import Foundation
+import XCTest
+@testable import OnymDiscovery
+
+/// UserDefaults round-trip with per-test suite isolation — same
+/// pattern as the sibling selection-store tests in the app target.
+final class SeatSelectionStoreTests: XCTestCase {
+    private var defaults: UserDefaults!
+    private var suiteName: String!
+    private var store: UserDefaultsSeatSelectionStore!
+
+    override func setUp() {
+        super.setUp()
+        suiteName = "app.onym.ios.discovery.selections.tests.\(UUID().uuidString)"
+        defaults = UserDefaults(suiteName: suiteName)
+        store = UserDefaultsSeatSelectionStore(defaults: defaults)
+    }
+
+    override func tearDown() {
+        defaults.removePersistentDomain(forName: suiteName)
+        defaults = nil
+        suiteName = nil
+        store = nil
+        super.tearDown()
+    }
+
+    private func selection(
+        seatType: String = "notary",
+        componentId: String = "onym:component:sample-notary",
+        providerId: String = "onym:component:onym-discovery",
+        manifestHash: String = "sha256:" + String(repeating: "a", count: 64),
+        offerId: String? = nil,
+        selectedAt: Date = Date(timeIntervalSince1970: 1_700_000_000)
+    ) -> ModuleSelection {
+        ModuleSelection(
+            seatType: seatType,
+            componentId: componentId,
+            providerId: providerId,
+            manifestHash: manifestHash,
+            offerId: offerId,
+            selectedAt: selectedAt
+        )
+    }
+
+    func test_activeSelection_isNilWhenNothingRecorded() {
+        XCTAssertNil(store.activeSelection(seatType: "notary"))
+        XCTAssertEqual(store.history(seatType: "notary"), [])
+    }
+
+    func test_record_roundTripsThroughUserDefaults() {
+        let recorded = selection(offerId: "free-tier")
+        store.record(recorded)
+
+        // A fresh store over the same suite sees the persisted record.
+        let reloaded = UserDefaultsSeatSelectionStore(defaults: defaults)
+        XCTAssertEqual(reloaded.activeSelection(seatType: "notary"), recorded)
+        XCTAssertEqual(reloaded.history(seatType: "notary"), [recorded])
+    }
+
+    func test_record_replacesActiveSelectionAndKeepsHistory() {
+        let first = selection(
+            componentId: "onym:component:notary-one",
+            selectedAt: Date(timeIntervalSince1970: 1_700_000_000)
+        )
+        let second = selection(
+            componentId: "onym:component:notary-two",
+            manifestHash: "sha256:" + String(repeating: "b", count: 64),
+            selectedAt: Date(timeIntervalSince1970: 1_700_000_100)
+        )
+        store.record(first)
+        store.record(second)
+
+        XCTAssertEqual(store.activeSelection(seatType: "notary"), second)
+        // Oldest-first history keeps the replaced selection.
+        XCTAssertEqual(store.history(seatType: "notary"), [first, second])
+    }
+
+    func test_selectionsAreScopedPerSeatType() {
+        let notary = selection(seatType: "notary")
+        let transport = selection(
+            seatType: "transport.message",
+            componentId: "onym:component:onym-courier"
+        )
+        store.record(notary)
+        store.record(transport)
+
+        XCTAssertEqual(store.activeSelection(seatType: "notary"), notary)
+        XCTAssertEqual(store.activeSelection(seatType: "transport.message"), transport)
+        XCTAssertEqual(store.history(seatType: "notary"), [notary])
+        XCTAssertEqual(store.history(seatType: "transport.message"), [transport])
+        XCTAssertNil(store.activeSelection(seatType: "blob.storage"))
+    }
+
+    func test_malformedPersistedBlobDegradesToEmpty() {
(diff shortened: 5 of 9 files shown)


rinat-enikeev:
Restacked as a single commit (a71e27d) on top of `service-consent-primitives` (#245, incl. its round-2 throwing `PinnedConsentStore` API) and rebased the PR base accordingly. The diff now shows only this branch's own wiring work: seat adapters with legacy fallback, SeatSelectionStore, boot wiring + UITest slots, adapter tests, project.yml. All review-round fixes are preserved in the squashed tree (2 original wiring commits plus the adaptations made across the merge rounds tracking #244/#245 revisions). Re-verified after restack: OnymDiscovery 91/91, OnymFoundation 51/51, DiscoverySeatAdaptersTests 13/13, full app build (iPhone 17 Pro sim). Merge order 245 → 246 → 247, squash recommended.

dev-onym (commented):
REQUEST_CHANGES — trust chain and fallback shape are sound, but "add, never take away" isn't what the code does, and boot ordering makes discovery a no-op for relayers.

- `Sources/OnymIOS/DiscoverySeatAdapters.swift:189,222,251,299` — discovery **replaces** legacy, it doesn't add. One surviving entry makes the whole result discovery-only: `RelayerRepository.refresh` (`Packages/OnymChain/.../RelayerRepository.swift:132`) stores it as `knownList` *and* auto-populates the first-run config from it; nostr/blossom `applyDefault` overwrite the published defaults. So a single third-party source silently drops relayers.json / nostr-relays.json from the picker — the opposite of the header comment. Union with `fallback.fetchLatest()` (dedupe by URL) instead of falling back only on empty.
- `Sources/OnymIOS/OnymIOSApp.swift:825,831` — `relayerRepository.start()` runs before `discoveryRepository?.start()`, so the relayer adapter's `entries(seatType:)` hits the actor before `rebuildAggregateFromRetained()` and sees `[]` → always legacy on cold boot. Move discovery's `start()` above it. Related: the seat repositories' one-shot `start()` has already completed before discovery's network refresh lands, so freshly fetched entries only take effect next launch — worth stating, or hooking the `snapshots` stream.
- `Sources/OnymIOS/DiscoverySeatAdapters.swift:79` — the review binds the operator key but not `componentId` or `seat`. A provider can list an operator's genuine, correctly-signed `blob.storage` manifest as a `notary` entry and the adapter maps it into the relayer list; the operator never signed consent to that seat. Add `reviewed.signedManifest.componentId == entry.componentId` plus a seat check (the vocabularies differ — `seat: "courier"` vs `seatType: "transport.message"` — so that mapping needs pinning down).
- `Sources/OnymIOS/DiscoverySeatAdapters.swift:119,279` — `firstEndpointURL` ignores `endpoints[].role`, yet the fixtures publish roles (`submit`, `api`, `http-fallback`). First-https-wins will pick a status/docs URL as a relayer submit URL or an authority `apiBaseURL`. Filter by role with a documented fallback.
- `Sources/OnymIOS/DiscoverySeatAdapters.swift:63-66` — manifests are fetched serially, once per adapter, per refresh, with no bound on catalog size; a pull-to-refresh with N entries is N sequential round-trips behind the relayer spinner. Cap the entry count (and `log()` the truncation) and fetch concurrently.
- `Packages/OnymDiscovery/.../SeatSelectionStore.swift:90` — unbounded append-only history and a non-atomic read-modify-write, right after #245 landed an *atomic capped* consent store. Same treatment here, or the blob grows forever and concurrent `record` calls lose entries.
- `Sources/OnymIOS/DiscoverySeatAdapters.swift:79` — `ServiceManifestReviewer.review` already takes `expectedOperatorKey:` (checked before signature verification); pass it instead of comparing afterwards.
- `Sources/OnymIOS/OnymIOSApp.swift` — `DiscoverySeatCatalog(repository:fetcher:)` is handed a second `URLSessionDiscoveryFetcher()` rather than the repository's; harmless today, but it's a separate URLSession/config.
- Tests only cover ≤1 entry per catalog: no multi-entry, no partial-failure (some entries review, some don't), no cross-provider duplicate-URL case. Those are exactly where the replace-vs-add semantics bite.
- No Swift toolchain in this environment — the 91/51/13 test counts and the simulator build are unverified here.

rinat-enikeev:
Thanks for the review — the replace-vs-add finding and the boot-ordering one are both real bugs, and the replace-vs-add in particular is the substantive catch of this round: one surviving discovery entry made the fetched list discovery-only, `RelayerRepository` stored that as the whole known list, first-run auto-populate installed it, and the nostr/blossom `applyDefault` overwrote the published defaults with it. All addressed; both branches force-pushed (this PR now at 3b0eb5c, #247 rebased on top at 8200d10). Point by point:

1. **Replace-vs-add (substantive)** — fixed. The three endpoint adapters now ALWAYS fetch the legacy list and UNION it with the surviving discovery entries: dedupe by endpoint URL, discovery entry wins the row (its reviewed attribution), duplicates *within* the discovery results (cross-provider) also collapse to the first, and a legacy outage doesn't subtract surviving discovery endpoints (with none, the failure propagates as before). One note on where the fix lives: #247's commit already carried this union (entangled with its consent gate), so rather than duplicating it I pulled the union semantics DOWN into this PR's commit — wiring owns the adapters' base semantics — and #247's commit now keeps only the consent gating (`hasActiveConsent` filter) on top. The authorities adapter deliberately keeps replace-on-success: it's a directory the user picks from (mandate-enforced downstream), not an auto-populating configuration; that's now documented on the type.

2. **Boot ordering (substantive)** — fixed. `discoveryRepository?.start()` now runs before `relayerRepository.start()` in `OnymIOSApp.swift`; `start()` rebuilds the retained aggregate synchronously inside the actor call before spawning the network refresh, so cold-boot adapter reads see it. The freshly-fetched-entries-surface-next-launch limitation is documented in a comment at the call site (per your note, a comment rather than snapshot-stream plumbing — the clean seam is named there if it's ever needed live).

3. **Seat/component binding** — done, all three parts: `expectedOperatorKey: entry.operatorKey` is now passed to `ServiceManifestReviewer.review` (replacing the hand-rolled post-review guard), `reviewed.signedManifest.componentId == entry.componentId` is checked, and an explicit `SeatManifestVocabulary` table pins catalog seatType → accepted manifest seats, built from the real published manifests (onym-relayer `operator-manifest.src.json`: seat `"notary"`; onym-discovery deploy templates: courier `"transport.message"`, blossom `"blob.storage"`) plus the historical aliases (`"courier"`, `"blossom"`); moderation accepts `"moderation"`; unknown seat types require exact match. Any mismatch skips the entry with a log line.

4. **Endpoint role filtering** — done. Checked the real templates: courier and blossom both declare `role: "read-write"` on their client-facing endpoints (the notary template predates the field). `firstEndpointURL(schemes:)` now prefers the first scheme-matching endpoint with role `"read-write"` and falls back to the first scheme-matching endpoint of any role, documented on the method.

5. **Bound + parallelize** — done. `DiscoverySeatCatalog.maxEntriesPerSeat = 16`, truncation logged; fetch+review runs in a `TaskGroup` with results reassembled into aggregate order, so output is deterministic regardless of completion order.

6. **SeatSelectionStore** — done, mirroring #245's consent store: `record()` is serialized by a process-wide `NSLock` (`SeatSelectionStoreSerialization`), and per-seat history is capped at the active selection plus the last 8 prior records, oldest evicted; other seats untouched. Cap test added in `SeatSelectionStoreTests`.

7. Folded into 3 — `expectedOperatorKey` exists on the reviewer and is now used.

8. **Fetcher reuse** — done. One `URLSessionDiscoveryFetcher` instance now serves both the `DiscoveryRepository` and the `DiscoverySeatCatalog` seam; no second URLSession.

9. **Tests** — added in `DiscoverySeatAdaptersTests`: multi-entry union with the legacy list, partial failure (digest-mismatched entry dropped, survivor + legacy serve), cross-provider duplicate URL + discovery-vs-legacy dedupe with discovery attribution winning, legacy-outage survival/propagation, seat-vocabulary mismatch skip, componentId mismatch skip, role preference + fallback, and the per-seat cap. Existing map tests now assert the union shape.

**#247 (discovery-settings-ui)** was rebased onto the new tip; the expected conflicts in `DiscoverySeatAdapters.swift` (and its tests / `OnymIOSApp.swift`) were resolved preserving both sides — this PR's base semantics (union, reviewer hardening, seat table, role preference, bounded parallel fetches) with #247's consent gate filtered on top; #247's now-redundant pure-union test was dropped in favor of the ones here, its consent-specific ones kept.

**Verified per branch** (iPhone 17 Pro simulator):
- #246 @ 3b0eb5c: OnymDiscovery 92, OnymFoundation 51, app-target DiscoverySeatAdaptersTests 21, full app build.
- #247 @ 8200d10 (rebased): OnymDiscovery 113, OnymFoundation 53, app-target DiscoverySeatAdaptersTests + RelayerSettingsFlowTests 45, full app build.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01GupFt7f1hWn4zSP2HJEXB8

dev-onym:
No commit made — everything the comment describes is **already in the tree at HEAD (`3b0eb5c`)**, which is the exact SHA the comment names as this PR's force-pushed tip. The working tree is clean, so there is no change to make and nothing to commit.

I checked all nine points against the source rather than trusting the commit message:

| # | Claim | Evidence |
|---|---|---|
| 1 | Union, not replace | `mergeDiscoveredWithLegacy` at `Sources/OnymIOS/DiscoverySeatAdapters.swift:291` — legacy always fetched, dedupe by URL, discovery wins the row, intra-discovery dupes collapse, legacy outage returns survivors and only propagates when `unique.isEmpty`. Wired into all three endpoint adapters (`:342`, `:379`, `:412`); authorities keeps replace-on-success with the rationale documented on the type (`:420-429`). |
| 2 | Boot ordering | `await discoveryRepository?.start()` at `Sources/OnymIOS/OnymIOSApp.swift:845`, before `relayerRepository.start()` at `:850`; the next-launch limitation is commented at `:837-844`. |
| 3 | Seat/component binding | `expectedOperatorKey: entry.operatorKey` (`:153`), componentId cross-check (`:156`), `SeatManifestVocabulary` table (`:51-64`) with `notary` / `transport.message`+`courier` / `blob.storage`+`blossom` / `moderation`, unknown types falling back to exact match. |
| 4 | Endpoint role | `preferredEndpointRole = "read-write"` (`:196`), `firstEndpointURL(schemes:)` prefers it and falls back to first-any (`:216-229`). |
| 5 | Bound + parallelize | `maxEntriesPerSeat = 16` (`:81`), truncation log (`:115`), `withTaskGroup` with index slots for deterministic aggregate order (`:121-132`). |
| 6 | SeatSelectionStore | `SeatSelectionStoreSerialization` `NSLock` (`SeatSelectionStore.swift:74`), cap of active + 8 (`:96`, `:120-126`), `test_record_capsHistoryPerSeatKeepingActivePlusNewest`. |
| 7 | Reviewer param | `expectedOperatorKey` exists at `ReviewedServiceManifest.swift:64` and is enforced at `:86`; no hand-rolled post-review guard remains in the adapter. |
| 8 | Fetcher reuse | One `URLSessionDiscoveryFetcher` at `OnymIOSApp.swift:95`, passed to both the repository (`:105`) and `DiscoverySeatCatalog` (`:115-116`). |
| 9 | Tests | 21 tests in `Tests/OnymIOSTests/DiscoverySeatAdaptersTests.swift`, matching the claimed count — union, partial failure, dedupe-with-discovery-attribution, legacy outage, vocabulary skip, componentId skip, role preference *and* fallback (both halves in one test at `:387-422`), per-seat cap. |

One caveat on verification: this container has neither `xcodebuild` nor `swift`, so I confirmed by reading the source, not by reproducing the 92 / 51 / 21 test counts. The claimed counts are consistent with what's on disk (21 adapter tests).

If you intended for me to make an additional change beyond what's in `3b0eb5c`, let me know which — as written, the comment reads as a report of work that's already landed.

dev-onym (commented):
**REQUEST_CHANGES** — the crypto/review chain is solid and well tested, but the merge feeds paths that *activate* endpoints, not just list them.

- **Discovery endpoints are auto-activated with no consent step.** `DiscoverySeatAdapters.swift:291` merges catalog entries into the same list that `RelayerRepository.refresh` installs wholesale on first launch (`RelayerRepository.swift:130-138`, strategy `.random`) and that `NostrRelaysRepository.applyDefault` / `BlossomServersRepository.applyDefault` overwrite the config with whenever `hasUserInteracted == false` (`NostrRelaysRepository.swift:78`, `BlossomServersRepository.swift:75`). So any operator listed in a provider catalog becomes a live notary / message relay / blob host for every user who never customized those screens — before the consent UI (#247) exists, and with nothing written to `SeatSelectionStore`. The PR's "discovery can only add sources" invariant holds for *known lists* but not for *active configuration*. Either keep discovery entries out of the auto-populate/`applyDefault` path (surface them only in the picker), or land this after consent.
- **`isDefault: true`** at `DiscoverySeatAdapters.swift:377` and `:410` badges third-party catalog entries as Onym-published defaults. Even if #1 is accepted as intended, this label is wrong.
- **Legacy fetch is serialized behind up to 16 manifest fetches** (`:296-300`): `fetchLegacy()` only starts after `reviewedEntries` completes, and `URLSessionDiscoveryFetcher` allows 60 s per request. One unresponsive operator host now delays the published relayer list — and first-launch auto-populate — by a minute, where today it is one GitHub round-trip. Kick both off concurrently (`async let legacy = fetchLegacy()`).
- **Authorities adapter subtracts.** `:469` replaces the legacy directory whenever ≥1 discovery listing survives, so a catalog with one moderation entry hides every published authority — the opposite of the merge rule used for the other three seats. Merge by `componentId` instead; the "it's a directory, not a config" rationale argues for merging too.
- **Manifest endpoint URIs bypass the profile URI rules.** `firstEndpointURL` (`:216`) checks only the scheme, so `https://user:pass@host:8443/x?q=1` is accepted as a relayer/blossom URL, while every other URL in OnymDiscovery goes through `DiscoveryFormat.isValidURI` (`DiscoveryModels.swift:748`). Route endpoint URIs through the same check.
- **Dedupe is raw `URL` equality** (`:296`): `wss://relay.example` and `wss://relay.example/` won't collapse, yielding a duplicate row and a second connection to the same relay. Normalize (scheme+host lowercased, trailing slash stripped) before keying.
- **Verify the moderation seat vocabulary.** `:56` accepts only `seat: "moderation"`, but `SignedServiceManifest`'s own docs name `"authority"` as the seat string for that role (`ServiceManifest.swift:65-67`). If deployed authority manifests say `authority`, every moderation entry silently fails review. The other three seats got aliases; this one should too, or confirm against the published manifest.

Non-blocking: `SeatSelectionStore` retention math and eviction ordering are correct, and the concurrency/lock rationale matches the sibling stores.

rinat-enikeev:
All seven points addressed — amended in place (single-commit discipline kept), new SHAs below. Point by point:

**1. Auto-activation without consent — the substantive catch of this round, thank you.** You're right, and the invariant comment was wrong about itself: "add, never take away" held for the *known list* while the known list is precisely what `RelayerRepository.refresh` installs wholesale on first launch and what `applyDefault` overwrites for every `!hasUserInteracted` user — so at this layer, pre-consent-UI, any catalog operator became a live notary/relay/blob host for default-config users with nothing recorded anywhere. Rather than landing this after #247, the fix moves the consent-gate seam DOWN into this PR: the three endpoint adapters now take an optional `hasActiveConsent: (@Sendable (String) -> Bool)?` closure.

- **nil** (no consent system present — this PR standalone, and how `OnymIOSApp` wires it here): discovery entries are **excluded from `fetchLatest` entirely** — pure legacy passthrough, no catalog review fan-out even. Catalog entries stay reachable only through the separate catalog-entries surface (`DiscoveryRepository.entries(seatType:)`) that consent-aware pickers read.
- **provided**: only entries whose componentId the closure approves merge into the known list.

So #246 standalone can never activate a catalog endpoint, and #247's commit reduces to *providing* the closure (backed by the pinned consent store its module-consent flow writes) plus the UI — its adapter-level gate diff is gone because the seam now lives here. Tests cover both halves: nil-closure asserts the catalog is never even consulted; closure-provided asserts consented-only union.

**2. `isDefault: true` → `false`.** Agreed — that flag means "Onym-published default" (the badge the settings rows render), and a third-party catalog entry is not one, whatever list it entered. Both nostr and blossom mappings fixed, tests updated.

**3. Serialization behind manifest fetches** — fixed. The legacy fetch now starts first and runs concurrently with the review fan-out (`async let` inside the merge helper), so a slow operator host costs only itself; the published list (and first-launch auto-populate) is never serialized behind up to 16 manifest fetches. (With the nil-closure wiring in this PR the fan-out doesn't run at all.)

**4. Authorities subtract** — fixed to merge by `componentId`, discovery attribution winning duplicate ids, published directory never subtracted; legacy outage still serves surviving discovery listings, and an empty-discovery legacy failure propagates as before. The adapter deliberately keeps **no** consent gate: the list is a directory the user picks from, listing activates nothing, and moderation consent remains the signed-mandate flow downstream of the pick — now stated in the adapter comment.

**5. Endpoint URIs through the profile rules** — done. `DiscoveryFormat.isValidURI` gained an allowed-scheme parameter (default stays `https`; the nostr seat passes `wss`) and is now `public` so the app-target adapters can use the same §7 check as everything else in OnymDiscovery. `firstEndpointURL` routes every candidate through it, so `https://user:pass@host:8443/x?q=1` is rejected, not adopted. New package tests pin the wss variant; a new adapter test pins userinfo/port/query rejection.

**6. Dedupe normalization** — done: scheme+host lowercased, one trailing slash stripped, applied on both sides of the merge (and to intra-discovery duplicates). `wss://Relay.Example/` and `wss://relay.example` now collapse to one row and one connection; test added.

**7. Moderation seat vocabulary — verified against the live manifest.** The deployed authority manifest at https://authority.onym.app/manifest.json declares `seat: "moderation"` (`componentId: onym:component:onym-authority`), so the current acceptance is correct for what's actually published — the `ServiceManifest.swift` doc comment naming `"authority"` is the drifted side. Since the foundation docs do use "authority" for the seat, it's now tolerated as an alias (`"moderation": ["moderation", "authority"]`), with the live-manifest citation pinned in the vocabulary comment and a test for the alias.

**Verified** (both branches, iPhone 17 Pro simulator): OnymDiscovery + OnymFoundation package tests, app-target discovery suites, full app build. #247 (`discovery-settings-ui`) rebased on top, its commit message updated to reflect that it now only supplies the closure + UI.

- #246 `discovery-wiring`: `fdafb7566aaf0d4c937dc567b7a473d13134b05f`
- #247 `discovery-settings-ui`: `f3813b04a1dfd34ed0db257d133913650dbfac89`

dev-onym:
Verified the branch against all seven points. **The work is already present at the cited SHA — I made no commit, because there is nothing left to change.**

State of the worktree: `HEAD` = `fdafb75` = `origin/discovery-wiring` = the `#246` SHA in your comment, working tree clean.

Point-by-point verification against the tree:

1. **Consent gate seam.** All three endpoint adapters carry `let hasActiveConsent: (@Sendable (String) -> Bool)?` and `guard let hasActiveConsent else { return try await fallback.fetchLatest() }` before any catalog access — `DiscoverySeatAdapters.swift:377`, `:398`, `:430`, `:448`, `:484`, `:502`. `OnymIOSApp.swift` passes no closure at any of the six `wrapping(...)` call sites, so this PR standalone is pure legacy passthrough. Tests: `DiscoverySeatAdaptersTests.swift:687` (stub catalog `XCTFail`s if consulted — pins "never even fans out"), `:707`, `:721`, and `:738` (consented-only union).
2. **`isDefault: false`** on both discovery-derived mappings — `:463` (nostr), `:515` (blossom), with the "Onym-published default" rationale in the comment. Tests assert it at `DiscoverySeatAdaptersTests.swift:543` and `:584`; the legacy stubs still carry `isDefault: true`.
3. **Concurrency.** `async let legacyFetch = fetchLegacy()` is the first statement of `mergeDiscoveredWithLegacy` (`:332`), before `await discovered()` — legacy is not serialized behind the fan-out.
4. **Authorities merge** by `dedupeKey: \.componentId` (`:553`), discovery rows first so duplicates collapse to the discovery attribution, published directory never subtracted; no consent gate, with the "listing activates nothing" rationale at `:529-535`. Tests `:818`, `:835` (incl. empty-discovery legacy failure propagating).
5. **URI rules.** `DiscoveryFormat.isValidURI(_:scheme:)` is `public` with `scheme` defaulting to `https` (`DiscoveryModels.swift:750`); `firstEndpointURL` routes every candidate through it (`DiscoverySeatAdapters.swift:252`). Package tests pin the wss variant (`DiscoveryModelsTests.swift:257-266`); adapter test `:771` pins userinfo/port/query rejection.
6. **Dedupe normalization** — `endpointDedupeKey` lowercases scheme+host and strips one trailing slash (`:349-358`), applied to both merge sides and intra-discovery duplicates. Test `:795`.
7. **Seat vocabulary** — `"moderation": ["moderation", "authority"]` with the live-manifest citation (`:77-82`); alias test at `:869`.

One limitation to flag: this container has no Swift toolchain (`swift`/`xcodebuild` not found), so I could not re-run the package/app test suites or the build — my verification is source-level only. The simulator verification in your comment stands as the build evidence.

dev-onym (commented):
REQUEST_CHANGES — the wiring and per-entry review are solid, but the one path that is actually live in this PR (authorities) lets a discovery provider take over a published authority's listing.

- **`Sources/OnymIOS/DiscoverySeatAdapters.swift:553` — discovery wins a componentId collision in the un-gated authorities merge.** `dedupeKey: \.componentId` with discovery rows placed first means any enabled provider can publish an entry with an already-published authority's componentId and replace its `apiBaseURL` and `operatorPublicKeyBase64`. Those are exactly the two fields that carry trust: `AuthorityManifestFetcher.swift:71` pins the manifest key from the listing, and `ModerationAuthorityClient.swift:350` sends the user's signed mandate to `listing.apiBaseURL`. Worse, `ModerationRepository.swift:428` resolves the *active mandate's* authority by `listings.first(where: componentId == ...)`, so the takeover hits an already-consented authority — forcing an attacker-controlled re-consent gate and redirecting mandate traffic. This is a trust downgrade vs. today's pinned `authorities.json`, and `test_authorities_duplicateComponentIdCollapsesToDiscoveryRow` locks the behavior in. Fix: legacy wins on collision for authorities (discovery may only contribute *unseen* componentIds) — seed `seen` from the legacy result before appending discovery rows, and invert that test.
- **`DiscoverySeatAdapters.swift:553` — this is the PR's only live behavior change.** All three endpoint adapters get `hasActiveConsent: nil` at every boot site (`OnymIOSApp.swift:139,257,365,452`), so relayers/nostr/blossom are pure passthrough; the un-gated seat is the only one that ships. Given the finding above, consider gating authorities behind the same seam until the consent UI lands, or at least merging legacy-first.
- **`DiscoverySeatAdapters.swift:198` — `CatalogEntry.status` is ignored.** A provider-flagged `{"state": "warning"|"review"}` entry (DiscoveryModels.swift:444) merges into the directory indistinguishable from a clean one. At minimum drop `warning` entries from the known list, or carry the flag forward.
- **`DiscoverySeatAdapters.swift:587` — hand-rolled hex→base64 duplicates existing, verified code.** The reviewer already pins the manifest key to the entry's, and `SignedServiceManifest.operatorPublicKeyHex` (ServiceManifest.swift:71) plus `DiscoveryFormat.operatorKeyHex` give the same value; prefer widening the latter's access over a second parser.
- **`DiscoverySeatAdapters.swift:247` — scheme check is tautological and case-brittle.** `scheme` is derived from the URL being validated, so only `schemes.contains` restricts it; and since `URLComponents.scheme` preserves case, a manifest publishing `WSS://relay.example` fails `isValidURI`'s exact `==` and the entry is silently dropped. Lowercase in the comparison inside `DiscoveryModels.swift:750`.
- **Dead code:** `SeatManifestFields.endpointURIs` (`DiscoverySeatAdapters.swift:289`) has no callers, and `ModuleSelection`/`SeatSelectionStore` are referenced only by their own tests — fine if the Settings PR lands next, but flag if that slips.
- Verified as claimed: `discoveryRepository?.start()` does rebuild the retained aggregate before `relayerRepository.start()` (actor `start()` is synchronous up to the spawned refresh), per-entry review failures are correctly isolated, and the `entriesBeyondPerSeatCap` / legacy-failure-still-serves paths match the tests.

rinat-enikeev:
Round-3 (post-restack) fixes are in — `discovery-wiring` amended to db29ceb, `discovery-settings-ui` (#247) rebased on top as 2b19d04. Point by point:

**1. Authority-listing takeover — the substantive catch, and a genuine trust downgrade.** You're right on every link of the chain: with discovery rows first in the componentId dedupe, any enabled provider could replace a published authority's `apiBaseURL` and `operatorPublicKeyBase64` — exactly the two trust-carrying fields (`AuthorityManifestFetcher` pins the manifest key from the listing, `ModerationAuthorityClient` sends the signed mandate to `listing.apiBaseURL`) — and because `ModerationRepository` resolves the *active mandate's* authority by componentId, the takeover reached an already-consented authority: attacker-controlled re-consent gate plus redirected mandate traffic, strictly worse than today's pinned `authorities.json`. Fixed in two layers, both of which I'd state explicitly:

- **Legacy wins on collision** — `mergeDiscoveredWithLegacy` gained a `legacyWinsOnCollision` mode the authorities adapter uses: the legacy keys are subtracted from the discovery rows, so discovery contributes only componentIds the published directory doesn't list; the published row (endpoint + key included) survives untouched. `test_authorities_duplicateComponentIdCollapsesToDiscoveryRow` is INVERTED into `test_authorities_duplicateComponentIdKeepsLegacyRow`, pinning the legacy row against a colliding catalog entry that publishes an attacker endpoint.
- **Gate seam** — took your stronger option too, since this was the PR's only live behavior change. The authorities adapter now carries `discoveryEnabled: (@Sendable () -> Bool)?` mirroring the endpoint adapters' `hasActiveConsent` seam: nil (this PR standalone — all boot sites) is a pure legacy passthrough with **no catalog fan-out at all** (`test_authorities_nilDiscoveryEnabledServesLegacyOnlyWithoutCatalogFanout`); #247, which ships the discovery surface, enables it — and the legacy-wins rule governs the merge once it's on. So #246 alone changes no live behavior anywhere, and even with #247 a catalog entry can never shadow a published authority.

**2. `CatalogEntry.status` honored.** `reviewedEntries` now drops provider-flagged `warning` entries outright, before they cost a manifest fetch or a per-seat cap slot (`test_relayers_warningStatusEntryIsExcluded`). `review`-flagged entries pass (`test_relayers_reviewStatusEntryStillMerges`): I checked the endpoint shapes — `AuthorityListing`, `RelayerEndpoint`, `NostrRelayEndpoint`, `BlossomServerEndpoint` — and none has a field to carry the flag on, so surfacing review state stays a consent/picker concern for the settings-UI layer, which reads it off `attributed.entry.status`; noted as such in the code.

**3. Hex→base64 via the verified helpers.** The hand-rolled parser is gone: `operatorKeyBase64(fromOnymKey:)` is now `DiscoveryFormat.operatorKeyHex` + `Data(lowercaseHex:)`, both widened to `public` with a one-line rationale each — the same minimal treatment `isValidURI` got. No second parser left in the app target.

**4. `isValidURI` scheme case.** The comparison lowercases the parsed scheme (RFC 3986 §3.1), so `WSS://relay.example` validates instead of silently dropping — covered at the format level (`HTTPS://` and `WSS://` asserts in `DiscoveryModelsTests`) and end-to-end (`test_nostr_uppercaseSchemeEndpointStillMaps`). On the tautology: with the case fix, `isValidURI`'s internal scheme equality is indeed a no-op when `firstEndpointURL` passes the URI's own scheme — `schemes.contains` is the restricting check; I've documented that division of labor at the call site rather than restructuring the signature.

**5. Dead code.** `SeatManifestFields.endpointURIs` turned out NOT to be dead — #247's consent-flow `apply` closures feed it to `ModuleConsentAcceptance.requiredEndpointURL` for both the relayer and nostr seats — so it stays, now with a doc comment naming that consumer. `ModuleSelection` / `SeatSelectionStore` stay for the same reason: the Settings PR consumes both, and it's stacked directly on this one.

**Verified** (iPhone 17 Pro simulator):
- #246 @ db29ceb: OnymDiscovery 93, OnymFoundation 51, app-target DiscoverySeatAdaptersTests 34 + RelayerSettingsFlowTests 17 — all passing; app builds.
- #247 @ 2b19d04 (rebased, conflict in the boot-wiring comment resolved; authorities `discoveryEnabled: { true }` wired at both boot sites): OnymDiscovery 114, OnymFoundation 53, app-target suites 52 — all passing; app builds.

dev-onym:
All five items from the comment are already present in this worktree's HEAD (`db29ceb`), and the tree is clean — there is nothing left to change and therefore no commit to make. I verified each point against the actual code rather than the description:

1. **Authority-listing takeover** — `mergeDiscoveredWithLegacy` has the `legacyWinsOnCollision` mode (`Sources/OnymIOS/DiscoverySeatAdapters.swift:361-382`, legacy keys subtracted from discovery rows), and the authorities adapter passes `legacyWinsOnCollision: true` plus carries the `discoveryEnabled: (@Sendable () -> Bool)?` seam that short-circuits to `fallback.fetchLatest()` before any catalog fan-out (`:588-622`). Tests `test_authorities_duplicateComponentIdKeepsLegacyRow` and `test_authorities_nilDiscoveryEnabledServesLegacyOnlyWithoutCatalogFanout` exist (the latter `XCTFail`s inside the catalog stub, so it really pins "no fan-out"); the old `…CollapsesToDiscoveryRow` name is gone. Both boot sites in `OnymIOSApp.swift` call `wrapping(_:catalog:)` without `discoveryEnabled`, i.e. nil.
2. **`CatalogEntry.status`** — `reviewedEntries` filters `state == "warning"` before the cap and the manifest fetch (`:143-163`), with `test_relayers_warningStatusEntryIsExcluded` / `test_relayers_reviewStatusEntryStillMerges`.
3. **Hex→base64** — `operatorKeyBase64(fromOnymKey:)` is `DiscoveryFormat.operatorKeyHex` + `Data(lowercaseHex:)` (`:656-661`); both are `public` with the one-line rationale (`DiscoveryModels.swift:694-697`, `DiscoveryTrust.swift:450-453`). No second parser in the app target.
4. **`isValidURI` scheme case** — `components.scheme?.lowercased()` with the RFC 3986 §3.1 note (`DiscoveryModels.swift:752-758`); `HTTPS://` / `WSS://` asserts in `DiscoveryModelsTests.swift:236,263` and `test_nostr_uppercaseSchemeEndpointStillMaps` end-to-end. The division-of-labor note is at the `firstEndpointURL` call site (`DiscoverySeatAdapters.swift:266-276`).
5. **Not-dead code** — `SeatManifestFields.endpointURIs` carries the doc comment naming `ModuleConsentAcceptance.requiredEndpointURL` (`:311-315`); `SeatSelectionStore` / `ModuleSelection` remain in `Packages/OnymDiscovery/Sources/OnymDiscovery/SeatSelectionStore.swift`.

I did not re-run the suites — this is a Linux container without an iOS simulator, so the build/test numbers in the comment can't be reproduced here. If you expected a further edit on top of `db29ceb`, tell me which behavior is still off and I'll make it.

dev-onym (commented):
COMMENT — trust chain and fallback semantics are solid; two merge-semantics issues to settle before #247 wires the consent gate.

- `Sources/OnymIOS/DiscoverySeatAdapters.swift:381` — for the three endpoint adapters, the discovery row wins a URL collision, so a catalog entry listing an **already-published** endpoint takes over that row's `name`, `networks`, and `isDefault`. Effect: a third-party entry can rename an Onym-published relay and strip its "Default" badge (`NostrRelaySettingsView.swift:103`), or restate a testnet-only relayer's networks. The authorities adapter already argues legacy-must-win for exactly this reason; the same argument applies here since nothing about the discovery row is more trustworthy than the published one for a URL the published list already vouches for. Suggest legacy-wins everywhere, or keep the discovery row only for URLs legacy doesn't list.
- `Packages/OnymChain/Sources/OnymChain/RelayerRepository.swift:131` + `DiscoverySeatAdapters.swift:112` — first-launch auto-populate installs the *whole* merged list with `strategy: .random`. With `maxEntriesPerSeat = 16` discovery rows against ~2 published ones, once #247 supplies `hasActiveConsent` the published relayers receive a small minority of submissions. Worth either excluding discovery rows from auto-populate or bounding them relative to the legacy list — consenting to a provider isn't the same as consenting to it winning the random draw 8:1.
- `Sources/OnymIOS/OnymIOSApp.swift:855` — all four adapters are inert in this PR (both gate closures nil), yet `start()` seeds the default source and performs a network refresh on every launch. That's new outbound traffic with no user-visible feature and no opt-in surface yet. Intentional for this stack, or should the repository also stay dormant until #247?
- `DiscoverySeatAdapters.swift:112` — the per-seat cap is applied over the aggregate in source order, so a single enabled source that sorts first can consume all 16 slots and crowd out other providers' entries for that seat. A per-provider sub-cap would make the bound fairer.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryModels.swift:758` — `allowedScheme` is compared against a lowercased scheme but is not itself lowercased; a caller passing `"WSS"` silently matches nothing. One-line fix, cheap insurance on a newly public API.
- `SeatSelectionStore.swift:108` — `history` is documented "oldest first" but returns insertion order and `activeSelection` is the last *inserted*, not the max `selectedAt`. Fine for the current single writer; either sort by `selectedAt` or soften the doc before other callers arrive.

Nice work on the parts that matter most: the authority operator key comes from the verified catalog entry rather than the fetched manifest, the componentId/seat cross-checks are hard-enforced, `warning` entries are dropped before they cost a fetch, and the legacy fetch races ahead of the review fan-out so a slow operator can't stall auto-populate. Test coverage of the failure paths is genuinely thorough.

PR #247, opened 2026-08-13
Discovery Settings UI: providers, module consent, offers

rinat-enikeev:
## Summary

PRs 4–6 of the discovery sequence, **stacked on #246** (base branch set accordingly). Restacked 2026-08-15 as a **single commit** on top of the restacked `discovery-wiring` — the diff now shows only this branch's own Settings/consent/offers work, with all review-round fixes preserved in the squashed tree.

- **DISCOVERY section** in Settings: provider list with status chips (OK/FETCHING/FAILED/INTEGRITY/UNCONFIRMED), pinned-key fingerprints, permanent remove with confirm; **Add Discovery Provider** — https URL → TOFU screen with the operator fingerprint rendered large → confirm pins the key. Errors mapped from the typed trust/fetch errors, integrity failures distinguished by a typed flag.
- **Module consent flow** (loading → reviewing → applying → done, plus a terminal failed state for entries that can never be consented here): manifest reviewed against the catalog-pinned digest via `ServiceManifestReviewer`, operator-key/seat/componentId cross-checks against the verified catalog entry, attribution + relationship disclosure, linked terms via MarkdownDocumentView, manifest-hash card, one explicit accept → consent pin, then selection record, then seat-specific apply closure. Moderation seat excluded at three layers (picker filter, flow refusal, persistence throw).
- **Consent is the only path for catalog entries** — enforced at the adapter level: the discovery-backed known-list fetchers include a discovery-derived endpoint only when an active `PinnedConsentRecord` exists for its componentId, and serve the legacy list otherwise; consented catalogs merge into the legacy lists. The published-list dedupe is keyed off the consented entries' pinned manifest endpoints — no fetch-time registry, no read-timing dependency.
- **Offers**: FREE/model badges + detail view; free selectable through `EntitlementProviding` (incl. `FreeTierEntitlementProvider.init(reviewing:consentStore:)` so first consent can grant a manifest's free offer before any record exists), paid disabled with "purchasing not yet available".
- "FROM CATALOG" sections in the relayer and nostr pickers with consent-state chips (CONSENTED / TERMS CHANGED / REVIEW by hash match), fed by a live stream of the discovery aggregate. All new UI behind nil-default factory params — the app is unchanged when discovery is absent.

## Restack note

The branch previously carried many merge commits from the review rounds; it is now exactly one commit containing the five original review-round commits squashed (initial UI; entitlements-at-review-time + honest apply failures; surfaced notes + closed consent-gate holes; adapter-level consent gate + registry-free dedupe; legacy-list merge + terminal cross-checks + drain lifecycle), **fixed forward onto #245's round-2 throwing `PinnedConsentStore` API**: accept paths adopt `try`, gating/entitlement reads are fail-closed via `try?`, and a corrupt consent store surfaces as a non-transient error on the consent surface instead of "Try again".

**Merge order: #245 → #246 → #247, squash-merge recommended for each.**

## Testing (re-run after restack)

OnymDiscovery package 112/112, OnymFoundation package 53/53, app-target `DiscoverySeatAdaptersTests` + `RelayerSettingsFlowTests` 38/38, full app build for iPhone 17 Pro simulator. Testable logic lives in OnymDiscovery / the app test target since OnymSettings has no test target. Remaining from the plan: PR 7 (UI-test fakes + instrumented flows), blossom catalog section deliberately not included.

### DIFF
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySettingsFlow.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySettingsFlow.swift
@@ -1,0 +1,210 @@
+import Foundation
+import Observation
+
+/// `@Observable @MainActor` view-model for the Discovery Providers
+/// Settings screen. Drains `DiscoveryRepository` snapshots into local
+/// `state`; intents dispatch to the repository for source mutations.
+///
+/// Also owns the add-provider sub-flow (URL → fetched preview → TOFU
+/// confirm), because its outcome mutates the same source list this
+/// screen renders.
+///
+/// Lives in OnymDiscovery (not OnymSettings, which renders it) so its
+/// state machine is unit-testable — OnymSettings has no test target.
+@MainActor
+@Observable
+public final class DiscoverySettingsFlow {
+    /// Where the add-provider sheet currently stands. The preview is
+    /// fully verified but **nothing is pinned** until the user confirms
+    /// the operator key fingerprint (`DiscoveryRepository.addSource`
+    /// contract).
+    public enum AddPhase {
+        case idle
+        case fetching
+        case confirming(DiscoveryProviderPreview)
+        case added
+    }
+
+    public private(set) var state: DiscoveryState = .empty
+
+    // MARK: Add-provider sub-flow
+
+    public var addDraft: String = ""
+    public private(set) var addPhase: AddPhase = .idle
+    public private(set) var addError: String?
+
+    private let repository: DiscoveryRepository
+    private var snapshotTask: Task<Void, Never>?
+    /// The in-flight add-provider fetch, kept so `resetAdd` can cancel
+    /// it (internal so tests can await its completion).
+    @ObservationIgnored private(set) var addTask: Task<Void, Never>?
+    /// The in-flight TOFU confirm. NOT cancelled by `resetAdd` — the
+    /// user already confirmed the pin, so the persist + targeted
+    /// refresh must run to completion; only the sheet-phase mutation
+    /// is generation-guarded. Internal so tests can await it.
+    @ObservationIgnored private(set) var confirmTask: Task<Void, Never>?
+    /// Stale-completion guard for the add-provider fetch: bumped by
+    /// every fetch AND every reset, so a completion from a superseded
+    /// fetch (cancelled or not — `addSource` may not observe the
+    /// cancellation) can never resurrect a dismissed sheet's phase.
+    private var addGeneration = 0
+    /// Double-tap guard for the Confirm button: set the moment the tap
+    /// is accepted (NOT when `.added` lands), because the phase stays
+    /// `.confirming` while the confirm task runs and a second tap in
+    /// that window would run the whole pin + targeted-refresh pipeline
+    /// again. Cleared when the confirm task finishes (`defer` in the
+    /// task — the guard tracks the PIPELINE, not the sheet, so backing
+    /// out of the preview and fetching a fresh URL must not leave
+    /// Confirm a permanent no-op) and by `resetAdd` for the next sheet.
+    @ObservationIgnored private var isConfirmInFlight = false
+
+    public init(repository: DiscoveryRepository) {
+        self.repository = repository
+    }
+
+    /// Begin draining repository snapshots. Idempotent — safe to call
+    /// from `.task` on every appear.
+    public func start() {
+        guard snapshotTask == nil else { return }
+        snapshotTask = Task { [weak self] in
+            guard let self else { return }
+            for await snapshot in self.repository.snapshots {
+                self.state = snapshot
+            }
+        }
+    }
+
+    public func stop() {
+        snapshotTask?.cancel()
+        snapshotTask = nil
+    }
+
+    // MARK: - Read helpers
+
+    /// How many verified catalog entries a source currently contributes.
+    public func entryCount(providerId: String) -> Int {
+        state.aggregate.count { $0.source.providerId == providerId }
+    }
+
+    // MARK: - Intents
+
+    /// Confirmed removal. Permanent by design: the providerId joins
+    /// `removedDefaultProviderIds`, so no future seeding restores it —
+    /// only an explicit re-add does.
+    public func tappedRemove(providerId: String) {
+        Task { await repository.removeSource(providerId: providerId) }
+    }
+
+    public func tappedRefresh() {
+        Task { await repository.refresh() }
+    }
+
+    /// Awaitable variant for `.refreshable` (pull-to-refresh keeps its
+    /// spinner until the pass completes).
+    public func refresh() async {
+        await repository.refresh()
+    }
+
+    /// "Confirm…" on an UNCONFIRMED source (the seeded default before
+    /// its TOFU confirmation): reuse the add-provider path against the
+    /// source's own manifest URL — fetch + verify, show the operator
+    /// key fingerprint, and only pin on explicit confirm
+    /// (`confirmAddSource` replaces the unpinned source in place).
+    public func tappedConfirmSource(providerId: String) {
+        guard let status = state.sources.first(where: { $0.id == providerId }) else { return }
+        addDraft = status.source.manifestURL.absoluteString
+        tappedFetchProvider()
+    }
+
+    /// "Fetch" in the add-provider sheet: validate the typed URL, fetch
+    /// + verify the manifest, and move to the TOFU confirmation step.
+    public func tappedFetchProvider() {
+        let trimmed = addDraft.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard let url = Self.validate(trimmed) else {
+            addError = String(localized: "Enter a valid https:// URL")
+            return
+        }
+        addError = nil
+        addPhase = .fetching
+        addGeneration += 1
+        let generation = addGeneration
+        addTask?.cancel()
+        addTask = Task { [weak self] in
+            do {
+                let preview = try await self?.repository.addSource(manifestURL: url)
+                guard let self, self.addGeneration == generation, let preview else { return }
+                self.addPhase = .confirming(preview)
+            } catch {
+                guard let self, self.addGeneration == generation else { return }
+                self.addError = DiscoveryUserMessage.message(for: error)
+                self.addPhase = .idle
+            }
+        }
+    }
+
+    /// TOFU confirm: pin the previewed operator key, persist the
+    /// source, and run a TARGETED refresh so its catalogs land in the
+    /// aggregate. `refreshSource` (not `refresh()`) on purpose: a full
+    /// pass already in flight captured its source list before this add
+    /// and a coalescing `refresh()` would silently skip the new source.
+    ///
+    /// The sheet-phase mutation is guarded by `addGeneration` (same as
+    /// the fetch path): dismissing the sheet mid-confirm must not set
+    /// `.added` on a flow the user already walked away from. The pin
+    /// itself still completes — the user explicitly confirmed it.
+    public func tappedConfirmAdd() {
+        guard case .confirming(let preview) = addPhase, !isConfirmInFlight else { return }
+        isConfirmInFlight = true
+        let generation = addGeneration
+        confirmTask = Task { [weak self] in
+            guard let self else { return }
+            // The double-tap guard lasts exactly as long as the
+            // pipeline: cleared on completion so a later Confirm (new
+            // preview after Back → fresh fetch) is live again.
+            defer { self.isConfirmInFlight = false }
+            let source = await self.repository.confirmAddSource(preview)
+            if self.addGeneration == generation {
+                self.addPhase = .added
+            }
+            await self.repository.refreshSource(providerId: source.providerId)
+        }
+    }
+
+    /// Back out of the TOFU confirmation without pinning anything.
+    public func tappedCancelPreview() {
+        if case .confirming = addPhase {
+            // Backing out supersedes this preview the same way a
+            // sheet dismissal does: bump the generation so an
+            // in-flight confirm's completion can never surface
+            // `.added` over whatever the user does next (a fresh
+            // URL's fetch → confirm cycle included). The pin itself
+            // still completes if a confirm was already accepted.
+            addGeneration += 1
+            addPhase = .idle
+        }
+    }
+
+    /// Reset the add-provider sub-flow (sheet dismissed). Invalidates
+    /// any in-flight fetch: the task is cancelled AND the generation
+    /// bump guarantees its completion can't set `addPhase` afterwards,
+    /// so a dismissed sheet never resurrects as `.confirming`.
+    public func resetAdd() {
+        addGeneration += 1
+        addTask?.cancel()
+        addTask = nil
+        addDraft = ""
+        addPhase = .idle
+        addError = nil
+        isConfirmInFlight = false
+    }
+
+    /// Discovery manifests are https-only (profile URI rules) — this
+    /// is stricter than the relayer picker's http allowance on purpose.
+    public static func validate(_ raw: String) -> URL? {
+        guard let url = URL(string: raw),
+              url.scheme?.lowercased() == "https",
+              let host = url.host(), !host.isEmpty
+        else { return nil }
+        return url
+    }
+}
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentAcceptance.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentAcceptance.swift
@@ -1,0 +1,90 @@
+import Foundation
+import OnymFoundation
+
+/// Why a generic module consent could not be recorded.
+public enum ModuleConsentError: Error, Equatable, Sendable {
+    /// The moderation seat's consent is a *signed mandate*
+    /// (OnymModeration's flow) — it must never be reduced to this
+    /// generic local pin.
+    case moderationSeatExcluded
+}
+
+/// Why a recorded consent could not be *applied* to the seat's legacy
+/// configuration. Thrown by the seat-specific apply closures so the
+/// flow can tell the user the truth: the consent pin exists, but the
+/// service was not added.
+public enum ModuleApplyError: Error, Equatable, Sendable {
+    /// The consented manifest lists no endpoint with the scheme the
+    /// seat requires (e.g. no https endpoint for a notary, no wss
+    /// endpoint for a message relay).
+    case noUsableEndpoint
+}
+
+/// The accept path of the generalized module consent flow, as a pure
+/// helper so its ordering is unit-testable without any UI: pin the
+/// reviewed manifest, then record the selection provenance that
+/// references it.
+public enum ModuleConsentAcceptance {
+    /// The one seat type this generic consent must refuse.
+    public static let excludedSeatType = "moderation"
+
+    /// Record consent to a reviewed manifest picked from a discovery
+    /// catalog.
+    ///
+    /// Ordering is deliberate: the consent pin is written **first**,
+    /// the `ModuleSelection` second — the selection references the
+    /// consented hash, so an interruption between the two leaves a
+    /// consent without a selection (harmless history) rather than a
+    /// selection whose consent was never pinned. Applying the derived
+    /// endpoint to the seat's legacy configuration is the caller's
+    /// final step, after both records exist.
+    @discardableResult
+    public static func accept(
+        reviewed: ReviewedServiceManifest,
+        source: SourceAttribution,
+        offerId: String?,
+        consentStore: any PinnedConsentStore,
+        selectionStore: any SeatSelectionStore,
+        acceptedAt: Date = Date()
+    ) throws -> (record: PinnedConsentRecord, selection: ModuleSelection) {
+        let manifest = reviewed.signedManifest
+        guard manifest.seat != excludedSeatType else {
+            throw ModuleConsentError.moderationSeatExcluded
+        }
+        let record = try consentStore.accept(
+            reviewed,
+            sourceLabel: source.sourceLabel,
+            offerId: offerId,
+            acceptedAt: acceptedAt
+        )
+        let selection = ModuleSelection(
+            seatType: manifest.seat,
+            componentId: manifest.componentId,
+            providerId: source.providerId,
+            manifestHash: manifest.manifestHash,
+            offerId: offerId,
+            selectedAt: acceptedAt
+        )
+        selectionStore.record(selection)
+        return (record, selection)
+    }
+
+    /// The endpoint a seat's apply step writes into its legacy
+    /// configuration: the first of `endpointURIs` (published order)
+    /// whose URL scheme matches `scheme`, case-insensitively.
+    ///
+    /// Throws `ModuleApplyError.noUsableEndpoint` instead of returning
+    /// nil so a manifest without a usable endpoint surfaces as an
+    /// error the consent flow must show — never as a silent no-op
+    /// behind a "Service added" screen.
+    public static func requiredEndpointURL(
+        in endpointURIs: [String],
+        scheme: String
+    ) throws -> URL {
+        guard let url = endpointURIs
+            .compactMap(URL.init(string:))
+            .first(where: { $0.scheme?.lowercased() == scheme.lowercased() })
+        else { throw ModuleApplyError.noUsableEndpoint }
+        return url
+    }
+}
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift
@@ -1,0 +1,405 @@
+import Foundation
+import Observation
+import OnymFoundation
+
+/// Everything a per-seat picker needs to render its "From catalog"
+/// section, bundled so the app's composition root can hand one value
+/// to each settings flow. Optional everywhere it's consumed — when the
+/// app runs without discovery (UI-test harness), the pickers render
+/// exactly as before.
+@MainActor
+public struct DiscoveryModulePicker {
+    /// Verified catalog entries for this picker's seat type (the app
+    /// pre-binds the seat).
+    public let entries: () async -> [AttributedCatalogEntry]
+    /// The active pinned consent for a componentId, if any — drives
+    /// the consent-state text on catalog rows.
+    public let activeConsent: (_ componentId: String) -> PinnedConsentRecord?
+    /// Build the consent flow for one entry. The app pre-binds the
+    /// seat-specific `apply` closure that writes the endpoint into the
+    /// legacy configuration.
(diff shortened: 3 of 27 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the free-offer selection path is wired to a lookup that is empty at the moment it's consulted, so the headline "free selectable" behavior can't work on a first consent.

- **Free offers are never selectable on first consent** (`Sources/OnymIOS/OnymIOSApp.swift:575` + `Packages/OnymSettings/Sources/OnymSettings/ModuleConsentFlow.swift:204`): `FreeTierEntitlementProvider(consentStore:)` resolves an offer via `consentStore.activeRecord(componentId:)?.consentedManifest()` (`Packages/OnymFoundation/Sources/OnymFoundation/Entitlement.swift`), but `computeEntitlements` runs *before* any consent exists for that component. `activeRecord` → nil → lookup → nil → `entitledOfferIds` is always empty on the path that matters. Consequences: every offer radio renders disabled (`ModuleConsentView.swift:229`), the "Paid offers can't be selected yet" footnote (`ModuleConsentView.swift:209`) shows even for a free-only manifest, `selectedOfferId` stays nil so the pin and `ModuleSelection` record `offerId: nil`, and `DiscoveryCatalogSection.consentedOffer` can therefore never show a badge. Fix: build the provider with `FreeTierEntitlementProvider(offerLookup:)` over the manifest currently under review (or fall back to the reviewed manifest when no active record exists).
- **Silent no-op apply reported as success** (`OnymIOSApp.swift:597-600`, `:629-632`; `ModuleConsentFlow.swift:233`): if the manifest carries no `https`/`wss` endpoint, `apply` returns without adding anything, yet the flow still shows "Service added — Its endpoint is now in your configured list", after consent + selection are already persisted. Re-entering the picker then shows CONSENTED with no endpoint. `apply` should be able to report failure (throwing or `-> Bool`) and the surface should say so.
- **INTEGRITY chip paired with a generic message** (`Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryUserMessage.swift:9-30`): `isIntegrityFailure` returns true for `policyUnavailable`, `entryManifestMismatch`, `entryManifestInvalid`, `sourceConflict`, but `message(for:)` has no cases for them, so those render red "INTEGRITY" next to "Couldn't refresh the provider." — the least informative string for the most serious class. Add cases (and a test row; `DiscoveryUserMessageTests` asserts the classification for these but never their message).
- **Stale add-provider state survives dismissal** (`DiscoverySettingsFlow.swift:83-96` with `DiscoverySettingsView.swift` `onDismiss: { flow.resetAdd() }`): dismissing while `.fetching` runs `resetAdd()`, then the in-flight task sets `addPhase = .confirming(preview)`; reopening the sheet lands directly on a TOFU confirm screen for the abandoned URL. Cancel/ignore the fetch task in `resetAdd()` (or tag it with a generation counter).
- Nit (`ModuleConsentFlow.swift:238`): the `catch` around `ModuleConsentAcceptance.accept` maps every error, including `moderationSeatExcluded`, to "Couldn't record consent. Try again." — retry will never succeed for that one.

Good: the accept ordering (consent pin before selection) is both deliberate and test-pinned; the operator-key and seat cross-checks against the verified catalog entry (`ModuleConsentFlow.swift:180-192`) are the right hard checks, and the nil-default factory seams do keep the discovery-less app untouched.

dev-onym (commented):
REQUEST_CHANGES — the flows are well-factored and the trust checks are careful, but the new Discovery screen drops surfacing the repository already computes, and the consent gate has holes.

- `Packages/OnymSettings/Sources/OnymSettings/DiscoverySettingsView.swift:137` — `DiscoverySourceStatus.notes` is never rendered. Those notes are the only surfacing of "Needs confirmation: confirm this source's operator key", skipped catalogs, forward-jumps, policy transitions and §9 `result_incomplete` (`DiscoveryRepository.swift:249`, `:365`). An unconfirmed or partly-skipped source now shows chip + "0 catalog entries" and no explanation — exactly the "silently empty source" the repository comments say must never happen.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySettingsFlow.swift:84` — `tappedRefresh()` has no caller (no button, no `.refreshable`). A source badged FAILED/INTEGRITY has no retry path short of relaunching the app, and an UNCONFIRMED seeded default can never be advanced from this screen.
- `DiscoverySettingsFlow.swift:116` + `DiscoveryRepository.swift:214` — `tappedConfirmAdd` persists the source then calls `refresh()`, which *coalesces onto an in-flight pass* whose source list was captured before the add. If the user adds a provider while `start()`'s refresh is running, the new source is never fetched, yet the sheet says "Its catalogs are being fetched and verified now" and the row renders OK / 0 entries.
- `Packages/OnymSettings/Sources/OnymSettings/ModuleConsentView.swift:147` — Accept isn't gated on an entitled offer. A manifest carrying only paid offers leaves `selectedOfferId == nil`, Accept stays enabled, and the service is consented + written into the legacy config with `offerId: nil` — adopting a paid service while the footnote says purchasing isn't available.
- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsView.swift:35` vs `:38` — the same discovery entries reach PUBLISHED through `DiscoveryBackedKnownRelayersFetcher` (tap-to-add, no consent, no attribution) and FROM CATALOG (consent flow). Duplicate rows, and the consent gate is one tap away from being bypassed. Same for nostr.
- `Packages/OnymSettings/Sources/OnymSettings/ModuleConsentFlow.swift:194-198` — operatorKey and seat are cross-checked against the entry, but `componentId` isn't. A manifest whose `componentId` differs from the entry's records consent under one id while `DiscoveryCatalogSection` looks it up by the entry's id: the row stays "REVIEW" permanently and each tap appends another consent record.
- `DiscoveryCatalogSection.swift:48` vs `ModuleConsentView.swift:review` — the §4.2 `status` warning/review chip appears on the picker row but not on the consent surface where the user actually accepts (and `status.uri` is never surfaced anywhere). The disclosure should be strongest at the accept step.
- `ModuleConsentFlow.swift:235` — after `ModuleApplyError.noUsableEndpoint` the step returns to `.reviewing` with Accept re-enabled; each retry deterministically fails again but re-runs `accept`, appending a fresh consent record + `ModuleSelection` every tap. Guard on "consent already pinned for these bytes" and retry only `apply`.
- `DiscoverySettingsFlow.swift:116` — the confirm path lacks the `addGeneration` guard the fetch path got: dismissing the sheet mid-confirm still sets `addPhase = .added`, so reopening Add shows the "Provider added" screen.
- `DiscoverySettingsView.swift:106` and `:157` — `settings.discovery.source.<id>` is set on both `SwipeToDeleteRow` and the nested `sourceRow`; nested duplicate identifiers make the PR-7 UI-test queries ambiguous.
- Nits: `OfferBadge.swift:14` `offer.model.uppercased()` renders an unlocalized wire token; `DiscoverySettingsView.swift:181` `"\(count) catalog entries"` isn't a plural rule; `DiscoveryCatalogSection.swift:38,101` calls `activeConsent` + `consentedOffer`, each decoding all of UserDefaults per row per render.

dev-onym (commented):
REQUEST_CHANGES — the consent gate this PR builds is bypassed by two paths that leave it closed only in the happy case; the rest is solid work.

**Consent-gate holes**
- `Packages/OnymChain/Sources/OnymChain/RelayerRepository.swift:131` — first-launch auto-populate installs *the whole fetched known list* when `!hasUserInteracted`, and since #246 that list is discovery-derived (`DiscoveryBackedKnownRelayersFetcher`). So on a fresh install every catalog relayer lands in the configured endpoints with **zero** taps — no attribution, no pinned consent record, no `ModuleSelection`. Same shape for nostr (`NostrRelaysRepository.applyDefault`, reached via `refresh()` when `!hasUserInteracted`) and blossom. The one-tap hole is closed while the zero-tap one is wide open; the PR body's "the consent flow is the only path for catalog entries" isn't true yet.
- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:83` + `Sources/OnymIOS/DiscoverySeatAdapters.swift:218` — the dedupe reads the registry *once*, at `.task`, but the registry is only written after `fetchLatest()` completes in this process. `knownList` is restored from `store.loadCachedKnownList()` at repo init, so: open Relayer settings at boot → cached discovery URLs render in the tap-to-add published list, registry still empty, `unconfiguredKnownList` (`:160`) filters nothing. If the fetch resolves later in the session the screen never re-reads it. Same on any launch where the fetch fails. Subscribe to the registry (or key the dedupe off entry-derived URLs from `catalogEntries`, which needs no registry at all).

**Correctness**
- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryRepository.swift:462` — `confirmAddSource` replaces the source with a fresh `DiscoverySource`, so `lastAccepted` (sequence + digest, the rollback chain) resets to `[:]` while `store.removeRetainedSnapshots` is *not* called. This PR makes that path user-reachable two ways (Add Provider with an already-added URL; `tappedConfirmSource`), and after it a provider can serve an older snapshot and have it accepted. Carry `lastAccepted` forward when the providerId already exists.
- `DiscoveryRepository.swift:457` — same call replaces `userLabel` with `manifestURL.host()`, so confirming the seeded default turns "Onym Discovery" into "discovery.onym.app" everywhere `sourceLabel` is disclosed (`DiscoveryCatalogSection` subtitle, consent attribution banner). Preserve the existing label on in-place confirm.

**Should fix**
- `RelayerSettingsFlow.swift:80` / `NostrRelaySettingsFlow.swift:80` — the pickers snapshot the aggregate once instead of draining `repository.snapshots` like `DiscoverySettingsFlow` does. At boot the discovery refresh usually lands *after* the screen appears, so FROM CATALOG renders empty until the user navigates away and back.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift:200,205` — both guards set `errorMessage` while leaving `step == .loading`, and the loading view's only affordance is "Retry", which can never succeed for a moderation entry or a missing manifest URI. Give these a terminal state.
- `DiscoveryRepository.swift:256` — `refreshSource` writes the *global* `fetchStatus`, so one added source failing flips the whole screen to `.failed` (and races an in-flight full pass's status).
- `DiscoverySettingsFlow.swift:146` — `tappedConfirmAdd`'s guard only clears once the task sets `.added`; a double tap runs `confirmAddSource` + `refreshSource` twice.

Tests are genuinely good — the interleaved-refresh and retry-after-apply-failure cases are the right ones to have pinned.

dev-onym (commented):
REQUEST_CHANGES — solid consent architecture and test coverage, but the consent gate + published-list dedupe combine to silently delete the legacy relayer list.

- `Sources/OnymIOS/DiscoverySeatAdapters.swift:213` — the adapter *replaces* rather than merges: as soon as one catalog notary is consented, `endpoints` is non-empty so the GitHub-published relayers are never fetched. Combined with the new dedupe at `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:188`, `unconfiguredKnownList` becomes empty and the picker renders "All published relayers added." (`RelayerSettingsView.swift:219`) — the user loses the entire published list after a single consent and can only get those relayers back by typing custom URLs. Replace-not-merge predates this PR, but the consent gate makes it reachable in the normal path. Merge legacy + consented (dedupe by URL) instead of falling back only when empty.
- Same at `:254` / `:294` for nostr/blossom: one consented `transport.message` entry makes the legacy `nostr-relays.json` defaults disappear from the known list, which is also what `applyDefault`/first-launch auto-populate reads.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift:245,249,258` — the operator-key / seat / componentId cross-check failures leave `step == .loading`, so the view shows the error card **with a Retry button**. The manifest is digest-pinned, so a refetch can only yield the same mismatched bytes; these are as terminal as the moderation-seat case the PR models with `.failed` (`:222`, `:229`). Route them to `.failed` for consistency with the flow's own "no Retry that can't succeed" rule.
- `DiscoverySettingsFlow.swift:67` — `guard let self else { return }` is *outside* the `for await`, so the task holds a strong `self` for its lifetime and the repository stream never finishes. `stop()` is never called anywhere in the app (`.task { flow.start() }` only), so every push of Discovery Settings leaks a flow plus a live continuation in `DiscoveryRepository.continuations`; each `publish()` then fans out to all stale subscribers. Same accumulation for `catalogTask` at `RelayerSettingsFlow.swift:69` / `NostrRelaySettingsFlow.swift:65` (weak self there, but the task and its subscription still never end). Either call `stop()` from an `.onDisappear`, or drain inside a `.task` so SwiftUI cancels it.

Smaller notes:
- `ModuleConsentFlow.swift:258` region: after a failed load, `selectedOfferId` from a previous review isn't cleared before `computeEntitlements` re-runs; `canAccept` then depends on the stale id surviving in `entitledOfferIds`. Harmless today, fragile if offers change between retries.
- `DiscoveryUserMessage.swift` classifies `snapshotExpired` as an integrity failure, so a provider that simply stopped republishing gets the red INTEGRITY chip rather than FAILED — worth double-checking that's the intended signal.
- `DiscoverySettingsFlowTests.testDoubleTapConfirmRunsThePipelineOnce` relies on a 100 ms sleep to let a hypothetical second task surface; consider asserting on a deterministic signal instead.

dev-onym (commented):
**COMMENT** — solid, well-tested work; three issues worth addressing before merge, none blocking the architecture.

- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySettingsFlow.swift:167` — `tappedCancelPreview()` returns to `.idle` but never clears `isConfirmInFlight` (only `resetAdd()` does, on sheet dismiss). Tap "Pin Key & Add Provider", then "Back" during the confirm's persist + targeted refresh: the flag stays set, so every later Confirm in that same sheet session hits the `!isConfirmInFlight` guard at `:153` and silently does nothing — a user who then fetches a *different* provider gets a dead Confirm button with no feedback until they dismiss the sheet. Same path also lets the in-flight confirm set `.added` afterwards (`tappedCancelPreview` doesn't bump `addGeneration`, unlike `resetAdd`), so "Back" can bounce to "Provider added".
- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoveryRepository.swift:441` + `:472` — re-confirming an **already-pinned** provider through the Add sheet is a silent key rotation: `addSource` verifies with `pinnedOperatorKeyHex: nil` (pure TOFU) and `confirmAddSource` overwrites `pinnedOperatorKeyHex` with whatever the URL now serves, while carrying `lastAccepted` forward. The new in-place-update behavior is right for the seeded-default confirm, but nothing distinguishes "first use" from "this provider's key changed" — and `AddDiscoveryProviderView.swift` copy explicitly frames the screen as first-use ("It's pinned on confirm"). Suggest: if `currentSource(providerId)?.pinnedOperatorKeyHex` exists and differs, surface a key-change warning (or refuse) rather than rendering the fingerprint as a fresh pin.
- `Sources/OnymIOS/DiscoverySeatAdapters.swift:235-236` (same at `:279`, `:326`) — the consent gate is applied *after* `reviewedEntries` has already fetched every catalog entry's manifest. On every known-list refresh the app contacts the operator URL of services the user never consented to (and never will see, since the gate drops them). Filtering `catalog.entries(seatType)` by `hasActiveConsent` before the fetch/review loop gets the same result without the unconsented network contact. Pre-existing shape, but the gate is new and makes the wasted fetch pure cost.

Smaller notes:
- `Packages/OnymFoundation/Sources/OnymFoundation/PinnedConsentRecord.swift:69` — `consentedEndpointURLs()` dedupe against the published list is exact `URL` equality; a trailing-slash or case difference between the manifest URI and the GitHub list re-exposes the bare row. Normalizing (or comparing on host+path) would make `RelayerSettingsFlow.unconfiguredKnownList` robust.
- `Packages/OnymSettings/Sources/OnymSettings/AddDiscoveryProviderView.swift` confirm step — the Confirm button has no in-flight state (unlike Fetch's `ProgressView`), which is what makes the double-tap/Back window at issue 1 user-visible in the first place.

Verified good: `refreshSource` per-catalog merge keeps `fetchStatus` and interleaved-pass state correct; terminal `.failed` cross-checks (operator key / seat / componentId) are genuinely unretryable and tested; `alreadyPinned` retry guard prevents duplicate consent records; drain `stop()` → continuation removal is covered by `subscriberCount`; moderation exclusion holds at all three layers.

rinat-enikeev:
Restacked as a single commit (1b7c96d) on top of the restacked `discovery-wiring` (#246). The diff now shows only this branch's own Settings/consent/offers work. All 5 review-round commits are preserved in the squashed tree (initial UI; entitlements at review time + honest apply failures; surfaced notes + consent-gate holes closed; adapter-level consent gate + registry-free dedupe; legacy-list merge + terminal cross-checks + drain lifecycle), fixed forward onto #245's round-2 throwing store API (try/try? adoption, fail-closed gating, corrupt store surfaced as non-transient on the consent surface). Foundation/repository changes that were meanwhile upstreamed into merged #244 / rebased #245 were dropped from this diff rather than duplicated. Re-verified after restack: OnymDiscovery 112/112, OnymFoundation 53/53, app-target adapter + relayer-flow tests 38/38, full app build (iPhone 17 Pro sim). Merge order 245 → 246 → 247, squash recommended.

dev-onym (commented):
**REQUEST_CHANGES** — solid state machines and test coverage; one trust-boundary gap in the consent gate plus a few smaller issues.

- `Sources/OnymIOS/OnymIOSApp.swift:138` — the known-list consent gate is keyed on **componentId only**, but consent is pinned to *exact bytes* (`PinnedConsentRecord.manifestHash`). After an operator republishes a manifest under the same componentId with a different endpoint, the catalog entry's digest changes, `reviewEntry` happily reviews the new bytes, and `hasActiveDiscoveryConsent` still returns true from the *old* record — so an endpoint the user never reviewed enters the relayer/nostr/blossom known lists and first-launch auto-populate. The UI has a whole "TERMS CHANGED" state for exactly this situation (`DiscoveryCatalogSection.swift:88`), so it's an anticipated state that the gate doesn't enforce. The seam needs the entry digest (or the record's hash) passed through, not just the componentId.
- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:191` — same root cause on the dedupe: `discoverySourcedKnownURLs` is built from the *pinned* record's bytes (`consentedEndpointURLs()`), so after a terms change the merged known-list row carries the **new** URL, isn't in the set, and renders as a bare tap-to-add published row alongside the "TERMS CHANGED" catalog row — the duplicate the dedupe exists to prevent.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySettingsFlow.swift:154` — `isConfirmInFlight` is set on tap and cleared only in `resetAdd()`. "Back" (`:167`) is enabled while the confirm task runs, so: Confirm → Back → new URL → Fetch → `.confirming` → **"Pin Key & Add Provider" is a permanent no-op** for the rest of the sheet session, with no error shown. Clear it when `confirmTask` finishes (and/or bump `addGeneration` in `tappedCancelPreview`).
- `Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift:326` — the retry-after-apply-failure guard keys only on `manifestHash`, so if the user changes `selectedOfferId` after an apply failure and taps Accept again, the apply is retried but the pinned record and `ModuleSelection` keep the *original* offerId — the recorded commercial terms silently disagree with what the screen shows selected.
- `Packages/OnymDiscovery/Tests/OnymDiscoveryTests/ModuleConsentFlowTests.swift` — no test covers the `PinnedConsentStoreError.corruptStore` accept branch (`ModuleConsentFlow.swift:345`), which the restack note calls out as the headline fix-forward; the fake store's `load()` can never throw. A throwing fake would pin that message and its non-retryable framing.
- Nit, `Sources/OnymIOS/OnymIOSApp.swift:138` — the closure does a full defaults read + decode of *all* consent records per catalog entry per fetch (up to 16 × 4 seats); one load per fetch pass would do.

Not blocking, verified as claimed: the accessibility ids in `DiscoverySettingsView.swift:117` don't nest duplicates (`SwipeToDeleteRow` only derives `<id>.delete`), the merge keeps the discovery URL on collision, and the `.failed` terminal states are consistently non-retryable.

rinat-enikeev:
Fifth review addressed in a5f1bd5 (amended into the single commit, force-with-lease). Point by point:

**1. Stale consent gate (digest not compared) — the substantive catch of this round, and credit where due: consent-to-bytes is the whole model here, and the one place that turns consent into live endpoints was comparing componentIds while everything around it (the pinned record, the reviewer, the TERMS CHANGED chip) speaks in exact bytes.** Fixed exactly as suggested: the entry's manifest digest is threaded through the seam. The three endpoint adapters now take a `makeConsentGate` factory returning a `(componentId, manifestDigest) -> Bool` check and pass `entry.manifest.digest` per entry; the app's gate (`OnymIOSApp.swift`) answers `true` only when the ACTIVE `PinnedConsentRecord`'s `manifestHash` equals that digest (fail-closed on a corrupt store, as before). After a republish, the old record admits nothing — the entry drops out of the known lists and auto-populate until the user walks the TERMS CHANGED re-consent path. Note this reshapes `DiscoverySeatAdapters.swift` (a #246-commit file), so the seam change now rides in this PR's diff. Tests: `test_relayers_consentGateExcludesRepublishedManifestWithStaleConsent` (digest mismatch excluded despite an active record; matching digest still merges).

**2. Terms-changed duplicate row.** Verified closed by fix 1: `RelayerSettingsFlow.discoverySourcedKnownURLs` derives from pinned bytes, and with the digest-comparing gate the only discovery URLs that can reach the known list are ones whose entry digest equals the pinned hash — i.e. exactly the URLs the dedupe set covers. A terms-changed entry's new URL never enters the known list at all, so no bare published row can render next to the TERMS CHANGED catalog row. Regression test at the screen level: `test_unconfiguredKnownList_termsChangedEntryDoesNotOfferNewURLAsPublishedRow` (real adapter + app-shaped gate + repository + flow; asserts the record/digest mismatch that drives the TERMS CHANGED chip AND that the new URL is absent from the tap-to-add list).

**3. `isConfirmInFlight` never cleared.** Both halves applied: the confirm task clears the flag in a `defer` (the guard tracks the pipeline, not the sheet), and `tappedCancelPreview` bumps `addGeneration` so the superseded confirm's completion can never surface `.added` over the next cycle. Test walks the exact sequence: `testConfirmThenBackThenNewFetchCanConfirmAgain` (Confirm → Back → URL → Fetch → Confirm → `.added`, plus the no-resurrect assertion after Back).

**4. Retry-after-apply-failure with a changed offer.** Took the simpler-correct option: the already-pinned probe now keys on `manifestHash` AND `selectedOfferId`, documented at the check. A retry with a changed selection takes the full accept path, and the store's accept already deactivates-and-appends, so the old record is superseded cleanly — the pinned record and the `ModuleSelection` both carry the offer on screen. Test: `testRetryWithChangedOfferRecordsTheNewOffer` (free-a apply-fails → switch to free-b → retry → active record + selection say free-b, one active record).

**5. Corrupt-store accept branch.** Test added — and in writing it, aligned the framing: the branch now lands in the terminal `.failed` step (rendered without any retry affordance) instead of `.reviewing`, since the store refuses every accept until the corrupt history is explicitly cleared — the same no-retry-that-can't-succeed rule the entry cross-checks follow, and what the branch's own "not transient" comment already promised. `testCorruptStoreOnAcceptIsTerminalFailedWithHonestMessage`: throwing fake store → the "couldn't be read / Nothing was changed" message, `.failed`, no selection written, `tappedRetry` inert.

**6. Per-fetch consent-store load.** Folded into fix 1's seam shape: the gate is a factory invoked once at the top of each `fetchLatest` pass, which loads the store once into a `[componentId: manifestHash]` dict; per-entry checks are dictionary hits. Staleness is bounded to a single pass — each pass re-invokes the factory, so no cross-pass cache to invalidate. Test: `test_relayers_consentGateFactoryRunsOncePerFetchPass`.

Verified, all green: OnymDiscovery package tests (117), OnymFoundation package tests (53), app-target `DiscoverySeatAdaptersTests` + `RelayerSettingsFlowTests` (55), full app build on the iPhone 17 Pro simulator.

dev-onym (commented):
COMMENT — solid, well-tested state machines and a genuinely tight consent gate; findings below are niche correctness/perf issues, none blocking.

- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:191` — the published-list dedupe compares **raw `URL` equality** against `consentedEndpointURLs()`, but the adapter merge dedupes with `endpointDedupeKey` (lowercased scheme/host, trailing slash stripped) and **legacy wins collisions**. If a consented catalog service's endpoint also appears in the legacy list under a normalization-equivalent-but-different string (`https://notary.example/api` vs the manifest's `https://Notary.Example/api/`), the known list carries the *legacy* string, `discoverySourcedKnownURLs` doesn't match, and the bare tap-to-add row renders next to the "FROM CATALOG" row — exactly the duplicate this set exists to prevent. Suggest exporting `endpointDedupeKey` (or an equivalent normalizer next to `PinnedConsentRecord.consentedEndpointURLs()`, `PinnedConsentRecord.swift:87`) and comparing keys on both sides. The existing test uses byte-identical URLs, so it passes either way.
- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:98` / `NostrRelaySettingsFlow.swift:92` — `applyCatalog` calls `discovery.activeConsent(componentId)` once per unique componentId, and the production binding (`OnymIOSApp.swift:674`, `:713`) is `pinnedConsentStore.activeRecord(...)`, i.e. a **full defaults read + decode of every record (manifest bytes included) per component**, on the MainActor, on every catalog stream tick. The memoization the comment describes only removes the per-row-per-render cost. One `load()` reduced into a `[componentId: record]` dictionary — the same shape `makeDiscoveryConsentGate` already uses — would make this one decode per refresh.
- `Packages/OnymSettings/Sources/OnymSettings/RelayerSettingsFlow.swift:191` — while the catalog stream hasn't delivered its first value (or `discovery` is wired but the aggregate is still empty), `discoverySourcedKnownURLs` is empty, so a consented service does briefly render as a bare published row before the catalog section appears. Cosmetic/transient, but it's the same state the dedupe targets.
- `Packages/OnymSettings/Sources/OnymSettings/DiscoverySettingsView.swift:214` — the `FETCHING` chip keys off the **global** `flow.state.fetchStatus`, so during any refresh every source reads FETCHING, including ones not being fetched, and it masks the `UNCONFIRMED` state (the branch below it) for a seeded default. Worth keying off per-source status if `DiscoverySourceStatus` can carry it.
- `Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift:149` — `displayName` and `termsURL` each run a full `JSONSerialization.jsonObject` over `rawBytes` on every access, and both are read from the SwiftUI body; that's two manifest re-parses per body evaluation. Computing them once when `reviewed` is set would be cheaper and keeps them stable.
- Nit: `DiscoveryModulePicker.makeConsentFlow` never passes `onAccepted` in the app wiring, so the documented "presenter can dismiss and refresh its consent badges" hook is dead in production — the refresh happens via `.sheet(onDismiss:)` instead. Either wire it or drop the parameter so the doc comment doesn't describe behavior that doesn't exist.

PR #249, opened 2026-08-15
Add OnymOnboarding package: flow state machine, store, step scaffolding

rinat-enikeev:
PR 1 of the first-launch onboarding plan (iOS Onboarding module). **Package-only** — the app target is unchanged: no `project.yml` edit, no registration in `OnymIOSApp`, no RootView gate. The package compiles and tests standalone, exactly like `OnymDiscovery`.

## What's here

New local package `Packages/OnymOnboarding` (zero dependencies — not even OnymFoundation was needed):

- **`OnboardingStore`** — protocol + `UserDefaultsOnboardingStore` under `app.onym.ios.onboarding.completed`, plus `OnboardingGate.shouldOnboard(store:isExistingUser:)`: completed flag OR the injected existing-user signal (the app will pass "any `hasUserInteracted` or a mandate present") means no onboarding, so upgraders are grandfathered. The probe is a pure read and is only consulted when the flag is absent.
- **`OnboardingFlow`** (`@MainActor @Observable`, modeled on `ModuleConsentFlow`) — steps `welcome → discoveryConfirm → messageTransport → blobTransport → notary → moderation → done`, `advance()/skip()/back()/complete()`, `stepIndex/stepCount` for the indicator. Every step skippable except moderation while the authority directory has entries (injected `() async -> Bool` probe, resolved by `start()`) and the terminal `done`. Outcomes recorded per step as `StepOutcome.consented(componentId:)/skipped/notApplicable` — the flow models step results abstractly and never links against OnymDiscovery/OnymModerationUI. `complete()` is guarded to the `done` step, writes the flag, fires `onCompleted`.
- **`OnboardingStepScaffold` + `OnboardingView`** — the established consent-view skeleton in plain SwiftUI (no OnymDesign dependency yet); `stepContent: (OnboardingStep) -> AnyView` and `stepIndicator: (Int, Int) -> AnyView` are injected slots with built-in placeholders; Welcome and Done have real built-in content (Done renders the per-step outcome summary + backup nudge). Accessibility ids `onboarding.<step>.<element>`. `interactiveDismissDisabled` presentation semantics documented on `OnboardingView`.
- **27 unit tests** (isolated per-test UserDefaults suites): step order, back/skip paths, moderation skippability both ways, completion flag + `onCompleted`, grandfathering truth table, outcome recording.

## Verification

`xcodebuild test -scheme OnymOnboarding -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` — 27/27 green.

## What lands next

- **PR 2 (unblockers, in parallel on `onboarding-unblockers`)**: app-target seams this package will consume — blossom catalog picker + apply closure, relayer auto-populate suppression.
- **PR 3 (wiring)**: `project.yml` registration, RootView `fullScreenCover` gate, `SettingsStepIndicator` made public and passed through the indicator slot, real step content (discovery confirm, catalog pickers via `ModuleConsentFlow`, `ModerationConsentFlow(mode: .onboarding)`) injected through `stepContent`, `--reset-keychain` clearing the flag.

### DIFF
--- a/Packages/OnymOnboarding/Package.swift
+++ b/Packages/OnymOnboarding/Package.swift
@@ -1,0 +1,25 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
+    name: "OnymOnboarding",
+    platforms: [.iOS("18.0")],
+    products: [
+        .library(name: "OnymOnboarding", targets: ["OnymOnboarding"]),
+    ],
+    targets: [
+        // Deliberately dependency-free: the flow models each step's
+        // outcome abstractly and receives every collaborator as an
+        // injected closure, so the state machine never links against
+        // OnymDiscovery / OnymModerationUI / OnymDesign. The app's
+        // composition root (PR 3) supplies the step content and the
+        // step indicator from those packages.
+        .target(
+            name: "OnymOnboarding"
+        ),
+        .testTarget(
+            name: "OnymOnboardingTests",
+            dependencies: ["OnymOnboarding"]
+        ),
+    ]
+)
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
@@ -1,0 +1,181 @@
+import Foundation
+import Observation
+
+/// The seven first-launch steps, in presentation order.
+public enum OnboardingStep: String, CaseIterable, Equatable, Hashable, Sendable {
+    case welcome
+    case discoveryConfirm
+    case messageTransport
+    case blobTransport
+    case notary
+    case moderation
+    case done
+}
+
+/// What happened at one step, modeled abstractly — the flow never
+/// links against the packages whose surfaces produce these outcomes
+/// (OnymDiscovery's module consent, OnymModerationUI's mandate flow).
+/// The app-layer step content (PR 3) reports back through this enum.
+public enum StepOutcome: Equatable, Sendable {
+    /// The user consented at this step. `componentId` names the chosen
+    /// component when the step's surface knows one (catalog picks);
+    /// nil for consents without a component identity (e.g. confirming
+    /// the seeded discovery source).
+    case consented(componentId: String?)
+    /// The user skipped the step; the seat falls back to today's
+    /// defaults.
+    case skipped
+    /// The step had nothing to decide (informational steps: welcome,
+    /// done, moderation-with-empty-directory).
+    case notApplicable
+}
+
+/// State machine for the first-launch onboarding sequence. Modeled on
+/// `ModuleConsentFlow` / `ModerationConsentFlow`: `@MainActor
+/// @Observable`, every collaborator injected as a closure so the
+/// machine is unit-testable and the package stays dependency-free.
+///
+/// Presentation contract (PR 3): RootView presents `OnboardingView` as
+/// a `fullScreenCover` with `.interactiveDismissDisabled(true)` — the
+/// only exits are `complete()` on the Done step and the per-step Skip
+/// affordances; there is no swipe-to-dismiss.
+@MainActor
+@Observable
+public final class OnboardingFlow {
+    public private(set) var step: OnboardingStep = .welcome
+    /// Per-step record of what the user decided. Written by
+    /// `recordOutcome` (step content reporting a consent), `skip()`
+    /// (`.skipped`), and `advance()` (backfills `.notApplicable` for
+    /// steps that were walked past without a decision). Revisiting a
+    /// step via `back()` keeps the prior outcome until a new one is
+    /// recorded over it.
+    public private(set) var outcomes: [OnboardingStep: StepOutcome] = [:]
+    /// Whether the moderation directory has entries — resolved once by
+    /// `start()` via the injected probe. Until (or unless) the probe
+    /// answers true, moderation behaves as skippable/informational,
+    /// matching today's gate semantics (empty directory ⇒ the gate
+    /// reads operational and never blocks).
+    public private(set) var moderationDirectoryHasEntries = false
+
+    private let store: any OnboardingStore
+    private let moderationDirectoryNonEmpty: () async -> Bool
+    /// Called after `complete()` writes the flag, so the presenter can
+    /// dismiss the cover and hand control to the moderation gate.
+    private let onCompleted: @MainActor () -> Void
+    /// The in-flight directory probe, internal so tests can await it.
+    @ObservationIgnored private(set) var moderationProbeTask: Task<Void, Never>?
+
+    public init(
+        store: any OnboardingStore,
+        moderationDirectoryNonEmpty: @escaping () async -> Bool,
+        onCompleted: @escaping @MainActor () -> Void = {}
+    ) {
+        self.store = store
+        self.moderationDirectoryNonEmpty = moderationDirectoryNonEmpty
+        self.onCompleted = onCompleted
+    }
+
+    // MARK: - Step indicator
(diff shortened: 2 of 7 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — solid, well-tested state machine, but the "unskippable moderation" guarantee doesn't actually hold.

- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:135` — `advance()` ignores `isSkippable`. On `moderation` with a non-empty directory, the primary "Continue" button (`OnboardingView.swift:67` → `advance()`) walks straight to `done` and backfills `.notApplicable`. The only thing `isSkippable` blocks is `skip()`, so the step that's documented as mandatory is bypassable with one tap. Either gate `advance()` on a recorded outcome for mandatory steps (`guard !isMandatory(step) || outcomes[step] != nil`), or have the view disable the primary until content reports a consent. Worth a test: `advance()` on moderation with entries is a no-op without an outcome.
- `OnboardingFlow.swift:114` + `:100` — probe race. `moderationDirectoryHasEntries` starts `false` and is only corrected when the async probe lands (`start()` is kicked from `.task`, `OnboardingView.swift:47`). Until then moderation reads skippable, so a fast tapper — or any slow directory read — gets a Skip button on a step that should block. Consider a tri-state (`nil` = unresolved) treated as unskippable, or await the probe before rendering moderation.
- `OnboardingStore.swift:23` and `:5` — `key` is internal and the protocol has no reset, but the PR body and `OnboardingStoreTests.testUsesTheDocumentedKey` both promise the app's `--reset-keychain` path clears it. That path (`Sources/OnymIOS/OnymIOSApp.swift:1314`) is in the app target and can't see the key, so PR 3 will hardcode the string — exactly what the pinning test was meant to prevent. Add `func resetOnboarding()` to the protocol (or make `key` public).
- `OnboardingView.swift:131` vs `:86` — comment says "PR 3 may replace the mark via `stepContent`", but `content(for:)` hardcodes `WelcomeStepContent`/`DoneStepContent` and never consults the injected slot for those steps. Fix the comment, or fall back to the built-ins only when `stepContent` is nil.
- `Tests/OnymOnboardingTests/OnboardingFlowTests.swift:168` — `probes` is mutated from a non-isolated `() async -> Bool` closure; compiles under the package's Swift 5 mode but will be an error under strict concurrency. An actor-isolated counter or `@MainActor` closure keeps it future-proof.

Nits (non-blocking): `isSkippable(.welcome)` returns `true` while the view always hides Skip there (`OnboardingView.swift:79`) — harmless, but the flow and view disagree on the same question. And there's no persistence of partial progress: a kill mid-flow restarts at `welcome`, which is fine now but means PR 3's already-applied consents get re-presented.

rinat-enikeev:
All six findings addressed in dbe7773 — point by point:

1. **advance() bypassing unskippable moderation — the substantive catch, thank you.** The guarantee genuinely didn't hold: Continue backfilled `.notApplicable` and walked straight past a mandatory moderation step. Fixed with `isMandatory(step)` (moderation while the directory has entries — or unresolved, see 2) and a guard in `advance()`: `guard !isMandatory(step) || outcomes[step] != nil else { return }`. The view is the first layer of the same rule — the scaffold takes `primaryDisabled` and `OnboardingView` passes `flow.isMandatory(step) && flow.outcomes[step] == nil`. New tests: `testAdvanceOnMandatoryModerationWithoutOutcomeIsANoOp` (also asserts no outcome is backfilled by the refused advance) and `testAdvanceOnMandatoryModerationWithRecordedConsentAdvances`.

2. **Probe race.** `moderationDirectoryHasEntries` is now `Bool?` (nil = unresolved) and fail-closed: moderation reads unskippable AND mandatory until the probe answers `false`. The scaffold shows a small `ProgressView` in the Skip slot (`onboarding.moderation.skip_pending`) while `moderationProbeResolved` is false. Both transitions tested: `testModerationFailsClosedWhileProbeUnresolved` (unresolved → resolved-empty opens skip) and `testResolvedNonEmptyDirectoryStaysMandatory` (unresolved → resolved-non-empty stays locked).

3. **Reset seam.** `resetOnboarding()` added to the `OnboardingStore` protocol; the UserDefaults impl removes the key. The key-pinning test stays (`testUsesTheDocumentedKey`), plus `testResetOnboardingClearsTheFlag` / `testResetReopensTheGate`.

4. **stepContent never consulted for Welcome/Done.** Signature changed to `(OnboardingStep) -> AnyView?`; `content(for:)` consults the injected slot first for every step and falls back to the built-in Welcome/Done bodies and middle-step placeholders on nil. Comment fixed.

5. **Strict concurrency.** The idempotency test's captured-var counter is now a private `ProbeCounter` actor.

6. **Nits.** `isSkippable(.welcome)` is now `false` (with `testWelcomeIsNotSkippable`), and the flow's doc comment states that partial progress is deliberately not persisted — walk position restarts at `welcome` after a kill while individual consents persist in their own stores; revisit in PR 3 if the wired flow warrants resume.

`xcodebuild test -scheme OnymOnboarding -destination 'platform=iOS Simulator,name=iPhone 17 Pro'`: **34/34 green** (23 flow + 11 store).

dev-onym (commented):
**COMMENT** — clean, well-tested state machine; nothing blocking in-package, but the probe wedge and CI gap should be settled before PR 3.

- `Packages/OnymOnboarding/Package.swift:1` — CI never runs these 27 tests. `release.yml` has hand-written per-package steps only for `OnymDiscovery` (:195) and `OnymFoundation` (:222), and the package isn't in `project.yml`, so nothing in CI builds or tests it. The PR says "tests standalone, exactly like `OnymDiscovery`" — but OnymDiscovery has a CI step and this doesn't. Add the step here rather than in PR 3, or the tests are decorative for two PRs.
- `OnboardingFlow.swift:147` — the probe has no timeout and no failure channel. `() async -> Bool` can't report an error, and fail-closed is permanent: if the app's directory read hangs, moderation is unskippable (`isSkippable` false), un-advanceable (`isMandatory` true, no outcome), and only Back works — the user is wedged with no forward exit. Fail-closed *until resolved* is right; fail-closed *forever* isn't. Either bound it inside `start()` (`withTimeout` → treat as `true` but re-armable) or document that the injected probe must be deadline-bounded and non-throwing at the injection site. The tests only cover probes that answer instantly.
- `OnboardingView.swift:67`, `OnboardingView.swift:210` — same wedge is reachable in this PR alone: with a non-empty directory, `PlaceholderStepContent` never calls `recordOutcome`, so the built-in view dead-ends at moderation. Fine as scaffolding, but worth a `// PR 3` note on the placeholder so nobody previews it and files a bug.
- `OnboardingFlow.swift:200` — `complete()` isn't idempotent on `onCompleted`. It's guarded to `.done` but not to *not-yet-completed*, so a double-tap on Start fires the dismiss callback twice (`store.markOnboardingCompleted()` is harmless the second time). A `didComplete` flag costs one line; SwiftUI double-taps on a button that dismisses its own cover are a real class of bug.
- `OnboardingView.swift:147` — free-standing prose uses `Text("literal")` (LocalizedStringKey, resolved against `Bundle.main`, no `.module`) while everything else uses `Text(verbatim:)` over `String(localized:)`. Consistent with OnymDiscovery's existing posture, so not a defect — just pick one form within the file.
- Nits, take or leave: `stepCount` counts `done`, so the terminal screen reads "Step 7 of 7" while it's really the confirmation (`OnboardingFlow.swift:100`); and Welcome renders an empty Back/Skip `HStack` — ~28pt of dead space below the primary button.

The fail-closed tri-state and the `advance()` mandatory guard are both right, and the truth-table tests for grandfathering plus the "probe not consulted when flag set" test are exactly the ones that matter for upgraders.

PR #250, opened 2026-08-15
Onboarding unblockers: blossom picker, auto-populate seam, live blossom base URL, step indicator

### LINKED ISSUE #251: ChatImageLoader should honor attachment.server for downloads
Flagged in the review of #250 (review 4943389109): nothing reads `ChatImageAttachment.server` (or the video/voice equivalents) on the download path — `ChatImageLoader` / `ChatVideoLoader` / `ChatVoiceLoader` fetch every blob through the shared `BlossomClient`, which since #250 resolves its base URL live from the first configured server.

Pre-existing gap, but live resolution makes it bite sooner: switching the active Blossom server immediately breaks fetches of blobs that were uploaded to (and stamped with) the previous server. The attachment already carries the authoritative location — the loaders should build/bind a client for `attachment.server` (via `BlossomClient.bound(toServer:)`, added in #250) and fall back to the configured client only for legacy rows where `server` is nil.

The upload side is already consistent: since #250 the send path resolves the URL once per send, stamps it, and uploads through a client bound to it; `retry` re-uploads to the stamped server. Downloads are the remaining half.

rinat-enikeev:
Four pieces of the iOS onboarding plan (PR 2 of the sequence, plus the design prerequisite), refined across two review rounds. **Live-behavior summary**: the two *seams* (blossom catalog picker wiring, relayer auto-populate policy) default to current behavior until PR 3 wires them — but pieces of this PR ARE live immediately: the Blossom base URL resolves per operation (server changes apply to the next upload/download instead of the next launch), downloads honor each attachment's server stamp, and the Blossom settings screen gains a FROM CATALOG section, an ACTIVE badge, and a Make Active affordance.

### 1. Blossom catalog picker + make-active semantics (unblocks onboarding step 4, blob transport)
The "blob.storage" seat was the one without a consent picker. `blossomCatalogPicker` in `OnymIOSApp` mirrors the relayer/nostr pickers; the consent apply is extracted into `BlossomCatalogConsent.apply` (unit-tested): required **https** endpoint from the manifest → `blossomServersRepository.addEndpoint(..., makeActive: true)`. **Make-active semantics** (added after review — a plain append behind the seeded default left the pick silently inert since uploads/downloads only target the FIRST endpoint): `addEndpoint(_:makeActive:)` inserts/moves the endpoint to the head, and `makeActive(url:)` promotes any configured row (no-op for unknown/already-first URLs, which doesn't count as user interaction). `BlossomRelaySettingsFlow`/`View` gain the optional `discovery:` seam (default nil), a FROM CATALOG section, an ACTIVE badge on the first row, and a Make Active button on the rest.

### 2. Relayer auto-populate suppression seam (unblocks onboarding step 5, notary)
`RelayerRepository` takes `autoPopulatePolicy: @Sendable () -> Bool` (default `{ true }` — today's behavior; the app passes nothing yet). When false, a first-launch fetch updates `knownList` (and the disk cache) only — no endpoint install, no `hasUserInteracted` flip — deferred, not cancelled. PR 3 sets it to "false while onboarding is incomplete".

### 3. Live blossom base URL + server-stamp contract (LIVE in this PR; unblocks step 4's "active immediately" copy)
`DynamicBaseURLBlossomClient` resolves its base URL per operation from the repository's first endpoint (fallback: app default) — server changes apply to the next upload/download, no relaunch. The stamp contract holds on both halves:
- **Write**: each send resolves the URL once, stamps it into the attachments, and uploads every blob through `BlossomClient.bound(toServer:)` (new protocol requirement, default `self`), so stamp and blobs can't diverge mid-send; `retry` re-uploads to the stamped server.
- **Read**: all three loaders (image/video/voice) download through the client bound to `attachment.server`, falling back to the live client for legacy nil-server rows — so switching servers never 404s already-uploaded history (own or peers'). Superseded and closed #251.
- `bound(toServer:)` rejects schemeless/hostless stamps and falls back to live resolution.
The `--ui-loopback` in-memory client path is untouched.

### 4. `SettingsStepIndicator` public (design prerequisite)
Public struct + init(step:count:) + body in OnymDesign for the onboarding package's step progress; EXTRACTION-NOTES.md updated.

Also: the three transport settings flows' catalog plumbing is deduplicated into an internal `DiscoveryCatalogState` (OnymSettings), behavior-identical.

## Verification (all green, after round-3 fixes)
- App-target OnymIOSTests: 1222 passed, 0 failures (3 pre-existing skips) — includes BlossomCatalogConsentTests, ChatMediaLoaderServerBindingTests, DynamicBaseURLBlossomClientTests, RelayerRepositoryTests (auto-populate seam), SendMessageInteractorTests (mid-send binding), DiscoverySeatAdaptersTests
- OnymDiscovery package tests: 117 passed; OnymFoundation: 53 passed (OnymChain has no package test target — covered in the app suite)
- Full app build (iPhone 17 Pro simulator): succeeded

### DIFF
--- a/Packages/OnymChain/Sources/OnymChain/RelayerRepository.swift
+++ b/Packages/OnymChain/Sources/OnymChain/RelayerRepository.swift
@@ -51,14 +51,28 @@ public struct RelayerState: Equatable, Sendable {
 public actor RelayerRepository {
     private let fetcher: any KnownRelayersFetcher
     private let store: any RelayerSelectionStore
+    /// Seam over the first-launch auto-populate: consulted right
+    /// before a fetch would install the whole published list into an
+    /// untouched configuration. Returning `false` defers it — the
+    /// fetched list still lands in `knownList` (and the on-disk
+    /// cache), but the configuration and `hasUserInteracted` stay
+    /// untouched, so a later fetch (or an explicit user pick, e.g.
+    /// during onboarding) can still populate. Defaults to always-on,
+    /// preserving the historical behavior.
+    private let autoPopulatePolicy: @Sendable () -> Bool
 
     private var cached: RelayerState
     private var continuations: [UUID: AsyncStream<RelayerState>.Continuation] = [:]
     private var startTask: Task<Void, Never>?
 
-    public init(fetcher: any KnownRelayersFetcher, store: any RelayerSelectionStore) {
+    public init(
+        fetcher: any KnownRelayersFetcher,
+        store: any RelayerSelectionStore,
+        autoPopulatePolicy: @escaping @Sendable () -> Bool = { true }
+    ) {
         self.fetcher = fetcher
         self.store = store
+        self.autoPopulatePolicy = autoPopulatePolicy
         self.cached = RelayerState(
             configuration: store.loadConfiguration(),
             knownList: store.loadCachedKnownList(),
@@ -101,6 +115,10 @@ public actor RelayerRepository {
     /// `.random`, and `hasUserInteracted` flips to `true`. The flag is
     /// sticky — subsequent fetches never re-auto-populate, so a user
     /// who explicitly clears the list isn't fought by the next refresh.
+    /// `autoPopulatePolicy` gates the install: when it returns `false`
+    /// the fetched list only lands in `knownList` (configuration and
+    /// flag untouched), leaving the choice to a later fetch or an
+    /// explicit user pick.
     public func refresh() async throws {
         // Mark in-flight so the picker stops showing whatever stale
         // status it had and renders the spinner.
@@ -128,7 +146,7 @@ public actor RelayerRepository {
 
         let current = cached.configuration
         let updatedConfig: RelayerConfiguration
-        if !current.hasUserInteracted && !list.isEmpty {
+        if !current.hasUserInteracted && !list.isEmpty && autoPopulatePolicy() {
             updatedConfig = RelayerConfiguration(
                 endpoints: list,
                 primaryURL: nil,
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift
@@ -1,0 +1,46 @@
+import Foundation
+import OnymTransportBlossom
+
+/// Client selection for a stamped attachment download — the trust
+/// boundary between the peer's metadata and the user's network.
+///
+/// The `server` stamp is an OPTIMIZATION hint for multi-server
+/// consistency — "this blob lives on that one of YOUR servers" — never
+/// an instruction from the peer. It decodes straight off the wire and
+/// bubbles auto-load on render, so honoring an arbitrary stamp would
+/// let a hostile sender choose the download host: an unauthenticated
+/// GET of a per-message-unique path (a read receipt, plus IP/UA) to a
+/// third party the moment the recipient scrolls the thread. The stamp
+/// is therefore honored ONLY when its normalized origin
+/// (scheme+host+port) matches one of the endpoints the USER has
+/// configured/consented to; anything else — unknown hosts, malformed
+/// stamps, legacy nil stamps — downloads through the live client
+/// pointed at the user's own active server.
+enum BlossomServerStampPolicy {
+    /// The client a stamped attachment downloads through: `live` bound
+    /// to the stamp when the stamp's origin is in `allowedServers`,
+    /// plain `live` otherwise.
+    static func client(
+        forStamp server: String?,
+        allowedServers: [URL],
+        live: any BlossomClient
+    ) -> any BlossomClient {
+        guard let server,
+              let stamped = URL(string: server),
+              let stampedKey = originKey(stamped),
+              allowedServers.contains(where: { originKey($0) == stampedKey })
+        else { return live }
+        return live.bound(toServer: server)
+    }
+
+    /// Normalized scheme+host+port comparison key. `nil` when the URL
+    /// has no scheme or host — such a stamp is never comparable and
+    /// never honored.
+    static func originKey(_ url: URL) -> String? {
+        guard let scheme = url.scheme?.lowercased(),
+              let host = url.host()?.lowercased(), !host.isEmpty
+        else { return nil }
+        let port = url.port.map { ":\($0)" } ?? ""
+        return "\(scheme)://\(host)\(port)"
+    }
+}
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift
@@ -18,8 +18,18 @@ import OnymTransportBlossom
 /// pointed at the user's own active server.
 enum BlossomServerStampPolicy {
     /// The client a stamped attachment downloads through: `live` bound
-    /// to the stamp when the stamp's origin is in `allowedServers`,
-    /// plain `live` otherwise.
+    /// to the MATCHED allowlist URL when the stamp is https and its
+    /// origin is in `allowedServers`, plain `live` otherwise.
+    ///
+    /// This is the PEER-influenced trust level: https is required (a
+    /// peer must not be able to steer traffic to cleartext or odd
+    /// schemes), the origin must be one the user configured, and the
+    /// request is built from the allowlist's own URL — the stamp's
+    /// origin proves membership, but its path/query are still
+    /// peer-chosen bytes and must never shape the request, even
+    /// against the user's own server. The user's own hand-typed
+    /// endpoints (send/retry binding) live at the trusted level in
+    /// `bound(toServer:)` instead, where http local-dev URLs are fine.
     static func client(
         forStamp server: String?,
         allowedServers: [URL],
@@ -27,20 +37,24 @@ enum BlossomServerStampPolicy {
     ) -> any BlossomClient {
         guard let server,
               let stamped = URL(string: server),
+              stamped.scheme?.lowercased() == "https",
               let stampedKey = originKey(stamped),
-              allowedServers.contains(where: { originKey($0) == stampedKey })
+              let matched = allowedServers.first(where: { originKey($0) == stampedKey })
         else { return live }
-        return live.bound(toServer: server)
+        return live.bound(toServer: matched.absoluteString)
     }
 
-    /// Normalized scheme+host+port comparison key. `nil` when the URL
+    /// Normalized scheme+host+port comparison key; default ports
+    /// (443/https, 80/http) fold away so `https://a.example` and
+    /// `https://a.example:443` are the same origin. `nil` when the URL
     /// has no scheme or host — such a stamp is never comparable and
     /// never honored.
     static func originKey(_ url: URL) -> String? {
         guard let scheme = url.scheme?.lowercased(),
               let host = url.host()?.lowercased(), !host.isEmpty
         else { return nil }
-        let port = url.port.map { ":\($0)" } ?? ""
+        let defaultPort: Int? = scheme == "https" ? 443 : (scheme == "http" ? 80 : nil)
+        let port = url.port.flatMap { $0 == defaultPort ? nil : ":\($0)" } ?? ""
         return "\(scheme)://\(host)\(port)"
     }
 }
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
@@ -87,6 +87,11 @@ public actor ChatImageLoader {
 
     private func downloadAndDecrypt(_ attachment: ChatImageAttachment) async throws -> Data {
         let key = attachment.sha256
+        // Resolve the allowlist BEFORE the inflight lookup: the await
+        // is a suspension point, and the check-and-insert below must
+        // stay contiguous (actor-atomic) or two concurrent requests
+        // for the same sha both miss `inflight` and double-download.
+        let allowedServers = await allowedStampServers()
         if let existing = inflight[key] { return try await existing.value }
         // Fetch from the server STAMPED into the attachment — the one
         // the sender actually uploaded to — but ONLY when it is one of
@@ -97,7 +102,7 @@ public actor ChatImageLoader {
         // hosts and legacy nil stamps use the live client.
         let client = BlossomServerStampPolicy.client(
             forStamp: attachment.server,
-            allowedServers: await allowedStampServers(),
+            allowedServers: allowedServers,
             live: blossomClient
         )
         let task = Task<Data, Error> {
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
@@ -12,12 +12,22 @@ import OnymTransportBlossom
 /// message never touches the network.
 public actor ChatImageLoader {
     private let blossomClient: any BlossomClient
+    /// The user's configured Blossom endpoints — the ONLY servers an
+    /// attachment's `server` stamp may route a download to (see
+    /// `BlossomServerStampPolicy`). Defaults to empty: stamps are
+    /// ignored unless the composition root wires the configured set.
+    private let allowedStampServers: @Sendable () async -> [URL]
     private let cacheDir: URL
     private var memory: [String: UIImage] = [:]
     private var inflight: [String: Task<Data, Error>] = [:]
 
-    public init(blossomClient: any BlossomClient, cacheDirectory: URL? = nil) {
+    public init(
+        blossomClient: any BlossomClient,
+        cacheDirectory: URL? = nil,
+        allowedStampServers: @escaping @Sendable () async -> [URL] = { [] }
+    ) {
         self.blossomClient = blossomClient
+        self.allowedStampServers = allowedStampServers
         self.cacheDir = cacheDirectory ?? FileManager.default
             .urls(for: .cachesDirectory, in: .userDomainMask)[0]
             .appendingPathComponent("OnymChatImages", isDirectory: true)
@@ -79,12 +89,17 @@ public actor ChatImageLoader {
         let key = attachment.sha256
         if let existing = inflight[key] { return try await existing.value }
         // Fetch from the server STAMPED into the attachment — the one
-        // the sender actually uploaded to — not whatever server is
-        // currently configured (the live client's base URL moves the
-        // moment the user picks a different server). Legacy rows
-        // without a stamp fall back to the live client.
-        let client = attachment.server.map { blossomClient.bound(toServer: $0) }
-            ?? blossomClient
+        // the sender actually uploaded to — but ONLY when it is one of
+        // the user's own configured servers: the stamp arrives off the
+        // wire, so it is a hint for multi-server consistency among
+        // servers the USER trusts, never a peer-chosen download host
+        // (BlossomServerStampPolicy has the full rationale). Unknown
+        // hosts and legacy nil stamps use the live client.
+        let client = BlossomServerStampPolicy.client(
+            forStamp: attachment.server,
+            allowedServers: await allowedStampServers(),
+            live: blossomClient
+        )
         let task = Task<Data, Error> {
             let blob = try await client.download(sha256: attachment.sha256)
             return try ChatImageCrypto.open(
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift
@@ -78,7 +78,13 @@ public actor ChatImageLoader {
     private func downloadAndDecrypt(_ attachment: ChatImageAttachment) async throws -> Data {
         let key = attachment.sha256
         if let existing = inflight[key] { return try await existing.value }
-        let client = blossomClient
+        // Fetch from the server STAMPED into the attachment — the one
+        // the sender actually uploaded to — not whatever server is
+        // currently configured (the live client's base URL moves the
+        // moment the user picks a different server). Legacy rows
+        // without a stamp fall back to the live client.
+        let client = attachment.server.map { blossomClient.bound(toServer: $0) }
+            ?? blossomClient
         let task = Task<Data, Error> {
             let blob = try await client.download(sha256: attachment.sha256)
             return try ChatImageCrypto.open(
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
@@ -14,11 +14,19 @@ import OnymTransportBlossom
 /// play it directly.
 public actor ChatVideoLoader {
     private let blossomClient: any BlossomClient
+    /// See `ChatImageLoader.allowedStampServers` — the only servers a
+    /// wire-decoded `server` stamp may route a download to.
+    private let allowedStampServers: @Sendable () async -> [URL]
     private let cacheDir: URL
     private var inflight: [String: Task<URL, Error>] = [:]
 
-    public init(blossomClient: any BlossomClient, cacheDirectory: URL? = nil) {
+    public init(
+        blossomClient: any BlossomClient,
+        cacheDirectory: URL? = nil,
+        allowedStampServers: @escaping @Sendable () async -> [URL] = { [] }
+    ) {
         self.blossomClient = blossomClient
+        self.allowedStampServers = allowedStampServers
         self.cacheDir = cacheDirectory ?? FileManager.default
             .urls(for: .cachesDirectory, in: .userDomainMask)[0]
             .appendingPathComponent("OnymChatVideos", isDirectory: true)
@@ -37,11 +45,13 @@ public actor ChatVideoLoader {
         if FileManager.default.fileExists(atPath: dest.path) { return dest }
         if let existing = inflight[key] { return try await existing.value }
 
-        // Fetch from the server STAMPED into the attachment (where the
-        // sender uploaded), falling back to the live client for legacy
-        // rows without a stamp — see ChatImageLoader.
-        let client = attachment.server.map { blossomClient.bound(toServer: $0) }
-            ?? blossomClient
+        // Stamp honored only within the user's configured server set —
+        // never a peer-chosen host. See BlossomServerStampPolicy.
+        let client = BlossomServerStampPolicy.client(
+            forStamp: attachment.server,
+            allowedServers: await allowedStampServers(),
+            live: blossomClient
+        )
         let task = Task<URL, Error> {
             let blob = try await client.download(sha256: attachment.sha256)
             let plaintext = try ChatImageCrypto.open(
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
@@ -43,13 +43,18 @@ public actor ChatVideoLoader {
         let key = attachment.sha256
         let dest = cacheFileURL(key)
         if FileManager.default.fileExists(atPath: dest.path) { return dest }
+        // Resolve the allowlist BEFORE the inflight lookup — the await
+        // suspends, and the check-and-insert below must stay contiguous
+        // (actor-atomic) or two concurrent requests double-download and
+        // double-write `dest`. See ChatImageLoader.
+        let allowedServers = await allowedStampServers()
         if let existing = inflight[key] { return try await existing.value }
 
         // Stamp honored only within the user's configured server set —
         // never a peer-chosen host. See BlossomServerStampPolicy.
         let client = BlossomServerStampPolicy.client(
             forStamp: attachment.server,
-            allowedServers: await allowedStampServers(),
+            allowedServers: allowedServers,
             live: blossomClient
         )
         let task = Task<URL, Error> {
@@ -57,7 +62,9 @@ public actor ChatVideoLoader {
             let plaintext = try ChatImageCrypto.open(
                 blob: blob, key: attachment.encKey, expectedSha256Hex: attachment.sha256
             )
-            try plaintext.write(to: dest, options: .completeFileProtection)
+            // `.atomic`: a concurrent reader (AVPlayer) must never open
+            // a partially-written file.
+            try plaintext.write(to: dest, options: [.atomic, .completeFileProtection])
             return dest
         }
         inflight[key] = task
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift
@@ -37,7 +37,11 @@ public actor ChatVideoLoader {
         if FileManager.default.fileExists(atPath: dest.path) { return dest }
         if let existing = inflight[key] { return try await existing.value }
 
-        let client = blossomClient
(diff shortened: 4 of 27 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — solid, well-tested seams, but the new Blossom catalog pick is silently inert because uploads only ever use the *first* endpoint.

- `Sources/OnymIOS/OnymIOSApp.swift:786` — consent `apply` calls `addEndpoint`, which **appends** (`BlossomServersRepository.swift:96`). First launch already seeds `.seed` (`BlossomServersRepository.swift:37`), so a catalog-picked server lands at index 1 while `currentEndpoints().first` — the upload/download target (`OnymIOSApp.swift:430`) — stays Onym Official. There is no reorder/make-primary affordance in `BlossomRelaySettingsView`/`Flow`, so the user's pick never takes effect unless they also remove the default. That defeats piece 1 + piece 3's stated "active immediately" goal for onboarding step 4. Either insert the consent-picked endpoint at the head, or add an explicit "make active" mutation and have the picker use it. (`removeEndpoint` is the only existing escape hatch and it's undiscoverable from the catalog row.)
- `Packages/OnymSettings/Sources/OnymSettings/BlossomRelaySettingsView.swift:49` — the footnote is now half-updated: "Changes apply to the next upload or download. Uploads target the first configured server." Given the point above, that second sentence is the whole story for catalog picks and reads as a shrug. Worth stating how to make a non-first server active.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:333` — the PR claims the stamp "always agrees with where the blobs were actually uploaded," but the stamp is sampled *before* the uploads, and `DynamicBaseURLBlossomClient` re-resolves independently per upload (`DynamicBaseURLBlossomClient.swift:29`). A config change mid-send makes the stamp disagree, and in the multi-attachment loop (`SendMessageInteractor.swift:627`) two blobs of one message can land on different servers. Narrow window, but the doc comment overstates the guarantee — either resolve once and thread the URL into the client call, or soften the comment.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift` — nothing reads `ChatImageAttachment.server`; downloads go through the shared client, which now resolves live. Pre-existing, but per-operation resolution makes it bite sooner: switching servers immediately breaks fetches of already-uploaded blobs. Please file the "loader should honor `attachment.server`" follow-up rather than leaving it implied by the new comments.
- `Packages/OnymSettings/Sources/OnymSettings/BlossomRelaySettingsFlow.swift:33-118` — ~60 lines (`catalogEntries`, `applyCatalog`, `activeConsent`, `consentedOffer`, `refreshCatalog`, task plumbing) are a byte-for-byte copy of `NostrRelaySettingsFlow.swift:29-114`. Third copy is coming; a small `DiscoveryCatalogState` helper would collapse all of them.
- `Tests/OnymIOSTests/BlossomRelaySettingsFlowTests.swift` — no coverage for the new catalog surface (stream install, memoization, `refreshCatalog`). Consistent with the untested Nostr precedent, so not a blocker, but the consent-`apply` closure in `OnymIOSApp.swift:772` — the one piece with real behavior — is untested and is exactly where the ordering bug above lives.

Non-blocking and correct as written: `RelayerRepository.autoPopulatePolicy` (`RelayerRepository.swift:149`) with its three new tests, `DynamicBaseURLBlossomClient` + tests, and the `SettingsStepIndicator` publicization with the EXTRACTION-NOTES move.

rinat-enikeev:
All findings addressed in 83e67eb, point by point:

**1. Inert catalog pick (blocker)** — fixed with explicit make-active semantics. `BlossomServersRepository.addEndpoint(_:makeActive:)` inserts a new endpoint at the head or moves an existing one there (metadata still updates, others keep relative order), and a standalone `makeActive(url:)` promotes any configured row — a no-op for unknown or already-first URLs, which deliberately does NOT flip `hasUserInteracted`. The consent `apply` now calls `addEndpoint(..., makeActive: true)`: a consented pick becomes the active (first) server — the evident intent of picking a blob-storage module — with the seeded default kept right behind it. The apply step is extracted into `BlossomCatalogConsent.apply` so the ordering is unit-tested (see 6). `BlossomRelaySettingsView` shows an ACTIVE badge on the first row and a "Make Active" button on every other row (`blossom.make_active.<url>`).

**2. Footnote** — now: "Uploads and downloads target the active server — the first in the list. Picking a server from the catalog makes it active; tap Make Active on any other row to switch. Changes apply to the next upload or download."

**3. Stamp guarantee** — made true rather than softened. New `BlossomClient.bound(toServer:)` protocol requirement (default `self`, so fixed-URL clients and the UI-test fake are untouched); `DynamicBaseURLBlossomClient` returns its inner client pinned to that URL. Every send path resolves the URL once, stamps it, and passes the bound client into `uploadAndFanOut` — so all blobs of a multi-attachment message land on the stamped server even if the configuration changes mid-send. Bonus correctness on the same seam: `retry` now re-uploads to the server already stamped into the message's attachments (falling back to the live client for legacy nil-server rows), so a config change between a failed send and its retry can't strand the descriptor either. Doc comments updated to state the now-true guarantee.

**4. Download-side follow-up filed**: #251 — "ChatImageLoader should honor attachment.server for downloads", referencing this review.

**5. Duplication** — collapsed into `DiscoveryCatalogState` (OnymSettings): entries stream drain, one-shot refresh, memoized consent/offer lookups, and the relayer flow's consented-URL dedupe set. All three flows (Relayer/Nostr/Blossom) consume it; their public APIs are unchanged and behavior is identical (existing tests untouched and green).

**6. Tests** — `BlossomCatalogConsentTests` (5): pick lands FIRST with seed preserved, first-https-in-published-order, no-https throws `noUsableEndpoint` leaving config untouched, name fallback, re-apply-after-demotion promotes without duplicating. `BlossomServersRepositoryTests` (+6): make-active insert/move/no-op semantics. `BlossomRelaySettingsFlowTests` (+4): catalog stream install, `refreshCatalog` re-read, nil-discovery stays empty, make-active intent. `DynamicBaseURLBlossomClientTests` (+2): `bound(toServer:)` pins ignoring the resolver, unparseable falls back to live. `SendMessageInteractorTests` (+1): mid-send server change — stamp and both album blobs use the send-start URL, zero unbound uploads.

Verification: OnymIOSTests 1217 passed / 0 failures (3 pre-existing skips), OnymDiscovery 117, OnymFoundation 53, full app build (iPhone 17 Pro sim) succeeded.

dev-onym (commented):
**REQUEST_CHANGES** — the per-send binding is solid, but nothing on the read side consumes the `server` stamp this PR works to keep accurate, and making the base URL live turns that into a reachable regression.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift:80` (same at `ChatVideoLoader.swift:42`, `ChatVoiceLoader.swift:40`): all three loaders download through the shared `blossomClient` and ignore `attachment.server`. With `DynamicBaseURLBlossomClient` that now resolves live, the moment a catalog pick lands (auto `makeActive`) or the user taps Make Active, every not-yet-cached blob — the user's own history, and every message from a peer on a different server — 404s. Previously this needed a relaunch to bite; now it's one tap, and onboarding step 4 makes it the happy path. The fix is the seam you just added: `attachment.server.map { blossomClient.bound(toServer: $0) } ?? blossomClient` in `downloadAndDecrypt`. Without it, `SendMessageInteractor`'s "receivers fetch from the same server the sender uploaded to" comment (`SendMessageInteractor.swift:25`) is not true of any receiver.
- `Packages/OnymSettings/Sources/OnymSettings/BlossomRelaySettingsView.swift:49`: the new footnote promises "Uploads **and downloads** target the active server" — accurate for the client, but combined with the above it's user-facing copy for the broken behavior. Worth landing the loader fix before this copy ships.
- `Packages/OnymTransportBlossom/Sources/OnymTransportBlossom/DynamicBaseURLBlossomClient.swift:39`: `URL(string:)` only returns nil for a few inputs — a schemeless stamp like `blossom.example` parses fine and builds a client against a relative URL, so every request fails instead of falling back to live resolution. Guard on `url.scheme != nil && url.host != nil`. `DynamicBaseURLBlossomClientTests.swift:96` only covers `""`.
- `Packages/OnymSettings/Sources/OnymSettings/DiscoveryCatalogState.swift:22`: `public` class + 6 public members with no consumer outside OnymSettings — `internal` matches how the flows' own catalog plumbing was scoped, and this repo tracks unjustified `public` deliberately (`Packages/OnymDesign/EXTRACTION-NOTES.md`).
- `Packages/OnymSettings/Sources/OnymSettings/DiscoveryCatalogSection.swift:6`: doc still says "shared by the per-seat pickers (relayer, nostr)" — blossom is now a third consumer.
- PR body: "both seams default to current behavior, so this PR changes nothing observable live" is not accurate — piece 3 changes live behavior unconditionally, and the Blossom settings screen gains a FROM CATALOG section, an ACTIVE badge, and a Make Active button. The body also predates `makeActive:`/`BlossomServersRepository.makeActive` (commit 83e67eb) and still describes a plain `addEndpoint` append; please refresh it before merge.

Verified good: `bound(toServer:)` covers every upload path and the retry path pins to the stamped server (`SendMessageInteractor.swift:968`); `makeActive` no-ops don't flip `hasUserInteracted` and are tested; the `autoPopulatePolicy` seam leaves `knownList`/cache intact and defaults to prior behavior; the three-flow `DiscoveryCatalogState` extraction is behavior-preserving.

rinat-enikeev:
Round 3 addressed in e366b5f, point by point:

**1. Read side honors the stamp** — all three loaders (`ChatImageLoader.downloadAndDecrypt`, `ChatVideoLoader.fileURL`, `ChatVoiceLoader.fileURL`) now select their client via `attachment.server.map { blossomClient.bound(toServer: $0) } ?? blossomClient`. All three attachment types carry `server: String?`, so stamped rows fetch from the sender's upload server and legacy nil-server rows keep the live client. The catalog-pick / Make Active tap no longer strands any not-yet-cached blob. This fully supersedes #251 — I checked for other Blossom fetch paths and the three loaders are the only `BlossomClient` consumers (group avatars/names travel inline in message payloads, not via Blossom), so #251 is **closed** with nothing left to narrow to. The `SendMessageInteractor` comment about receivers fetching from the sender's server is now true of actual receivers.

**2. Footnote** — kept as-is; it's accurate as of the same commit since the download half now holds.

**3. Bound-URL guard** — `bound(toServer:)` now requires `url.scheme != nil` and a non-empty `url.host()`, falling back to live resolution otherwise. New test covers `"blossom.example"` (schemeless) and `"https://"` (hostless) alongside the existing `""` case.

**4. `DiscoveryCatalogState`** — demoted to internal (class + all members); no consumer outside OnymSettings.

**5. `DiscoveryCatalogSection` doc** — now lists "(relayer, nostr, blossom)".

**6. PR body** — refreshed: it now states plainly that piece 3 is live in this PR (per-operation base URL, stamp-bound downloads, and the Blossom screen's FROM CATALOG / ACTIVE / Make Active affordances), and describes the `addEndpoint(_:makeActive:)` / `makeActive(url:)` semantics instead of the stale plain-append text. Only the two seams (picker wiring for onboarding, `autoPopulatePolicy`) remain default-off until PR 3.

Tests added: `ChatMediaLoaderServerBindingTests` (4 — stamped→bound for image/video/voice, nil→live) and the schemeless/hostless bound-guard test. Verification: OnymIOSTests 1222 passed / 0 failures (3 pre-existing skips), OnymDiscovery 117, OnymFoundation 53, full app build (iPhone 17 Pro sim) succeeded.

dev-onym (commented):
**REQUEST_CHANGES** — the read-side stamp honoring turns a peer-controlled string into the download host, with no scheme/allowlist constraint.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift:86` (same at `ChatVideoLoader.swift:43`, `ChatVoiceLoader.swift:41`): `attachment.server` is decoded straight off the wire (`ChatImageAttachment.swift:47`, `case server` in its `CodingKeys`), so a sender chooses the host. Bubbles auto-load on render (`ChatBubbleCell.swift:598`, `AlbumGridView.swift:157`), so a hostile peer can stamp `https://tracker.example` and get an unauthenticated GET of a per-message-unique path the moment you scroll the thread — a read receipt plus IP/UA disclosure to a third party. Before this PR every download went to the user's own configured server, so this is a new regression, not pre-existing. Payload content is still safe (AEAD open with the sender's key), but the fetch itself is the leak.
- Suggested fix, roughly in order of cost: only honor the stamp when the URL is in `blossomServersRepository.currentEndpoints()` (or the seeded/consented set) and otherwise fall back to the live client; or gate stamped-host fetches behind the existing trust/consent model for the sender. The write half (`SendMessageInteractor` per-send binding) is self-stamped and unaffected.
- `Packages/OnymTransportBlossom/Sources/OnymTransportBlossom/DynamicBaseURLBlossomClient.swift:44`: the guard requires only `scheme != nil` + non-empty host, so `http://…` (cleartext, and the repo has no ATS exception plist I can find, so it'd just fail) and odd schemes/ports pass. Require `scheme == "https"` — the catalog apply path already does exactly that (`BlossomCatalogConsent.swift`, `requiredEndpointURL(in:scheme:"https")`).
- `SendMessageInteractor.swift:975-980`: `stampedServer` takes the first media stamp and only falls back to voice; correct today (one resolve per send) but a mixed-stamp legacy row would silently re-upload all blobs to one server. Worth a comment or per-blob binding.

Non-blocking, all verified fine: `makeActive`/`addEndpoint(makeActive:)` ordering and the `hasUserInteracted` no-op cases (`BlossomServersRepository.swift:95-140`) match their tests; the `autoPopulatePolicy` gate is a pure conjunct that leaves the else-branch untouched (`RelayerRepository.swift:148`); `DiscoveryCatalogState` extraction is behavior-identical and nested `@Observable` reads still track through the computed `catalogEntries`; `UITestBlossomClient` correctly inherits the `bound → self` default so `--ui-loopback` is untouched.

rinat-enikeev:
Round 4 addressed in a0433d4. First: the peer-controlled-host finding is a genuine security regression that round 3 introduced and this review caught before merge — `attachment.server` decodes off the wire, bubbles auto-load on render, so unconditional stamp honoring handed a hostile sender an unauthenticated GET (per-message-unique path = read receipt, plus IP/UA) to any host of their choosing on scroll. Good catch; thank you.

**1. Allowlisted stamp honoring** — implemented your first (cheapest) option, as a dedicated trust boundary: new `BlossomServerStampPolicy` in OnymChatsCore. A stamp routes a download only when its normalized origin — scheme+host+port, case-insensitive, path/trailing-slash ignored, differing port rejected — matches one of the user's configured Blossom endpoints; everything else (unknown hosts, malformed stamps, legacy nil) goes through the live client, with **zero** requests to the stamp. The trust rationale is stated in the policy's doc comment: *the stamp is an optimization hint for multi-server consistency among servers the USER trusts — "this blob lives on that one of YOUR servers" — never an instruction from the peer.* All three loaders take an `allowedStampServers: @Sendable () async -> [URL]` provider whose default is empty (secure default: stamps ignored unless wired); the composition root wires `blossomServersRepository.currentEndpoints()`, so the allowlist tracks the live configuration including catalog picks. (The app-target dedupe normalization keys on the full URL, so the policy has its own origin key rather than reusing it.) The self-stamped write half is unchanged.

**2. https-only binding** — `bound(toServer:)` now requires `scheme == "https"` exactly (plus non-empty host), matching `BlossomCatalogConsent`'s `requiredEndpointURL(in:scheme:"https")`; `http://` and `ftp://` tests added alongside the schemeless/hostless/empty cases. As you noted, cleartext http would just die on ATS anyway — now it falls back to live resolution instead of failing every request.

**3. `stampedServer` first-pick** — documented rather than per-blob-bound: since the per-send binding landed, every send resolves once so all of a message's attachments carry the same stamp and any is representative; a mixed-stamp message can only be a pre-guarantee legacy row, where re-uploading all blobs to the first stamp's server merely re-homes a legacy blob alongside its siblings. Comment says exactly this.

Tests added/updated (+4 net): stamped-but-unconfigured → live with no request to the stamp; unwired allowlist ignores stamps; normalized-origin matching (case/slash match honored, different port refused); the three bound-path loader tests now pass the allowlist; http/ftp bound-guard fallback. Verification: OnymIOSTests **1226 passed / 0 failures** (3 pre-existing skips), OnymDiscovery 117, OnymFoundation 53, full app build (iPhone 17 Pro sim) succeeded.

dev-onym:
I checked the branch against every claim in the comment. **The work is already complete and pushed — there is nothing left to change, so I made no commit.**

Commit `a0433d4` (identical on `origin/onboarding-unblockers`, 0 ahead / 0 behind) already contains all three items exactly as the comment describes:

**1. Allowlisted stamp honoring** — `Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift:19` is present, with the trust rationale in its doc comment (the stamp is an optimization hint among servers the *user* trusts, never a peer instruction). `client(forStamp:allowedServers:live:)` binds only when `originKey` (lowercased scheme + host + optional `:port`, path ignored) matches an allowed endpoint; everything else returns `live` without touching the stamp. All three loaders take `allowedStampServers: @escaping @Sendable () async -> [URL] = { [] }` (`ChatImageLoader.swift:22`, `ChatVideoLoader.swift`, `ChatVoiceLoader.swift`), and the composition root wires it at `Sources/OnymIOS/OnymIOSApp.swift:465` via the local `blossomServersRepository.currentEndpoints().map(\.url)` — captured as a local, not `self`, so init ordering is sound, and it reads the live actor state each call. These are the only three loader construction sites outside tests.

**2. https-only binding** — `DynamicBaseURLBlossomClient.bound(toServer:)` guards `url.scheme?.lowercased() == "https"` plus non-empty host, with the doc comment tying it to `requiredEndpointURL(in:scheme:"https")`; `test_boundToServer_nonHTTPSScheme_fallsBackToLiveResolution` covers `http://` and `ftp://`.

**3. `stampedServer` first-pick** — documented at `SendMessageInteractor.swift:973`, stating the per-send-resolve guarantee and the legacy mixed-stamp re-homing tradeoff.

Tests match too: the unconfigured-stamp/zero-request case, the unwired-allowlist case, normalized-origin matching (case + trailing slash honored, differing port refused), and the three bound-path tests updated to pass an allowlist.

One caveat on verification: this container has no Swift toolchain (`swift`/`xcodebuild` are absent), so I could not re-run the 1226-test suite or the simulator build — those numbers are from the session that produced `a0433d4`, not re-confirmed here. My check was a source-level audit of every claim, and each one holds.

dev-onym (commented):
**REQUEST_CHANGES** — the stamp/allowlist design is sound, but the new `await` in all three loaders breaks in-flight deduplication.

- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatImageLoader.swift:90-109` — **in-flight dedup regression.** `if let existing = inflight[key]` and `inflight[key] = task` used to be separated only by `let client = blossomClient` (synchronous → actor-atomic). The new `allowedServers: await allowedStampServers()` inserts a suspension point between the check and the insert, so two concurrent `image(for:)`/`plaintext(for:)` calls for the same sha both miss `inflight`, both spawn a download task, and the first `defer { inflight[key] = nil }` clears the other's entry. Concretely: a thread scrolling into view twice (or image + `plaintext` for a report) double-downloads the blob. Fix: resolve `await allowedStampServers()` *before* the `inflight` lookup so the check-and-insert stays contiguous.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatVideoLoader.swift:44-61`, `ChatVoiceLoader.swift:44-61` — same race, worse consequence: two tasks now write the same `dest` via `plaintext.write(to: dest, options: .completeFileProtection)` (no `.atomic`), so a concurrent reader (AVPlayer / voice playback) can open a partially-written file. Add `.atomic` here regardless of the dedup fix — `ChatImageLoader.writeDisk` already does.
- `Tests/OnymIOSTests/ChatMediaLoaderServerBindingTests.swift` — no concurrency test; a `TaskGroup` with two simultaneous `plaintext(for:)` asserting `downloads.count == 1` would have caught the above. Worth adding alongside the fix.
- `Packages/OnymTransportBlossom/Sources/OnymTransportBlossom/DynamicBaseURLBlossomClient.swift:44-51` — `bound(toServer:)` rejects non-https, but `resolveBaseURL` doesn't. If a user configures an `http://` server, every send stamps that URL, `bound` falls back to `self`, and the per-send pin silently degrades to live resolution — exactly the mid-send divergence the pin exists to prevent. Either reject non-https at config/stamp time too, or note that http endpoints are unpinnable.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift:26-38` — the allowlist is correct as a trust boundary, but it makes cross-user delivery depend on server overlap: if A picks a catalog blob server (now their *active* server, per `BlossomCatalogConsent.apply`) and sends media, B — whose list is the default seed — falls back to their own active server and 404s. Not a regression (stamps were ignored entirely before), but PR 3's onboarding step 4 will make custom picks common, so the failure mode is about to get real. Worth a follow-up issue for a "try stamp origin, fall back to active" or an explicit unavailable-media state.
- `Packages/OnymSettings/.../BlossomRelaySettingsView.swift:49` — the four-sentence footnote is doing a lot; consider trimming now that the ACTIVE badge and Make Active buttons make the rule visible.

Non-blocking and correct as written: `DiscoveryCatalogState` dedupe (nested `@Observable` reads still track through the computed `catalogEntries`), `makeActive`/`addEndpoint(makeActive:)` ordering and `hasUserInteracted` semantics, the retry-uses-first-stamp reasoning, and the `autoPopulatePolicy` default preserving today's behavior.

rinat-enikeev:
Round 5 addressed in 439e3f3. The suspension-point catch is a sharp one — the dedup invariant depended on nothing but the *absence* of an `await` between the inflight check and insert, which round 4's allowlist call quietly violated in all three loaders. Exactly the kind of actor-reentrancy bug that survives every sequential test. Credit where due.

**1. Dedup restored** — `await allowedStampServers()` moved BEFORE the inflight lookup in `ChatImageLoader.downloadAndDecrypt`, `ChatVideoLoader.fileURL(for:)`, and `ChatVoiceLoader.fileURL(for:)`, so the check-and-insert is contiguous (actor-atomic) again. Each site carries a comment pinning the invariant so the next parameter added doesn't repeat this. And `.atomic` added to the video/voice `plaintext.write` regardless of the dedup fix — `[.atomic, .completeFileProtection]`, matching `ChatImageLoader.writeDisk` — so AVPlayer/voice playback can never open a partially-written file.

**2. Concurrency tests** — per-loader TaskGroup tests (3): two simultaneous requests for the same sha against a fake with 200ms download latency, asserting exactly one recorded download (and, for video/voice, both callers reading the complete written file). These fail against the round-4 code.

**3. resolveBaseURL asymmetry** — took the reject option: the app's `resolveBlossomBaseURL` is now https-only (first https endpoint, else the app default), matching `bound(toServer:)` and `requiredEndpointURL(in:scheme:"https")`. The comment states the rule: an http endpoint would stamp a URL the per-send pin can never bind to, silently degrading to live resolution — unpinnable is unusable for sends, so non-https entries are skipped rather than half-honored.

**4. Follow-up filed**: #254 — cross-user delivery depends on server overlap once custom picks are common (PR 3's onboarding makes it real); sketches both the "stamp origin within policy, else fall back" option and the explicit unavailable-media state, with the constraint that any solution keeps the no-automatic-request-to-peer-chosen-hosts property. References this review.

**5. Footnote** — trimmed to two sentences ("Uploads and downloads use the active server. Changes apply to the next upload or download."); the ACTIVE badge and Make Active buttons carry the mechanics.

Verification: OnymIOSTests **1229 passed / 0 failures** (3 pre-existing skips; +3 concurrency tests), OnymDiscovery 117, OnymFoundation 53, full app build (iPhone 17 Pro sim) succeeded.

dev-onym (commented):
REQUEST_CHANGES — solid, well-tested work; one live regression where the UI's ACTIVE badge and the actual upload target can disagree.

- `Sources/OnymIOS/OnymIOSApp.swift:435` — `resolveBlossomBaseURL` skips non-https endpoints, but `BlossomRelaySettingsFlow.validate` (`Packages/OnymSettings/Sources/OnymSettings/BlossomRelaySettingsFlow.swift:134`) deliberately accepts `http://` "for local dev / loopback", and pre-PR the base URL was `endpoints.first?.url` regardless of scheme. **Regression**: a user who adds/promotes `http://192.168.1.10:3000` now has every upload/download silently go to `https://blossom.onym.app` instead of their own box — encrypted blobs, but the upload metadata lands at Onym rather than where the user pointed it.
- `Packages/OnymSettings/Sources/OnymSettings/BlossomRelaySettingsView.swift:127` — the ACTIVE badge is unconditional `idx == 0`, so in the case above the UI asserts a server is active that nothing ever contacts, and "Make Active" on it is a no-op in effect. Badge/resolver must agree: either compute the badge from the same first-https rule, or drop `http` from `validate` and reject/flag existing http rows.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift:31` — allowlist matches on origin only, then binds to the **raw stamp string**, path and query included (`baseURL.appendingPathComponent(sha256)` at `Packages/OnymTransportBlossom/.../BlossomClient.swift:98`). A peer can therefore choose the request path on your own server. Low impact, but binding to the matched allowlist `URL` instead of `server` closes it for free.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/BlossomServerStampPolicy.swift:41` — `originKey` doesn't normalize default ports, so `https://a.example` and `https://a.example:443` are different keys and such a stamp is never honored. Fail-safe direction, but worth folding in.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift:968` — the retry binding (first-stamp derivation, `?? blossomClient` legacy fallback) is new logic with no test asserting the re-upload goes through the *stamped* client; existing retry tests only cover status/dedup. `ServerTaggingBlossomClient` already gives you the hook.
- `Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift:576` — `SettingsStepIndicator` is now public with zero consumers in-repo. Fine as a stated prerequisite, just noting EXTRACTION-NOTES' "justified by a consumer outside the package" rule isn't satisfied until PR 3.

Verified as correct: `message.media` (`ChatMessage.swift:120`) does fall back to single image/video, so the retry stamp derivation covers non-album messages; the allowlist `await` is correctly hoisted above the `inflight` check-and-insert in all three loaders; `bound(toServer:)`'s default `self` keeps `UITestBlossomClient` and the loopback path unchanged; `DiscoveryCatalogState` is the only loader/catalog construction site and observation propagates through the computed `catalogEntries`.

rinat-enikeev:
Round 6 addressed in 1254c99. The badge/target desync is a real catch — round 5 fixed the pin asymmetry by tightening the wrong side, leaving Settings asserting ACTIVE on a server nothing ever contacted while an http local-dev user's uploads silently went to Onym. The organizing principle, now stated at both sites in code: **https-only enforcement belongs on PEER-influenced paths (download stamps, catalog applies); the user's own hand-typed endpoints are trusted as typed and their scheme is never second-guessed.**

**1. Local-dev http restored** — `resolveBlossomBaseURL` is plain `first?.url` again (pre-PR semantics), with a comment stating the trust rule. The per-send pin works for that URL too: `bound(toServer:)` is documented as the TRUSTED binding level (caller vouches: own configuration, or retry's own earlier stamp) and its guard only rejects unparseable/schemeless/hostless strings — `http://192.168.1.10:3000` binds exactly as configured, so uploads AND downloads to a local box work again exactly as pre-PR. The protocol doc carries a matching TRUST NOTE pointing peer-influenced callers at the policy.

**2. Badge/resolver agreement** — by construction again: badge = `idx == 0`, resolver = `first`. Verified, no UI change needed.

**3. Bind to the matched allowlist URL** — `BlossomServerStampPolicy` now binds `matched.absoluteString`, never the raw stamp: the stamp's origin proves allowlist membership, but its path/query are peer-chosen bytes and no longer shape the request even against the user's own server. Test: a stamp carrying `/evil/prefix?track=1` produces a request through the allowlist URL only. (The policy also keeps https-only on this peer path — with an explicit test that an http stamp is never honored even when the matching http endpoint IS configured; such rows simply ride the live client.)

**4. Default-port normalization** — `originKey` folds 443/https and 80/http, tested in both directions (`:443` stamp vs portless allowlist entry, and the reverse).

**5. Retry-binding tests** — via the `ServerTaggingBlossomClient` hook: a failed-at-fan-out send retried after the resolver moves to a new URL re-uploads through the client bound to the STAMPED server (provider's later value never used, zero unbound uploads); and a hand-crafted legacy nil-stamp row (blob written straight into a dedicated outbox directory) re-uploads through the live client with zero bound uploads.

**6. `SettingsStepIndicator`** — acknowledged: the public surface is a stated prerequisite and EXTRACTION-NOTES' outside-consumer rule is satisfied only when PR 3 (imminent — the OnymOnboarding package wiring) lands its consumer. Happy to fold the publicization into PR 3 instead if you'd rather keep the rule airtight per-PR.

Verification: OnymIOSTests **1234 passed / 0 failures** (3 pre-existing skips; +5 net new tests), OnymDiscovery 117, OnymFoundation 53, full app build (iPhone 17 Pro sim) succeeded.

dev-onym (commented):
**APPROVE** — the trust split (peer stamps allowlist-gated in `BlossomServerStampPolicy`, own config trusted in `bound(toServer:)`) holds up; findings below are non-blocking.

Verified beyond the diff:
- `direction: .outgoing` is set only in `SendMessageInteractor` (5 sites, no receive/forward path), so `retry`'s unvalidated `bound(toServer: stampedServer)` can only ever bind to a self-authored stamp. The trusted level is actually trusted.
- `BlossomServersRepository.refresh()` re-checks `hasUserInteracted` *after* the network await (`BlossomServersRepository.swift:75`), so a boot fetch in flight can't clobber an onboarding catalog pick. Same for the relayer seam.
- `message.media` (`ChatMessage.swift:120`) returns album→image→video and is empty for voice, so the `?? voiceAttachment?.server` fallback in `SendMessageInteractor.swift:987` is reachable and correct.

Findings:
- `BlossomServerStampPolicy.swift:35` — the allowlist gate means a peer whose Blossom server isn't in *your* configured list falls back to your own server and 404s. That matches pre-PR behavior (stamps were never honored), so it's not a regression, but the PR body's "switching servers never 404s already-uploaded history (own or **peers'**)" is only true for own history. Worth correcting the description; a follow-up for user-visible "media unavailable on your servers" signalling would be good.
- `BlossomServersRepository.swift:162` — `resetToDefault()` replaces the whole list, dropping a catalog-consented server, while the pinned consent record is untouched. The new FROM CATALOG row then still renders as consented/active with no endpoint behind it. Same divergence the other seats have, but newly reachable for blossom now that it has a picker.
- `BlossomClient.swift:51` + `SendMessageInteractor.swift:92` — the protocol default `bound(toServer:) -> self` makes the "stamp and blobs can't diverge" invariant convention-only: the interactor's *default* `blossomClient` is a fixed-base `URLSessionBlossomClient`, so `SendMessageInteractor(blossomServerURL: { custom })` with the defaulted client silently stamps one server and uploads to another. Not live (the app always injects the dynamic client), but the failure is invisible. Consider making `URLSessionBlossomClient.bound` rebuild itself with the new base URL rather than inheriting the no-op.
- `BlossomServersRepository.swift:128` — `makeActive` flips `hasUserInteracted = true`, so merely reordering permanently opts the user out of published-default refreshes (`refresh()` guards on that flag). Consistent with add/remove, but reordering is a weaker signal than editing.
- `BlossomRelaySettingsFlow.swift:83` — `stop()` is internal while `start()` is public (matches Nostr; `RelayerSettingsFlow.swift:85` is public). PR 3's onboarding package will call `start()` from outside OnymSettings and won't be able to pair it, leaking the snapshot drain.
- `ChatVideoLoader.swift:44-49` / `ChatVoiceLoader.swift` — moving the allowlist `await` above the inflight check fixes the dedup race, but the `fileExists(dest)` check now sits *before* a suspension point, so a request arriving just as another finishes can still redo one download. Harmless (hash-verified, `.atomic` write), just not the full atomicity the comment implies.
- `DiscoveryCatalogState.swift:96` — `consentedEndpointURLs` is computed on every catalog install for all three seats, but only `RelayerSettingsFlow` reads it. Trivial waste from the dedup; fine to leave.

PR #252, opened 2026-08-15
Make the discovery surface UI-testable: fixture-backed fakes behind --ui-discovery

rinat-enikeev:
First half of onboarding-plan PR 4, with independent value on its own: it makes the discovery surface UI-testable. PR 3's onboarding UI test will build on it.

## Problem

Under `--ui-testing` / `--ui-loopback`, `discoveryRepository` is nil, cascading nils through the catalog, the seat pickers, and the Settings flow — the whole discovery surface (provider management, TOFU confirm, catalog entries, module consent) could not be exercised in UI tests. `UITestFakes.swift` had fakes for every other seam but none for discovery.

## What this adds

- **`UITestDiscoveryFetcher` + `InMemoryDiscoveryStore`** (`Sources/OnymIOS/UITestDiscoveryFakes.swift`, `#if DEBUG` like its sibling `UITestFakes.swift`). The fetcher serves the byte-pinned OnymDiscovery conformance fixtures — provider manifest, `snapshot-1`, and the destination manifest — keyed by exact URL, 404 otherwise. Their signatures verify against the fixture operator key embedded in the provider manifest, so the app runs the **real `DiscoveryTrust` pipeline** (TOFU preview → key pin → snapshot verification → digest-pinned destination manifest) entirely offline. The destination manifest is included so a full add-provider → TOFU → catalog-entries → consent walk works with no network.
- **Fixtures embedded as raw string literals**, not bundled resources — decided deliberately: every digest and signature is over *exact bytes*; a `#"…"#` literal pins those bytes in source (the fixtures are single-line JSON with no backslashes), and the app target gains no resource plumbing through `project.yml` for test-only data. The repository is constructed with the fixture-era `now` (`2026-08-14T00:00:00Z`, same as `Fixture.now` in OnymDiscoveryTests) so the tests keep passing after the real clock passes the fixtures' expiry dates.
- **New launch flag `--ui-discovery`** (honoured only alongside `--ui-testing`): constructs `DiscoveryRepository` with the fakes instead of nil. Without the flag, `--ui-testing` behavior is byte-for-byte unchanged (discovery stays nil), so every existing UI test is untouched.
- **`AppLauncher.launchFresh(discovery:)`** — default `false`, existing call sites unchanged.
- **`DiscoverySettingsUITests`** + `DiscoverySettingsScreen` page object: Settings → DISCOVERY → the seeded default source renders UNCONFIRMED → Confirm… opens the TOFU sheet → fingerprint asserted **equal to the fixture operator key** (`ea4a 6c63 e29c 520a`) → pin → back on the list, the row shows the pinned key and the verified snapshot's catalog entry count.

## Verification

- Full app `build-for-testing` succeeds.
- New `DiscoverySettingsUITests` passes on iPhone 17 Pro simulator.
- Existing `RelayerSettingsUITests/test_firstLaunch_configuredListAutoPopulatedFromManifest` passes (default `--ui-testing` behavior unchanged).
- OnymDiscovery package tests green (iOS simulator destination).

Note for reviewers: `discovery.onym.app` is now live (serving operator key `42b0da00…`), which is exactly why the fakes matter — an early test run on a shared simulator picked up a real fetch and pinned the live key instead of the fixture; the fakes keep the UI tests deterministic and offline regardless of the deployment.

### DIFF
--- a/Sources/OnymIOS/OnymIOSApp.swift
+++ b/Sources/OnymIOS/OnymIOSApp.swift
@@ -105,16 +105,35 @@ struct OnymIOSApp: App {
         // One fetcher instance serves both the repository and the seat
         // catalog below — same URLSession, same size caps; no second
         // session to configure and drift.
-        let discoveryFetcher = URLSessionDiscoveryFetcher()
+        let discoveryFetcher: any DiscoveryFetching
         let discoveryRepository: DiscoveryRepository?
         #if DEBUG
-        discoveryRepository = args.contains("--ui-loopback") || args.contains("--ui-testing")
-            ? nil
-            : DiscoveryRepository(
-                fetcher: discoveryFetcher,
-                store: UserDefaultsDiscoveryStore()
+        if args.contains("--ui-testing") && args.contains("--ui-discovery") {
+            // `--ui-discovery` (with `--ui-testing`): the discovery
+            // surface runs against byte-pinned fixture documents served
+            // offline — real signatures, real `DiscoveryTrust` checks,
+            // fixture-era `now` so the tests outlive the fixtures'
+            // expiry dates. See `UITestDiscoveryFakes.swift`. Without
+            // the flag, `--ui-testing` keeps discovery nil exactly as
+            // before, so every existing UI test is untouched.
+            let fetcher = UITestDiscoveryFetcher()
+            discoveryFetcher = fetcher
+            discoveryRepository = DiscoveryRepository(
+                fetcher: fetcher,
+                store: InMemoryDiscoveryStore(),
+                now: { UITestDiscoveryFixtures.now }
             )
+        } else {
+            discoveryFetcher = URLSessionDiscoveryFetcher()
+            discoveryRepository = args.contains("--ui-loopback") || args.contains("--ui-testing")
+                ? nil
+                : DiscoveryRepository(
+                    fetcher: discoveryFetcher,
+                    store: UserDefaultsDiscoveryStore()
+                )
+        }
         #else
+        discoveryFetcher = URLSessionDiscoveryFetcher()
         discoveryRepository = DiscoveryRepository(
             fetcher: discoveryFetcher,
             store: UserDefaultsDiscoveryStore()
--- a/Sources/OnymIOS/UITestDiscoveryFakes.swift
+++ b/Sources/OnymIOS/UITestDiscoveryFakes.swift
@@ -1,0 +1,134 @@
+#if DEBUG
+import Foundation
+import OnymDiscovery
+
+/// Discovery fakes for the `--ui-discovery` launch flag (honoured only
+/// alongside `--ui-testing`). Sibling of `UITestFakes.swift` — same
+/// rules: production sources because `OnymIOSApp.init` constructs them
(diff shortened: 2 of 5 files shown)


ux-onym (commented):
I verified the substance rather than taking the description on faith: the three embedded literals are byte-identical to the package fixtures (sha256 match), the destination manifest hashes to exactly the `sha256:6536b1a7…` digest the snapshot entry pins, `1_786_665_600` is 2026-08-14T00:00:00Z (same as `Fixture.now`), the fixture URLs match `DiscoverySource.onymDefault`, all six accessibility identifiers exist in OnymSettings, and every `--ui-loopback` test also passes `--ui-testing` without `--ui-discovery`, so existing behavior is genuinely untouched. I could not compile (no Xcode on this Linux host).

**APPROVE** — wiring is correct and the fixture chain verifies end-to-end; findings below are non-blocking.

- `Sources/OnymIOS/DiscoverySeatAdapters.swift:218` — the frozen clock only covers the repository leg. `DiscoverySeatCatalog.reviewEntry` calls `ServiceManifestReviewer().review(…)` with the default `now: Date()` (real clock), and the destination manifest fixture's `validUntil` is `2026-12-31T23:59:59Z`. So the catalog-entry → module-consent walk the PR body says "works with no network" will start failing on 2027-01-01 with `ServiceManifestError.expired`, despite `UITestDiscoveryFixtures.now`. Today's test doesn't reach that path, but PR 3's onboarding test will. Worth threading an injectable `now` into `DiscoverySeatCatalog` under the flag, or at least noting the date-bomb.
- `Sources/OnymIOS/UITestDiscoveryFakes.swift:27` — "the embedded digests below fail the trust checks loudly on any drift" isn't right: the embedded triple is internally self-consistent (manifest key signs the snapshot, snapshot pins the destination digest), so if the package fixtures are regenerated these copies keep verifying happily and just silently diverge. The comment at line 25 names the fix (`UITestDiscoveryFixturesTests` asserting byte-identity with the package files) — that test is the thing that would make drift loud; consider adding it rather than documenting it as already-handled.
- `Tests/OnymIOSUITests/PageObjects/DiscoverySettingsScreen.swift:21` — `let row = discoveryRow` resolves `firstMatching` once, *before* scrolling. `firstMatching` picks a query by `.exists` at call time and falls back to `app.buttons[…]`; if the row isn't yet in the hierarchy it locks onto the buttons query for the whole swipe loop and the later `waitForExistence`. `SettingsScreen.tapClearMessages:144`, cited as the pattern, re-evaluates the computed property each iteration. Re-evaluating (`while !(discoveryRow.exists && …)`) would match the referenced convention and remove the flake window.
- `Sources/OnymIOS/OnymIOSApp.swift:103` — stale now that this PR lands them: "(UI-test discovery fakes land with the settings-UI PR)". Point it at `--ui-discovery` / `UITestDiscoveryFakes.swift`.
- Nit, `Tests/OnymIOSUITests/DiscoverySettingsUITests.swift:44` — `pinConfirmButton.tap()` has no `waitForExistence`; it's on the same sheet as the fingerprint so it's near-certainly present, but every other interaction in the test waits first.

PR #253, opened 2026-08-15
Discovery hygiene: fresh default-source comment, fix raw inflection markup

rinat-enikeev:
Two small cleanups around the discovery surface, zero overlap with `onboarding-unblockers`:

**Stale comment on `DiscoverySource.onymDefault`** — it claimed the deployment "is not live yet". `discovery.onym.app` is live; the comment now states the actual design (the default ships unpinned because pinning is the user's own trust-on-first-use act) and records the currently-served operator key `onym:key:42b0da00…853f24` / fingerprint `4d:a9:ec:c9:e8:6f:6e:97` for maintainers, explicitly marked reference-only, never to be pre-pinned.

**Raw inflection markup in the entry-count label** — UI tests saw `^[1 catalog entries](inflect: true)` rendered literally under the `-AppleLanguages` override. Root cause: the automatic-grammar-agreement string had no entry in any strings catalog (OnymSettings ships no localization resources, and the key was never added to `Resources/Localizable.xcstrings`), so `String(localized:)` fell through to the literal key and the markup reached the screen un-inflected. Fix: a real plural rule via the catalog — new `%lld catalog entries` key with en `one`/`other` and ru `one`/`few`/`many`/`other` variations, resolved through `Bundle.main` like every other app string — extracted into a public `DiscoveryEntriesLabel` helper so the hosted unit tests can pin the rendered forms (`1 catalog entry`, `2 catalog entries`, and never any `^[`/`inflect` leakage in any locale). The page object's stale comment is refreshed; its CONTAINS predicate still matches.

**Verified**: `xcodebuild test` on the OnymDiscovery package scheme (117/117 green) and the app scheme with `-only-testing:OnymIOSTests/DiscoveryEntriesLabelTests` (build + 2/2 green), iPhone 17 Pro simulator.

### DIFF
--- a/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySource.swift
+++ b/Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySource.swift
@@ -86,9 +86,15 @@ public struct DiscoverySource: Codable, Equatable, Hashable, Identifiable, Senda
     }
 
     /// The Onym-operated default provider, seeded on first run only
-    /// (see `DiscoverySourcesConfiguration`). Ships unpinned — the
-    /// deployment is not live yet, and pinning happens through the
-    /// same TOFU confirmation as any other source.
+    /// (see `DiscoverySourcesConfiguration`). Ships unpinned **by
+    /// design**: pinning is the user's own trust-on-first-use act, so
+    /// even the bundled default goes through the same TOFU
+    /// confirmation as any other source. The deployment is live; for
+    /// maintainers, the operator key it currently serves is
+    /// `onym:key:42b0da001104dd03052c7feddab9520c920c9e40d11b245c46c27cf6be853f24`
+    /// (fingerprint `4d:a9:ec:c9:e8:6f:6e:97` — sha256 of the 32 raw
+    /// key bytes, first 8 bytes colon-hex). Reference only — never
+    /// pre-pin it here.
     public static let onymDefault = DiscoverySource(
         providerId: "onym:component:onym-discovery",
         userLabel: "Onym Discovery",
--- a/Packages/OnymSettings/Sources/OnymSettings/DiscoverySettingsView.swift
+++ b/Packages/OnymSettings/Sources/OnymSettings/DiscoverySettingsView.swift
@@ -224,9 +224,20 @@ struct DiscoverySettingsView: View {
     }
 
     private func entriesLine(for status: DiscoverySourceStatus) -> String {
-        let count = flow.entryCount(providerId: status.source.providerId)
-        // Automatic grammar agreement — a real plural rule, not an
-        // English-only `== 1` branch.
-        return String(localized: "^[\(count) catalog entries](inflect: true)")
+        DiscoveryEntriesLabel.text(count: flow.entryCount(providerId: status.source.providerId))
+    }
+}
+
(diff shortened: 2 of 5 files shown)


ux-onym (commented):
**COMMENT** — the localization fix is sound and correctly wired; the new doc comment records a fingerprint in a format the app never displays.

- `Packages/OnymDiscovery/Sources/OnymDiscovery/DiscoverySource.swift:95` — the comment gives the fingerprint as `4d:a9:ec:c9:e8:6f:6e:97` ("sha256 of the 32 raw key bytes, first 8 bytes colon-hex"). That derivation is arithmetically correct (I verified: sha256 of those 32 bytes starts `4da9ecc9e86f6e97`), but it is **not** the app's fingerprint format. `DiscoverySource.fingerprint(ofKeyHex:)` at `DiscoverySource.swift:80` takes the first 16 hex chars of the key in groups of 4 — so the TOFU confirm screen and `ModuleConsentFlow.swift:177` will show `42b0 da00 1104 dd03`. A maintainer using this comment to eyeball-verify a live pin will see a mismatch and have no way to reconcile it. Either record `42b0 da00 1104 dd03` (matching what ships), or keep both and label which is which.
- `DiscoverySource.swift:94` — the key is written `onym:key:42b0…`, but `pinnedOperatorKeyHex` is documented as bare 64-lowercase-hex (`DiscoverySource.swift:63`). Since the point is maintainer reference against a pinned value, quoting the bare hex avoids a prefix-stripping step.
- `Packages/OnymSettings/Sources/OnymSettings/DiscoverySettingsView.swift:240` — outside the app bundle (SwiftUI previews, or any other host of `OnymSettings`) the key isn't in `Bundle.main`, so `String(localized:)` falls back to the formatted key → `1 catalog entries`. No markup leak, so the reported bug is genuinely fixed, but the plural is wrong in previews. Worth a line in the doc comment; not worth code.
- `DiscoverySettingsView.swift:239` — `DiscoveryEntriesLabel` is made `public` purely for test access. `@testable import OnymSettings` from the hosted test would avoid widening the package's API surface; if public is intentional, it reads better in its own file than appended to a view file.
- `Tests/OnymIOSTests/DiscoveryEntriesLabelTests.swift:32` — the exact-form test skips unless the host resolves `en`, and there's no ru counterpart, so the four new ru variations are shipped untested. `AppLauncher.launchFresh(language: "ru")` already exists (`Tests/OnymIOSUITests/Support/AppLauncher.swift:23`) and `RecoveryPhraseBackupUITests.swift:101` uses it — a `1/2/5` ru assertion would pin `запись/записи/записей` cheaply.
- `Resources/Localizable.xcstrings` — new key has both `en` and `ru` with populated variations, so `test_localizable_xcstrings_everyKey_hasEnAndRu` (`Tests/OnymIOSTests/LocalizationCatalogTests.swift:71`) stays green; `extractionState: "manual"` matches sibling keys, which is right since the call site lives in a package Xcode won't scan.
- `Tests/OnymIOSUITests/PageObjects/DiscoverySettingsScreen.swift:78` — CONTAINS on `"\(count) catalog entr"` is correct for the only caller (`DiscoverySettingsUITests.swift:57`, default `language: "en"`), but it silently stops matching if that test is ever run under the ru override. Fine as-is; the refreshed comment now says so.

PR #255, opened 2026-08-15
Onboarding wiring: RootView gate, step surfaces, first-launch composition

rinat-enikeev:
PR 3 of the onboarding sequence (after #249's state machine and #250's unblockers): first-launch, Apple-new-device-style onboarding, fully wired.

## What / why

A fresh install now walks a seven-step full-screen cover before landing in the app. Every step surface is the *same* flow/repository/picker code the Settings screens use — a choice made during onboarding and one made later in Settings run identical paths, so nothing consents, pins, or applies differently.

- **RootView gate** — `OnboardingGate.shouldOnboard` (decided once in `OnymIOSApp.init`, pure read) drives a `fullScreenCover` + `.interactiveDismissDisabled(true)` over the tab bar, same pattern as the moderation consent cover. While onboarding is up the moderation gate's own cover is suppressed (the walk's step 6 *is* that surface); on completion the flow sets the flag, the cover dismisses on the new `OnboardingFlow.isCompleted` signal, and `moderationGateFlow.consentCompleted()` runs so the gate re-checks instead of re-prompting.
- **Grandfathering** — `OnboardingLaunch.isExistingUser` (new, unit-tested): any `hasUserInteracted` across relayer/Nostr/Blossom configurations, or any mandate record, means no onboarding, ever. Seeded-but-untouched defaults deliberately do **not** count.
- **`AppDependencies.makeOnboardingFlow`** — `(@MainActor () -> OnboardingFlow)?`, nil under `--ui-testing` unless `--ui-onboarding`, so every existing UI test launches exactly as before (`--reset-keychain` also clears the completion flag through the store). A sibling `makeOnboardingStepContent` slot carries the step bodies.
- **Notary auto-populate** — the relayer repository's #250 `autoPopulatePolicy` seam now answers false while onboarding is incomplete; `complete()` kicks one `refresh()`, which installs the published defaults iff the user skipped the step (an explicit pick already flipped `hasUserInteracted`).

## Steps

1. **Welcome** — OnymMark brand framing, "your keys, your services"; identity bootstrap stays silent. Unskippable, Continue only.
2. **Service Directory** — TOFU-confirm of the seeded default source through `DiscoverySettingsFlow`'s own confirm path (fetch → fingerprint hero → pin on explicit confirm), plus "add your own provider URL". Skippable → legacy lists.
3. **Message Transport** — seeded Onym Official preselected; discovery catalog entries consent through `ModuleConsentFlow` (sheet, exactly like Settings); custom URL adds without consent.
4. **File Storage** — Blossom catalog picker from #250; a consented pick becomes the ACTIVE first server via `BlossomCatalogConsent.apply` (addEndpoint makeActive).
5. **Notary** — published list + catalog + custom URL over an auto-populate-suppressed configuration; skip installs published defaults after completion.
6. **Moderation** — the mandate flow embedded via the newly extracted public `ModerationConsentContent` (same copy, same accessibility ids as the blocking cover); mandatory while the directory has entries (tri-state fail-closed probe), informational + skippable when empty.
7. **Done** — summary card read back from the repositories (names + endpoint/fingerprint/manifest-hash details) and the backup-phrase nudge pointing at Settings → Back Up Recovery Phrase (deliberately not inlined — that flow takes biometrics).

Progress renders through `SettingsStepIndicator` (count 7). Accessibility ids follow `onboarding.<step>.<element>`.

`OnboardingFlow`'s #249 semantics are untouched except one addition: `isCompleted`, flipped only inside `complete()`, as the presenter's dismissal signal (covered by a new flow test).

Small enablers: `ModuleConsentView`, `DiscoveryCatalogSection`, and the Nostr/Blossom flows' `stop()` are now public; `ModerationConsentContent` extracted from `ModerationConsentView` (pure move, view delegates to it). `project.yml` registers OnymOnboarding for the app + unit-test targets. All new user-facing copy is in `Resources/Localizable.xcstrings` with en + ru (including #249's step titles/subtitles, which had no catalog entries yet).

## Verification

- `OnymOnboarding` package suite: **35 tests, 0 failures** (incl. new `isCompleted` test)
- `OnymDiscovery` package suite: **117 tests, 0 failures**
- `OnymFoundation` package suite: **53 tests, 0 failures**
- App unit suite (`OnymIOSTests`): **1243 tests (3 skipped), 0 failures** (incl. new `OnboardingLaunchTests`)
- App target builds (iPhone 17 Pro simulator, Xcode 26)
- UI-test posture: every existing suite launches via `AppLauncher` with `--ui-testing` and no `--ui-onboarding`, so `makeOnboardingFlow` is nil and no cover presents; representative run of `RelayerSettingsUITests` (exercises the relayer config + auto-populate path under `--ui-testing`): **4 tests, 0 failures** — no onboarding cover presented. The full UI matrix runs in the release workflow as usual.

PR 4 (the UI-test walk of the flow + a `--skip-onboarding` convenience for other suites) follows.

### DIFF
--- a/Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationConsentView.swift
+++ b/Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationConsentView.swift
@@ -18,19 +18,8 @@ public struct ModerationConsentView: View {
     public var body: some View {
         NavigationStack {
             ScrollView {
-                VStack(alignment: .leading, spacing: 0) {
-                    switch flow.state.step {
-                    case .loadingDirectory:
-                        loading
-                    case .pickingAuthority:
-                        picker
-                    case .reviewingManifest, .signing:
-                        review
-                    case .done:
-                        done
-                    }
-                }
-                .padding(.bottom, 32)
+                ModerationConsentContent(flow: flow)
+                    .padding(.bottom, 32)
             }
             .background(OnymTokens.surface.ignoresSafeArea())
             .navigationTitle(navigationTitle)
@@ -69,6 +58,41 @@ public struct ModerationConsentView: View {
         switch flow.state.step {
         case .done: return "Done"
         default: return "Cancel"
+        }
+    }
+
+}
+
+/// The consent surface's step bodies (directory pick → manifest review
+/// → sign → done), extracted from `ModerationConsentView` so the
+/// first-launch onboarding moderation step can embed the exact same
+/// surface — same copy, same accessibility identifiers, same flow —
+/// inside its own scaffold instead of a second implementation.
+/// `ModerationConsentView` remains the standalone presentation
+/// (NavigationStack + dismissal rules) over this content.
+///
+/// The embedder owns the flow's lifecycle: call `flow.start()` /
+/// `flow.stop()` around this view's appearance the way
+/// `ModerationConsentView` does.
+public struct ModerationConsentContent: View {
+    private let flow: ModerationConsentFlow
+
+    public init(flow: ModerationConsentFlow) {
+        self.flow = flow
+    }
+
+    public var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            switch flow.state.step {
+            case .loadingDirectory:
+                loading
+            case .pickingAuthority:
+                picker
+            case .reviewingManifest, .signing:
+                review
+            case .done:
+                done
+            }
         }
     }
 
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
@@ -64,6 +64,11 @@ public final class OnboardingFlow {
     /// answers `false`. Racing the probe must never open a skip window
     /// that a resolved `true` would have closed.
     public private(set) var moderationDirectoryHasEntries: Bool?
+    /// Flips exactly once, inside `complete()` — the observable signal
+    /// the presenter dismisses its cover on. `onCompleted` alone can't
+    /// carry the dismissal: it is bound at the composition root, which
+    /// has no handle on the presenting view's state.
+    public private(set) var isCompleted = false
 
     /// Whether the directory probe has answered — the view shows a
     /// progress state on the skip affordance while this is false.
@@ -203,6 +208,7 @@ public final class OnboardingFlow {
             outcomes[step] = .notApplicable
         }
         store.markOnboardingCompleted()
+        isCompleted = true
         onCompleted()
     }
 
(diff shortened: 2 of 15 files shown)


dev-onym (commented):
**COMMENT** — solid, well-documented wiring; no confirmed blocker, but two behavioral gaps and several polish items worth addressing before merge.

- `Sources/OnymIOS/RootView.swift:169` (onboarding cover) + `:118` (consent cover) — two `.fullScreenCover`s on one chain, and on completion `onboardingFlow = nil` and `consentPresentation = …` are set in the *same* update. That's exactly the dismiss-then-present-in-one-transaction hazard the neighbouring comment at `:119-128` warns about ("the likely outcome is no cover at all — an ungated app"). If the gate still reads `.needsConsent`/`.needsReconsent` at that instant (mandate snapshot not yet recomputed), the consent cover can silently fail to present; it only self-heals on the next gate change or foreground. Safer: re-drive presentation after dismissal (`.onDisappear` on the cover, or a runloop hop) rather than in the `isCompleted` handler.
- `Sources/OnymIOS/OnboardingLaunch.swift:26-36` — grandfathering ignores the strongest signals: an existing identity keychain, existing groups, and `DiscoverySource.hasUserInteracted`. An upgrading user whose relayer known-list fetch never succeeded (offline/blocked, so `RelayerRepository.refresh():149` never flipped the flag) and whose directory is empty reads as a fresh install and gets the full seven-step walk. Adding "identity exists" to the probe closes it cheaply and stays a pure sync read.
- `Sources/OnymIOS/OnymIOSApp.swift:212-218` — because #249 deliberately doesn't persist walk position, a user who force-quits mid-onboarding on every launch leaves `hasCompletedOnboarding()` false forever, so relayer auto-populate never runs. Mostly moot behind a blocking cover, but it's a permanent suppression keyed on a flag only `complete()` writes — worth a comment or a launch-count escape hatch.
- `Sources/OnymIOS/OnboardingSteps.swift` (`OnboardingModerationContent`) — the embedded bodies are tuned for a full-screen `NavigationStack`: `loading` and `done` carry `.padding(.top, 120)` (`ModerationConsentView.swift:104`, `:191`) and the picker its own 16pt insets, cancelled by `.padding(.horizontal, -16)`. Negative padding will let card backgrounds bleed past the scaffold gutters and breaks the moment either side's insets change. Prefer an inset parameter on `ModerationConsentContent`.
- `Sources/OnymIOS/OnboardingSteps.swift` (`OnboardingNotaryContent.configured`) — passes `selected: true` for every row, so all configured notaries render a filled checkmark; nostr/blossom use `idx == 0`. Inconsistent affordance.
- `Sources/OnymIOS/OnymIOSApp.swift` (`loadSummary`, discovery row) — uses `source.operatorKeyFingerprint`, not `pinnedOperatorKeyHex`. A user who *skipped* the TOFU confirm still sees a fingerprint under "Service Directory", which reads as confirmed. Read the pinned state so the summary is actually checkable, as the PR claims.
- `Sources/OnymIOS/OnboardingSteps.swift` + `RootView.swift:41` — `makeFlow()` / `makeOnboardingFlow?()` are invoked inside `init` and discarded by `State(initialValue:)` on every body evaluation, allocating a store-reading flow per render. Harmless now; capture the factory and build in `.task` if step bodies get heavier.
- Tests: `OnboardingLaunchTests` covers the probe well, but the two load-bearing new behaviors — gate-cover suppression while onboarding, and the auto-populate policy flipping on `complete()` — have no automated coverage. The policy closure at `OnymIOSApp.swift:215` is unit-testable as-is.

PR #256, opened 2026-08-15
Onboarding UI tests: seven-step walk, persistence, --skip-onboarding declared default

rinat-enikeev:
PR 4 of 4 — completes the onboarding sequence (#249 state machine → #250 unblockers → #255 wiring → this).

## What / why

- **`--skip-onboarding`, declared** — `AppLauncher.launchFresh` now passes `--skip-onboarding` by default (and gains `onboarding: true` → `--ui-onboarding`). Every existing suite's no-onboarding posture is stated in its launch arguments instead of being an incidental consequence of the factory's nil-under-`--ui-testing` default — a future change to the app's default can't silently put a cover in front of every test. The app honours the flag as an explicit veto that outranks `--ui-onboarding`, and as a plain-DEBUG convenience for skipping the walk on a fresh simulator.
- **`OnboardingUITests.test_onboardingWalk_allSevenSteps_thenNeverAgain`** — the full walk, offline and deterministic, on the #252 discovery fixtures (repository clock parked at 2026-08-14, so the snapshot's 2026-09-12 expiry can't rot the test):
  1. **welcome** — asserts no Skip affordance, Continue.
  2. **discoveryConfirm** — Verify & Confirm on the seeded source, fingerprint hero asserted byte-exact against the fixture operator key (`ea4a 6c63 e29c 520a`), Pin Key & Confirm, provider-confirmed state.
  3. **messageTransport** — seeded Onym Official row present; consents to the fixture courier module through the shared ModuleConsent sheet (real digest-pinned review, free offer preselected).
  4. **blobTransport** — seeded ACTIVE pick kept.
  5. **notary** — auto-populate suppressed during onboarding, so the published UITest list is the offer; adds the testnet relayer and sees it configured.
  6. **moderation** — under `--moderation-needs-consent` the directory has the UITest Authority, so the step is mandatory: asserts Continue disabled and no Skip, then pick → hash-pinned review → I agree and sign → mandate-signed state unlocks Continue.
  7. **done** — summary card renders, Start dismisses onto the Chats tab.

  Then a **second launch without `--reset-keychain`**: the completion flag persisted, and the cover never re-presents even with `--ui-onboarding` still passed.
- **`test_skipOnboarding_neverPresentsCover`** — the declared veto pinned on a completely fresh install.
- New `OnboardingScreen` page object on the `onboarding.<step>.<element>` id convention from #255; consent-sheet and moderation ids reuse the shared surfaces' existing identifiers.

No fakes gaps needed filling: the #252 discovery fixtures (courier entry, `transport.message`) plus the existing moderation fakes covered every step; blossom/relayer catalog consent paths run the same shared sheet exercised on step 3.

## Verification (iPhone 17 Pro sim, Xcode 26)

- New `OnboardingUITests`: **2 tests, 0 failures** (walk 53s, skip-veto 5s)
- Affected existing suites re-run: `DiscoverySettingsUITests` **1/0**, `RelayerSettingsUITests` **4/0**, `RecoveryPhraseBackupUITests` **4/0**
- App unit suite (`OnymIOSTests`): **1243 tests (3 skipped), 0 failures**
- Full UI matrix left to the release workflow (local run of the remaining suites is slow; the AppLauncher change is argument-append-only, and the app-side gate change is inert without the new flags).

### DIFF
--- a/Sources/OnymIOS/OnymIOSApp.swift
+++ b/Sources/OnymIOS/OnymIOSApp.swift
@@ -200,13 +200,22 @@ struct OnymIOSApp: App {
             // opts in, so every existing UI test launches exactly as
             // before. Under the opt-in the fakes are a fresh world, so
             // the existing-user probe answers false by construction.
+            // `--skip-onboarding` is the explicit veto — AppLauncher
+            // passes it by default so a suite's no-onboarding posture
+            // is declared in its launch arguments, and it wins over
+            // `--ui-onboarding` should both ever appear.
             shouldOnboard = args.contains("--ui-onboarding")
+                && !args.contains("--skip-onboarding")
                 && OnboardingGate.shouldOnboard(store: onboardingStore, isExistingUser: { false })
         } else {
-            shouldOnboard = OnboardingGate.shouldOnboard(
-                store: onboardingStore,
-                isExistingUser: { OnboardingLaunch.isExistingUser() }
-            )
+            // `--skip-onboarding` also works for plain DEBUG runs —
+            // a developer convenience for skipping the walk on a
+            // fresh simulator without composing UserDefaults.
+            shouldOnboard = !args.contains("--skip-onboarding")
+                && OnboardingGate.shouldOnboard(
+                    store: onboardingStore,
+                    isExistingUser: { OnboardingLaunch.isExistingUser() }
+                )
         }
         #else
         shouldOnboard = OnboardingGate.shouldOnboard(
--- a/Tests/OnymIOSUITests/OnboardingUITests.swift
+++ b/Tests/OnymIOSUITests/OnboardingUITests.swift
@@ -1,0 +1,224 @@
+import XCTest
+
+/// End-to-end walk of the first-launch onboarding cover under
+/// `--ui-onboarding` + `--ui-discovery`: all seven steps, offline and
+/// deterministic.
+///
+/// The discovery fakes serve the byte-pinned OnymDiscovery conformance
+/// fixtures with the repository clock parked in the fixtures' validity
+/// window (2026-08-14; snapshot expiry 2026-09-12 can't rot the test),
+/// so step 2's TOFU confirm and step 3's module-consent walk run the
+/// full production trust pipeline. The moderation fakes serve the
+/// canned UITest Authority; `--moderation-needs-consent` starts the
+/// mandate store empty so step 6 exercises the real pick → review →
+/// sign path (mandatory: the directory has an entry, so Continue stays
+/// locked until the mandate is signed).
+///
+/// A second launch WITHOUT `--reset-keychain` then proves persistence:
+/// the completion flag survives, and the cover never re-presents even
+/// though `--ui-onboarding` is still passed.
+final class OnboardingUITests: XCTestCase {
+
+    /// First 16 hex chars of the fixture operator key — duplicated
+    /// from `UITestDiscoveryFixtures.operatorKeyFingerprint` on
+    /// purpose (cross-bundle contract, same convention as
+    /// `DiscoverySettingsUITests`).
+    private let fingerprint = "ea4a 6c63 e29c 520a"
+
+    /// The fixture snapshot's one catalog entry, as the aggregate row
+    /// id embedded in the catalog rows' accessibility identifiers:
+    /// `<providerId>|<catalogId>|<componentId>`.
+    private let courierEntryId =
+        "onym:component:onym-discovery|public-all-seats|onym:component:onym-courier"
+
+    func test_onboardingWalk_allSevenSteps_thenNeverAgain() throws {
+        let app = XCUIApplication()
+        app.launchArguments = [
+            "--ui-testing",
+            "--reset-keychain",
+            "--mock-biometric",
+            "--ui-discovery",
+            "--ui-onboarding",
+            "--moderation-needs-consent",
+            "-AppleLanguages", "(en)",
+            "-AppleLocale", "en_US",
+        ]
+        app.launch()
+
+        let onboarding = OnboardingScreen(app: app)
+
+        // Step 1 — Welcome. Unskippable: no Skip affordance.
+        XCTAssertTrue(onboarding.title("welcome").waitForExistence(timeout: 10),
+                      "onboarding cover never presented. Hierarchy:\n\(app.debugDescription)")
+        XCTAssertFalse(onboarding.skip("welcome").exists,
+                       "welcome must not offer Skip")
+        onboarding.continueFrom("welcome")
+
+        // Step 2 — Discovery TOFU confirm. The seeded default source is
+        // unpinned; Verify & Confirm fetches the fixture manifest and
+        // shows the fixtures' operator-key fingerprint, byte-exact.
+        XCTAssertTrue(onboarding.title("discoveryConfirm").waitForExistence(timeout: 5))
+        XCTAssertTrue(onboarding.discoveryConfirmButton.waitForExistence(timeout: 5),
+                      "seeded source's Verify & Confirm never appeared")
+        onboarding.discoveryConfirmButton.tap()
+        XCTAssertTrue(onboarding.discoveryFingerprint.waitForExistence(timeout: 5),
+                      "TOFU fingerprint never appeared")
+        XCTAssertEqual(onboarding.discoveryFingerprint.label, fingerprint,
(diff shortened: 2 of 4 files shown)


dev-onym (commented):
REQUEST_CHANGES — the walk test is solid, but the skip-veto test doesn't exercise the veto it claims to pin.

- **`Tests/OnymIOSUITests/OnboardingUITests.swift:180` — `test_skipOnboarding_neverPresentsCover` is vacuous.** The UI-testing gate is `args.contains("--ui-onboarding") && !args.contains("--skip-onboarding") && …` (`Sources/OnymIOS/OnymIOSApp.swift:207`). `AppLauncher.launchFresh()` doesn't pass `--ui-onboarding`, so `shouldOnboard` is already `false` from the first conjunct — deleting `--skip-onboarding` entirely leaves this test green. It pins the factory's nil default, exactly the incidental posture the PR set out to replace. Pass **both** flags (`launchFresh(onboarding: true)` plus an explicit `--skip-onboarding`, or a hand-built arg list) so the veto's precedence is what's asserted.
- **`Tests/OnymIOSUITests/Support/AppLauncher.swift:33` — `onboarding:` is dead API.** No call site passes it; the walk test hand-rolls its launch arguments (it needs `--moderation-needs-consent`). Either use it in the fixed veto test above or drop the parameter until a caller exists.
- **`Tests/OnymIOSUITests/OnboardingUITests.swift:131` — `XCTAssertFalse(primary("moderation").isEnabled)` passes vacuously when the element is absent.** XCUIElement returns `false` for `isEnabled` on a non-resolving query, so an identifier rename turns the mandatory-step assertion into a no-op. Assert `.exists` first (the neighbouring `skip(...)` check is fine — absence *is* the claim there).
- **`Tests/OnymIOSUITests/PageObjects/OnboardingScreen.swift:72-78` — the query loop in `catalogRow` never fires.** It probes `.exists` synchronously at call time, before the discovery fetch lands, so all three miss and it always returns the `app.buttons[...]` fallback that the caller then waits on. The comment implies element-type tolerance that isn't there; drop the loop (matching `DiscoverySettingsScreen`'s plain `app.buttons[...]`) or make each branch wait.
- **`Tests/OnymIOSUITests/OnboardingUITests.swift:100-103` — step 4 asserts only the title.** The PR describes "seeded ACTIVE pick kept", but nothing checks `configuredRow(step: "blobTransport", url:)`; a step that silently rendered the empty-state card would still pass. One `configuredRow` assertion closes it — the helper already exists and the identifier is emitted at `Sources/OnymIOS/OnboardingSteps.swift:690`.

Verified correct: scaffold/step/notary/consent/moderation identifiers all match their emitters (`OnboardingStepScaffold.swift:74,100,110`, `OnboardingSteps.swift:332,454,476,562,890,1003`, `DiscoveryCatalogSection.swift:81`, `ModuleConsentView.swift:193,377,47`, `ModerationConsentView.swift:168,293,317`), the `<providerId>|<catalogId>|<componentId>` entry id matches `DiscoveryRepository.swift:50`, and the fixture ids resolve (`UITestFakes.swift:316`). The app-side gate change is inert for release builds and for existing suites.

rinat-enikeev:
Addressed all five findings from the review in 580f28f (amended commit, force-pushed). Point by point:

1. **Vacuous veto test — real catch, fixed.** Verified: `launchFresh()` never passed `--ui-onboarding`, and the gate is `args.contains("--ui-onboarding") && !args.contains("--skip-onboarding") && …`, so the first conjunct alone kept the cover away — the test passed even with `--skip-onboarding` deleted, i.e. the assertion survived the disappearance of the very thing it guarded. `test_skipOnboarding_neverPresentsCover` now hand-builds its launch arguments mirroring the walk test's posture plus the veto, passing BOTH `--ui-onboarding` and `--skip-onboarding`, so it pins the veto's precedence specifically. The doc comment explains why both flags are load-bearing.

2. **Dead `onboarding:` parameter — removed.** Since the fixed veto test needs both flags at once (which `launchFresh`'s either/or append couldn't express), the hand-built list was the simpler route, so the parameter stayed dead and is gone. `launchFresh` now unconditionally appends `--skip-onboarding`; onboarding suites build their own argument lists. Doc comments in `AppLauncher` and `OnboardingScreen` updated to match.

3. **Vacuous `isEnabled` on moderation Continue — real catch, fixed.** Verified: a non-resolving `XCUIElement` reports `isEnabled == false`, so `XCTAssertFalse(primary("moderation").isEnabled)` passed whether the button was disabled or simply absent — again an assertion that passes when its subject disappears. Now `waitForExistence(timeout: 5)` is asserted first, then not-enabled. The neighbouring `skip("moderation").exists == false` check is left as is — absence is the claim there.

4. **`catalogRow` element-type loop — simplified.** Verified: the loop probed `.exists` synchronously, before the caller's `waitForExistence` had ever let the row land, so on a fresh step it always fell through to the `app.buttons[...]` fallback — the implied cross-element-type tolerance never actually executed. Replaced with the plain `app.buttons[...]` query (matching `DiscoverySettingsScreen`) and dropped the misleading comment.

5. **Step 4 empty-state blind spot — assertion added.** The blob-transport step now asserts `configuredRow(step: "blobTransport", url: "https://blossom.onym.app")` (the identifier emitted at `OnboardingSteps.swift:690` for the seeded ACTIVE endpoint) before continuing, so an empty-state render fails instead of sliding by on the title alone.

The two vacuous-assertion catches (1 and 3) were the substantive ones — both were checks that kept passing when the behaviour they existed to pin was removed. Thanks for running them to ground.

Verification on iPhone 17 Pro simulator: `build-for-testing` succeeded; `test_onboardingWalk_allSevenSteps_thenNeverAgain` passed (56.7 s), `test_skipOnboarding_neverPresentsCover` passed (4.7 s).

dev-onym (commented):
**APPROVE** — app-side gate change is correct and inert for existing suites; all new identifiers verified against the app/package sources. Findings are test-hardening and doc drift.

- `Tests/OnymIOSUITests/OnboardingUITests.swift:80` — after `consentDismiss.tap()`, `continueFrom("messageTransport")` only checks `exists`/`isEnabled`. The step title and primary button still *exist* behind a presented sheet (`Sources/OnymIOS/OnboardingSteps.swift:536`), so the tap can race the sheet's dismissal animation and its `onDismiss → consentDismissed()`, which is what records `.consented(componentId:)`. Worth a `waitForNonExistence` on `consentAccept` (or an `isHittable` poll) before continuing — this is the one plausible flake in the walk.
- `Tests/OnymIOSUITests/PageObjects/OnboardingScreen.swift:41` — same gap generalized: `continueFrom` asserts `isEnabled` but never `isHittable`. Adding it makes the page object safe for every sheet-adjacent step, not just step 3.
- `Tests/OnymIOSUITests/Support/AppLauncher.swift:47` — the "declared posture" invariant only covers AppLauncher call sites. Four suites build their own argument lists and still get no-onboarding incidentally: `ScreenshotUITests.swift:25`, `DeleteChatUITests.swift:12`, `SearchUITests.swift:14`, `MultiIdentityChatUITests.swift:33`. Adding `--skip-onboarding` to those `baseArgs` would actually close the hole the PR describes.
- `README.md:524-533` — the hook table still says "reads three launch arguments" and "`--ui-testing`: Required gate. Without it the App ignores the other two args." `--skip-onboarding` (`Sources/OnymIOS/OnymIOSApp.swift:214`) is the first hook that deliberately works *outside* `--ui-testing`, so that sentence is now wrong. (The table was already stale re: `--ui-loopback`/`--ui-discovery`/`--ui-onboarding`; this PR adds the flag that contradicts it.)
- `Sources/OnymIOS/OnymIOSApp.swift:214` — the plain-DEBUG branch has no coverage: every UI test passes `--ui-testing`, so only the line 207-209 conjunct is exercised by `test_skipOnboarding_neverPresentsCover`. Low risk (single `&&`), noting for completeness.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:29` — no `continueAfterFailure = false` in a `setUp`; seven other suites set it (e.g. `RecoveryPhraseBackupUITests.swift:26`). On a 53s seven-step walk, a mid-walk failure will cascade through the remaining steps and bury the first real error in CI logs.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:104` — step 4 asserts the configured row exists but not that it's the ACTIVE pick, which is what the PR body claims is being pinned. If a future change reorders the seeded blossom list, the test still passes.

PR #257, opened 2026-08-15
Settings: Restart Onboarding — re-run seat selection, keep identity and chats

rinat-enikeev:
Adds **Settings → SETUP → Restart Onboarding**: re-run the first-launch seat selection while keeping everything else.

## What / why

- **Settings row + confirmation** — new SETUP section above DATA (house style: card row + footnote, alert like the message-cache clear but single-step, because nothing is deleted). The alert states the contract explicitly: identity, chats, and messages are kept; transport / file storage / notary / moderation selection re-runs; current choices stay until changed. Row and section hide when onboarding isn't wired (UI harness without `--ui-onboarding`) — same optional-factory pattern as the moderation/discovery rows.
- **Overriding grandfathering without weakening it** — the cold-boot gate's existing-user probe would (correctly) suppress onboarding for any configured user, so restart cannot simply clear the flag and re-run the gate. Design: `OnboardingStore` gains a persisted restart marker (`app.onym.ios.onboarding.restartRequested`). `requestRestart()` sets it and clears the completion flag in one intent; `OnboardingGate.shouldOnboard` now checks the marker **before** the flag and the probe. An explicit request outranks grandfathering; upgraders never carry the marker, so their cold-boot behavior is byte-for-byte unchanged (pinned by tests).
- **Presentation** — new `OnboardingRestartController` (@Observable) in the composition root: Settings raises `pendingRestart`, RootView observes it and presents a fresh walk via the same factories the launch path uses. The factories now exist whenever onboarding is *available* (decoupled from the new `presentOnboardingAtLaunch` flag), not only on launches that onboard.
- **During the re-run** — nothing is deleted up front: the step surfaces already render the applied configuration (seeded/ACTIVE rows, configured lists). The moderation step now records a standing active mandate as satisfying the step — re-consent happens only if the user changes authority through the existing pick → review → sign path (ModerationConsentFlow's own switch semantics; nothing new invented). First runs are unaffected (no mandate → unchanged mandatory behavior).
- **Auto-populate** — the #250 relayer `autoPopulatePolicy` re-arms while a restart walk is active (the policy reads the persisted marker — on steady-state launches the marker is the *only* valid suppressor, since grandfathered users never carry the completion flag) and restores on completion, identical to a first run.
- **Mid-walk kill** — the marker persists until `markOnboardingCompleted()` clears it, so the next launch resumes onboarding through the normal gate; chats/tabs remain intact underneath the cover, and the completion path (gate poke + relayer refresh) is the first-run path.

## Verification

- `OnymOnboarding` package suite: **41 tests, 0 failures** (+6 new: marker/flag transitions, restart-outranks-grandfathering, persist-until-completion, upgrader behavior pinned)
- App unit suite (`OnymIOSTests`): **1248 tests (3 skipped), 0 failures** (+5 new `OnboardingRestartControllerTests`, incl. the Settings-action → cover-presentation signal)
- Settings-touching UI suites re-run: `ClearMessageCacheUITests` **1/0**, `DiscoverySettingsUITests` **1/0** (row correctly absent under the harness)
- App builds (iPhone 17 Pro sim, Xcode 26)

## Overlap note

Deliberately does **not** touch `AppLauncher` or the onboarding UI-test files — those belong to the open #256. This PR edits `OnymIOSApp.swift`'s gate block (availability split + policy), which #256 also touches for `--skip-onboarding`; whichever lands second takes a small, mechanical rebase — the two changes compose (`--skip-onboarding` vetoes launch presentation; availability stays keyed on `--ui-onboarding`).

### DIFF
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingStore.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingStore.swift
@@ -1,30 +1,49 @@
 import Foundation
 
-/// Persistence seam for the one bit the onboarding gate needs: has the
-/// user completed (or been grandfathered past) the first-launch flow.
+/// Persistence seam for the two bits the onboarding gate needs: has
+/// the user completed (or been grandfathered past) the first-launch
+/// flow, and did they explicitly ask to run it again.
 public protocol OnboardingStore: Sendable {
     /// Whether the completion flag is set.
     func hasCompletedOnboarding() -> Bool
-    /// Set the completion flag. Idempotent.
+    /// Set the completion flag. Idempotent. Also ends any pending
+    /// restart request — completion is completion, however the walk
+    /// was entered.
     func markOnboardingCompleted()
-    /// Clear the completion flag, so the next launch onboards again.
-    /// The app's `--reset-keychain` test path calls this instead of
-    /// hardcoding the storage key. Idempotent.
+    /// Clear the completion flag AND any restart request, so the next
+    /// launch onboards again as if fresh. The app's `--reset-keychain`
+    /// test path calls this instead of hardcoding the storage keys.
+    /// Idempotent.
     func resetOnboarding()
+    /// Whether the user explicitly asked to re-run onboarding
+    /// (Settings → Restart Onboarding) and hasn't completed it since.
+    /// Persisted: an app killed mid-restart-walk resumes on next
+    /// launch. This is the signal that OUTRANKS grandfathering in
+    /// `OnboardingGate.shouldOnboard` — an existing user's configured
+    /// state must suppress onboarding they never asked for, never
+    /// onboarding they explicitly requested.
+    func isRestartRequested() -> Bool
+    /// Record an explicit restart request: sets the restart marker and
+    /// clears the completion flag in one intent, so every reading of
+    /// the store agrees the walk is due. Idempotent.
+    func requestRestart()
 }
 
-/// Production `OnboardingStore`. Single boolean under
-/// `app.onym.ios.onboarding.completed` — cleared by the app's
-/// `--reset-keychain` test path alongside the other first-run state.
+/// Production `OnboardingStore`. Two booleans —
+/// `app.onym.ios.onboarding.completed` and
+/// `app.onym.ios.onboarding.restartRequested` — cleared together by
+/// the app's `--reset-keychain` test path alongside the other
+/// first-run state.
 ///
 /// `@unchecked Sendable` for the same reason as the sibling
 /// UserDefaults stores (e.g. `UserDefaultsSeatSelectionStore`):
 /// `UserDefaults` is documented thread-safe but not formally
 /// `Sendable`.
 public struct UserDefaultsOnboardingStore: OnboardingStore, @unchecked Sendable {
-    /// Internal so tests can assert the exact key the app's reset path
-    /// must clear.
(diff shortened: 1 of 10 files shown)


dev-onym (commented):
**COMMENT** — design is sound (marker-outranks-grandfathering is the right call and well-pinned by tests); findings are UX/robustness, no correctness blocker found.

- `Resources/Localizable.xcstrings:626` — the `"Review your service choices again"` entry is dead. `SettingsRow.subtitle` is a plain `String` rendered as `Text(subtitle)` (`Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift:393`), i.e. verbatim, no key lookup. Every other row subtitle in `SettingsView.swift` (lines 128, 230, 285) is deliberately absent from the catalog for the same reason. The Russian string will never render; either drop the entry or route the subtitle through a localized value. The other five new entries (title, `SETUP`, footnote, alert title/body/button) are `LocalizedStringKey`/`Text` and are fine.
- `Sources/OnymIOS/OnboardingLaunch.swift:44` + `Packages/OnymOnboarding/.../OnboardingView.swift:44` — **a restart has no off-ramp.** `requestRestart()` clears the completion flag and writes the persisted marker *before* the cover is up, and the walk's only exit is `complete()` (no cancel; `interactiveDismissDisabled`). A user who taps the row to look, or force-quits, is now onboarding-due on **every** launch until they walk all the way to Done — and for a grandfathered user that state was previously unreachable. The alert is the sole undo. Consider writing the marker on first `advance()` rather than on confirm, or giving the restart-entered walk a Cancel that restores the prior flag.
- `Packages/OnymModerationUI/.../ModerationConsentView.swift:239` — `rowSubtitle` marks "Your current authority" only `case .reconsent`, and the picker blurb for `.onboarding` reads as a first-time choice. So on a restart the moderation step is the one surface that does *not* "render the applied configuration" the PR body promises — the user sees an unmarked list and no indication which authority they're already under, while the step is silently auto-satisfied underneath.
- `Sources/OnymIOS/RootView.swift:134-136` — `consumeRestart()` runs *before* the `onboardingFlow == nil` / factory guards. If either guard ever fails, the in-session signal is dropped while the persisted marker + cleared flag survive: no cover this session, then an unexplained onboarding walk at the next cold boot. Unreachable today (Settings is unreachable while the launch cover is up), but move the consume below the guards — it costs nothing.
- `Sources/OnymIOS/OnboardingSteps.swift:948` — the guard tests `outcomes[.moderation]` but `recordOutcome` writes `outcomes[flow.step]` (`OnboardingFlow.swift:165`), so correctness rests entirely on `OnboardingView` rendering exactly one step at a time. It does today (`scaffold(for: flow.step)`), but a future paged/preloaded step container would silently record a moderation `componentId` into a neighbouring step's outcome. A `recordOutcome(_:for:)` overload, or a `guard onboarding.step == .moderation`, would pin it.
- Test coverage: the two behavioral changes with the most surface area — the moderation auto-record (`OnboardingSteps.swift:948`) and the RootView present-on-`pendingRestart` path — are the ones without tests. `OnboardingRestartControllerTests` covers the controller/store contract only; the "Settings-action → cover-presentation" claim in the PR body is `test_requestRestart_raisesThePresentationSignal`, which asserts the flag, not the presentation. No UI test taps `settings.restart_onboarding_row`.
- Not verified here: build and test results (Linux worktree, no Xcode). The store/gate logic, key names, and reset paths I checked by reading; conformer count is 1 (`UserDefaultsOnboardingStore`), so the protocol additions break nothing else.

PR #258, opened 2026-08-15
Onboarding: moderation step loses its skip path entirely

rinat-enikeev:
Moderation-authority selection is no longer skippable during onboarding — in any directory state.

## What / why

#249/#255 made the moderation step mandatory when the authority directory has entries (tri-state probe, fail-closed while unresolved) but offered a Skip when the directory was empty. This removes the skip path entirely:

- **Directory non-empty, or probe unresolved** — unchanged mandatory semantics: the user must pick an authority and complete the pick → review → sign consent to advance. `skip()` and `advance()` both refuse until the consent outcome is recorded; the probe stays fail-closed (unresolved = mandatory).
- **Directory EMPTY** — nothing is selectable, so hard-blocking would brick onboarding. The step becomes **Continue-only informational**: no Skip button, no `.skipped` outcome. The plain Continue is an acknowledgment, recorded as the new distinct `StepOutcome.unavailable`, so downstream state never reads "moderation wasn't available yet" as "the user opted out" — opting out of moderation is not a thing. The Done summary renders it as "Unavailable" (en + ru).

Changed at the layer that declares skippability — `OnboardingFlow.isSkippable` in the OnymOnboarding package; the view's Skip slot keys off it, so no view-layer change was needed beyond the summary label. `isMandatory`, the probe, and #249's other hardened semantics (welcome/done unskippable, advance gating, `complete()` only from done) are untouched.

Tests: the moderation skip-path tests became must-consent tests; added the empty-directory Continue-only/`.unavailable` test and a never-skippable sweep across all three probe states (unresolved / resolved-empty / resolved-non-empty).

## Verification

- `OnymOnboarding` package suite: **42 tests, 0 failures**
- App unit suite (`OnymIOSTests`): **1248 tests (3 skipped), 0 failures**
- App builds clean (iPhone 17 Pro sim, Xcode 26)
- The onboarding walk UI test lives in the open #256 and is not on main yet, so this change is covered by the unit suites here.

## #256 / #257 interaction

- **#256** (onboarding UI tests): its walk runs under `--moderation-needs-consent` (non-empty directory) and already asserts Continue-disabled + **no Skip** on the moderation step — that assertion becomes universally true and survives unchanged. This PR touches `OnboardingFlow.swift` / `OnboardingView.swift` / the flow tests, none of #256's files; expect at most a clean rebase. No follow-up tweak needed.
- **#257** (restart onboarding): merged into this base already. Its standing-mandate behavior composes cleanly — on a restart walk the active mandate records `.consented(...)`, which satisfies the (now stricter) step without re-signing; switching authority still overwrites via the normal consent path.

### DIFF
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
@@ -26,8 +26,16 @@ public enum StepOutcome: Equatable, Sendable {
     /// defaults.
     case skipped
     /// The step had nothing to decide (informational steps: welcome,
-    /// done, moderation-with-empty-directory).
+    /// done).
     case notApplicable
+    /// The step's obligation exists but could not be offered — today
+    /// exactly moderation with an EMPTY authority directory: there is
+    /// no authority to pick, so the user acknowledges with Continue.
+    /// Distinct from `.skipped` (a choice the user declined to make)
+    /// and `.notApplicable` (nothing to decide at all) so downstream
+    /// state never mistakes "moderation wasn't available" for
+    /// "the user opted out" — opting out of moderation is not a thing.
+    case unavailable
 }
 
 /// State machine for the first-launch onboarding sequence. Modeled on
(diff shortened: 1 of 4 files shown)


dev-onym (commented):
**APPROVE** — core state-machine change is correct; findings are stale-doc/UI-polish, none blocking.

Verified: `advance()` checks the mandatory guard *before* backfilling, so `.unavailable` can only be written on a resolved-empty directory (`OnboardingFlow.swift:188-197`); fail-closed `isMandatory` is untouched; both `DoneStepContent` switches handle the new case exhaustively (`OnboardingView.swift:191-206`); `"Unavailable"` lands in the app catalog alongside the existing `Chosen`/`Default` package strings, which is the established resolution path. No Swift toolchain in this environment, so I could not re-run the suites.

- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:63` — `showsSkipProgress` is now dead-ended: `isSkippable(.moderation)` is a constant `false`, so the spinner occupies a slot that can never be replaced by a Skip button — it appears, then vanishes into nothing when the probe resolves. The comment at :64-66 ("moderation's skippability is unknown") is no longer true; what's actually unresolved is *mandatoriness* (Continue enabled/disabled). Either drive the pending indicator off the Continue gate or drop it.
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:70-73` — `moderationDirectoryHasEntries` doc still reads "unskippable/mandatory until the probe answers `false`" and "must never open a skip window"; with the skip path gone only mandatoriness varies. Same for :81-82 ("the view shows a progress state on the skip affordance"), which points at the stale affordance above.
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:199-201` — `skip()` doc: "a no-op on an unskippable step (moderation with a non-empty directory, done)" — moderation is now unskippable in *every* state, and `welcome` is missing from the list.
- `Sources/OnymIOS/OnboardingSteps.swift:913` — directly contradicts the new rule: "With an empty directory the surface is informational and the step skippable." Should say Continue-only / `.unavailable`.
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:127` — the moderation subtitle still says "Choose a moderation authority and review its terms" in the empty-directory case where nothing is choosable. Mitigated by the body's "No authorities are published yet." (`Packages/OnymModerationUI/Sources/OnymModerationUI/ModerationConsentView.swift:143-145`), but that subtitle used to be paired with a visible Skip and now with a bare Continue.
- Test gap (optional): nothing exercises the Done-summary mapping for `.unavailable`; `symbol`/`label` are private, so the new case's icon/label are unasserted while the flow-level outcome is well covered.
- PR body nit: "#256 … is not on main yet" is stale for this base — `Tests/OnymIOSUITests/OnboardingUITests.swift` is present at `fdc3af0`. Its step-6 assertions (:126-140, Continue disabled + no Skip under `--moderation-needs-consent`, which seeds an entry-bearing directory with an empty mandate store) survive unchanged, as claimed.

PR #259, opened 2026-08-15
Onboarding redesign 1/3: three-step default path, services hub, recovery step

rinat-enikeev:
First of three stacked PRs implementing the [Onym Onboarding redesign](https://claude.ai/design/p/f3d5bdf0-84de-429c-9289-e7311f4b6c6b?file=Onym+Onboarding.dc.html): collapse the seven protocol-named wizard screens into a three-step default path, with every choice still available behind the "Choose services myself" branch.

## Step model (OnymOnboarding)
- `welcome → identity → services → moderation → recoveryPhrase → done`
- `indicatorPosition(for:)` numbers only identity/services/moderation ("step N of 3"); welcome, recovery and done are unnumbered
- Only `recoveryPhrase` is skippable, labeled **Remind me later**; moderation keeps its never-skippable, fail-closed probe rules verbatim
- Per-step primary titles: *Create my identity* / *I've written it down* / *Start messaging*

## App layer
- **Services step**: "Recommended setup" card preselected (seeded defaults + published lists, unchanged semantics — same as skipping every old step); "Choose services myself" opens a hub sheet whose pushed screens host the *exact* surfaces the old wizard steps rendered, driven by the same Settings flows (discovery TOFU, nostr, blossom, relayer). Sub-surface consents record onto the services step.
- **Identity step** surfaces the silent key bootstrap as a checklist.
- **Recovery step** runs the same `RecoveryPhraseBackupFlow` Settings uses (biometric gate, verify quiz) from a sheet — promoted from the old Done-step footnote.
- Moderation surface untouched; its step raw value is stable so `onboarding.moderation.*` accessibility ids don't churn.

Package tests adapted mechanically (44 green). Visual polish (plain-language seat names, In use / Suggested sections, chips, terms re-skin, counts summary) is PR 2; new behavior tests + the UI walk update are PR 3.


dev-onym (commented):
**REQUEST_CHANGES** — the step model and hub composition are sound; the consent-recording semantics and a few state/lifecycle details need fixing before merge.

- `Sources/OnymIOS/OnboardingSteps.swift:254-268` — the identity checklist is a static claim, not bound to state. It renders green "Identity key created" / "Recovery phrase generated" unconditionally, while the bootstrap is a fire-and-forget `try? await identityRepository.bootstrap()` (`OnymIOSApp.swift:1279`) whose race is real enough that `RecoveryPhraseBackupFlow` carries `isReady` to guard it. On a slow or failed bootstrap the step asserts something false and then hands the user a recovery step that can't reveal. Bind to the identity snapshot (spinner → check → error).
- `OnboardingSteps.swift:337-338,357-358` — tapping "Recommended setup" records `.notApplicable`, so an explicit affirmative choice is indistinguishable from walking past the screen (`OnboardingView.swift:241` maps both to "Default"). This also contradicts the doc this PR writes at `OnboardingFlow.swift:31-34`, which lists "accepting the recommended setup" under `.consented(componentId: nil)`. Record `.consented(nil)`.
- `OnboardingSteps.swift:341-350` — `onDone` only records when `outcomes[.services] == nil`; after a prior Recommended tap it leaves `.notApplicable` while flipping `usingCustom = true`, so the two pieces of state disagree.
- `OnboardingSteps.swift:319` — `usingCustom` is view-local `@State` with no rehydration from `onboarding.outcomes[.services]`. Navigate to moderation and Back and the "Selected" chip snaps to Recommended even though the user configured custom services. Derive it from the outcome instead.
- `OnboardingSteps.swift:540` — "Use the recommended setup instead" only rewrites the outcome; consent pins the sub-surfaces already persisted stay in place. The affordance reads as a revert but isn't one — either reword or actually undo.
- `OnboardingSteps.swift:1385-1390` — nothing sets `showBackup = false` on completion. `RecoveryPhraseBackupView`'s DoneScreen calls `tappedDoneFromCompletion()`, which resets the flow to `.intro` (`RecoveryPhraseBackupFlow.swift:169-172`), so finishing the quiz drops the user back at the intro screen inside a still-presented sheet; swipe-down is the only exit. Inherited from Settings, but onboarding makes it load-bearing — dismiss on `.done`.
- `OnboardingFlow.swift:146` + primary title "I've written it down" — nothing gates the primary on having revealed the phrase, and advancing without revealing backfills `.notApplicable` → "Default", not `.skipped` → "Later". A user who never saw the words is recorded the same as one who has them. Either gate the primary until `backedUp`, or backfill `.skipped`.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:60-107` still walks `discoveryConfirm` / `messageTransport` / `blobTransport` as top-level steps; those surfaces now live behind the hub sheet, so the UI suite is red on this commit. PR 3 fixing it doesn't help bisects — consider skipping the affected tests here.
- `OnboardingSteps.swift:379-382` — recommended-card values are hardcoded strings, not read from the seeded defaults they claim to describe. Fine if PR 2 replaces them; flagging so it isn't lost.
- Nit `OnboardingStepScaffold.swift:~145` — `indicator.frame(maxWidth: .infinity).padding(.top, 24)` still applies to the now-empty indicator on welcome/recovery/done, leaving a blank 24pt band plus VStack spacing. Worth conditioning the slot.

rinat-enikeev:
Thanks — all findings addressed in e3c1644 (propagated up the stack; the walk test now lives in #261):

- **Identity checklist** now binds to the bootstrap via an injected `identityReady` (idempotent `repository.bootstrap()` await): spinner while resolving, green checks on a snapshot, an explicit failure card with retry otherwise.
- **Recommended tap records `.consented(componentId: nil)`**, matching the `StepOutcome` doc; walking past still backfills `.notApplicable`, so affirmative choice and drive-by stay distinguishable.
- **Hub Done** keeps a sub-surface's componentId consent and only fills `.consented(nil)` over nil/`.notApplicable` — the two pieces of state can no longer disagree.
- **`servicesChoice` moved onto the flow** (`ServicesSetupChoice`); the step body derives the Selected chip from it, so Back/forward can't snap it back to Recommended.
- **"Use the recommended setup instead"** now records the consent and carries a footnote stating that already-added services stay configured — the affordance no longer reads as a revert. (Actual undo would mean unwinding consent pins across three repositories; happy to do that as a follow-up if you'd rather have it.)
- **Backup sheet dismisses** on the DoneScreen's `.done → .intro` transition instead of stranding the user at the re-presented intro.
- **"I've written it down" is outcome-gated**: new `requiresOutcomeToAdvance` disables the primary (and guards `advance()`) until the reveal records the outcome; "Remind me later" stays the honest escape. The #261 walk asserts disabled-before-reveal and unlocks it through the real backup sheet.
- **Old seven-step walk is `XCTSkip`'d on this PR** so the commit is green for bisects; #261 replaces it with the redesigned walk.
- **Recommended-card values**: left as the fixed promised set, now documented as such in code — the Done summary is the live, checkable view (its values read from the repositories). Shout if you'd rather have the card read live state too.
- **Indicator band**: scaffold takes `showsIndicator` and drops the slot (and its 24pt padding) entirely on welcome/recovery/done.

Local state: 47 package tests, 1248 app unit tests, 2/2 onboarding UI tests green.

dev-onym (commented):
**REQUEST_CHANGES** — flow/state machine is clean and well-tested; two view-state issues and a coverage gap.

- `Sources/OnymIOS/OnboardingSteps.swift:1428` — `sawPhrase`/`backedUp` are view-local `@State`, but the step content is rebuilt (new `AnyView` type per step) on every navigation. Walk: reveal + verify → Continue → Back from Done. The status card resets to amber "Not backed up yet" while the primary "I've written it down" is enabled (the flow-level outcome survived) — a fresh `makeBackupFlow()` is also constructed, so re-reveal re-prompts biometrics. This is exactly the bug `servicesChoice` was hoisted onto the flow to avoid (`OnboardingFlow.swift:103`); recovery needs the same treatment. Derive both booleans from flow state.
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:190` — the reveal and the verified-done path both record `.consented(componentId: nil)`, so nothing can distinguish "saw the words" from "verified them". The Done summary then reports the recovery phrase as "Chosen" for a user who only revealed. Either add a distinct outcome/flag or don't claim backup on reveal alone.
- `Sources/OnymIOS/OnboardingSteps.swift:289` — the doc comment says "a failed bootstrap must not hand the user a recovery step that can't reveal", but nothing enforces it: `requiresOutcomeToAdvance(.identity)` is `false` and the primary is never disabled, so `.failed` users walk straight on to services and a recovery step whose reveal will fail. Either gate Continue on `phase == .ready` or drop the claim from the comment.
- `Sources/OnymIOS/OnboardingSteps.swift:425` — the recommended card renders "Selected" + blue border from `servicesChoice`, which defaults to `.recommended`, yet the consent is only recorded on tap. Tapping it is a visual no-op (nothing on screen changes), and a user who just hits Continue gets `.notApplicable` backfilled despite the screen having said "Selected" the whole time. Suggest seeding the outcome when the step is first shown, or driving the chip from the recorded outcome.
- `Sources/OnymIOS/OnboardingSteps.swift:413` — hub sub-surfaces all record onto the single `.services` outcome, so the "keep the specific consent" branch preserves whichever seat the user happened to consent to last. Arbitrary but harmless given the app's Done summary reads from repositories; worth a comment saying so.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:27` — skipping the full walk leaves zero end-to-end coverage that the cover presents, completes, and doesn't re-present; the surviving test only asserts absence under `--skip-onboarding`. Understood that PR 3 restores it, but the completion-flag regression is unguarded across two merges — consider a minimal welcome → … → Start smoke test in this PR.
- Nits: hub sub-surfaces still emit `onboarding.messageTransport.*` / `onboarding.notary.*` / `onboarding.discoveryConfirm.*` ids naming steps that no longer exist (`OnboardingSteps.swift:1002`, `:1129`, `:1246`) — rename now, before PR 3 writes tests against them. `OnboardingFlowTests.swift:126` `testBackKeepsThePriorOutcomeUntilOverwritten` no longer tests back-then-skip preservation; it now duplicates `testRecordingAgainOverwrites`.

Verified separately: `IdentityRepository` is an actor and `bootstrap()` is idempotent, so `identityReady` racing the `WindowGroup` task is safe; the `indicator:` builder's optional branch types fine; no localization resources exist, so the `Button("Skip")` → `Text(verbatim:)` change is bundle-neutral.

rinat-enikeev:
Round 2 addressed in 826c549 (propagated through #260/#261):

- **Recovery state hoisted onto the flow** — new `RecoveryBackupState` (`none`/`revealed`/`verified`), written by the step content and read back by it, so the status card survives Back from Done exactly like `servicesChoice` does. (The rebuilt `makeBackupFlow()` still re-prompts biometrics on "View it again" — left as-is deliberately: re-viewing the phrase is worth a biometric check.)
- **Revealed vs verified**: the Done summary's recovery row now reads from `recoveryBackupState` — "Backed up" only after the verify quiz, "Revealed" after a reveal alone, "Later" otherwise. The consent outcome stays the primary-unlock signal; the claim and the label no longer overreach it.
- **Identity is now enforced, not just documented**: `requiresOutcomeToAdvance` includes `.identity`, and the step content records `.notApplicable` only once `identityReady` confirms a snapshot — a failed bootstrap pins the user on the failure card + retry instead of walking them into an unrevealable recovery step.
- **Services outcome seeded on first show** (`.consented(nil)`): the card says "Selected" from the first frame, so Continue is now recorded as accepting the recommendation, and tapping the preselected card is an honest no-op rather than a hidden state change.
- **Last-consent-wins across hub seats** documented at the `onDone` branch, with the note that the Done summary reads live repository state so the arbitrariness is harmless.
- **Smoke test added to this PR**: `test_onboardingSmoke_defaultPathCompletes_thenNeverAgain` covers present → recommended path → signed mandate → "Remind me later" → Start → relaunch-no-re-present, so the completion-flag contract is guarded across both merges. #261 keeps the full hub walk (and now also asserts the identity gate).
- **Nits**: hub sub-surface ids renamed to `onboarding.services.{directory,messageDelivery,mediaDelivery,groupIntegrity}.*` (#261's page object and walk updated to match); `testBackKeepsThePriorOutcomeUntilOverwritten` now tests the actual back-then-overwrite path (reveal consent → Back from Done → skip overwrites) instead of duplicating `testRecordingAgainOverwrites`.

Green locally: 49 package tests, 1248 app unit tests, 3/3 onboarding UI tests (smoke + full walk + veto).

dev-onym (commented):
REQUEST_CHANGES — the flow's new recovery state is dead in production, and the Done step still tells verified users to go back up their phrase in Settings.

**Blocking**

- `Sources/OnymIOS/OnboardingSteps.swift:1605` — the app's `OnboardingDoneContent` renders the amber "Back up your recovery phrase … from Settings → Back Up Recovery Phrase" nudge **unconditionally**, unchanged from when recovery was a footnote. A user who just revealed *and verified* on the new recovery step lands on Done and is told to go do it. Gate it on `flow.recoveryBackupState` (hide when `.verified`, soften to "finish verifying" when `.revealed`).
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:120` — `recoveryBackupState` has no reader at the app layer (grep: only `servicesChoice` is consumed). Its only consumer is the package's `DoneStepContent`, which `OnboardingStepContentBuilder.content(for:)` overrides for `.done` — so the `revealed` vs `verified` distinction the PR builds and tests is unreachable in the shipped app. Either wire it into `OnboardingDoneContent` (same fix as above) or drop the claim from the doc comment.

**Non-blocking**

- Accessibility-id rename is half-applied inside the same views: catalog prefixes moved to `onboarding.services.*` but the custom-URL fields still pass the old ones — `OnboardingSteps.swift:1010` (`onboarding.messageTransport`), `:1137` (`onboarding.blobTransport`), `:1249` (`onboarding.notary`). Same screen, two naming schemes.
- `Tests/OnymIOSUITests/PageObjects/OnboardingScreen.swift:47-67,106` — `discoveryConfirmButton`, `discoveryFingerprint`, `discoveryPinButton`, `discoveryAdded`, `notaryPublishedRow` still target the removed `onboarding.discoveryConfirm.*` / `onboarding.notary.published.*` ids. Now that their only caller is the `XCTSkip`ped test, they're accessors that silently never match. Fine if PR 3 owns them; a `// stale until PR 3` marker would stop the next reader trusting them.
- `Sources/OnymIOS/OnboardingSteps.swift:1520` — `OnboardingRecoveryContent`'s `onChange` calls `onboarding.recordOutcome` without the `onboarding.step == .recoveryPhrase` guard that `OnboardingIdentityContent` (`:340`) uses for the same "async callback writes to the *current* step" hazard. Hard to trigger through the modal sheet, but the asymmetry invites a future regression.
- `Sources/OnymIOS/OnboardingSteps.swift:266` — a permanently failing bootstrap leaves identity with a disabled Continue, no skip, and no completion path: onboarding becomes a dead end behind `interactiveDismissDisabled`. Fail-closed is right, but "Try again" is the only affordance; worth confirming that's the intended terminal state.
- `OnboardingView.swift:254` — with recovery handled by the `recoveryBackupState` branch, the `.skipped → "Later"` arm is now unreachable (only recovery is skippable). Harmless, but it reads as live logic.

Flow-level model, gating matrix, and the 44 adapted package tests look sound; `advance()`'s generalization from `isMandatory` to `requiresOutcomeToAdvance` keeps the moderation fail-closed rule intact.

rinat-enikeev:
Round 3 addressed in 4b0ee32 (merged through #260/#261):

**Blocking**
- `OnboardingDoneContent` now takes `recoveryBackupState` — the flow property has its app-layer reader — and gates the nudge on it: **hidden** once revealed and verified, **softened** to "Finish verifying your recovery phrase" after a reveal alone, and the full Settings nudge only when the step was deferred. (On the polish branch's redesigned Done, the gated nudge sits above the everything-stays-editable footnote.)

**Non-blocking**
- Custom-URL fields now pass the renamed `onboarding.services.{messageDelivery,mediaDelivery,groupIntegrity}` prefixes — one naming scheme per screen.
- Page-object accessors renamed to the live ids (matching what #261's walk already targets), with a header note that the hub accessors are only exercised by the tests PR's walk while this PR runs the smoke test.
- `OnboardingRecoveryContent.onChange` got the same `step == .recoveryPhrase` guard the identity content uses.
- The package Done label's `.skipped` arm is folded into the default with a comment — it existed for exhaustiveness only, since recovery (the sole skippable step) is special-cased above it.
- **Identity dead end**: confirmed intentional. A permanently failing bootstrap means no keys, and nothing past the step (services consent signing, mandate signing, the reveal) can work without them — so fail-closed with retry as the only affordance is the honest terminal state. If we later add a support/diagnostics path, the failure card is where it would hang.

Green at the top of the stack: 49 package tests, 3/3 onboarding UI tests (smoke + full walk + veto), app builds throughout.

dev-onym (commented):
**COMMENT** — solid restructure with careful state hoisting; a few honesty/coverage gaps worth addressing before the stack lands.

- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:109` — `requiresOutcomeToAdvance(.identity)` is unconditionally true, but the package's built-in content for `identity` is `PlaceholderStepContent`, which never records an outcome. Without an app-supplied `stepContent` the walk dead-ends at identity: Continue permanently disabled, no Skip, only Back. That contradicts the "the built-in keeps the package standalone" claim in the same file and breaks package previews. Either have the placeholder record `.notApplicable` on appear, or gate `identity` on a flow-level flag the host sets.
- `Sources/OnymIOS/OnboardingSteps.swift:437` — `onDone` sets `servicesChoice = .custom` unconditionally. Opening the hub and closing it without touching anything flips the card to "Choose services myself / Selected" while the recommended defaults are still exactly what gets installed — the chip then contradicts reality. Consider only flipping to `.custom` when a sub-surface actually recorded something (which the `componentId` check right below already detects).
- `Tests/OnymIOSUITests/OnboardingUITests.swift:65` — the smoke test's moderation assertions from the old walk are gone: `Continue` disabled before the mandate and `skip("moderation")` absent. Those were the only view-level guard on the fail-closed moderation rule (`primaryDisabled` wiring); the package tests cover `advance()`/`isSkippable` but not the button binding. Two PRs with that unguarded is the riskiest part of the split — re-adding those two asserts to the smoke test is cheap.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:65` — `continueFrom("services")` is tapped with no preceding `title("services").waitForExistence`, unlike every other step in the test. Flaky (race against the identity→services transition) and it means the new services screen's rendering is never asserted at all.
- `Sources/OnymIOS/OnboardingSteps.swift:778,865` — discovery's unavailable/empty cards still read "Continue — the app uses its built-in defaults" and "or continue with the built-in defaults", but that surface is now a pushed hub screen with no Continue. This is the same stale-affordance copy fixed for nostr/blossom/notary in this diff (`OnboardingSteps.swift:1036,1163,1261`) — discovery was missed.
- `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift:118,122` — `servicesChoice` and `recoveryBackupState` are `public var` while every other outcome goes through `recordOutcome`. The "never downgrades: `verified` stays `verified`" invariant on line 122 is enforced only at the single call site in `OnboardingRecoveryContent`; nothing stops a future caller writing `.none` back. `private(set)` plus a `noteRecoveryBackup(_:)` that takes the max would make the doc comment true.

Notes, not blockers: `identityReady`'s retry path is sound — `IdentityRepository.bootstrap()` is genuinely idempotent and retryable (`IdentityRepository.swift:69`). I could not run `swift test` to confirm the "44 green" claim (no Swift toolchain in this environment).

PR #260, opened 2026-08-15
Onboarding redesign 2/3: plain-language polish, role chips, summary

rinat-enikeev:
Second PR of the onboarding-redesign stack (base: #259). Visual/copy pass matching the [design spec](https://claude.ai/design/p/f3d5bdf0-84de-429c-9289-e7311f4b6c6b?file=Onym+Onboarding.dc.html):

- Every seat screen opens with a plain-language explainer of the *job* (delivery, media, group integrity, directory) instead of the protocol noun; sections are now **In use / Suggested by your directories / Add your own**.
- Relay rows show **Primary/Backup** role chips with the "top relay is tried first" footnote; the media service shows **Active**.
- The Done step gets the design's success hero and a "Your setup" card: per-seat icon, chosen service, checkable detail (URL / fingerprint / manifest hash) and service counts. Summary titles now read *Message delivery / Media delivery / Group integrity / Directory / Reports & safety*.
- Moderation step gains the "an authority never sees your messages" footnote.

Deferred (noted for follow-ups): the welcome screen's "I have a recovery phrase" restore entry (needs `AddIdentitySheet`/`IdentitiesFlow` plumbing made public), and the deep terms re-skin (can/can't/appeals layout) inside `ModerationConsentContent` — the manifest data model doesn't carry structured clause lists yet.


dev-onym (commented):
**REQUEST_CHANGES** — the copy pass drops Russian localization for every string it touches and re-introduces an English-only plural branch the codebase explicitly warns against.

- `Resources/Localizable.xcstrings` is hand-maintained with 100% `ru` coverage (338/338, `extractionState: manual`). Every string this PR *replaces* has a `ru` entry — `ADD CUSTOM URL`, `FROM CATALOG`, `DEFAULT`, `ACTIVE`, `Message Transport`, `File Storage` — and **none** of the replacements do (`IN USE`, `ADD YOUR OWN`, `SUGGESTED BY YOUR DIRECTORIES`, `YOUR SETUP`, `PRIMARY`, `BACKUP`, `Message delivery`, `Media delivery`, `Group integrity`, `Reports & safety`, `Directory`, `Agreed`, plus all four new explainer paragraphs and both footnotes). Net effect under `ru`: previously-translated chrome regresses to English. Catalog entries need adding in this PR.
- `OnymIOSApp.swift:1004-1010` — `serviceCount` is exactly the `== 1` branch that `DiscoverySettingsView.swift:229-236` documents as wrong ("a real per-locale plural rule, not an English-only `== 1` branch"); Russian needs one/few/many/other. It also splits into two keys (`"1 service"` / `"%lld services"`), so the plural key's `one` variant can never be reached. Use a single `String(localized: "\(count) services")` with catalog plural variations, per `DiscoveryEntriesLabel`. Same at `OnymIOSApp.swift:1048-1050` for sources.
- `OnboardingSteps.swift:1198` — media rows past the first still show the old `DEFAULT` chip, so the "role chip" pass is half-applied on the seat the PR body says was converted; notary rows (`:1320`) pass `badge: nil` and get no chip at all. Pick one vocabulary across seats.
- `OnboardingSteps.swift:1012`, `:1138`, `:1255` — the `IN USE` label renders above `configured`, which is the empty-state card ("No relay configured yet…") when nothing is configured. Label should be inside the non-empty branch.
- `OnboardingSteps.swift:1653` — the doc comment two lines up still claims the summary is "checkable", but value and detail are now merged into one proportional-font line with `truncationMode(.middle)`; a fingerprint or manifest hash is elided mid-string and no longer monospaced, which is where checkability actually lived. Either keep the detail on its own monospaced line or drop the "checkable" claim.
- `OnboardingSteps.swift:1456` — "…the terms you're about to read" is pinned below `ModerationConsentContent` for every `Step`, including `.signing` and `.done`, where the terms have already been read and signed. Gate on `flow.state.step`.
- `OnymIOSApp.swift:1033-1038` — when no relayer endpoints exist the row reads "Group integrity / Published defaults … Default"; the trailing chip repeats the value.

Nit: `OnboardingSteps.swift:822` puts the directory explainer only in `idle(_:)`, and that seat never got the `IN USE` label, so it's the one screen not matching the "In use / Suggested / Add your own" structure the PR describes.

rinat-enikeev:
Addressed in 79c0696 (on #259, its own strings) + a912e73 (this PR):

- **Russian coverage restored.** The catalog gets ru entries for every string the stack introduces — 67 for #259's flow restructure (committed on that branch so each PR stays localization-complete) and 17 here (section labels, role chips, the four explainers, footnotes, "Agreed"). Terminology follows the existing glossary (личность / фраза восстановления / релей / релеер / нотариус / каталог сервисов), «вы» form throughout. Verified live: the onboarding cover renders fully in Russian under `-AppleLanguages (ru)`.
- **Real plurals.** `serviceCount` and the sources trailing now go through single interpolated keys with catalog plural variations (`%lld services` / `%lld sources`, one/few/many/other for ru) — the `== 1` branch and the dead `one` variant are gone, per the `DiscoveryEntriesLabel` pattern. #261 adds `OnboardingCountLabelsTests` pinning the rendered forms against the app bundle, same as `DiscoveryEntriesLabelTests`.
- **One chip vocabulary.** Relays PRIMARY/BACKUP; media ACTIVE/BACKUP (upload target + fallbacks); notaries PRIMARY/BACKUP (`endpoints.first` is the `RelayerConfiguration` fallback, so first-is-primary is honest). The stray DEFAULT chip and the chip-less notary rows are gone.
- **IN USE** renders only above a non-empty list on all four seats — including the directory screen, which now carries the label and matches the In use / Suggested / Add your own structure.
- **Checkable summary**: the detail (URL / fingerprint / manifest hash) is back on its own monospaced line under the value, matching the doc comment's claim.
- **Moderation footnote** ("…the terms you're about to read") is gated to `loadingDirectory`/`pickingAuthority`/`reviewingManifest` and disappears at `signing`/`done`.
- **Group-integrity summary row** no longer pairs "Published defaults" with a repeating "Default" trailing (trailing is nil when empty).

Green: 1250 app unit tests (incl. the two new plural pins), 49 package tests, 3/3 onboarding UI tests at the top of the stack.

dev-onym (commented):
REQUEST_CHANGES — the new role vocabulary states failover behavior the transports don't actually implement, and the renamed catalog header renders on top of the old one.

- `Sources/OnymIOS/OnboardingSteps.swift:1072` + `:1060` — "The top relay is tried first; the rest are fallbacks" and PRIMARY/BACKUP chips are false for Nostr: `NostrMessageTransport.publish` fans out to **every** connected relay concurrently (`NostrMessageTransport.swift:117-135`), and boot connects to all of them (`OnymIOSApp.swift:1443`). A user reading "BACKUP" will believe those relays only see traffic when the top one is down; they see every event. This is a privacy claim, so it needs to be correct — either say relays are used together, or reserve Primary/Backup for seats that really do fail over.
- `Sources/OnymIOS/OnboardingSteps.swift:1331` — same chip on the notary seat, but the auto-populated config uses `strategy: .random` (`RelayerRepository.swift:153`) and `.primary` resolves `primaryURL`, not `endpoints.first` (`RelayerEndpoint.swift:170-180`). The inline comment at :1327-1330 is only true in the nil-primary/`.primary` corner case. Chip should read the actual `strategy`/`primaryURL`, or drop it.
- `Sources/OnymIOS/OnboardingSteps.swift:1203` — media "BACKUP": uploads and downloads only ever target `currentEndpoints().first` (`OnymIOSApp.swift:542`, `BlossomCatalogConsent.swift:11`); the rest are only an allowlist for peer-stamped download hosts, not fallbacks for your uploads.
- `Sources/OnymIOS/OnboardingSteps.swift:1082`, `:1144`, `:1266` — `DiscoveryCatalogSection` unconditionally renders its own `SettingsSectionLabel("FROM CATALOG")` (`DiscoveryCatalogSection.swift:48`), so each seat now shows "SUGGESTED BY YOUR DIRECTORIES" directly above "FROM CATALOG". Pre-existing duplication, but the rename makes it visibly inconsistent — parameterize the label (default keeps Settings unchanged).
- `Packages/OnymSettings/.../NostrRelaySettingsView.swift:128`, `BlossomRelaySettingsView.swift:141` — Settings still says "DEFAULT" while onboarding now says PRIMARY/ACTIVE, and the Done footnote sends users there ("Settings → Services"). Worth a follow-up issue if not in this PR's scope.

Verified fine: ru coverage is complete for every new key including real `few`/`many` plural forms; `String(localized: "\(count) services")` maps to the `%lld services` catalog entry; the moderation `switch` is exhaustive over `ModerationConsentFlow.Step`; `onboarding.done.summary` / `backup_nudge` identifiers survive the restructure (`OnboardingScreen.swift:134`); hero circle is `accessibilityHidden`; divider inset 50 matches icon 24 + spacing 12 + padding 14.

rinat-enikeev:
Addressed in 7273381 (+ 2c47809 on #259 for the hub footnote it introduced):

- **Nostr is fan-out, and now says so.** PRIMARY/BACKUP chips and the "top relay is tried first" footnote are gone; the explainer and footnote state the real behavior — every message is sent through all of your relays at once, they're used together, not as fallbacks, and any one that's up is enough to reach you. No row claims reduced traffic visibility anymore.
- **The notary chip is strategy-true.** It reads the actual `RelayerConfiguration`: under `.primary` exactly the endpoint `selectURL` resolves (matching `primaryURL`, or `endpoints.first` when none is pinned — the documented fallback) is marked PRIMARY; under `.random` (the auto-populated default) no row is special, and the footnote says what actually happens: when you create a group, one of these notaries is picked at random for it, and everyone in that group then uses the same one.
- **Media** rows past the first lose their chip; the footnote states that uploads go through the active service only and the rest are hosts you trust for downloading what people send you.
- **Double header fixed at the source**: `DiscoveryCatalogSection` takes an optional `label` (defaulting to "FROM CATALOG" so Settings is unchanged); the hub screens pass nil under their own "SUGGESTED BY YOUR DIRECTORIES".
- **Same honesty applied to a string you didn't flag**: the hub's "Onym uses the fastest one that answers" footnote (from #259) was equally false — replaced on that branch with the per-category truth (fan-out / download allowlist / per-group notary pick).
- **Catalog kept tight**: the three false-claim keys removed, six honest replacements added with ru; coverage stays 100%.
- **Settings vocabulary follow-up** filed as an issue (DEFAULT chips on the Nostr/Blossom Settings screens vs the new onboarding labels).

Green at the top of the stack: build + 3/3 onboarding UI tests.

dev-onym (commented):
**APPROVE** — copy and structure changes are sound; the transport claims match actual behaviour and localization is complete.

Verified:
- `OnboardingSteps.swift:1381-1399` — `isPrimary` exactly mirrors `RelayerConfiguration.selectURL` (`RelayerEndpoint.swift:171-186`): primary only under `.primary`, `primaryURL` match required, `endpoints.first` fallback. Displayed list is `configuration.endpoints` (`:1340`), so the `index == 0` fallback indexes the same array. Correct.
- All 17 new/changed keys (`IN USE`, `PRIMARY`, `Agreed`, `Message delivery`, `Group integrity`, `%lld services`, `%lld sources`, …) resolve in `Localizable.xcstrings` with `en` + `ru`, and the ru plurals carry `one/few/many/other`. No stale key left behind by the summary-title rename.
- `DiscoveryCatalogSection.swift:40` default `label: "FROM CATALOG"` keeps all four Settings call sites source-compatible; all three onboarding sites pass `label: nil`.
- `OnboardingEndpointRow` memberwise arg order (`badge, selected, badgeColor, last`) is respected at every call site; `OnymTokens.green/onAccent/text2` all exist as `Color`.
- Moderation `switch` covers all five `ModerationConsentFlow.Step` cases; `.task`/`.onDisappear`/`.onChange` now hang off the wrapping `VStack`, so flow lifecycle is unchanged.

Findings:
- `OnymIOSApp.swift:1032-1042` — the Group integrity row shows `relayerEndpoints.first?.name` + its URL as the checkable detail, but the auto-populated default strategy is `.random`, where no single endpoint is the notary. The notary screen honestly says "Onym picks one of these at random"; the summary silently contradicts it. Same shape at `:1014-1019` for message delivery, where every relay carries every message. Consider a value like "Random from N" (or reusing the seat's own footnote wording) when `strategy == .random`.
- `OnboardingSteps.swift:1076`, `:1231`, `:1374` — the three plural-assuming footnotes render with a single configured endpoint too ("Every relay in this list…", "Any others in the list…", "picks one of these at random"). Gating on `endpoints.count > 1` would avoid the one-item awkwardness; the seed config is usually one relay.
- `OnboardingSteps.swift:1718` — the summary seat icon isn't `.accessibilityHidden(true)`, unlike the success hero at `:1699-1710`. VoiceOver will read "antenna radiowaves left and right" before the row title on all five rows.
- `OnboardingSteps.swift:1221`, `:1362` — green-on-green-14% at 10pt bold is roughly 3:1; the old `text2`-on-`surface3` chip was well above 4.5:1. Consistent with `AboutView.swift:224`, so system-wide, but it is a contrast regression on this screen.
- `OnymIOSApp.swift:1053` — Directory row's trailing reads "N sources" while the new section header says "Suggested by your **directories**". Pick one noun.
- PR body is stale: it advertises "Primary/Backup role chips with the 'top relay is tried first' footnote" for relays, but the merged code deliberately does the opposite (`badge: nil`, fan-out footnote, `OnboardingSteps.swift:1058-1064`). Worth fixing before merge so the description doesn't outlive the honesty fix in `7273381`.
- `OnboardingSteps.swift:112` — `badgeColor` defaults to `text2`, but every badge-bearing row now passes `.green`; the default is dead. Non-blocking.
- Not verifiable here: no Swift toolchain in this environment, so nothing was compiled and no snapshot/UI tests were run.

PR #261, opened 2026-08-15
Onboarding redesign 3/3: UI walk through the services hub and gated recovery reveal

rinat-enikeev:
Final PR of the onboarding-redesign stack (base: #260). Restores full UI coverage for the redesigned flow:

- `test_onboardingWalk_customServicesPath_thenNeverAgain`: welcome → identity → services, taking the **Choose services myself** hub (Directory TOFU pin first, then Message delivery with a real module-consent walk, Media delivery, Group integrity from the published list) → Reports & safety (mandatory pick → review → sign; Continue disabled beforehand) → recovery phrase → done, then the persistence relaunch.
- The recovery step's review-round semantics are exercised end-to-end: primary ("I've written it down") asserted **disabled before the reveal**, the real backup sheet runs (mock biometric → intro → tap-to-reveal), and the primary unlocks after.
- Hub rows carry accessibility identifiers; the page object grows services/hub/recovery helpers.
- Package-level behavior tests for the new flow rules (outcome-gate matrix, `servicesChoice` persistence, indicator coverage) landed with the review-fix commit on #259.

All green locally: 47 OnymOnboarding package tests, 1248 app unit tests, 2/2 onboarding UI tests on iPhone 17e.

### DIFF
--- a/Sources/OnymIOS/OnboardingSteps.swift
+++ b/Sources/OnymIOS/OnboardingSteps.swift
@@ -471,6 +471,7 @@ struct OnboardingServicesHub: View {
                     OnboardingSectionLabel(text: "NEEDED TO SEND MESSAGES")
                     VStack(spacing: 0) {
                         hubRow(
+                            id: "messageDelivery",
                             symbol: "antenna.radiowaves.left.and.right",
                             color: OnymAccent.purple.color,
                             title: "Message delivery",
@@ -484,6 +485,7 @@ struct OnboardingServicesHub: View {
                         }
                         divider
                         hubRow(
+                            id: "mediaDelivery",
                             symbol: "photo.on.rectangle.angled",
                             color: OnymAccent.pink.color,
                             title: "Media delivery",
@@ -497,6 +499,7 @@ struct OnboardingServicesHub: View {
                         }
                         divider
                         hubRow(
+                            id: "directory",
                             symbol: "magnifyingglass",
                             color: OnymAccent.blue.color,
                             title: "Directory",
@@ -514,6 +517,7 @@ struct OnboardingServicesHub: View {
                     OnboardingSectionLabel(text: "OPTIONAL")
                     VStack(spacing: 0) {
                         hubRow(
+                            id: "groupIntegrity",
                             symbol: "checkmark.seal",
                             color: OnymAccent.green.color,
                             title: "Group integrity",
@@ -566,6 +570,7 @@ struct OnboardingServicesHub: View {
     }
 
     private func hubRow(
+        id: String,
         symbol: String,
         color: Color,
         title: LocalizedStringKey,
@@ -600,6 +605,7 @@ struct OnboardingServicesHub: View {
             .contentShape(Rectangle())
         }
         .buttonStyle(.plain)
+        .accessibilityIdentifier("onboarding.services.hub.\(id)")
     }
 }
 
--- a/Tests/OnymIOSTests/OnboardingCountLabelsTests.swift
+++ b/Tests/OnymIOSTests/OnboardingCountLabelsTests.swift
@@ -1,0 +1,46 @@
+import XCTest
+
+/// Pins the rendered forms of the onboarding Done summary's count
+/// trailings ("%lld services" / "%lld sources"). Same contract as
+/// `DiscoveryEntriesLabelTests`: pluralization must come from the
+/// catalog's per-locale plural variations, not an English-only `== 1`
+/// branch — these hosted tests resolve the keys against the real app
+/// bundle (`TEST_HOST` = OnymIOS.app).
+final class OnboardingCountLabelsTests: XCTestCase {
+
+    private func services(_ count: Int) -> String {
+        String(localized: "\(count) services")
+    }
+
+    private func sources(_ count: Int) -> String {
+        String(localized: "\(count) sources")
+    }
+
+    /// Whatever the run locale, the raw key must never leak and the
+    /// count must be present.
+    func test_labels_neverLeakRawKeys() {
+        for count in [0, 1, 2, 5, 21] {
+            for label in [services(count), sources(count)] {
+                XCTAssertFalse(label.contains("%lld"),
+                               "raw format key leaked: \(label)")
+                XCTAssertTrue(label.contains("\(count)"),
+                              "count missing from label: \(label)")
+            }
+        }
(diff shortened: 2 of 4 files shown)


ux-onym (commented):
COMMENT — production change is a safe identifier addition; the walk is well-built, but two sheet-dismissal handoffs can flake and one assertion can silently pass with a sheet still up.

- `Tests/OnymIOSUITests/OnboardingUITests.swift:118` → `PageObjects/OnboardingScreen.swift:73`: `consentDismiss.tap()` is followed immediately by `hubBack()`, which taps `app.navigationBars.buttons.element(boundBy: 0)`. The module-consent sheet is its own `NavigationStack` with a toolbar (`Packages/OnymSettings/.../ModuleConsentView.swift:25,44`), so during the dismiss animation two nav bars are in the hierarchy and index 0 is ordering-dependent — the tap can land on the sheet's Cancel or a disappearing element. Wait for the sheet to go (`consentAccept`/`consentDismiss` non-existence) before `hubBack()`, and/or scope the back tap to the detail's own bar (`app.navigationBars["Message delivery"]`).
- `Tests/OnymIOSUITests/OnboardingUITests.swift:207-209`: after `app.swipeDown`, `recoveryPrimary.isEnabled` is queried without confirming the backup sheet closed. The step's primary sits behind the sheet and still reports `isEnabled == true`, so a failed swipe-to-dismiss passes this assertion and then fails confusingly at `continueFrom("recoveryPhrase")` (tap on a covered button). Add `XCTAssertTrue(app.staticTexts["reveal.word.1"].waitForNonExistence(timeout: 5))` first. The swipe itself is also gesture-dependent — `RevealScreen` is a `ScrollView` (`Packages/OnymRecovery/.../RecoveryPhraseBackupView.swift`), so dismissal relies on the scroll being at top; it works, but the explicit non-existence check is what makes the failure legible.
- `Tests/OnymIOSUITests/PageObjects/OnboardingScreen.swift:14-16`: doc comment is now stale — "exercised by the redesigned walk in the tests PR of this stack — on this PR only the smoke test runs". This *is* that PR; drop the forward reference.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:215`: the custom path's payoff (courier module consented, testnet notary added, TOFU-pinned directory) is only checked as `doneSummary.exists`. Asserting one summary row value — e.g. the `uitest-testnet-relayer.example` notary — would make the walk prove the picks actually landed rather than that the step rendered.
- `Sources/OnymIOS/OnboardingSteps.swift:648`: `onboarding.services.hub.use_recommended` (which flips `servicesChoice` back mid-hub) has an identifier but no coverage in either UI test; the walk always exits via Done. Optional, but it's the one hub affordance with a state side effect left untested.
- Nits: `Tests/OnymIOSUITests/OnboardingUITests.swift:366` adds a stray blank line before the closing brace; `PageObjects/OnboardingScreen.swift:80-84` and `:136-140` use `descendants(matching: .any)` full-hierarchy scans where `app.otherElements[...]` would match the same container far more cheaply.

The core assertions are sound: the disabled-before checks (`moderation` Continue at :162, recovery primary at :199) are safe against races since both gates only ever open, the Directory-first ordering is required for the other seats' catalogs, and the no-`--reset-keychain` relaunch genuinely proves persistence.

dev-onym (commented):
APPROVE — the walk matches the real surfaces: every identifier the page object queries exists in `OnboardingSteps.swift`, and the fixture constants (`ea4a 6c63 e29c 520a`, `public-all-seats|onym:component:onym-courier`, `uitest-testnet-relayer.example`, `blossom.onym.app`, `uitest-authority`) all check out against `UITestDiscoveryFakes.swift` / `UITestFakes.swift` / `BlossomClient.swift:72`. The `%lld services` / `%lld sources` keys do carry en/ru plural variations in `Resources/Localizable.xcstrings`, and `OnymIOSTests` is host-backed (`project.yml:208`), so the catalog test resolves for real.

- `Tests/OnymIOSUITests/PageObjects/OnboardingScreen.swift:73` — `hubBack()` taps `navigationBars.buttons.element(boundBy: 0)`. Index-based back is silently wrong the moment a hub detail gains a leading toolbar item, and the failure surfaces far downstream (a missed pop looks like "row never appeared"). The pushed details set `.navigationTitle(Text(title))` off a root titled "Your services" (`OnboardingSteps.swift:675,745`), so `app.navigationBars.buttons["Your services"]` is available and unambiguous. Matches an existing idiom (`ChatFlowScreens.swift:228`), so not a blocker.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:207` — `app.swipeDown(velocity: .fast)` dismisses the backup sheet, but the reveal screen is a `ScrollView` (`RecoveryPhraseBackupView.swift:234`) and the gesture originates at the app's center, inside that scroll view. At the top of content it should hand off to the sheet's dismiss, but on shorter screens / a partially scrolled state it just scrolls, and the next assertion ("primary should be enabled after the reveal") reports a misleading cause. The sheet exposes no close identifier, so at minimum assert the sheet is gone (`reveal.tap_button` / `reveal.word.1` non-existence) before asserting the primary.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:215` — the custom walk configures a courier module, a Blossom pick and a notary, then only asserts `doneSummary` *exists*. Nothing checks the summary counts reflect those picks, and the relaunch checks only the completion flag — so a regression that drops the custom configuration on hub Done still passes green. One count/label assertion on the Done card (the very strings `OnboardingCountLabelsTests` pins) would close the loop the PR title promises.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:78,150` — the branch taps `servicesCustomCard`, but `servicesChoice` only flips to `.custom` in `onDone` (`OnboardingSteps.swift:443`). Asserting the "Selected" chip moved to the custom card after `hubDone.tap()` would guard that transition; right now no UI test covers it.
- `Tests/OnymIOSTests/OnboardingCountLabelsTests.swift:38` — the English plural assertions are the actual regression guard (the `one` variant reachability), and `XCTSkip` makes them evaporate silently if the host resolves non-English. The UI tests force locale via `-AppleLanguages`; a unit test can't, so consider `XCTFail` instead of skip, or drop the guard and pin against a fixed-locale `Locale`.
- `Tests/OnymIOSUITests/OnboardingUITests.swift:366` — stray blank line before the closing brace.

PR #264, opened 2026-08-15
Onboarding: center Back under the primary button

rinat-enikeev:
Follow-up to the merged onboarding stack (#259–#261). The scaffold's bottom row (`HStack { Back, Spacer, Skip }`) left Back stranded in the corner — it read as page chrome, not as the screen's alternative action.

Secondary actions now stack centered under the primary: the skip affordance (or its probe progress) first, Back below, both as quiet full-width-centered text buttons per the design. Accessibility identifiers are unchanged; verified visually in the simulator and 3/3 onboarding UI tests pass.

| Before | After |
|---|---|
| Back pinned bottom-left | Back centered under Continue |

### DIFF
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingStepScaffold.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingStepScaffold.swift
@@ -112,22 +112,27 @@ public struct OnboardingStepScaffold<Content: View, Indicator: View>: View {
                 .disabled(primaryDisabled)
                 .accessibilityIdentifier("onboarding.\(step.rawValue).primary")
 
-                HStack {
-                    if let backAction {
-                        Button("Back", action: backAction)
-                            .accessibilityIdentifier("onboarding.\(step.rawValue).back")
-                    }
-                    Spacer()
-                    if let skipAction {
-                        Button(action: skipAction) { Text(verbatim: skipTitle) }
-                            .accessibilityIdentifier("onboarding.\(step.rawValue).skip")
-                    } else if showsSkipProgress {
-                        ProgressView()
-                            .controlSize(.small)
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
**COMMENT** — clean, matches the consent-view precedent; two layout nits worth a look before merge.

- `OnboardingStepScaffold.swift:122,127,133` — `.frame(maxWidth: .infinity)` is applied *outside* the `Button`, unlike the primary which applies it to the label (`:108`). A `VStack` already centers its children, so these frames don't change the visual result; what they don't do is widen the tap target — with the frame outside the button, hit testing still follows the label unless you add `.contentShape(Rectangle())` or move the frame inside the label closure. If a full-width tap target was the intent, it isn't achieved; if it wasn't, the modifier is dead weight. The reference implementation (`ModerationConsentView.swift:295`) just uses a bare `Button("Back")` in the `VStack`.
- `OnboardingStepScaffold.swift:119-135` — Back's vertical position now depends on the skip slot. On the moderation step `showsSkipProgress` starts true and `skipAction` is nil (`OnboardingView.swift:67,71,72`), so when the directory probe resolves to *non-skippable* the whole slot vanishes and Back jumps up ~30pt — under the user's finger, mid-read. The old `HStack` kept Back pinned regardless of what the trailing slot did. Reserving the slot height (e.g. keep the branch and render `Color.clear.frame(height:)`, or give the row a `minHeight`) would preserve that stability.
- `OnboardingStepScaffold.swift:5` — doc comment still reads "text Back/Skip"; order is now Skip-then-Back. Trivial, but the header is the file's spec.
- Identifiers (`.primary` / `.skip` / `.skip_pending` / `.back`) are unchanged and `OnboardingScreen.swift:31` queries by identifier only, so the UI-test claim checks out — nothing in the page object depends on position.
- No behavior change to `skipTitle`, `primaryDisabled`, or the indicator slot; the `else if showsSkipProgress` precedence is preserved exactly as before.

PR #265, opened 2026-08-15
Onboarding: phrase lifetime, honest hub footnote, cancellation guard

rinat-enikeev:
Ports three findings from onym-android #215's review that apply equally to the merged iOS flow:

1. **Recovery phrase lifetime / re-auth** — `OnboardingRecoveryContent` held one `RecoveryPhraseBackupFlow` for the step's whole life, so after a reveal the sheet could be dismissed and re-opened onto the revealed words with no biometric re-auth, and the plaintext phrase stayed resident. The flow is now created per presentation and released on dismiss (`flow = nil`): every entry starts at Intro behind the biometric gate; the flow-held `recoveryBackupState` keeps the status card truthful across navigation.
2. **Honest hub footnote** — "remove it from the lists above if you've changed your mind" was false (onboarding rows have no remove affordance). Now: "…you can remove it later in Settings → Services." (en + ru catalog entry swapped.)
3. **Cancellation guard** — `identityReady`'s `try?` swallows `CancellationError`, so a navigation-cancelled probe could paint the failure card. The identity task now bails on cancellation instead of reporting a false failure.

3/3 onboarding UI tests green.

### DIFF
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
@@ -219,6 +219,20 @@ public final class OnboardingFlow {
         }
     }
 
+    /// Whether the recorded outcome actually satisfies an
+    /// outcome-gated step. `.skipped` deliberately does NOT: skipping
+    /// records a decision, but revisiting the step must re-gate its
+    /// primary — an earlier "Remind me later" must never leave
+    /// "I've written it down" enabled over a phrase that was never
+    /// revealed. The honest exits from a re-gated step are the same
+    /// as the first visit: satisfy the gate, or skip again.
+    public func outcomeSatisfiesGate(_ step: OnboardingStep) -> Bool {
+        switch outcomes[step] {
+        case nil, .skipped?: return false
+        default: return true
+        }
+    }
+
     // MARK: - Lifecycle
 
     /// Kick the async moderation-directory probe. Idempotent.
@@ -250,7 +264,7 @@ public final class OnboardingFlow {
     /// (the view also disables it; this guard is the second layer).
     public func advance() {
         guard let next = neighbor(offset: 1) else { return }
-        guard !requiresOutcomeToAdvance(step) || outcomes[step] != nil else { return }
+        guard !requiresOutcomeToAdvance(step) || outcomeSatisfiesGate(step) else { return }
         if outcomes[step] == nil {
             outcomes[step] = step == .moderation && moderationDirectoryHasEntries == false
                 ? .unavailable
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
@@ -61,7 +61,7 @@ public struct OnboardingView: View {
             // the step content records one (moderation's consent, the
             // recovery reveal); `flow.advance()` carries the same
             // guard as the second layer.
-            primaryDisabled: flow.requiresOutcomeToAdvance(step) && flow.outcomes[step] == nil,
+            primaryDisabled: flow.requiresOutcomeToAdvance(step) && !flow.outcomeSatisfiesGate(step),
             primaryAction: { primaryTapped() },
             skipTitle: Self.skipTitle(for: step),
             skipAction: flow.isSkippable(step) ? { flow.skip() } : nil,
--- a/Packages/OnymOnboarding/Tests/OnymOnboardingTests/OnboardingFlowTests.swift
+++ b/Packages/OnymOnboarding/Tests/OnymOnboardingTests/OnboardingFlowTests.swift
@@ -443,6 +443,29 @@ final class OnboardingFlowTests: XCTestCase {
         XCTAssertEqual(flow.step, .done)
     }
 
+    /// A `.skipped` outcome must NOT satisfy the gate: after
+    /// "Remind me later" → Done → Back, the recovery step's primary
+    /// ("I've written it down") re-locks — the recorded skip survives
+    /// (back-preservation), but it must never enable a claim about a
+    /// phrase that was never revealed. Regression for the gate hole
+    /// dev-onym found on the Android port.
+    func testSkipThenBackReGatesTheRecoveryPrimary() async {
+        let flow = await makeResolvedFlow()
+        walk(flow, to: .recoveryPhrase)
+        flow.skip()
+        XCTAssertEqual(flow.step, .done)
+        flow.back()
+        XCTAssertEqual(flow.step, .recoveryPhrase)
+        XCTAssertEqual(flow.outcomes[.recoveryPhrase], .skipped,
+                       "the skip decision survives Back")
+        XCTAssertFalse(flow.outcomeSatisfiesGate(.recoveryPhrase),
+                       "a skip must not satisfy the reveal gate")
+        flow.advance() // refused — the primary is re-locked
+        XCTAssertEqual(flow.step, .recoveryPhrase)
+        flow.skip() // skipping again remains the honest exit
+        XCTAssertEqual(flow.step, .done)
+    }
+
     /// The outcome-gate matrix: identity and recovery always;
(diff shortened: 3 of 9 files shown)


rinat-enikeev:
Added 3rd ported finding from onym-android #215 round 2: a `.skipped` outcome no longer satisfies the outcome gate — after "Remind me later" → Done → Back, the recovery primary re-locks instead of staying enabled over a never-revealed phrase (new `outcomeSatisfiesGate`, used by both `advance()` and the button state; regression test included).

ux-onym (commented):
**REQUEST_CHANGES** — the re-auth gating is right, but `flow = nil` does not actually release the flow, so the "phrase no longer stays resident" claim doesn't hold and each reveal now leaks a flow.

- `Sources/OnymIOS/OnboardingSteps.swift:1646` — `flow = nil` drops the only *view* reference, but `RecoveryPhraseBackupFlow.start()` (`Packages/OnymRecovery/.../RecoveryPhraseBackupFlow.swift:79-84`) spawns `snapshotTask` with `Task { [weak self] in guard let self else { return } … for await snapshot in self.repository.snapshots }`. After `guard let self`, `self` is held **strongly** for the task body, and `snapshots` (`IdentityRepository.swift:276`) is an `AsyncStream` that never finishes — so the flow retains itself forever. The plaintext phrase in `.reveal(phrase:…)` stays resident exactly as before, and because a fresh flow is now built on every "View it again" tap, each reveal adds another immortal flow plus another snapshot subscriber. Pre-existing hazard, but this PR turns one leak into N.
- `Packages/OnymRecovery/Sources/OnymRecovery/RecoveryPhraseBackupView.swift:42` — this is the one flow view in the repo that calls `flow.start()` without a paired `.onDisappear { flow.stop() }`; every sibling has it (`OnboardingSteps.swift:777,1046,1201,1341,1547`, `ModerationConsentView.swift:41`, `BlossomRelaySettingsView.swift:80`). Adding it there fixes the Settings presentation too; the minimal in-PR alternative is `onDismiss: { flow?.stop(); flow = nil }`. Consider also having `stop()` reset `step = .intro` so the phrase is dropped even if something else holds a reference.
- `Sources/OnymIOS/OnboardingSteps.swift:1591-1652` — the rest of the per-presentation rework is sound: both state writes happen in one action so the sheet never renders with `flow == nil`, `onDismiss` covers interactive dismissal, and the optional-pattern `onChange` cases are correctly guarded (`nil` old/new can't match `.done?`/`.intro?`). `recoveryBackupState` on `onboarding` keeps the status card honest across the flow's shorter life.
- `Sources/OnymIOS/OnboardingSteps.swift:333-337` — cancellation guard is correct and correctly placed after the `await`; leaving `phase == .checking` is fine since `.task(id: attempt)` re-enters at `.checking`.
- `Resources/Localizable.xcstrings:7117` — key swap is clean: old key fully removed, no stale references, en+ru both present (the catalog has only those two locales). Russian «в Настройки → Сервисы» is grammatically accusative-where-prepositional-is-expected, but it matches the existing string at :6887, so consistent.
- `Packages/OnymOnboarding/.../OnboardingFlow.swift:227-233` — `outcomeSatisfiesGate` is correct and can't deadlock: `recoveryPhrase` is the only skippable step (`isSkippable`), so a re-gated step always retains "Remind me later" as an exit, and `.unavailable` still satisfies the moderation gate. Test at `OnboardingFlowTests.swift:452` covers the skip→back→re-lock path.
- No test covers the flow-lifetime change (view-level); the UI-test claim in the body is unverified here.

rinat-enikeev:
Addressed in 656bf68 — you were right that `flow = nil` was theater:

- **`stop()` now tears down AND scrubs**: cancels the snapshot/verify tasks (breaking the snapshot task's strong self-retain over the never-finishing stream — the actual deallocation blocker you identified), drops the cached identity, and rewinds to `.intro` so the plaintext phrase is gone even if something else still holds a reference. One deliberate exception, documented in-line: the clipboard clear task is NOT cancelled — it captures only the pasteboard and the phrase (no self, no retain), and cancelling it would skip the early in-app clear; it self-completes, with the OS-enforced expiry as the real guarantee either way.
- **`RecoveryPhraseBackupView` gained the missing `onDisappear { flow.stop() }`** pairing — the fix lands for the Settings presentation too, not just onboarding. The onboarding sheet's `onDismiss` also stops before releasing (belt and suspenders across both dismissal paths).
- **The missing lifetime coverage exists now**: `test_stop_scrubsRevealedPhraseBackToIntro` (revealed step → stop → intro; re-arm goes back through auth) and `test_stop_breaksSnapshotRetainCycle_flowDeallocates` (weak-ref proof the flow deallocates after stop — this test fails without the cancel, so the N-leaks regression can't come back silently).
- Piggybacked one small port from the Android round-5 review since it touches the same PR's honesty theme: the bare `ACTIVE` key's ru is feminine (АКТИВНАЯ — correct for the identity card it also serves); the two masculine sites (onboarding media row, Blossom Settings) now use a keyed `ACTIVE_SERVICE_CHIP` entry (en "ACTIVE", ru «АКТИВНЫЙ»).

Green: 1252 app unit tests (incl. the two new lifetime tests, 21/21 in the flow class), 3/3 onboarding UI tests.

ux-onym (commented):
**COMMENT** — the three ports are correct and the scrub/retain-cycle fix is the right shape; a few small gaps worth a follow-up, none blocking.

- `Packages/OnymRecovery/.../RecoveryPhraseBackupFlow.swift:119` — `tappedContinueFromIntro`'s `Task { await self.authenticate() }` is untracked and *not* cancelled by `stop()`, so it strongly retains the flow past dismissal (until Face ID resolves) and then writes `step` on a scrubbed flow. Outcome happens to be benign — `currentIdentity` is nil by then, so it lands on `.authFailed`, not a re-populated `.reveal` — but the "stop() lets the flow deallocate" guarantee has this one hole. Storing it as `authTask` and cancelling alongside `snapshotTask` would close it.
- `Packages/OnymRecovery/.../RecoveryPhraseBackupFlow.swift:95` — scrub verified: `clipboardClearTask` genuinely captures only `[pasteboard, clipboardClearDelay]` + `phrase` (line 145), so leaving it uncancelled is correct and doesn't retain self. `repository.snapshots` is a real `AsyncStream` (IdentityRepository.swift:276), so cancellation does terminate the drain.
- `Tests/OnymIOSTests/RecoveryPhraseBackupFlowTests.swift:182` — a single `await Task.yield()` before the `XCTAssertNil` is thin: the cancelled snapshot task needs its stream-cancellation handler to resume *and* the body to unwind on the main actor. Consider polling (`for _ in 0..<20 { if leaked == nil { break }; await Task.yield() }`) to avoid a CI-only flake.
- `Sources/OnymIOS/OnboardingSteps.swift:1656` — `RecoveryPhraseBackupView` re-captures the flow into its own `@State` at init, so the per-presentation design depends on SwiftUI destroying the sheet content on dismiss. It does today; `.id(ObjectIdentifier(flow))` on the content would make that identity explicit rather than incidental.
- `Packages/OnymOnboarding/.../OnboardingFlow.swift:229` — `.skipped` re-gating is right, but it also erases a *real* reveal: reveal → advance → back → "Remind me later" → back again leaves the card saying "Revealed — write it down" while the primary is disabled. Gating on `recoveryBackupState != .none` for `.recoveryPhrase` would keep the card and the button telling the same story. Not a dead end (skip still exits), so fine to defer.
- `Sources/OnymIOS/OnboardingSteps.swift:670` — the new footnote is honest about *when* removal is possible, but Settings has no "Services" group (rows are "Nostr Relays", "Blossom Relays", "Relayer", … — no such label anywhere in `OnymSettings`). It matches the existing Done-step string at line 1827, so this is pre-existing debt, not a regression; worth one issue to either name the section or change both strings.
- `Sources/OnymIOS/OnboardingSteps.swift:337` and the ACTIVE chip both check out: the cancel guard leaves `phase == .checking` only for a task that `.task(id:)` will re-run on reappear, and `ACTIVE_SERVICE_CHIP` reaches `Text`/`LocalizedStringKey` in both call sites (`badge` is `LocalizedStringKey?`, line 110), with the feminine `АКТИВНАЯ` correctly left to `IdentityCarouselCard.swift:186`.

PR #266, opened 2026-08-15
Onboarding: TOFU-pin the seeded discovery source on recommended-setup accept

rinat-enikeev:
## The gap

The merged onboarding redesign (#259–#261) never pins the seeded default discovery source on the "Recommended setup" path. The seed ships unpinned **by design** (pinning is the user's own TOFU act), and only the manual hub walk ("Choose services myself" → Directory → Verify & Confirm) ever pinned it. `DiscoveryRepository` hard-skips unpinned sources on refresh, so for every user who took the recommended path:

- the catalog aggregate stayed empty — "SUGGESTED BY YOUR DIRECTORIES" never populated in the hub or Settings;
- the recommended card and the Done summary still presented "Onym Discovery" as part of the active setup, with a source count implying it worked.

Same gap dev-onym flagged on the Android port (onym-android #215, round 4); this is the iOS mirror of that fix.

## Pin on accept

**Accepting the recommended setup IS the explicit TOFU confirm.** The recommended card names "Onym Discovery" as part of what's being accepted, so leaving the services step forward with the recommended choice now programmatically runs the same fetch → verify → pin path the manual "Verify & Confirm" drives, then a targeted `refreshSource` so catalogs populate.

Trust rationale (documented at the implementation site, `SeededDiscoveryConfirmation`):
- only the SEEDED source's manifest URL is ever fetched — never a user-typed one;
- the manifest is signature-verified before anything is pinned, and the pin is refused if the URL serves a different providerId than the seeded identity;
- every later refresh still verifies against the pinned key;
- out-of-band fingerprint verification remains what the custom path is for.

Mechanics: new `confirmSeededDiscovery` closure on `OnboardingStepContentBuilder`, implemented in `OnymIOSApp` against the real repository (no-op when discovery is absent). Triggered from the services step's `onDisappear` — it fires after `flow.step` has already moved, so a forward-leave check (`step > .services`) catches exactly the moment the acceptance becomes final, and Back-navigation (where the user may still switch to custom) doesn't fire it. Idempotent, so a double fire is harmless; failure never blocks the walk (offline first run: the seed just stays unpinned).

## Honest "Not confirmed" summary

The Done summary's directory row previously showed `operatorKeyFingerprint` as detail (silently nil when unpinned) with a trailing "N sources" either way. An unpinned source is now visibly distinct: trailing **"Not confirmed"** (ru: «Не подтверждён») instead of the count, and no detail line; pinned keeps fingerprint + count. Strings appended to `Localizable.xcstrings` (en + ru).

Also checked OnymDiscovery's security-load-bearing comments for references to the deleted `discoveryConfirm` onboarding step as the TOFU site — none found; no package change needed (or made).

## Tests

- `SeededDiscoveryConfirmationTests` (new, OnymIOSTests): real trust pipeline over the byte-pinned `--ui-discovery` fixtures — pin + aggregate population, idempotence when already pinned, offline failure leaves the seed unpinned.
- `test_onboardingSmoke_defaultPathCompletes_thenNeverAgain` now asserts the pinned fixture fingerprint appears on the Done summary — the end-to-end proof the recommended path auto-pinned.
- Full `OnymIOSTests` suite and all three `OnboardingUITests` pass locally (iPhone 17e simulator).

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -7130,6 +7130,23 @@
           }
         }
       }
+    },
+    "Not confirmed": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Not confirmed"
+          }
+        },
+        "ru": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Не подтверждён"
+          }
+        }
+      }
     }
   },
   "version": "1.0"
--- a/Sources/OnymIOS/OnboardingSteps.swift
+++ b/Sources/OnymIOS/OnboardingSteps.swift
@@ -44,6 +44,12 @@ struct OnboardingStepContentBuilder {
     /// exists — the identity step's checklist binds to this instead of
     /// asserting success it can't know about.
     let identityReady: @MainActor () async -> Bool
+    /// TOFU-pins the SEEDED default discovery source when the user
+    /// leaves the services step on the recommended path — accepting
+    /// the recommended setup is the explicit confirm (see
+    /// `SeededDiscoveryConfirmation`). Idempotent; a no-op when
+    /// discovery is absent or the source is already pinned.
+    let confirmSeededDiscovery: @MainActor () async -> Void
     let loadSummary: @MainActor () async -> [OnboardingSummaryRow]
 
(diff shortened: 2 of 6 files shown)


rinat-enikeev:
drifted

PR #267, opened 2026-08-16
Gate sessions sign as the mandate's identity, not the selected one

rinat-enikeev:
## The bug

With 2+ identities on one device, switching identities broke the moderation gate: `GateCheckRepository.performAttempt` presented `userKey: record.mandate.user` (the identity that consented) but signed via `IdentityRepository.signWithStellarKey`, which uses the **currently-selected** identity. The enforcement service verifies strictly and refused every such session — `signature invalid: identity signature did not verify` in the deployed interface's logs — so the app hard-blocked on "Verification required". Retry rebuilds the same mismatched pair and can never succeed; re-consent successfully minted fresh mandates server-side without unblocking the gate. Foregrounding triggers the cadence re-check, so the repro was simply: switch identity, background, reopen.

(The "disappeared chats" from the same incident were the flip side of the identity switch — chats are owner-scoped by `ownerIdentityIDString` — not data loss.)

## The fix

- `ModerationSigner` gains `sign(_ message:, as userKey:)` — signature by the key the payload names, regardless of selection.
- `IdentityRepository.signWithStellarKey(_:matchingPublicKeyHex:)` resolves the named Stellar key across all stored identities; throws the new `IdentityError.noIdentityForKey` when it's absent (removed or quarantined).
- Signed **as** the identity their payload names: gate session, case response, appeal, report.
- Kept on plain `sign(_:)` because they present `userKeyID()` in the same flow (self-consistent): enrollment, mandate countersign, recovery redemption, authority-client credentials.

## Tests

- `testGateSessionIsSignedAsTheMandateIdentity` — the gate session must request the mandate's key; a regression to selected-identity signing records the `<selected>` sentinel and fails.
- `IdentityRepositorySignAsTests` — two-identity repository: signature verifies against the *named* identity's Ed25519 key while another is selected; unknown key throws `noIdentityForKey`.
- Full moderation suites (GateHardening, ModerationRepository, DeviceRecovery, ModerationAuthorityClient): 102 tests, 0 failures.

## Follow-up worth its own issue

`sessionRefusalReason` maps any 401/403 onto `backendRefused`, whose copy says "check the date and time" — this incident produced that misleading screen for an identity mismatch. The server could distinguish `signature_invalid` from timestamp staleness so the client can suggest switching to the enrolled identity.

### DIFF
--- a/Packages/OnymIdentity/Sources/OnymIdentity/IdentityError.swift
+++ b/Packages/OnymIdentity/Sources/OnymIdentity/IdentityError.swift
@@ -7,6 +7,9 @@ public enum IdentityError: Error, Equatable {
     case storedSnapshotInvalid(reason: String)
     case invalidMnemonic
     case identityNotLoaded
+    /// No stored identity's Stellar public key matches the requested
+    /// hex — it was removed, or quarantined by a fresh-install verdict.
+    case noIdentityForKey(String)
     case sdkFailure(String)
 }
 
@@ -25,6 +28,8 @@ extension IdentityError: LocalizedError {
             return "Invalid recovery phrase"
         case .identityNotLoaded:
             return "No identity is loaded — bootstrap or restore first"
+        case let .noIdentityForKey(hex):
+            return "No stored identity has the public key \(hex)"
         case let .sdkFailure(message):
             return "OnymSDK call failed: \(message)"
         }
--- a/Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift
+++ b/Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift
@@ -271,6 +271,35 @@ public actor IdentityRepository: InvitationEnvelopeDecrypting, InvitationEnvelop
         return try privateKey.signature(for: message)
     }
 
+    /// Ed25519 detached signature over `message` with the signing key
+    /// of whichever stored identity's derived Stellar public key is
+    /// `publicKeyHex` — NOT the currently-selected identity. The
+    /// moderation layer uses this for mandate-scoped sessions: the
+    /// mandate names the identity that consented, and that identity
+    /// must sign even while another one is selected in the picker.
+    /// Throws `.noIdentityForKey` when no stored identity matches
+    /// (removed, or quarantined by a fresh-install verdict).
+    public func signWithStellarKey(
+        _ message: Data,
+        matchingPublicKeyHex publicKeyHex: String
+    ) async throws -> Data {
+        try await ensureLoaded()
+        let target = publicKeyHex.lowercased()
+        for id in orderedIDs {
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
**APPROVE** — the gate fix is correct and well-targeted; the notes below are follow-ups, not blockers.

- `GateCheckRepository.swift:305` — `noIdentityForKey` is a *local, permanent* failure but falls into `.unreachable`, so it serves the cached state for `policy.offlineGrace` and then blocks with `.offlineGraceExpired` ("you're offline"). If the mandate's identity was removed or quarantined, no retry can ever succeed and the user gets connectivity copy for an identity problem — the same misleading-screen class as the `backendRefused` follow-up already noted in the PR body. Worth mapping it to `.enrollmentLost` (or a new reason) and adding a test; nothing covers the throw path today.
- `ModerationRepository.swift:1186`, `:1230`, `:1523` — these three are self-consistent already: `caseStanding(requireOwnership: true)` (`:957`) and `performFileReport` (`:1451-1452`) both hard-guard `userKeyID() == mandate.user`, and the new-holder branch sets `submittingUser = userKeyID()` (`:1110`). So `as:` resolves to the selected identity in every reachable case — the changes are no-ops that document intent, and the *user-visible* divergence bug (respond/appeal/report all throw "active identity does not own…" after an identity switch) is untouched. Fine to land, but the PR body reads as if those flows are fixed; they aren't.
- `ModerationRepository.swift:744` — `activeMandateRecord()` is `records.first { $0.isActive }`, and consent deactivates every other record device-wide (`:693`), so the gate is device-scoped, not selection-scoped. Before this PR that mismatch failed closed (hard block); now it succeeds silently against whichever identity last consented. Failing-closed is gone in the direction that matters: if identity B was banned and A later consented, B chats under A's clean mandate. Probably its own issue, but it's a consequence of this change, not pre-existing behavior.
- `IdentityRepository.swift:295` — when the hex matches but `keychain.read(id)` returns nil, the loop `continue`s and the caller sees `noIdentityForKey`, i.e. "no such identity" for what is really a keychain read miss on a known identity. Prefer `throw IdentityError.identityNotLoaded` there so the two failures stay distinguishable in logs.
- `IdentityModerationSigner.swift:25` — ad-hoc `onym:key:` prefix stripping duplicates `AuthorityKey.publicKey(fromReference:)` / `DiscoveryModels`'s helper, and the `else` branch silently treats any other reference form as raw hex. Reusing the shared parser would reject a malformed reference with a clearer error.
- Tests read well — `KeyCapturingSigner`'s `<selected>` sentinel (`GateHardeningTests.swift:271`) is a genuine regression trap, and `IdentityRepositorySignAsTests` verifies against the real Ed25519 key rather than a stub. I reviewed statically; I did not build or run the suites myself, so the 102/0 figure is from the PR body.

PR #268, opened 2026-08-16
Store-open failures: log, back up aside, never delete

rinat-enikeev:
## The incident (2026-08-16)

A device lost its chat history twice in one day: once at the TestFlight-over-debug install, once at the debug-over-TestFlight reinstall. Forensics on the container showed the four `.store` files recreated while month-old cache files survived — the wipe-and-retry in the store inits fired, silently. Two facts rule out the "schema mismatch" story the old comments assumed:

- Release-over-Debug from **identical sources** on a simulator clone with months-old stores opens cleanly — no wipe.
- `Groups.store` was wiped again **14 minutes into the same debug binary's session** — same build, same schema.

So the trigger is environmental: a transiently unreadable store (Complete file protection during a locked or prewarmed launch is the prime suspect) throws from `ModelContainer(...)` exactly like corruption does, and the old policy answered *any* throw with `removeItem`. Silent, permanent data loss. The catch blocks logged nothing, so the true error was never recorded anywhere.

## The policy, now shared

`PersistentStoreOpener` (OnymFoundation, all three store packages already depend on it):

- **Log first**: every open failure becomes an os_log fault with the store name and the real error — the next incident is diagnosable from Console.app.
- **Unreadable ≠ broken**: if the store file exists but yields no bytes (file-protection lock), the error is rethrown with the disk untouched; the app-level in-memory fallback covers the session and the data is intact next launch.
- **Back up, never delete**: an openable-but-incompatible store trio is renamed to `.bak` siblings (one generation, newest incident wins) before the fresh-store retry.

`SwiftDataGroupStore`, `SwiftDataMessageStore`, `SwiftDataInvitationStore`, and `SwiftDataIntroRequestStore` all route through it, replacing four copies of the wipe boilerplate.

## Tests

`PersistentStoreOpenerTests`: healthy store reopens with data intact and no backup manufactured; garbage store is moved aside byte-for-byte (not deleted) with a usable fresh store in its place; a second incident replaces the backup. Existing store-consumer suite (GroupStateRefreshTests) still green.

## Not in this PR

The recovered `.bak` currently has no re-import path — it makes loss recoverable by hand, not automatic. And the root-cause hunt for the launch context that makes the open fail (prewarming vs. locked background launch) needs the fault log this PR adds; today there is nothing to read.

### DIFF
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/SwiftDataMessageStore.swift
@@ -21,34 +21,17 @@ public actor SwiftDataMessageStore: MessageStore {
     /// `Application Support/OnymIOS/Messages.store`, with
     /// `FileProtectionType.complete` on the directory.
     ///
-    /// Schema-mismatch at `ModelContainer(...)` init wipes the
-    /// on-disk store and retries once. Pre-1.0 install base, no real
-    /// users to preserve — matches the policy on `SwiftDataGroupStore`.
+    /// Open failures go through `PersistentStoreOpener`: logged,
+    /// moved aside as `.bak` (never deleted), retried once — and left
+    /// completely untouched when the file is merely unreadable
+    /// (locked-device launch).
     public init() throws {
-        let appSupport = FileManager.default.urls(
-            for: .applicationSupportDirectory, in: .userDomainMask
-        )[0]
-        let storeDir = appSupport.appendingPathComponent("OnymIOS", isDirectory: true)
-        try FileManager.default.createDirectory(
-            at: storeDir,
-            withIntermediateDirectories: true,
-            attributes: [.protectionKey: FileProtectionType.complete]
+        let url = try PersistentStoreOpener.storeDirectory()
+            .appendingPathComponent("Messages.store")
+        let container = try PersistentStoreOpener.openContainer(
+            schema: Schema([PersistedMessage.self]),
+            url: url
         )
-        let url = storeDir.appendingPathComponent("Messages.store")
-        let schema = Schema([PersistedMessage.self])
-        let config = ModelConfiguration(schema: schema, url: url, cloudKitDatabase: .none)
-        let container: ModelContainer
-        do {
-            container = try ModelContainer(for: schema, configurations: [config])
-        } catch {
-            for suffix in ["", "-shm", "-wal"] {
-                try? FileManager.default.removeItem(
-                    at: url.deletingPathExtension().appendingPathExtension("store\(suffix)")
-                )
-            }
-            try? FileManager.default.removeItem(at: url)
-            container = try ModelContainer(for: schema, configurations: [config])
-        }
         self.container = container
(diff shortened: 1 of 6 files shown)


dev-onym (commented):
**COMMENT** — the core fix (log, never delete, don't touch unreadable stores) is right and correctly wired through all four stores; a few edge cases in the recovery path undercut the "recoverable by hand" promise.

- `PersistentStoreOpener.swift:98` — the `moveItem` failure fallback **deletes** the live store, reintroducing exactly the silent data loss this type exists to prevent. Concrete path: the pre-existing `.bak` can't be removed (line 90 swallows that error), so `moveItem` fails with "file exists", and the only copy of the user's messages is unlinked. In-memory-forever is already the accepted outcome on the unreadable branch (line 64) and is non-destructive — prefer a uniquely-suffixed backup name, or rethrow, over `removeItem`.
- `PersistentStoreOpener.swift:84-92` — backups can end up **mixed-generation**. Each suffix is skipped via `continue` when its source is absent, but the stale `.bak` for that suffix is left behind. Incident 1 leaves `X.store-wal.bak`; incident 2 (checkpointed, no `-wal` on disk) replaces only `X.store.bak` — a hand recovery then pairs a new main DB with a month-old WAL, which SQLite will happily replay. `testSecondIncidentReplacesTheBackup` (`PersistentStoreOpenerTests.swift:70-77`) creates precisely this state and only asserts on the main file. Fix: delete all three `.bak` siblings up front, then move whatever exists.
- `PersistentStoreOpener.swift:78` — `read(upToCount: 1)` returns `nil` at EOF, so a **0-byte** `.store` (interrupted first create) is classified "unreadable" → rethrow → in-memory session, every launch, forever, with nothing that ever repairs or backs it up. Special-case zero length as readable, or distinguish the EPERM throw at line 76 from an empty read.
- `PersistentStoreOpenerTests.swift` — the unreadable→rethrow→disk-untouched branch, the one behavior that actually fixes the incident, has **no test**. `chmod 0o000` on the store file makes `FileHandle(forReadingFrom:)` throw and gives a deterministic simulator-side test: assert `openContainer` throws and no `.bak` appears.
- `PersistentStoreOpenerTests.swift:72` — `_ = try PersistentStoreOpener.openContainer(...)` keeps a live `ModelContainer` on the path whose files the test then deletes and rewrites; test 1 uses `autoreleasepool` for exactly this reason. Wrap both incidents the same way or this will flake on WAL checkpoint-at-deinit.
- Housekeeping (non-blocking): a full-size `Messages.store.bak` lives in Application Support forever — backed up to iCloud, counted against device storage, with no reaper and (per the PR) no re-import. Worth a follow-up to drop the `.bak` once the fresh store has survived N clean launches.

Verified independently: `import OnymFoundation` and the package dependency are present in all four store files; all four production callers (`OnymIOSApp.swift:434,446,659,676`) use `try?` with a real in-memory fallback, so the new rethrow degrades rather than crashes; no wipe-on-open sites remain elsewhere; the glob-based `OnymIOSTests` target in `project.yml:199` picks up the new test file with no registration needed. I did not build or run the suite — no Xcode toolchain in this environment.

PR #269, opened 2026-08-16
Park out-of-order chat messages; drain when their group catches up

rinat-enikeev:
## Why 11 messages came back instead of the whole history

After the 2026-08-16 store wipes, the device's relaunch did replay the full inbox — the REQ has no `since`, no `limit`, and no persisted watermark, and the relay still holds ~35k events. But relays deliver stored events newest-first, so most chat messages arrived *before* the `GroupInvitationPayload` that re-materializes their group (or the `MemberAnnouncementPayload` that re-adds their sender), hit the unknown-group / unknown-sender guards in `persistChatMessage`, and were silently discarded mid-replay. Only the slice that happened to follow its group's materialization survived.

## The change

- `ChatMessageParkingLot` — a session-scoped, capacity-bounded actor in OnymInbox. Messages that hit either guard park (keyed by group id) instead of dropping.
- `materializeGroup` and `applyAnnouncement` drain the lot for their group after updating the repository, re-driving each parked message through the **normal** persist path — every trust check (roster membership, insider-spoof Ed25519 match, variant/group-type gate) runs unchanged at drain time. Parking defers verification; it never bypasses it. Still-blocked entries re-park; each drain removes its batch first, so there is no loop.
- Session-scoped on purpose: the replay that delivers an orphaned message delivers its group in the same subscription, and a launch that dies mid-replay just gets the full unbounded replay again on the next connect.

Net effect: any app relaunch is now a genuine backfill of whatever history the relay retains — which is also the recovery path for the wiped device.

## Tests

- `test_chatMessage_beforeInvitation_parksThenPersistsOnMaterialize` — message for an unknown group parks, then persists when the (chain-verified Tyranny) invitation lands.
- `test_chatMessage_unknownSender_parksThenPersistsWhenRosterCatchesUp` — known group, unknown sender parks, then persists when a re-delivered invitation carries the fuller roster.
- All existing dispatcher/fanout suites pass unchanged — including `unknownGroup_drops`/`senderNotInRoster_drops`, which assert nothing is *persisted* (parking keeps that true).

## Notes for review

- Delivered-receipt latching (`deliveredAckSent`) applies to drained messages exactly as to normal arrivals, so a large backfill acks each recovered message once — same volume as if they'd arrived in order.
- The `applyAnnouncement` group lookup matches on group id without an owner filter (pre-existing); the drain re-checks owner inside `persistChatMessage`, so a drain can only persist rows for owners whose group row actually exists.

### DIFF
--- a/Packages/OnymInbox/Sources/OnymInbox/ChatMessageParkingLot.swift
+++ b/Packages/OnymInbox/Sources/OnymInbox/ChatMessageParkingLot.swift
@@ -1,0 +1,57 @@
+import Foundation
+import OnymChatsCore
+import OnymIdentity
+
+/// Holding pen for chat messages that arrived before the local state
+/// they depend on.
+///
+/// Relays replay the stored inbox newest-first, so after a store loss
+/// (or on any fresh device) chat messages routinely land *before* the
+/// `GroupInvitationPayload` that materializes their group or the
+/// `MemberAnnouncementPayload` that adds their sender to the roster.
+/// The dispatcher used to drop such messages silently, which turned
+/// the relay's full-history replay into a one-shot lottery: only
+/// messages that happened to arrive after their group materialized
+/// survived. Parking them here and re-driving them when the group or
+/// roster catches up turns that replay into a real backfill.
+///
+/// Session-scoped by design: the replay that delivers an orphaned
+/// message delivers its group in the same subscription, and a launch
+/// that dies mid-replay gets the full replay again on the next
+/// connect (the inbox REQ is unbounded).
+public actor ChatMessageParkingLot {
+    struct Entry: Sendable {
+        let groupIDHex: String
+        let payload: ChatMessagePayload
+        let ownerIdentityID: IdentityID
+        let senderEd25519PublicKey: Data
+    }
+
+    /// Oldest entries are evicted first past this cap — a bound on
+    /// replay-burst memory, not a correctness knob. At the default
+    /// cap a full drop-and-repark cycle stays far below the size of
+    /// any realistic single-group backlog.
+    private let capacity: Int
+    private var entries: [Entry] = []
+
+    public init(capacity: Int = 2048) {
+        self.capacity = capacity
+    }
+
+    func park(_ entry: Entry) {
+        entries.append(entry)
+        if entries.count > capacity {
+            entries.removeFirst(entries.count - capacity)
+        }
+    }
+
+    /// Remove and return every parked message for `groupIDHex`, in
+    /// arrival order. The caller re-drives them through the normal
+    /// persist path; any that are still blocked re-park themselves.
+    func takeMatching(groupIDHex: String) -> [Entry] {
+        let matching = entries.filter { $0.groupIDHex == groupIDHex }
+        guard !matching.isEmpty else { return [] }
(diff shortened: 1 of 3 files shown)


dev-onym (commented):
**COMMENT** — the parking mechanism is sound (drain re-runs every trust check, no loop, dispatch is serialized per identity so there's no park/materialize race), but the stated goal "any relaunch is a genuine backfill" doesn't hold: the unknown-sender path has no working unblocker in a real replay.

- **`IncomingMessageDispatcher.swift:812` — announcements are still dropped out-of-order, so the unknown-sender park never drains in practice.** `applyAnnouncement` still returns early when the group isn't materialized yet. In a newest-first replay an announcement (sent *after* the invitation) is delivered *before* it, so for anyone who joined after the invitation snapshot: announcement dropped → roster never gains them → their messages park → re-park on every group drain → never persisted, and the same thing repeats next launch. The unknown-sender lot is unblockable for exactly the members it was built for. The parking idea needs to cover announcements too (or the roster needs a second source), otherwise only pre-snapshot members backfill.
- **`IncomingMessageDispatcherChatMessageTests.swift:405` — test 2 validates a path replays don't produce.** The roster "catches up" via a re-delivered invitation carrying a *fuller* roster; relays replay the original invitation bytes, so that roster is fixed. `applyAnnouncement`'s drain site (`IncomingMessageDispatcher.swift:853`) has no test at all, nor does the still-blocked→re-park→later-drain sequence.
- **`ChatMessageParkingLot.swift:41` — no dedup by message id, so the cap is effectively divided by relay count.** `NostrInboxTransport` merges each relay's stream with no cross-relay dedup (`NostrInboxTransport.swift:197-208`); `messageRepository.insert` absorbs the duplicates but the lot doesn't, so N relays park N copies. With 3 relays the 2048 cap holds ~680 distinct messages, against the ~35k-event replay this PR exists to recover. Key entries by `(owner, groupIDHex, messageID)`.
- **`ChatMessageParkingLot.swift:43` — FIFO eviction discards the wrong end.** Oldest-parked = earliest in a newest-first replay = the *most recent* messages. Past the cap the user keeps old history and loses the recent tail. Evicting by lowest `sentAtMillis` matches intent better; either way the doc claim at line 30-33 ("far below any realistic single-group backlog") is unsupported for the wiped-device case.
- **`ChatMessageParkingLot.swift:41`,`51` — eviction and parking are silent.** The bug being fixed was a silent drop; capacity eviction reintroduces one with no log or counter. Worth a log line at eviction, and at park when the entry is a re-park.
- **`ChatMessageParkingLot.swift:51` — `takeMatching` ignores owner.** A drain for identity B pulls identity A's entries for the same group, which then re-park at the tail (harmless, but it reshuffles eviction order and costs a full `currentGroups()` round-trip each). Key by `(ownerIdentityID, groupIDHex)`.
- **`IncomingMessageDispatcher.swift:934` — one full `currentGroups()` fetch per drained entry** (`:976`), plus a `receiptSender.send` per message. Draining a few thousand backfilled messages does that serially inside `materializeGroup`. Hoist the group lookup into the drain loop.

Nothing here regresses existing behavior — the drop-vs-park change is strictly safer, and the trust chain (`:1009` insider-spoof, `:1022` variant gate) genuinely still runs at drain time.

PR #270, opened 2026-08-16
Chat bubbles: tappable links and a Copy action on the long-press menu

rinat-enikeev:
## What

Two bubble affordances in one PR, as requested:

1. **Links in message text are tappable.** The body moves from `UILabel` to `LinkTextView`, a `UITextView` subclass that mimics the label everywhere except links: zero inner padding, Auto Layout self-sizing, zero intrinsic height when the body is empty (media-only bubbles), and — the load-bearing part — `point(inside:)` accepts only touches that land on a `.link` character. Everything else falls through to the cell, so table scrolling, the row's long-press menu, and swipe-to-reply behave exactly as before on body text. URLs are found by a shared `NSDataDetector` and rendered as an underline in the body's own color, readable on both the accent-filled outgoing bubble and the tinted incoming one. Long-pressing a link itself gets the system link preview/menu, which is the wanted behavior there.

2. **Copy on the long-press menu.** The existing context menu (previously Report-only, and only for reportable rows) now offers Copy for any non-system message with a non-blank body; Report keeps its gate unchanged. Copy puts the untrimmed body on the pasteboard.

## Tests

- `test_body_writesThroughToBodyView`, `test_body_urlGetsLinkAttribute_plainTextDoesNot` — the URL run carries `.link` with the parsed URL, the surrounding prose doesn't.
- Menu contract split: `test_longPress_unreportableMessageStillOffersCopyMenu` (has text → menu even when unreportable) and `test_longPress_unreportableEmptyBodyHasNoContextMenu` (nothing to copy or report → no menu).
- Full `ChatBubbleCellTests` + `ChatThreadViewControllerTests`: 90 tests, 0 failures — including the constraint-conflict and reuse suites, which cover the UILabel→UITextView layout swap.
- XCUITest exposure changed (bubble bodies are `textViews` now, not `staticTexts`): `ChatThreadScreen.waitForMessage` matches either, and the two direct `staticTexts` bubble assertions route through it.

## Related

`SearchUITests.test_search_findsMessage_andOpensChatAtIt` was failing on main — initially misattributed here to the onboarding redesign, actually a regression from #267 (the stub moderation seat's seeded mandate names a key no identity owns, so the identity-addressed gate signer throws and the app blocks on the verification screen). Fixed in #271.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatBubbleCell.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatBubbleCell.swift
@@ -103,7 +103,7 @@ final class ChatBubbleCell: UITableViewCell {
     private let maxWidthFraction: CGFloat = 0.75
 
     private let bubble = UIView()
-    private let bodyLabel = UILabel()
+    private let bodyView = LinkTextView()
     /// Image attachment view (shown when `message.imageAttachment != nil`).
     /// Sits above the caption; the blurhash placeholder renders first,
     /// then the decrypted image swaps in from `ChatImageLoader`.
@@ -333,14 +333,17 @@ final class ChatBubbleCell: UITableViewCell {
         // Reuse safety: a cell recycled mid-drag must start at rest.
         resetSwipeState()
         self.onSwipeToReply = onSwipeToReply
-        bodyLabel.text = message.body
         switch message.direction {
         case .outgoing:
             // Own messages: solid fill in the user's own accent (so
             // "your color" is consistent with the members list and
             // every group). Right-aligned, status glyph below.
             bubble.backgroundColor = UIColor(sender.accent.color)
-            bodyLabel.textColor = UIColor(OnymTokens.onAccent)
+            bodyView.setBody(
+                message.body,
+                font: .preferredFont(forTextStyle: .body),
+                color: UIColor(OnymTokens.onAccent)
+            )
             NSLayoutConstraint.deactivate(incomingConstraints)
             NSLayoutConstraint.activate(outgoingConstraints)
             bubbleBottomConstraint.isActive = false
@@ -350,7 +353,11 @@ final class ChatBubbleCell: UITableViewCell {
             // — distinguishable per person while keeping body text on
             // the regular text token readable.
             bubble.backgroundColor = UIColor(sender.accent.color.opacity(incomingTintAlpha))
-            bodyLabel.textColor = UIColor(OnymTokens.text)
+            bodyView.setBody(
+                message.body,
+                font: .preferredFont(forTextStyle: .body),
+                color: UIColor(OnymTokens.text)
+            )
             NSLayoutConstraint.deactivate(outgoingConstraints)
             NSLayoutConstraint.activate(incomingConstraints)
             statusBottomConstraint.isActive = false
@@ -883,20 +890,17 @@ final class ChatBubbleCell: UITableViewCell {
         bubble.layer.cornerCurve = .continuous
         contentView.addSubview(bubble)
 
-        bodyLabel.translatesAutoresizingMaskIntoConstraints = false
-        bodyLabel.numberOfLines = 0
-        bodyLabel.font = .preferredFont(forTextStyle: .body)
-        bodyLabel.adjustsFontForContentSizeCategory = true
+        bodyView.translatesAutoresizingMaskIntoConstraints = false
         // Hug the text horizontally with a decisive priority so a text
-        // bubble shrinks to fit its content. Without this the label sits at
-        // the default hugging (251), tying with the hidden media siblings
+        // bubble shrinks to fit its content. Without this the body view
+        // sits at its default hugging, tying with the hidden media siblings
         // (image/album/voice) that are also pinned edge-to-edge on the
         // bubble — autolayout resolves the tie inconsistently, so *some*
         // short text bubbles stretch toward the 75% cap. `.defaultHigh`
         // (750) beats those siblings while still yielding to the fixed
         // 75%-width constraint (1000) a real media message installs.
-        bodyLabel.setContentHuggingPriority(.defaultHigh, for: .horizontal)
-        bubble.addSubview(bodyLabel)
+        bodyView.setContentHuggingPriority(.defaultHigh, for: .horizontal)
+        bubble.addSubview(bodyView)
 
         attachmentImageView.translatesAutoresizingMaskIntoConstraints = false
         attachmentImageView.contentMode = .scaleAspectFill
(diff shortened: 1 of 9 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the link hit-testing accepts touches outside the text view, so taps in the bubble's padding can open a sender-controlled URL.

- `LinkTextView.swift:42` — `point(inside:)` never checks `bounds`. `hitTest` calls it with points outside the subview's frame too (that's how the "enlarge tap area" trick works), and `closestPosition(to:)` clamps any point to a valid position. So a tap in the bubble's 12pt side / 8pt bottom padding next to a body that ends in a link resolves to the link character and returns `true`: Safari opens from a tap that isn't on the text, and the row's long-press/swipe never sees it. Fix: `guard bounds.contains(point) else { return false }` first.
- `LinkTextView.swift:44-46` — the comment says trailing touches past the line end are rejected, but `.layout(.left)` does the opposite: it walks *back* to the preceding character. Tapping the empty area to the right of a link that ends a wrapped line still reports a hit. Compare the point against the character's `firstRect(for:)` (or the layout manager's bounding rect) instead of trusting `closestPosition`.
- `LinkTextView.swift:47` — at `beginningOfDocument`, `rangeEnclosingPosition(_, .character, .layout(.left))` returns nil, so a body that *starts* with a URL has a dead left half on its first glyph. Fall back to `.layout(.right)` when the left range is nil.
- `ChatThreadViewController.swift:996` — `UIPasteboard.general.string` syncs plaintext message bodies to the user's other devices via Universal Clipboard. `RecoveryPhraseBackupFlow.swift:258` already establishes the `setItems(..., options: [.localOnly: true])` pattern for sensitive text; message bodies deserve at least `localOnly` here (expiry is a product call).
- `Tests/OnymIOSTests/ChatBubbleCellTests.swift:51` — the load-bearing behavior of this PR, `point(inside:)` pass-through, has no test. Add cases: point on a link char → true; point on prose → false; point outside `bounds` → false; empty body → false. All three bugs above are unit-testable without a simulator UI.
- No test covers a long wrapping body's measured height. `UITextView` self-sizing is not `UILabel` self-sizing (intrinsic height depends on the container width resolved in a prior layout pass), and the existing suites only assert 320pt-wide single-line layouts. One `systemLayoutSizeFitting` assertion on a multi-line body vs. its single-line counterpart would cover the regression the label→text-view swap most plausibly introduces.
- `ChatBubbleCell.swift:342,356` — the two `setBody` calls differ only in color; hoist to `let bodyColor = ...` in the switch and call `setBody` once after it.
- `ChatFlowScreens.swift:245` — hand-rolled `RunLoop` polling replaces `waitForExistence`. `XCTNSPredicateExpectation` on an `.matching(...)` query, or just `bubble.waitForExistence(timeout:)` followed by a short `label` check, keeps the standard failure diagnostics. Non-blocking.

Not verified by me: I reviewed statically; the 90-test run and the pre-existing `SearchUITests` failure are as you report, and I didn't re-run them.

rinat-enikeev:
All review points addressed in 0f359d6:

- **Padding taps opening links / trailing-space false hits / dead first glyph**: all three were symptoms of position-based probing, so `point(inside:)` was rewritten to hit-test the links' `selectionRects(for:)` directly, behind a `bounds.contains(point)` guard. No `closestPosition`, no tokenizer — a hit is a touch inside a link range's rendered per-line rect, nothing else. (Worth noting: the suggested `firstRect(for:)`-per-character check also fails — TextKit 2 snaps in-link positions to the link *boundary*, which the new `test_pointOnLinkGlyph_hits` caught before the rewrite.)
- **Universal Clipboard**: Copy now writes via `setItems(..., options: [.localOnly: true])`, matching the recovery-phrase posture. No expiry — a message body is user-visible content, unlike the seed; flag if you disagree.
- **Hit-testing tests**: `LinkTextViewTests` pins six cases — link glyph hits; prose, out-of-bounds (three sides), empty-space right of a link's line, and empty body all miss.
- **Multi-line self-sizing**: `test_multilineBody_measuresTallerThanSingleLine` asserts a wrapping body measures >2× a one-liner via `systemLayoutSizeFitting` at 320pt.
- **`setBody` hoist**: single call after the direction switch.
- **`waitForMessage`**: `XCTNSPredicateExpectation` + `XCTWaiter` instead of the RunLoop poll.

Also, the "pre-existing `SearchUITests` failure" note in the PR description was wrong — it was a regression from #267 (the stub moderation seat's seeded mandate names a key no identity owns, so the identity-addressed signer throws and the gate blocks the app). Fixed separately in #271.

Suites: LinkTextViewTests, ChatBubbleCellTests, ChatThreadViewControllerTests all green.

dev-onym (commented):
**COMMENT** — solid approach and good hit-testing tests; a few things worth tightening before merge.

- `Tests/OnymIOSUITests/PageObjects/ChatFlowScreens.swift:249` — `app.textViews.matching(value == text)` matches *any* text view, including the composer `chat.input.textview`, whose `value` is the typed body. `handleSend` clears it (`ChatThreadViewController.swift:872`), so the only case where it still holds the text is a send-button tap that never registered — precisely what `thread.send(x); waitForMessage(x)` (`MultiIdentityChatUITests.swift:106`) is supposed to catch. Give `LinkTextView` an `accessibilityIdentifier` (e.g. `chat.bubble.body`) and match on that, or exclude the input's identifier.
- `LinkTextView.swift:85` — no scheme filter and no `UITextViewDelegate`, so a tap goes straight to the system opener for whatever `NSDataDetector` matched. Message bodies are sender-controlled and the app registers its own `onym://` scheme (`Info.plist:30`); `DeeplinkCapture`'s allowlist means the worst case is a join sheet rather than a silent action, but restricting `.link` runs to `http`/`https`/`mailto`/`tel` in `setBody` is a two-line hardening and drops the whole class.
- `LinkTextView.swift:27` + `:85` — `adjustsFontForContentSizeCategory` is set, but the view is driven exclusively by `attributedText` with a font baked in at configure time; on `UITextView` that flag adjusts the `font` property, not attributed runs, and nothing in `ChatThreadViewController` observes trait changes. The old `UILabel` re-scaled live. Please verify body text still grows on a Dynamic Type change without leaving/re-entering the thread.
- `ChatThreadViewController.swift:996-1003` — the comment claims "same posture as the recovery-phrase copy," but that write also carries an OS-enforced `expirationDate` (`RecoveryPhraseBackupFlow.swift:141`). Either add an expiry for E2E message text or reword so the comment doesn't overstate parity.
- `LinkTextView.swift:37` — an empty `UILabel` reports a full line height, not zero, so media-only bubbles likely lose ~1 line of height with this change. Probably an improvement, but it isn't "keep the label's behaviour" as the doc says — worth an eyeball on an image/voice bubble.
- `LinkTextView.swift:42` — a pan that *starts* on a link glyph now belongs to the text view, so swipe-to-reply and scroll-drag are dead on that run of characters. Acceptable, but not covered by the tests or mentioned in the class doc.
- `Tests/OnymIOSTests/LinkTextViewTests.swift:59` — comment references `.layout(.left)` and "the fallback," which the rendered-rect implementation no longer has. Stale after the review-fix commit.
- Nit: `"Copy"` already exists in `Resources/Localizable.xcstrings` (en/ru) — no localization gap.

PR #272, opened 2026-08-17
Multi-use invite links, retired only by the inviter

rinat-enikeev:
An invite link died on the first Approve — or Decline. Both revoked the per-invite intro key, which tore down the relay subscription the link points at, so every later joiner's request landed on a tag nobody was listening to. Nothing surfaced this on either side: the joiner sat on "Waiting for the host to approve…" forever, and the host never saw a row. Reported from the field as "the same link doesn't work twice in a row".

This is the #209–#213 stack rebased onto the SPM package layout (it was written against `Sources/OnymIOS/Group/…`; these files now live in `Packages/OnymGroup/`). Squashed to one feature commit for review, plus two commits addressing review findings.

## What changes

- **Approve and Decline no longer revoke.** Links are retired only by the inviter — "Generate new link" rotates, and a new invite list offers per-row Revoke. Every join is still gated by an explicit tap; nothing is auto-granted.
- **Re-join is safe.** A joiner already in `members` skips the chain anchor, so a replay or reinstall can't push a duplicate Merkle leaf. This also makes the long-standing double-tapped-Approve race safe.
- **Both approve and decline consume the whole collapsed sibling set**, re-derived from the live store at consume time.
- **Handled request ids are tombstoned durably**, keyed by the link they arrived on and pruned when that link is retired. Necessary because the subscription now outlives the approval and the Nostr REQ carries no `since`, so every reconnect replays requests already acted on.
- **Create-time offer keys are labelled per invitee** and listed with a per-row Revoke, alongside any superseded shared key, so no live key is unlistable.
- **The joiner gets a 90s `.unanswered` state** — a dead link is finally distinguishable from a thinking host. Non-terminal: a late approval still wins.

## Rebase decisions worth reviewing

Three conflicts needed judgement rather than mechanical resolution:

**The 24h TTL is gone, and this reverses a later decision.** `main` added `entryTTL` after this branch (issue #111, matching Android's `LIFETIME_MILLIS`); the branch deletes expiry entirely. Git auto-merged the deletion of the *filter* while the conflict resolution kept the *constant* — dead code claiming an expiry that no longer happened. I resolved toward the branch's design, because its whole UX assumes permanence (one reusable link, explicit rotate, per-row Revoke) and a silent 24h death recreates the exact complaint this fixes. **This is the one call I'd most like a second opinion on** — see Risks.

**The durable request store is kept.** `main` gained `SwiftDataIntroRequestStore` after this branch, which replaced it with an in-memory store. Kept main's durable store and moved the tombstones onto it.

**`alreadyInRoster` gates the anchor only.** I first also gated `broadcastJoin` and the admin's "X joined" row behind it; review showed that breaks the documented retry repair (see below).

## Review findings addressed

Two independent reviews of the diff. The load-bearing ones, all fixed in the follow-up commits:

- **Tombstone pruning wiped other identities' tombstones.** It was fed the *active* identity's snapshot but pruned a process-wide log, so any identity switch — and any locked-device launch, where the keychain read yields nothing — deleted every other identity's tombstones and replayed their handled requests as pending. Prune now names what **retired** (diffed from successive snapshots) rather than what survived, which cannot express either wipe.
- **The `alreadyInRoster` guard broke retry repair.** `approve()` is documented as re-runnable and persists the anchor *before* seal/send, so a transport failure leaves the joiner in `members` while no other member has heard of them — and the retry skipped the repair, parking their messages forever. The flag reads `members`; receivers dedupe on `memberProfiles`, and those diverge in exactly that case. Both calls are idempotent at the receiver, so they're unconditional again.
- **`consume()` only tombstoned durably when a row was present**, so the "laid ahead of a replay" case it documents was process-lifetime only.
- **`loadAll()` now filters tombstoned ids** — the tombstone commits before the row is deleted, so a crash in that window resurrected a handled request.
- The missing-intro-key drop now increments `decryptFailures` (it was the only decode path with no signal, and it's the common case for a retired link); rotate refreshes the invite list; the shared link's key is remembered rather than re-parsed (a non-`.ready` state listed the live link as revokable); the fallback store gets the durable log.

## Risks, stated plainly

- **A leaked link is now a standing invitation.** It used to be worth one join attempt. It's a spam/enumeration surface, not an access-control bypass — every joiner still needs an explicit Approve — but the only backstop is a human noticing and revoking.
- **Decline has no teeth, and the joiner has a "Try again" button.** There's no blocklist anywhere in the repo, and each retry is a fresh event id, so tombstones and the collapse key don't suppress it. A declined stranger can refill the queue indefinitely; the only remedy is rotating the link and redistributing it. Previously, decline locked them out. **Not fixed here** — a declined-joiner suppression list is a feature, and I'd rather scope it deliberately than bolt it on. This is the sharpest regression in the change.
- **Relay REQ slots grow monotonically.** Nothing expires and there's no group-scoped cascade, so a group created with 50 invitees holds 51 permanent subscriptions. The TTL used to reclaim these. A future leave/delete-group flow must cascade a revoke.
- **iOS/Android asymmetry.** Android still expires intro keys at 24h, so the same QR works forever with an iOS admin and dies overnight with an Android one — silently on both. Android needs the matching change before this ships widely.
- **Pre-upgrade keys all decode as `label == nil`**, i.e. as the shared link, so on upgrade `currentOrMint` may publish a per-invitee offer key and the first rotate revokes all of them. A migration marker would close this.
- `UserDefaultsHandledIntroRequestLog` stores `(intro pubkey, handled ids)` in UserDefaults — weaker file protection than the SwiftData store it mirrors, and backed up. Event ids aren't secret, but the pair reconstructs invite activity and derives the inbox tag.

**Android has the identical revoke-on-approve bug** (`JoinRequestApprover.kt:341` approve, `:461` decline). It needs its own PR; a faithful port needs the reusable-link/rotate/invite-list UI and tombstones to land coherently, not just deleting the two calls.

## Tests

1307 unit tests, 0 failures, on `iPhone 17 Pro` simulator. Both test bundles compile, so the UI-test target still builds against the changed views. New coverage for the tombstone half the rebase moved onto the durable store — including the cross-identity wipe and the no-row durability case — each verified to fail without its fix rather than pass vacuously.

Not verified: no device or multi-device run. Given the bug being fixed is a cross-device one, a two-device walkthrough (one link, two joiners, approve both) is worth doing before merge.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadViewController.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatThreadViewController.swift
@@ -134,6 +134,20 @@ final class ChatThreadViewController: UIViewController {
     /// user is already near the bottom.
     private var hasAppliedFirstSnapshot = false
 
+    /// Real, measured height per row id, filled in as rows are
+    /// displayed and served back through
+    /// `tableView(_:estimatedHeightForRowAt:)`.
+    ///
+    /// Without this the table answers every height question about a
+    /// row it hasn't displayed with the flat `estimatedRowHeight`, and
+    /// corrects it to the real height the moment the row is measured.
+    /// Each correction resizes the content and shifts the offset under
+    /// the user's thumb — which is the jump a fresh message produced:
+    /// inserting a row makes the table re-measure, so the whole thread
+    /// lurched every time one landed. Cached heights make the estimate
+    /// exact, so nothing moves but the intended scroll.
+    private var measuredRowHeights: [UUID: CGFloat] = [:]
+
     /// The message the composer is currently replying to, if any. Set
     /// by a swipe-to-reply on a bubble, cleared on cancel or after a
     /// send. Threaded into `onSendTapped` so the sent message carries
@@ -300,9 +314,23 @@ final class ChatThreadViewController: UIViewController {
         if !changedIDs.isEmpty {
             snapshot.reconfigureItems(changedIDs)
         }
+        // A reconfigured row may render at a different height (body
+        // edit, attachment resolving), so its cached measurement is
+        // stale until it's displayed again.
+        for id in changedIDs { measuredRowHeights[id] = nil }
+        pruneMeasuredHeights(keeping: snapshot.itemIdentifiers)
+
+        // A new row at the bottom that we're going to follow is applied
+        // without animation: the insert fade and the follow-scroll are
+        // two separate motions — the second only starting when `apply`'s
+        // completion fires, a third of a second after the row landed —
+        // and the pair read as a lurch. Committing the row instantly and
+        // animating the offset gives one continuous scroll instead.
+        let followsBottom = !isFirstApply && wasNearBottom && sorted.count > previousCount
+        let animates = !isFirstApply && !followsBottom
         // First apply is non-animated to avoid an initial-load
         // "fly-in" of every existing message.
-        dataSource.apply(snapshot, animatingDifferences: !isFirstApply) { [weak self] in
+        dataSource.apply(snapshot, animatingDifferences: animates) { [weak self] in
             guard let self else { return }
             if isFirstApply {
                 // Cold open: land at the bottom with no visible scroll.
@@ -313,10 +341,18 @@ final class ChatThreadViewController: UIViewController {
                 guard !sorted.isEmpty else { return }
                 self.jumpToBottomForColdOpen()
                 self.hasAppliedFirstSnapshot = true
-            } else if wasNearBottom && sorted.count > previousCount {
+            } else if followsBottom {
                 self.scrollToBottom(animated: true)
             }
         }
+    }
+
+    /// Forget measurements for rows no longer in the thread, so the
+    /// cache can't outgrow the list over a long session.
+    private func pruneMeasuredHeights(keeping ids: [UUID]) {
+        guard measuredRowHeights.count > ids.count else { return }
+        let live = Set(ids)
+        measuredRowHeights = measuredRowHeights.filter { live.contains($0.key) }
     }
 
     /// Push the founder's pending join requests for *this* group into
@@ -396,6 +432,8 @@ final class ChatThreadViewController: UIViewController {
         if !changed.isEmpty {
             snapshot.reconfigureItems(changed)
         }
+        for id in changed { measuredRowHeights[id] = nil }
+        pruneMeasuredHeights(keeping: snapshot.itemIdentifiers)
         dataSource.apply(snapshot, animatingDifferences: hasAppliedFirstSnapshot) { [weak self] in
             guard let self else { return }
             // The table sits at alpha 0 until the cold open reveals it,
@@ -651,13 +689,37 @@ final class ChatThreadViewController: UIViewController {
     }
 
     /// Scroll the table so the last row is visible.
+    ///
+    /// Targets the content offset directly rather than going through
+    /// `scrollToRow(at: .bottom)`. `scrollToRow` resolves its
+    /// destination from the row heights the table knows *at the time of
+    /// the call* — with self-sizing bubbles that includes estimates —
+    /// so it would animate to a position that stops being the bottom as
+    /// soon as the rows underneath it were measured, and the table
+    /// snapped the rest of the way. Laying out first and then animating
+    /// to the computed end of the content lands in one motion.
     func scrollToBottom(animated: Bool) {
-        let snapshot = dataSource.snapshot()
-        guard let lastSection = snapshot.sectionIdentifiers.last else { return }
-        let itemCount = snapshot.numberOfItems(inSection: lastSection)
-        guard itemCount > 0 else { return }
-        let lastIndex = IndexPath(row: itemCount - 1, section: 0)
-        tableView.scrollToRow(at: lastIndex, at: .bottom, animated: animated)
+        guard dataSource.snapshot().numberOfItems > 0 else { return }
+        // Resolve pending inserts/heights so `contentSize` below is the
+        // post-update one, not the size from before this row landed.
+        tableView.layoutIfNeeded()
+        let inset = tableView.adjustedContentInset
+        let target = max(
+            -inset.top,
+            tableView.contentSize.height + inset.bottom - tableView.bounds.height
+        )
+        guard abs(target - tableView.contentOffset.y) > 0.5 else { return }
+        guard animated else {
+            tableView.contentOffset.y = target
+            return
+        }
+        UIView.animate(
+            withDuration: 0.25,
+            delay: 0,
+            options: [.curveEaseOut, .beginFromCurrentState]
+        ) {
+            self.tableView.contentOffset.y = target
+        }
     }
 
     // MARK: - Table view (message list)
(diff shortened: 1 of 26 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the core fix is right and well-tested; one upgrade-path defect is cheap to close, and I'd push back on half the TTL removal.

**Blocking**

- `InviteIntroducer.swift:96` — the migration gap is worse than "a risk". Every pre-upgrade entry decodes as `label == nil`, and `listForOwner` sorts newest-first (`KeychainIntroKeyStore.swift:61`), so on upgrade `currentOrMint` adopts the *most recently minted legacy key* as the group's shared link — for a create-with-invitees group that is the last invitee's private offer key, now handed to everyone. The first `rotate` then revokes all of them (`:148`), killing every outstanding legacy invite at once. Cheap fix: stamp legacy rows with a sentinel label at first decode in `StoredIntroKey.toEntry()`, or have `currentOrMint` mint fresh when every unlabelled candidate predates the upgrade. Everything else here I'd merge.

**On the TTL question you asked about**

- Agreed for the *shared* link — it's visible, rotatable, and per-row revokable, so permanence is the honest model. I'd not extend it to the create-time offer keys (`CreateGroupInteractor.swift:746`). Those are 1:1 with a named invitee and inherently one-shot, yet now never expire, each hold a permanent relay REQ slot, and are only retirable from a list the user has to go find. Auto-revoke an offer key once *its* invitee is approved — the label already carries the inbox fingerprint, so the mapping exists — and the 51-permanent-subscriptions risk drops to 1 without weakening the fix.

**Should fix here**

- `JoinRequestApprover.swift:804` — the new `decryptFailures += 1` fires on every emission for the same row. `refresh` re-decodes the whole raw set per store publish, and a request against a retired link is never consumed, so one stale request inflates the counter without bound. The signal added for exactly this case can't distinguish one dead link from a thousand replays; count it once per id, or only on a first sighting.
- `IntroKeyEntry.swift:55` — 4-byte fingerprints as labels: two invitees colliding (or the same invitee added twice) yields two identical "Direct invites" rows and a duplicate `share_invite.revoke_button.<label>` a11y id at `ShareInviteView.swift:120`, so the user can't tell which Revoke kills which. Widen it or index-disambiguate.
- `decline` — I don't think "a suppression list is a feature" gets you out of it, because the button now does nothing durable. There's a cheap version entirely inside this change: tombstone the *collapse key* (joiner BLS ⊕ group) on decline rather than only event ids. No new UI, no blocklist model, and it restores decline's meaning until the real feature lands.

**Minor**

- `OnymIOSApp.swift:1589` — the tee only prunes on a live non-empty→smaller transition, so links retired while the app was closed or while another identity was active never prune. The log then shrinks only via `maxEntries`, which evicts *oldest-first* and can drop a live link's tombstone → that handled request returns as pending. The diff-based prune is the right shape; it just needs a reconcile pass at subscribe time too.
- `InviteIntroducer.swift:171` — "newest-first" holds for the keychain store but isn't a protocol guarantee (the in-memory double doesn't sort). Either sort in `liveInvites` or drop the claim.
- Android's 24h expiry landing after this ships means the same QR behaves differently per admin platform, silently on both. Agreed that's a release gate, not a PR gate — but it should block the release, not just be noted.

Two-device walkthrough before merge, as you said — one link, two joiners, approve both, then relaunch the host cold to exercise the tombstones against a real relay replay.

rinat-enikeev:
All addressed in f4f1e7a, including both design points — you were right on each.

**The blocking migration defect.** Worse than I'd written it up, and your read of the mechanism is exactly right: newest-first ordering means the adopted key is the *last invitee's* private offer key. Fixed by stamping `label_version` on every write, so its absence identifies a row this build never wrote. Legacy rows are never adopted as the shared link and never mass-revoked by rotate, while staying listed and individually revokable. The test drives your scenario — a legacy `label == nil` entry minted newest — and fails both assertions without the fix (adoption, then mass-revoke).

**Offer keys and the TTL.** I've taken this. Auto-revoke on approve, matched on the label's inbox fingerprint, so it only ever fires for the person that key was minted for. That's a better answer than the TTL was: the 24h expiry killed offer keys whether or not they'd been used, and left the shared link dying too. This kills exactly the spent ones and leaves the shared link permanent. On a 50-invitee group the standing-subscription count goes 51 → 1.

**Decline.** You're right that "it's a feature" didn't get me out of it — the button was doing nothing durable, and I shipped that knowingly. Your framing made the cheap version obvious: tombstone the collapse key rather than the event id, which is precisely what survives the retry. No blocklist model, no new UI, and the link stays live for everyone else. Declines now survive relaunch.

**The counter.** Fixed — counted once per request id. You're right that it was worse than no signal: unbounded growth off a single dead link, indistinguishable from real replay volume.

**Fingerprint collisions.** Accessibility ids now key on the intro pubkey rather than the label, which is unique by construction. I left the 4-byte fingerprint as the *display* label since it's the same shape the approval row shows; the ambiguity that mattered was the user not knowing which Revoke does what, and that's resolved.

**The reconcile.** Took this too, and it needed a new seam to be safe: a per-owner snapshot can't tell "retired" from "belongs to another identity", so I added `IntroKeyStore.allLivePublicKeys` and run one owner-agnostic reconcile at subscribe. Unattributed tombstones are kept — they have no link to check against. Your point about `maxEntries` evicting oldest-first and dropping a *live* link's tombstone is the part that makes this a correctness fix rather than hygiene.

**`liveInvites` ordering.** Sorts itself now rather than relying on the keychain store's behaviour.

**Android.** Agreed it's a release gate, not a PR gate. onymchat/onym-android#230 is the parity PR; I'm porting these same six fixes there now so the two don't diverge again, and I'll note the pairing on both. Nothing should ship until both are in.

1310 tests, 0 failures. Still no device run — the two-device walkthrough with a cold host relaunch is the right gate and I haven't done it.

dev-onym (commented):
COMMENT

The change is well-reasoned and the hard rebase calls (TTL removal, durable store) look right; findings below are real but none is a ship-blocker on its own.

- `Packages/OnymGroup/Sources/OnymGroup/KeychainIntroKeyStore.swift:80` — the new `allLivePublicKeys()` was inserted between `@discardableResult` and `deleteForOwner`, so the attribute now annotates the wrong function. `deleteForOwner` lost it, and `Sources/OnymIOS/OnymIOSApp.swift:1421` discards its `Int` → unused-result warning (build break under warnings-as-errors). Move the attribute back.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:775` — the PR body's risk section ("a declined stranger can refill the queue indefinitely… Not fixed here") is stale: `refresh` filters `declinedCollapseKeys`, so decline *is* durable suppression. Update the body — reviewers weighing the "sharpest regression" are reading a claim the code no longer matches.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:775` — flip side: a mistaken Decline is now permanent and invisible. There's no un-decline API on `HandledIntroRequestLog` and no UI listing declined joiners; the collapse key silently drops every future request from that person for that group. Deliberate scope-out is fine, but it deserves a line in the PR body and probably an issue.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:775` — declined joiners' retry rows are recorded but never consumed (the `continue` skips them before they reach any consume path), so they accumulate in `SwiftDataIntroRequestStore` and are re-decrypted on every emission until the retention sweep ages them out — a cheap CPU lever for exactly the declined-spammer the PR worries about. Consider consuming ids whose collapse key is declined right there.
- `Packages/OnymGroup/Sources/OnymGroup/HandledIntroRequestLog.swift:56` — `record` early-returns on a duplicate id, so a tombstone first laid unattributed (consume racing a replay) never gains its link attribution when the row later lands; it becomes permanently unprunable and only leaves via the 2,000-entry oldest-first eviction. Consider upgrading empty→real attribution on re-record.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:427` — approve revokes the offer key *before* consuming, so the tee's diff-prune can race the tombstone write: the tombstone lands attributed to an already-retired link and survives until the next startup reconcile. Harmless today (the dead link can't decode), but consuming before revoking closes it for free.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:159` — `countedDecodeFailures` grows without bound per distinct undecodable event id; a spammer on a retired link inflates it for the process lifetime. Cap or make it an LRU.

The verified-to-fail-first test discipline shows — the multi-use, sibling-consume, and cross-identity-wipe cases are all pinned. The two-device walkthrough the PR body asks for is still the right pre-merge gate, and Android's matching 24h-expiry change should be tracked before wide rollout.

PR #273, opened 2026-08-19
Onboarding: align copy with punchier design pass

rinat-enikeev:
## Summary
- Updates onboarding copy (welcome, identity, services, reports & safety, recovery phrase, done) to match the new "no account, ever" framing from the [Onym Onboarding design doc](https://claude.ai/design/p/f3d5bdf0-84de-429c-9289-e7311f4b6c6b?file=Onym+Onboarding.dc.html).
- Small layout touch-ups to match: welcome headline under the wordmark, a "You choose the referee" banner on the reports & safety authority picker, and a two-tone value/annotation treatment on the services summary line (was a single concatenated string).
- Adds the design's welcome-screen "I have a recovery phrase" affordance: a new sheet takes a 12/24-word mnemonic and calls `IdentityRepository.restore(mnemonic:)` — a code path that already existed (validated, used in tests) but had no production UI wired to it. This replaces whatever identity the app's launch-time bootstrap already minted with the one the entered phrase describes. `OnboardingFlow` gained an `identityOrigin` (`.minted`/`.restored`) so the identity step's checklist and title don't claim a key was "created" over one the user just typed in.
- Deliberately did **not** copy the design mock's terms/consent screen (kept the real per-authority, per-violation-class data instead of the mock's static "They can / can't" list) or its relay-fallback wording (kept the existing accurate fan-out-vs-fallback distinction between relays, media hosts, and notaries) — the mock is illustrative and its simplifications would have regressed copy that was previously made deliberately precise.
- Still not added: the design's "Checking the network" identity-step row, since it implies a live network probe that doesn't exist.

## Test plan
- [x] `xcodebuild -scheme OnymIOS -destination 'generic/platform=iOS Simulator' build` succeeds
- [x] Fresh-install screenshot confirms the welcome screen renders the new headline, punched-up bullets, and "I have a recovery phrase" button correctly
- [ ] Manual tap-through of the restore sheet (phrase entry → validation error → successful restore → identity step shows "restored" copy) — not automated in this environment; the sheet reuses the same `restore`/error-handling pattern as the already-shipped `RestoreIdentitySheet` in Settings
- [ ] Manual walk of the rest of the onboarding flow (services → reports & safety → recovery phrase → done)

### DIFF
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingFlow.swift
@@ -78,6 +78,20 @@ public enum ServicesSetupChoice: Equatable, Sendable {
     case custom
 }
 
+/// How the current identity came to exist. Lives on the flow (not the
+/// step view) for the same reason as `ServicesSetupChoice`: the
+/// `identity` step's content is rebuilt on every navigation, and its
+/// checklist copy must not fall back to "created" over an identity
+/// that was actually restored from a phrase the user typed in.
+public enum IdentityOrigin: Equatable, Sendable {
+    /// The default: a fresh key generated on this device.
+    case minted
+    /// The welcome step's "I have a recovery phrase" path replaced the
+    /// freshly-minted identity with one derived from an entered
+    /// mnemonic.
+    case restored
+}
+
 /// State machine for the first-launch onboarding sequence. Modeled on
 /// `ModuleConsentFlow` / `ModerationConsentFlow`: `@MainActor
 /// @Observable`, every collaborator injected as a closure so the
@@ -120,6 +134,11 @@ public final class OnboardingFlow {
     /// content as the backup sheet progresses. Never downgrades:
     /// `verified` stays `verified` across navigation and re-reveals.
     public var recoveryBackupState: RecoveryBackupState = .none
+    /// Written by the welcome step's restore sheet the moment a
+    /// mnemonic restore succeeds — read by the identity step's content
+    /// so its checklist doesn't claim a fresh key was "created" over
+    /// one the user just typed in.
+    public var identityOrigin: IdentityOrigin = .minted
     /// Flips exactly once, inside `complete()` — the observable signal
     /// the presenter dismisses its cover on. `onCompleted` alone can't
     /// carry the dismissal: it is bound at the composition root, which
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
@@ -129,15 +129,15 @@ public struct OnboardingView: View {
         case .welcome:
             return String(localized: "Messaging that answers to you — not to a company.")
         case .identity:
-            return String(localized: "This takes a second. Everything happens on your iPhone — nothing has been sent anywhere yet.")
+            return String(localized: "This takes a second, and it happens only on this iPhone. Nothing is sent anywhere — there's nowhere to send it to yet.")
         case .services:
-            return String(localized: "Onym runs on independent services for delivery, files and timestamps. Start with a set that works, or pick your own.")
+            return String(localized: "No homeserver to pick. Just independent services for delivery, files and timestamps — split up on purpose, so no single one holds the whole picture.")
         case .moderation:
             return String(localized: "If someone reports illegal content, an authority you choose reviews the report. Pick one to continue — you'll see exactly what it can do before you agree.")
         case .recoveryPhrase:
-            return String(localized: "These 12 words are the only way back into your account if you lose this iPhone. Onym cannot reset them for you.")
+            return String(localized: "These 12 words ARE your identity. Lose this iPhone without them, and even Onym can't get you back in.")
         case .done:
-            return String(localized: "Your identity is on this device and your services are connected.")
+            return String(localized: "Your identity lives on this device. No account, no admin, no company holds it but you.")
         }
     }
 
@@ -154,7 +154,7 @@ public struct OnboardingView: View {
     /// and it reads as a deferral, not a skip.
     static func skipTitle(for step: OnboardingStep) -> String {
         switch step {
-        case .recoveryPhrase: return String(localized: "Remind me later")
+        case .recoveryPhrase: return String(localized: "Remind me tonight")
         default: return String(localized: "Skip")
         }
     }
--- a/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
+++ b/Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift
@@ -55,7 +55,7 @@ public struct OnboardingView: View {
         OnboardingStepScaffold(
             step: step,
             title: Self.title(for: step, restoring: flow.identityOrigin == .restored),
-            subtitle: Self.subtitle(for: step),
+            subtitle: Self.subtitle(for: step, restoring: flow.identityOrigin == .restored),
             primaryTitle: Self.primaryTitle(for: step),
             // Outcome-gated steps render their primary disabled until
             // the step content records one (moderation's consent, the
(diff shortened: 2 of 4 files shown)


dev-onym (commented):
REQUEST_CHANGES

- **`Sources/OnymIOS/OnboardingSteps.swift:1855` — "Tap any line to change it" is false.** The done-summary rows (OnboardingSteps.swift:1774–1820) are plain `HStack`s with no button, tap gesture, or navigation. The old copy ("Tap into Settings → Services") pointed somewhere real; the new copy promises an interaction that doesn't exist and removes the only pointer to where changes actually happen.
- **`Sources/OnymIOS/OnboardingSteps.swift:500` — "2 relays" / "primary + backup" is fabricated.** The shipped seed is a single relay (`NostrRelaysConfiguration.seed`, NostrRelaysConfiguration.swift:30–42), and the doc comment there says V1 fans out to *every* configured relay — "there's no primary/strategy concept." The real installed list comes from a runtime-fetched release asset (KnownNostrRelaysFetcher.swift), so the count isn't knowable at compile time either. This also reintroduces exactly the fallback framing the PR body says it deliberately declined from the mock.
- **`Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:157` — "Remind me tonight" promises behavior that doesn't exist.** `skip()` just marks the step deferred; the follow-up is Settings nudging until backup (OnboardingFlow.swift:39). Nothing schedules a notification, tonight or otherwise. It also strands ~8 comments/tests that still say "Remind me later" (OnboardingView.swift:12, OnboardingStepScaffold.swift:28, OnboardingFlow.swift:39/89/168, OnboardingFlowTests.swift).
- **`Sources/OnymIOS/OnboardingSteps.swift:291` — "nowhere else, ever" overclaims and self-contradicts.** `Identity` carries a `recoveryPhrase` mnemonic (Identity.swift:36), and two steps later the flow tells the user "These 12 words ARE your identity" and asks them to write the key material down — at which point it demonstrably exists somewhere else. The old subtitle was accurate; suggest dropping the appended clause.
- Nit, `Sources/OnymIOS/OnboardingSteps.swift:1547`: the `.pickingAuthority`-gated banner appearing/disappearing mid-step causes a layout jump when terms load; consider a transition or keeping the header slot stable.

The layout changes (welcome headline, two-tone `summaryLine`, banner gating on the right flow state) are otherwise clean, and the PR's stated restraint about the mock's terms screen and relay wording is sound — findings 2 and 3 are the same overpromise pattern slipping in elsewhere.

dev-onym (commented):
REQUEST_CHANGES

- **Destructive restore reachable over live data, with copy claiming otherwise** — `Sources/OnymIOS/OnboardingSteps.swift:341`: the sheet says "Nothing else here is touched," but `IdentityRepository.restore(mnemonic:)` (`Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift:169`) wipes **every** existing identity and fires `identityRemoved`, which subscribers use to wipe identity-scoped chats/messages. Onboarding isn't fresh-install-only: Settings can re-enter it via `requestRestart()` (`Sources/OnymIOS/RootView.swift:269`), so a user with multiple identities and chat history can hit this button and silently lose everything. The code comment's "nothing else on this fresh install worth keeping" assumption is false on that path. Gate the sheet (or switch to a confirm-with-real-consequences flow) when identities/data already exist. Note the PR body's claim that this "reuses the same pattern as Settings' RestoreIdentitySheet" undersells the difference — Settings' sheet is additive (`submitAdd` → `add(mnemonic:)`), this one is the wipe-all call.
- **"Remind me tonight" promises a reminder that doesn't exist** — `Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:160`: skip just records `.skipped`; the only follow-up is the persistent Settings nudge (per `OnboardingFlow.swift:37-39`). No notification is scheduled for "tonight." Revert to "Remind me later" or actually schedule one.
- **"2 relays — primary + backup" contradicts the real seed** — `Sources/OnymIOS/OnboardingSteps.swift:628`: the hardcoded first-launch seed is a single relay (`NostrRelaysConfiguration.swift:33-42`, "Onym Official" only); the published-list fetch is a later, non-guaranteed network refresh. This copies the design mock's illustrative number into a factual claim — exactly the regression the PR body says it avoided elsewhere.
- **Restored 24-word phrase vs "These 12 words" copy** — `OnboardingView.swift:141` (and the step title "Save your recovery phrase"): after a 24-word restore, the recovery step still asserts 12 words, and asks the user to back up a phrase they just typed in. The step isn't `identityOrigin`-aware the way the identity step now is; consider origin-conditional copy or skipping/pre-satisfying the step on `.restored`.

Verified as sound: the bootstrap/restore ordering is safe (actor-serialized; `bootstrap()` only mints when the cache is empty, so either interleaving converges on the restored identity), `ModerationConsentFlow.state.step` has the `.pickingAuthority` case the new banner matches on, and the `restoring:` title override only affects the identity step.

rinat-enikeev:
Addressed in b24b3ff:

- **Destructive restore over live data** — the "I have a recovery phrase" affordance is now hidden whenever `OnboardingLaunch.isExistingUser()` reads true, so it can never appear on a Settings → Restart Onboarding walk (which keeps identity/chats/messages by design). Only reachable on a genuine fresh install, where the sheet's "nothing else is touched" claim is actually accurate.
- **"Remind me tonight"** — reverted to "Remind me later"; no reminder is scheduled.
- **"2 relays · primary + backup"** — reverted; the real first-launch seed (`NostrRelaysConfiguration.seed`) is a single relay, "Onym Official".
- **Recovery-phrase subtitle after a 24-word restore** — now origin-aware, drops the hardcoded "12 words" claim when `identityOrigin == .restored`.

Good catch on the restore-reachability bug — that would've been a real data-loss path.

dev-onym (commented):
APPROVE

- Verified the core claims: `IdentityRepository.restore(mnemonic:)` really is wipe-then-add and throws `IdentityError.invalidMnemonic` (`Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift:164-183`), matching the sheet's catch clauses (`Sources/OnymIOS/OnboardingSteps.swift:390-395`); `bootstrap()` is get-or-create on the actor (`IdentityRepository.swift:69-80`), so either ordering of the launch-time bootstrap and a restore converges on the restored identity — no wipe-over race.
- The destructive-restore gate is sound: `identityRestoreAllowed: !OnboardingLaunch.isExistingUser()` (`Sources/OnymIOS/OnymIOSApp.swift:1027`) reuses the same pure-read probe as the cold-boot gate (`Sources/OnymIOS/OnboardingLaunch.swift:71-82`), so a Settings → Restart walk (identity/chats kept) hides the button.
- Edge case, low severity: on a fresh install, consenting to a moderation mandate mid-walk then killing the app makes `isExistingUser()` true on the resumed walk (mandate store is part of the probe, `OnboardingLaunch.swift:80`), so the restore affordance vanishes on a genuinely fresh install's resume. Hiding is the safe direction, and the user would have to navigate back to welcome to notice — fine to ship, worth a comment near `OnymIOSApp.swift:1027`.
- PR body inaccuracy, not a code issue: the claimed "two-tone value/annotation treatment on the services summary line (was a single concatenated string)" isn't in the diff — `summaryLine(label:value:)` was already two-part before this PR (`151ce4d`'s `OnboardingSteps.swift:552`); the only change there is "relays" → "relay" (`Sources/OnymIOS/OnboardingSteps.swift:642`). Suggest correcting the description.
- Copy nit: `"These 12 words ARE your identity"` (`Packages/OnymOnboarding/Sources/OnymOnboarding/OnboardingView.swift:145`) — all-caps emphasis tends to survive poorly through localization; consider styled emphasis or rewording.
- Nit: title "Restoring your identity" (`OnboardingView.swift:121`) is present-progressive but the restore has already completed by the time the step shows; "Identity restored" would match the checklist's past-tense rows.
- Checked the rest: `.pickingAuthority` exists on `ModerationConsentFlow.Step` (`Packages/OnymModerationUI/.../ModerationConsentFlow.swift:24`); 12/24 word gate matches BIP-39 use here; `identityOrigin` on the flow (not the step view) is consistent with the stated `ServicesSetupChoice` rebuild rationale; new strings follow the file's `LocalizedStringKey`/`String(localized:)` conventions.

PR #274, opened 2026-08-19
Add OnymBackup: the device-backup domain, with no I/O and no crypto

rinat-enikeev:
First of the device-backup stack, and the smallest useful slice: the boundary objects of [`UI-Backup.md`](https://github.com/onymchat/onym-system/blob/main/backup/UI-Backup.md) plus the wire semantics its first implementation profile pins (onymchat/onym-system#37). Nothing here touches the network, the keychain, or a byte of user state, and nothing is reachable from the app yet — the sealing, adapter, and UI PRs that follow need a vocabulary to be written against, and this is it.

### Why this seat exists at all

Shipping the SwiftData stores to a server would leak the social graph: `PersistedGroup.id`, `PersistedMessage.groupID`, `sentAt`, `directionRaw` and `PersistedIntroRequest.id` (a raw Nostr event id) are plaintext columns, so a server holding many users' backups joins on group ID and rebuilds membership. The remedy is one opaque sealed archive rather than column scrubbing — which is what the types in this PR describe.

### The three that carry the weight

**`BackupTerms.regresses(against:)`** turns "terms bind forward only" from a promise into a check. It fails closed on every axis: an unparseable duration counts as a regression, and free-text scope fields count as a regression on *any* change, because nothing here can prove a reworded erasure scope is a widening rather than a narrowing. An operator improving its own wording therefore has to obtain fresh consent — the correct direction to be wrong in.

**`BackupOutcomeStatus`** keeps `submitted`, `accepted`, `retained` and `unknown` as distinct words, and `isRetention` deliberately excludes the first two: neither means the bytes are held. `unknown` is a real terminal state, reconciled by reference rather than promoted to success.

**`BackupError`** passes the operator's own code and message through `.rejected` instead of flattening an unrecognised refusal into a local guess. `exportWithheld` is in the enum so a client can *name* a violation, not because an operator may put us in that state.

`ErasureReceipt.decode` rejects an empty `excludedScope` as malformed — there is always something excluded, at minimum the copies held by everyone else in the conversations a snapshot contained.

### Worth a look in review

- `BackupFormat` duplicates four helpers `OnymFoundation` keeps internal behind `ServiceManifestFormat`. Widening a shared package's API for one consumer seemed worse than thirty lines of hex and digest handling — especially for a seat whose whole discipline is that two sides agree on *bytes* rather than share code. Say if you'd rather promote them.
- `BackupDuration` approximates years and months (365/30 days). Unfit for calendar arithmetic, fit for the one question asked: is this declared window longer than that one. The alternative was refusing to compare `P1M` against `P30D` at all.
- `BackupOperatorManifest` binds to the first `read-write` endpoint and stays bound. Multi-endpoint manifests are not an error but are not exercised.

### Verification

`swift build` clean for the package; `xcodebuild -scheme OnymIOS` **BUILD SUCCEEDED** with it linked into both targets. Tests land in the final PR of the stack, per convention.

### DIFF
--- a/Packages/OnymBackup/Package.swift
+++ b/Packages/OnymBackup/Package.swift
@@ -1,0 +1,21 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
+    name: "OnymBackup",
+    platforms: [.iOS("18.0")],
+    products: [
+        .library(name: "OnymBackup", targets: ["OnymBackup"]),
+    ],
+    dependencies: [
+        .package(path: "../OnymFoundation"),
+    ],
+    targets: [
+        .target(
+            name: "OnymBackup",
+            dependencies: [
+                "OnymFoundation",
+            ]
+        ),
+    ]
+)
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
@@ -36,13 +36,23 @@ public enum BackupError: Error, Equatable, Sendable {
     /// The operator requires a valid entitlement for this operation.
     case paymentRequired(componentId: String, offerIds: [String], entitlementIssuers: [String])
     /// An entitlement was presented and refused. Refresh without leaking
-    /// another credential to this operator.
-    case invalidEntitlement
+    /// another credential to this operator — the issuers it names are the
+    /// ones *this* operator accepts, and they go nowhere else.
+    case invalidEntitlement(entitlementIssuers: [String])
     /// Exceeds the operator's declared maximum sealed snapshot size.
     case snapshotTooLarge(maximumSealedSnapshotBytes: Int?)
     /// Retained count or byte limit reached. The operator reports it and
     /// must not silently drop an older snapshot to make room.
-    case quotaExceeded(retainedSnapshots: Int?, maximumRetainedSnapshots: Int?)
+    ///
+    /// All four figures are the operator's own claims about its own state,
+    /// carried so a person can be shown what is full rather than only that
+    /// something is.
+    case quotaExceeded(
+        retainedSnapshots: Int?,
+        maximumRetainedSnapshots: Int?,
+        retainedBytes: Int?,
+        limitBytes: Int?
+    )
     /// Bytes received do not compose the reference claimed. Never merged,
     /// never partially restored.
     case incompleteSnapshot
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
@@ -49,6 +49,13 @@ public enum BackupError: Error, Equatable, Sendable {
     /// Held once; no longer held.
     case retentionExpired
     /// Erasure acknowledged, completion pending. Not destruction.
+    ///
+    /// Never an operator response: erase always returns a receipt, and a
+    /// receipt is always an acknowledgment. This is *derived* locally by
+    /// comparing the receipt's `completionCommittedBy` against the clock,
+    /// so an operator that misses its own committed deadline is caught
+    /// from the receipt we already hold rather than from a status code it
+    /// would have to volunteer against its own interest.
     case erasureUnconfirmed(receipt: ErasureReceipt?)
     /// The operator withheld export. **Non-conforming** — see the note above.
     case exportWithheld
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
@@ -126,9 +126,18 @@ extension BackupError {
     /// same operation id (`UI-Backup-Object-HTTP.md` §10.2). `termsChanged`
     /// is not, because a retry would upload under terms nobody consented
     /// to.
+    ///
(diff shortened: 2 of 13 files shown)


dev-onym (commented):
**REQUEST_CHANGES**

- **`regresses(against:)` misses `retention.retentionClass` and `retention.snapshotsRetained`** — `Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:220`. A downgrade from `measurable` to `best-effort`, or a cut in snapshots retained, passes the check unnoticed. This contradicts the PR's own claim of failing closed on every axis; per the free-text-scope precedent, any change to these should be a regression. (`lawfulAccess.transparencyReporting` being dropped is also uncaught — lower stakes since it's optional, but same doctrine.)
- **`isLowercaseHex` accepts non-ASCII Unicode digits** — `Packages/OnymBackup/Sources/OnymBackup/BackupFormat.swift:17`. `Character.isNumber` is true for `٣`, `¾`, fullwidth digits, etc., so `isDigest` (line 33) passes a 64-char "digest" of such characters — exactly the "merely looks plausible" value the doc comment says must never reach a comparison. Check ASCII explicitly, e.g. `$0.isHexDigit && $0.isASCII && !$0.isUppercase`.
- **`canonicalTermsId` is not a stable canonical form** — `Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:191`. Round-tripping through `JSONSerialization` normalizes number formatting (`1.0` → `1`), float precision, and string escaping in platform-specific ways, so a conforming operator computing the digest with a real canonicalization (e.g. RFC 8785 JCS) can legitimately disagree with the locally computed `termsId`, and the manifest's `declaredTermsDigest` comparison fails for honest operators. This needs the spec's pinned canonicalization, not `.sortedKeys`.
- **`isRetriable` includes `.accessRefused`** — `Packages/OnymBackup/Sources/OnymBackup/BackupError.swift:121`. The property is documented as "retrying the *same* operation, unchanged, could succeed," but a bad signature retried verbatim fails identically; only stale-timestamp/nonce cases recover, and only with a *fresh* proof. Either narrow the doc or split the case.
- **Malformed sub-processor entries decode to empty strings** — `Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:135`. Non-string `role`/`jurisdiction` coerce to `""` instead of throwing `termsUnavailable`, the one place `decode` fails open.
- **`termsURL` accepts any-length hex** — `Packages/OnymBackup/Sources/OnymBackup/BackupOperatorManifest.swift:103`. Use the 64-char `isDigest` check rather than bare `isLowercaseHex`.
- Minor doc mismatch: `readWriteEndpoint` (`BackupOperatorManifest.swift:107`) skips malformed `read-write` entries and binds a later one, while the property doc says "we bind to the first read-write entry." Say which is intended.

The first three are worth fixing before the sealing/adapter PRs build on this vocabulary; the rest are take-or-leave. Overall structure, error taxonomy, and fail-closed posture are strong.

dev-onym (commented):
APPROVE — boundary types are sound; I verified the `SignedServiceManifest`/`ServiceOffer`/`ServiceManifestCanonical` bindings, platform versions, and the fail-closed logic against the actual code. Findings are non-blocking:

- `Packages/OnymBackup/Sources/OnymBackup/ErasureReceipt.swift:75` — `decode` throws `BackupError.rejected(code: "malformed_receipt")`, but `.rejected` is documented as "the operator's own code, preserved verbatim." A locally minted code under that case is indistinguishable from an operator refusal in logs and UI. A local case (`invalidReference`-style, or `localFailure`) would keep the enum's contract honest.
- `Packages/OnymBackup/Sources/OnymBackup/BackupProfile.swift:35` — `requiredOperations` gates on `upload/list/download/erase/export` only, yet `BackupOutcomeStatus.unknown` is documented as "reconciled by reference through `queryOutcome`." An operator that never implements outcome query passes the capability gate while making `unknown` permanently unreconcilable. If the spec folds query into `upload`, a comment saying so would help.
- `Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:247` — asymmetry worth confirming: *shortening* `retention.maximumRetentionPeriod` passes `regresses` silently, while any change to `snapshotsRetained` regresses. A shortened window also destroys history a person is relying on; if intentional (privacy-first), say so in the comment like the neighboring axes do.
- `Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:176` — `decode` defaults `endOfPayment.duringGrace` to `[]`, though the doc names `download`/`export`/`erase` as the conforming minimum. Terms declaring nothing during grace decode cleanly and nothing downstream checks the minimum.
- `Packages/OnymBackup/Sources/OnymBackup/BackupOutcome.swift:10` — `BackupOutcomeStatus` is `Codable` on its raw string, so an unrecognised wire status throws rather than landing in `unknown`. That's fine for a boundary type, but the adapter PR must map unknown statuses deliberately — worth a note here so it isn't decoded directly off the wire.

Things I checked that hold up: `BackupDuration.seconds` fails closed on non-ASCII digits (`Double.init` rejects them), fractional and malformed durations, and bare `P`; `isLowercaseHex` correctly excludes Unicode digits; `regresses` direction is correct on every duration axis I traced (longer retention/deadlines bad, shorter notice/grace bad); the `ServiceManifestCanonical.signingBytes(of:omitting:)` widening preserves the old entry point's behavior exactly.

PR #275, opened 2026-08-19
Backup sealing: seed-derived keys, Padmé padding, chunked AEAD

rinat-enikeev:
Second of the device-backup stack, on top of #274. Turns a plaintext archive into the exact bytes an operator stores, and back. Implements §5, §7 and §8 of the [object-HTTP profile](https://github.com/onymchat/onym-system/pull/37).

### The seed is the root, and getting it there took a seam

`StorageEncryption`'s key is pinned `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` — right for a stolen phone, disqualifying for a backup. A snapshot sealed under a key that cannot leave the device is unopenable on the replacement device, which is the only device that will ever need it.

So the root is the BIP39 seed. Passing it around would be worse than the problem, so:

- `IdentityRepository.deriveSeedScopedKey(info:)` reconstructs the seed, derives 32 bytes, zeroes it — same posture as `blsSecretKey()`.
- `OnymBackup` declares `BackupSeedDeriving`; the composition root conforms `IdentityRepository` to it (`Sources/OnymIOS/BackupSeedDerivation.swift`).
- `OnymIdentity` never learns about backup, `OnymBackup` never depends on identity, and the only thing crossing is 32 bytes for one `info` the caller named.
- The seed-taking overload of `BackupKeys.material` is **internal**, so production code can't reach for it.

An identity imported from raw key material throws `noRecoveryPhrase` rather than deriving from something else — substituting would produce a key unrecoverable from a phrase, i.e. an archive that can never be opened, discovered only by whoever tried.

`componentId` sits inside the access-key `info`, so two operators see unrelated public keys and neither can be tied to the Nostr, BLS, Stellar or inbox keys. Anti-correlation by derivation rather than by policy.

### Padmé, not a power-of-two ladder

Overhead caps near 12% instead of potentially doubling stored bytes. That matters here specifically: non-convergent keying forbids cross-holder dedup, so storage is strictly linear in holders, and buying a marginally coarser size signal by doubling everyone's bill is the wrong trade.

### Sealing and opening

Per-snapshot random salt → HKDF → per-snapshot key; 1 MiB plaintext chunks under AES-256-GCM with a counter nonce; digest accumulated over what is actually written, streamed throughout so a several-hundred-megabyte snapshot never sits in memory.

Opening **verifies the whole reference before decrypting anything**. Bytes that don't compose their reference never reach a decoder — otherwise a truncated file yields plausible plaintext for whatever prefix survived. Chunk order is also checked against the nonce counter, since per-chunk authentication says nothing about sequence.

### Tests are in this PR, deviating from the stack's convention

The rest of the stack keeps tests for a final PR. Sealing is the part where being wrong is *silent* — it looks fine until the day someone needs it back — so these land with the implementation:

```
testPadmeWorkedExample                      passed   (41,000,000 → 41,943,040, matching the profile §7 example exactly)
testPaddingOverheadStaysUnderTwelvePercent  passed
testSealOpenRoundTrip                       passed
testNonConvergentKeying                     passed
testTamperedChunkFailsAuthentication        passed
testAccessProofIsRequestBound               passed
** TEST SUCCEEDED **
```

`xcodebuild -scheme OnymIOS` also **BUILD SUCCEEDED** with everything linked.

### Worth a look in review

- `BackupArchiveWriter` writes records to a scratch file, then copies them in behind the finished header — the header names every entry and its digest, so it can't be written until the entries are known, and buffering an archive that may be hundreds of megabytes isn't an option. One extra pass over the bytes.
- Per-entry SHA-256 in the archive header is redundant against the AEAD on purpose: it lets a restore say *which* record failed rather than only that something did.
- `BackupSeedDeriving` is `async` because the seed's holder is an actor; a synchronous requirement would force either a blocking hop or a nonisolated accessor to secret state.
- Record structs (`BackupGroupRecord` and friends) are deliberately **not** here — they belong with the source/sink that produces them, in the next PR.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupAccessProof.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupAccessProof.swift
@@ -1,0 +1,97 @@
+import CryptoKit
+import Foundation
+import OnymFoundation
+
+/// The request-bound proof of possession every `/v1/` call carries
+/// (`UI-Backup-Object-HTTP.md` §8).
+///
+/// A holder *is* an Ed25519 public key at the operator — no account, no
+/// email, no support identifier, no password. That is what makes the
+/// absent reset path checkable rather than promised: there is nothing
+/// for an operator to reset.
+///
+/// The signature covers the method, the path, and a digest of the body,
+/// so a chunk upload cannot be replayed into a different index, a
+/// different upload, or a different operation. Each field is length-
+/// prefixed because a signature over bare concatenation can be
+/// reinterpreted by shifting a boundary between two adjacent fields that
+/// an attacker influences.
+public enum BackupAccessProof {
+    /// Domain separator, first field of every signed payload.
+    static let context = "onym-backup-v1"
+
+    public static let holderHeader = "X-Onym-Holder"
+    public static let timestampHeader = "X-Onym-Timestamp"
+    public static let nonceHeader = "X-Onym-Nonce"
+    public static let signatureHeader = "X-Onym-Signature"
+
+    /// The exact bytes signed, in the pinned order.
+    ///
+    /// The operator rebuilds this from the request it actually received
+    /// — never from a client-supplied copy — so anything we get wrong
+    /// here surfaces as a refusal rather than as a subtly weaker check.
+    public static func signingBytes(
+        method: String,
+        path: String,
+        holderReference: String,
+        timestamp: Date,
+        nonceHex: String,
+        body: Data
+    ) -> Data {
+        var out = Data()
+        appendLengthPrefixed(Data(context.utf8), to: &out)
+        appendLengthPrefixed(Data(method.uppercased().utf8), to: &out)
+        appendLengthPrefixed(Data(path.utf8), to: &out)
+        appendLengthPrefixed(Data(holderReference.utf8), to: &out)
+        appendLengthPrefixed(Data(BackupFormat.string(fromDate: timestamp).utf8), to: &out)
+        appendLengthPrefixed(Data(nonceHex.utf8), to: &out)
+        // The raw 32 digest bytes, not their hex — a body-less request
+        // uses the digest of the empty string rather than an empty field,
+        // so "no body" and "empty body" are the same signed value and
+        // neither is a shorter payload an attacker could exploit.
+        appendLengthPrefixed(Data(SHA256.hash(data: body)), to: &out)
+        return out
+    }
+
+    /// Sign a request. Returns the header set to send with it.
+    public static func headers(
+        method: String,
+        path: String,
+        body: Data,
+        material: BackupKeyMaterial,
+        now: Date = Date(),
+        nonce: Data? = nil
+    ) throws -> [String: String] {
+        // Every CSPRNG call in this codebase goes through SecureRandom,
+        // which throws rather than handing back a zero buffer. A
+        // predictable nonce is the one thing this proof cannot survive.
+        let nonceBytes = try nonce ?? SecureRandom.data(16)
+        let nonceHex = nonceBytes.map { String(format: "%02x", $0) }.joined()
+        // Second precision, no fractional seconds: the operator's
+        // reconstruction has to produce the same string we signed, and a
+        // formatter that emits milliseconds on one platform and not
+        // another would fail intermittently and only for some clocks.
+        let timestamp = Date(timeIntervalSince1970: now.timeIntervalSince1970.rounded(.down))
+        let bytes = signingBytes(
+            method: method,
+            path: path,
+            holderReference: material.holderReference,
+            timestamp: timestamp,
+            nonceHex: nonceHex,
+            body: body
+        )
+        let signature = try material.accessSigningKey.signature(for: bytes)
+        return [
+            holderHeader: material.holderReference,
+            timestampHeader: BackupFormat.string(fromDate: timestamp),
+            nonceHeader: nonceHex,
+            signatureHeader: signature.base64EncodedString(),
+        ]
+    }
+
+    private static func appendLengthPrefixed(_ field: Data, to out: inout Data) {
+        var length = UInt32(field.count).bigEndian
+        withUnsafeBytes(of: &length) { out.append(contentsOf: $0) }
+        out.append(field)
+    }
+}
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupAccessProof.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupAccessProof.swift
@@ -54,13 +54,28 @@ public enum BackupAccessProof {
     }
 
     /// Sign a request. Returns the header set to send with it.
+    ///
+    /// The nonce is not injectable from outside the package for the same
+    /// reason the snapshot salt is not: a reused nonce turns a
+    /// single-use proof into a replayable one. Fixtures that need a
+    /// fixed value use the internal overload.
     public static func headers(
         method: String,
         path: String,
         body: Data,
         material: BackupKeyMaterial,
+        now: Date = Date()
+    ) throws -> [String: String] {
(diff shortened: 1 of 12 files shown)


dev-onym (commented):
**REQUEST_CHANGES**

- **Seed-derivation oracle** — `IdentityRepository.deriveSeedScopedKey` (`Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift:277`) accepts an arbitrary caller-supplied `info` under the same salt (`app.onym.bip39`) as every other seed key. Calling it with `"nostr-secp256k1-v1"` or `"bls12-381-v1"` returns byte-for-byte the Nostr/BLS secret keys (`Bip39.swift:2200,2211`), so the "no path back to anything else" claim in the doc comment doesn't hold. Require/enforce a `backup-` info prefix (or an allowlist) inside the actor.
- **AES-GCM key+nonce reuse footgun** — `BackupSealer.seal`'s public `snapshotSalt: Data? = nil` parameter (`BackupSealer.swift:37`) lets a caller reuse a salt across two different plaintexts under the same root: same key, same counter nonces → full GCM confidentiality/integrity collapse. Nothing in-repo needs it (tests pass `nil`); make it internal or drop it. Same for the injectable `nonce` on `BackupAccessProof.headers` (`BackupAccessProof.swift:63`), though signature nonce reuse there is only replay-relevant.
- **Force-try on external conformer output** — `BackupKeys.material(deriving:)` (`BackupSeedDeriving.swift:48-50`) does `try!` on bytes returned by any `BackupSeedDeriving` conformer. The 32-byte contract is only documentation; a wrong-length return crashes the app instead of throwing. Guard `count == 32` and throw. (The mirrored `try!` in `BackupKeys.material(seed:)` at `BackupKeys.swift:117-119` is fine — HKDF output length is fixed there.)
- **Record kind never checked against header** — `BackupArchiveReader.forEachRecord` (`BackupArchive.swift:200-205`) validates `length` and digest against the header entry but trusts the framing's kind byte, ignoring `entry.kind`. Both `sha256` and the AEAD cover only the record bytes, so header and framing can disagree about kind; the header (the thing restores report from) should be authoritative — add `framing[framing.startIndex] == entry.kind`.
- **Unbounded chunk-size allocation** — `BackupOpener.open` (`BackupOpener.swift:69,84`) trusts the header's `chunkPlainBytes` (up to ~4 GiB) for its per-chunk read. Digest-verify-first limits this to a self-inflicted case, but capping it at `BackupSealer.chunkPlainBytes` is one guard.
- Minor: `BackupArchiveWriter` leaks the plaintext scratch file if `finish` is never reached (`BackupArchive.swift:104` — only `finish` removes it); consider cleanup in `deinit`. Double-`finish` throwing `.archiveUnreadable` (`BackupArchive.swift:117`) is a misleading error for a programmer bug.

The core crypto is right: Padmé matches the §7 worked example, verify-whole-digest-before-decrypt is correctly ordered, the counter-nonce ordering check is sound, and the seam keeping the seed inside the actor is well designed. The first two bullets are each a one-line fix but sit exactly where this PR says being wrong is silent.

rinat-enikeev:
All six in b40055b. Two were real, and on the first my doc comment asserted the opposite of the truth.

**Seed oracle.** Verified before fixing: `Bip39.deriveNostrKey` and `deriveBlsKey` use salt `app.onym.bip39` — the same one — so `deriveSeedScopedKey(info: "nostr-secp256k1-v1")` returned the Nostr secret key byte for byte. The comment claiming a caller "has no path back to anything else" was describing an intention, not the code. Now guarded by an **allowlist** of `info` prefixes checked before anything is read, with `seedScopeNotPermitted` as its own error. Allowlist rather than denylist deliberately: a denylist would let a future `Bip39` derivation become reachable the moment someone adds it, whereas an addition to the allowlist has a reviewer attached.

**Pinned salt.** Right, and worse than replay: same salt under the same archive root reproduces the same key, and the nonces are counters, so two different plaintexts get sealed under an identical (key, nonce) pair — GCM confidentiality *and* integrity gone. The public entry point now always draws from the CSPRNG; the pinning overload is internal for fixtures. Same treatment for the proof nonce, where the consequence is milder but the argument is the same.

**Force-try on conformer output.** Fixed as `keyMaterialInvalid`. You're right that the 32-byte contract is documentation and a conformer is code we don't own — turning someone else's bug into our crash isn't a trade worth making. (Agreed the mirrored `try!` in `material(seed:)` is fine; HKDF's output length is ours to guarantee.)

**Record kind.** Good catch, and the framing of it is what convinced me: neither the digest nor the AEAD covers that byte, so header and framing could disagree about what a record *is* while both verify. The header is what a restore reports from, so the header wins now.

**Chunk cap** and the two minors are in as described. The scratch file matters more than it looks — it holds *plaintext* archive records, so `deinit` cleanup closes a case where a composer that threw partway left decrypted state on disk. `finish` twice now reports `archiveWriterFinished` rather than masquerading as a corrupt archive.

Two new tests cover the fixes that would otherwise be silent — the pinned-salt shape that made key reuse possible, and a kind byte flipped under a still-valid digest:

```
testAccessProofIsRequestBound                  passed
testArchiveRoundTripAndKindMismatchRefused     passed
testNonConvergentKeying                        passed
testPaddingOverheadStaysUnderTwelvePercent     passed
testPadmeWorkedExample                         passed
testPinnedSaltWouldReuseKeyAndNonce            passed
testSealOpenRoundTrip                          passed
testTamperedChunkFailsAuthentication           passed
** TEST SUCCEEDED **
```

`xcodebuild -scheme OnymIOS` still **BUILD SUCCEEDED**.

dev-onym (commented):
COMMENT — sound design and crypto; a few non-blocking gaps, one of which contradicts the PR's own "silent failure" bar.

- `Packages/OnymBackup/Sources/OnymBackup/BackupSealer.swift:107` — `guard !plain.isEmpty else { break }` lets a seal "succeed" with fewer chunks than the header's `chunkCount` if the source file shrinks between `fileSize(of:)` and the read loop (or a short read occurs). The digest covers the bytes actually written, so `verify` passes and the failure surfaces only at restore time — exactly the silent-until-needed case this PR exists to prevent. After the loop, assert total plaintext consumed == `paddedSize` and throw at seal time instead of breaking.
- `Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift:174` — `headerLength` is an unbounded attacker-shaped u32 driving `read(upToCount: headerLength)`, up to a 4 GiB allocation attempt. `BackupOpener.open` guards its header fields for precisely this reason ("corrupted-then-resealed file… should not turn a u32 into a 4 GiB allocation"); the archive reader deserves the same cap.
- `Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift:89` — `append` doesn't check `finished`, so append-after-finish throws a generic closed-FileHandle error rather than the `.archiveWriterFinished` case whose doc comment claims to cover "written to after finishing." One-line guard.
- `Packages/OnymBackup/Sources/OnymBackup/BackupOpener.swift:47` — `verify` closes the file and `open` reopens it: a TOCTOU window where the bytes decrypted aren't the bytes verified. Per-chunk AEAD bounds the damage (a swap fails authentication), so this is defense-in-depth only; a comment noting the reliance on the AEAD would do.
- Test gaps: nothing exercises the chunk-reorder → nonce-counter refusal (`BackupOpener.swift:99-104`), wrong-`archiveRoot` open, or a `chunkCount` that overstates the file. The reorder check is the one uncovered branch in the opener.

Notes verified while reviewing: the Padmé math matches §7 (E=25, S=5, step 1 MiB → 41,943,040); the seed-oracle allowlist in `IdentityRepository.deriveSeedScopedKey` correctly blocks the shared-salt `nostr-*`/`bls12-*` infos; counter nonces are safe given the per-snapshot HKDF salt, and refusing salt injection at the public API is the right call.

None of these block merge; the first bullet is the one I'd want fixed before the source/sink PR builds on this.

PR #276, opened 2026-08-19
Backup composition and restore, over a source/sink port

rinat-enikeev:
Third of the device-backup stack, on top of #275. Reads local state through a port, writes a versioned archive, seals it — and on the way back, verifies, decodes, then writes through the app's stores so the restoring device re-encrypts everything under *its* at-rest key.

### Eligibility is structural, not remembered

`BackupComposer` holds a `BackupSourceProviding` and nothing else — no `IdentityRepository`, no keychain, no `StorageEncryption`. "Seed material is never eligible" ([UI-Backup.md §16.1](https://github.com/onymchat/onym-system/blob/main/backup/UI-Backup.md)) is therefore a property of the type graph rather than a discipline someone has to keep.

The test that matters most here opens a composed snapshot back to plaintext and asserts the seed, its mnemonic words, the access signing key and the archive root are all absent — with a positive assertion alongside, so it can't pass on an empty file.

### Media never comes from the device's caches

`ChatImageLoader` and friends write **decrypted** bytes into `Caches/Onym*` under filenames that are public Blossom content addresses. Copying those into a snapshot would hand an operator plaintext media keyed by an identifier anyone can look up — the single worst thing that could go in. Under `.includeCiphertext` the bytes come from the blob store and are verified against the address they claim before inclusion, and again on restore.

### Restore has three gates, in this order

1. Verify the whole reference — bytes that don't compose their digest never reach a decoder.
2. Decode and validate the *entire* archive.
3. Only then write, groups before messages.

Holding a decoded archive in memory is the cost. A half-restored device is the alternative, and a chat list with some groups and none of their messages looks like data loss because it is.

### Two judgement calls worth review

- **A missing blob is not a failure.** Blob retention is the blob operator's own contract and shorter than a backup's by design, so the descriptor still travels and `BackupRestoreSummary.unresolvedBlobs` reports what couldn't be resolved — rather than reporting plain success while the person's media is partly gone.
- **Restored consents land inactive.** Re-activating one would silently re-select an operator the person hasn't seen since the snapshot. They're also merged rather than replacing existing consents, so a choice made *after* the backup isn't revoked by restoring it.

### A bug the tests caught

Setting file protection via `setResourceValue` after creation throws where data protection is unavailable — every one of these tests failed with a misleading "couldn't be removed". It's now passed as a creation attribute, honoured on device and ignored where it can't apply. This would have failed on the simulator and on device-less CI only, so it was worth finding here.

### Verification

```
BackupCompositionTests   4 passed  (seed-absence, round trip, plaintext cleanup, blob address check)
BackupSealingTests       8 passed
** TEST SUCCEEDED **
```

### Still to come in the stack

The adapter and repository (`URLSessionBackupClient`, the payment-retry loop), then `OnymBilling`, then the UI. Nothing here is reachable from the app yet.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift
@@ -40,6 +40,14 @@ public struct BackupArchiveHeader: Sendable, Equatable, Codable {
     /// The archive's true length before padding. The padding is inside
     /// the seal, so this is the only place the real size is recorded.
     public let contentByteSize: Int
+    /// What this snapshot set out to carry for media.
+    ///
+    /// Recorded because a restore cannot otherwise tell "this archive
+    /// deliberately carries descriptors only" from "this archive should
+    /// have carried blobs and some are missing" — and reporting every
+    /// attachment of a descriptors-only snapshot as unresolved would be
+    /// alarming and wrong.
+    public let mediaPolicy: BackupMediaPolicy
     public let entries: [BackupArchiveEntry]
 }
 
@@ -68,6 +76,7 @@ public final class BackupArchiveWriter {
     private let handle: FileHandle
     private var entries: [BackupArchiveEntry] = []
     private var identityCount = 0
+    private var mediaPolicy: BackupMediaPolicy = .descriptorsOnly
     private var finished = false
 
     public init(scratchURL: URL) throws {
@@ -82,6 +91,10 @@ public final class BackupArchiveWriter {
 
     public func setIdentityCount(_ count: Int) {
         identityCount = count
+    }
+
+    public func setMediaPolicy(_ policy: BackupMediaPolicy) {
+        mediaPolicy = policy
     }
 
     /// Append one record. `bytes` is whatever the composer serialized
@@ -123,6 +136,7 @@ public final class BackupArchiveWriter {
             createdAt: createdAt,
             identityCount: identityCount,
             contentByteSize: recordBytes,
+            mediaPolicy: mediaPolicy,
             entries: entries
         )
         let encoder = JSONEncoder()
@@ -205,7 +219,16 @@ public struct BackupArchiveReader {
         defer { try? handle.close() }
         try handle.seek(toOffset: UInt64(recordsOffset))
 
+        var seenKinds: Set<UInt8> = []
         for entry in header.entries {
+            // Two records of one kind would be decoded in order and the
+            // later would win silently, so an archive that names groups
+            // twice could quietly discard half of them. Nothing writes
+            // that shape, which is exactly why it should be refused
+            // rather than tolerated.
+            guard seenKinds.insert(entry.kind).inserted else {
+                throw BackupError.incompleteSnapshot
+            }
             guard
                 let framing = try handle.read(upToCount: 5), framing.count == 5,
                 let kind = BackupArchiveEntryKind(rawValue: framing[framing.startIndex]),
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupArchive.swift
@@ -72,10 +72,10 @@ public final class BackupArchiveWriter {
 
     public init(scratchURL: URL) throws {
         self.scratchURL = scratchURL
-        FileManager.default.createFile(atPath: scratchURL.path, contents: nil)
-        try (scratchURL as NSURL).setResourceValue(
-            URLFileProtection.complete,
-            forKey: .fileProtectionKey
+        FileManager.default.createFile(
+            atPath: scratchURL.path,
+            contents: nil,
+            attributes: [.protectionKey: FileProtectionType.complete]
         )
         self.handle = try FileHandle(forWritingTo: scratchURL)
     }
@@ -130,8 +130,11 @@ public final class BackupArchiveWriter {
         encoder.dateEncodingStrategy = .iso8601
         let headerJSON = try encoder.encode(header)
 
-        FileManager.default.createFile(atPath: url.path, contents: nil)
-        try (url as NSURL).setResourceValue(URLFileProtection.complete, forKey: .fileProtectionKey)
+        FileManager.default.createFile(
+            atPath: url.path,
+            contents: nil,
+            attributes: [.protectionKey: FileProtectionType.complete]
+        )
         let output = try FileHandle(forWritingTo: url)
         defer { try? output.close() }
 
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupArchiveRecords.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupArchiveRecords.swift
@@ -1,0 +1,196 @@
+import Foundation
+
+/// The archive's wire schema.
+///
+/// These are **not** the app's domain models, and that separation is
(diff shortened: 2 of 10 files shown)


dev-onym (commented):
COMMENT

- **Consent merge can wipe the store on a failed load** — `Sources/OnymIOS/BackupSinks.swift:143`: `(try? consentStore.load()) ?? []` treats a throwing load (corrupt/transient) as an empty store, then `save(existing + additions)` at line 147 replaces the file with only the restored, inactive records — silently revoking every live consent. A failed load should abort the consent merge, not masquerade as "no existing consents". This contradicts the PR's own "merged, not replaced" guarantee.
- **`unresolvedBlobs` misreports under `.descriptorsOnly`** — `Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift:107`: the summary is `referenced − carried`, but a descriptors-only snapshot carries nothing, so *every* attachment is reported "unresolved" even though the blob store may still serve all of them. The doc comment at `BackupSource.swift:81` ("which the blob store no longer serves") claims a check the restorer never makes. Either thread the media policy through or rename/re-document as `blobsNotCarried`.
- **Summary counts can overstate what was written** — `BackupRestorer.swift:100` reports archive counts, but `AppBackupSink` silently `continue`s past rows whose shape it can't reconstruct (`BackupSinks.swift` group/message/invitation loops). A summary of "5 groups restored" when the sink wrote 3 is the same lie the PR body argues against. Have the sink report skips, or return written counts.
- **Network errors read as retention lapse** — `Sources/OnymIOS/BackupSources.swift:122`: `try? await blobClient.download` folds transient failures into "operator no longer serves it", so a flaky connection yields a media-thin `.includeCiphertext` snapshot with no compose-time signal. Consider distinguishing not-found from transport error, or surfacing a compose-side unresolved list.
- **Mnemonic assertions are vacuous** — `Tests/OnymIOSTests/BackupCompositionTests.swift:20,41`: the mnemonic string is never fed to the composer (the seed is `0x11 × 64`, unrelated), so those words could never appear; and the `… && word.count > 5` conjunction means six of the twelve words pass unconditionally. The seed/key-absence assertions are the real guard and they're solid — either derive the seed from that mnemonic or drop the word loop.
- Minor: a throw from `BackupSealer.seal` leaves a partial `pending-<id>` file behind — `BackupComposer.swift:92` isn't covered by the `defer` (ciphertext only, but it accumulates). Duplicate record kinds in an archive silently last-win in `BackupRestorer.swift:64`.

Solid structure otherwise: the source/sink port isolation, the three-gate restore ordering, the blob-address verification on both ends, and the creation-time file-protection fix all check out against the codebase (attachment schemas do expose `sha256`/`encKey`, including nested poster frames, which the recursive walk handles).

rinat-enikeev:
All six in c6cadb4. Two were the exact failure this PR's description argues against, committed by the code meant to prevent it.

**Consent merge.** You're right and it's the worst one here: `(try? load()) ?? []` turned a corrupt or transient read into "no existing consents", and the very next line saved — replacing every live consent with the restored, inactive ones. Silent revocation, performed by the merge that exists to prevent silent revocation. A failed load now aborts. The asymmetry decides it: losing the *restored* consents is recoverable, because the person is re-asked; losing the *live* ones is not.

**`unresolvedBlobs` under `.descriptorsOnly`.** Also right, and it would have been actively alarming — telling someone their media was gone while the blob store still held all of it. Rather than rename the field, I threaded the intent through: the archive header now records the media policy the snapshot was composed under, so a restore can distinguish "deliberately descriptors-only" from "meant to carry blobs and some are missing". That seemed better than a name that describes the bug accurately.

**Summary counts.** Fixed at the source rather than papered over — `BackupSinkProviding` methods now return what they *wrote*, and the shortfall surfaces as `skipped` by kind instead of being absorbed. A test drives a sink that refuses a row and asserts the summary reports 1 written, 1 skipped.

**Transport vs. retention.** Good catch. `blobCiphertext` now answers with a `BackupBlobAvailability` enum: `404`/`410` is `.gone` and expected; a dropped connection or 5xx throws, so the snapshot fails and is retried rather than silently shipping half the media and calling itself complete.

**Vacuous mnemonic assertions.** Embarrassing, and doubly so — the mnemonic was never fed to the composer *and* the `word.count > 5` conjunction let six of the twelve words pass regardless. The seed is now derived from that mnemonic via `Bip39.seedFromMnemonic`, and every word is checked unconditionally. Agreed the seed/key-absence assertions were the real guard; they now have company that can actually fail.

**Minors.** A throwing seal no longer leaves a partial `pending-<id>` behind. Duplicate record kinds are refused rather than last-winning — nothing writes that shape, which is exactly why tolerating it would mean quietly discarding half a person's groups.

```
BackupCompositionTests   6 passed  (+2: descriptors-only reporting, sink-written counts)
BackupSealingTests       8 passed
** TEST SUCCEEDED **
```

dev-onym (commented):
APPROVE

Verified against the working tree: all referenced symbols exist (`BackupError.incompleteSnapshot`, `BackupKeys.material`, `Bip39.seedFromMnemonic`, `BlossomError.badStatus`), store protocol signatures match the sink/source call sites exactly (`MessageStore.list(groupID:ownerIDString:)`, `PinnedConsentStore.load() throws`, `ChatGroup`/`ChatMessage` inits), `entry.kind` really is `UInt8` so the duplicate-kind guard is sound, and the sink/source are unwired as claimed. The three-gate restore ordering, the consent merge-not-replace with throwing `load()`, and the seed-absence test with its positive control all hold up. No Swift toolchain in this environment, so I could not re-run the tests the PR reports passing.

Non-blocking findings:

- `Sources/OnymIOS/BackupSources.swift:117` — `guard let blobClient else { return .gone }` folds a mis-wired composition root into "store doesn't hold it." Under `.includeCiphertext` with a nil client, every blob silently drops and the snapshot reports complete — the exact silent-media-thin outcome the comment directly below forbids for transport errors. Consider throwing (or `preconditionFailure`) when the policy is `.includeCiphertext`, or make the client non-optional on that path.
- `Sources/OnymIOS/BackupSinks.swift:166-176` — `restore(consents:)` returns `additions.count`, so consents that already exist locally count toward `held > written` (`BackupRestorer.swift:108`) and land in `summary.skipped`, which `BackupRestoreSummary` documents as "this build does not understand part of the snapshot." A second restore, or restoring onto a device that kept its consents, false-alarms. Return `restored.count` (rows reconstructed) rather than rows newly added, or track dedup separately.
- `Packages/OnymBackup/Sources/OnymBackup/BackupSource.swift:56-60` — the "Writes go through the app's stores…" paragraph belongs to `BackupSinkProviding` but is glued onto `BackupBlobAvailability`'s doc comment, so it renders as that enum's documentation.
- `Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift:155` — a blob whose bytes fail their own address check is silently `continue`d, indistinguishable from `.gone` until restore-time `unresolvedBlobs`. A store serving wrong bytes for a content address is an integrity signal worth at least distinguishing in the composer's result, even if the snapshot proceeds.
- `Sources/OnymIOS/BackupSources.swift:42` — `identityCount` counts distinct group owners only; an identity with invitations but no groups isn't counted. Cosmetic given the header field is display-only, but the doc says "identities this snapshot covers."

The judgement calls flagged in the PR body (missing blob ≠ failure; restored consents land inactive and merge) are both implemented as described and are the right calls.

PR #277, opened 2026-08-20
Backup adapter and repository: the wire, and what it refuses to believe

rinat-enikeev:
Fourth of the device-backup stack, on top of #276. `URLSessionBackupClient` speaks the [object-HTTP profile](https://github.com/onymchat/onym-system/blob/main/backup/UI-Backup-Object-HTTP.md); `BackupRepository` decides what to do with the answers.

Most of this is refusals, which is the shape of the seat: the operator sits outside the trust boundary, so what the client declines to believe matters more than what it accepts.

### What the client refuses

- **Redirects, always.** A signed request covers one path at one origin; an operator that could redirect would be choosing where our credentials go.
- **Oversized bodies.** The only soft bound on a list response is `maximumRetainedSnapshots` — the same operator's figure, with no signature behind it — so the cap is enforced on our side (§13). Sealed streams are bounded by the reference's own byte count, so an over-long stream is cut off rather than written to a device that may not have the room and only then failing a digest check.
- **Unrecognised status words.** They become `unknown`, never `retained`. A future operator inventing a word must not read as success to a client that has never heard of it.
- **Unmodelled error codes**, which are preserved verbatim rather than flattened into a local guess.

The manifest is deliberately **not** fetched or verified here — manifest review already exists in the consent flow, and a second implementation of one trust decision is how the two drift apart. The client is handed the manifest the person consented to and binds to its read-write origin.

### Three edges in the repository

**The payment retry sends the same bytes.** A refusal keeps the sealed snapshot and hands it back; the retry reuses the operation id and digest. Re-composing would mint a new salt, defeat `already_retained`, and charge someone for a second copy of the same history. The test asserts both preflights carry one operation id.

**The pending-operation record is written before the upload**, not after. A record written only on success is exactly the record that's missing when a response is lost. A lost response resolves to neither success nor failure — it stays pending until `queryOutcome` answers, and reconciliation runs *before* composing, since stacking a second snapshot on an unresolved first is how someone pays to store the same history twice.

**Changed terms stop the upload and don't re-pin.** Fresh consent is a screen, not an inference.

### Worth a look in review

- `BackupStateStore` throws on a corrupt state file rather than returning an empty state — reading "no backup configured" from corruption would drop the pinned terms and the pending list, and the next upload would go out under terms nobody re-consented to.
- Terms are verified twice over: the served document's computed `termsId` must equal the manifest's `declaredTermsDigest`, *and* its signature must verify against the manifest's operator key.
- Three of these tests initially failed because my fixture fabricated a `termsId` while the client computes it from the document. That was the pinning working, and the fixtures now derive the real digest.

### Verification

```
BackupAdapterTests      10 passed
** TEST SUCCEEDED **
```

The DEBUG `--backup-base-url` override lands with the wiring PR — nothing constructs this client from the app yet, so a resolver here would be dead code.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
@@ -17,7 +17,12 @@ import OnymFoundation
 public actor BackupComposer {
     private let source: any BackupSourceProviding
     private let mediaPolicy: BackupMediaPolicy
-    private let workingDirectory: URL
+    /// Where sealed snapshots and scratch files live.
+    ///
+    /// Immutable and `Sendable`, so callers outside the actor can read
+    /// it without a hop — the repository needs it to clean up bytes for
+    /// an operation the composer knows nothing about.
+    public let workingDirectory: URL
 
     public init(
         source: any BackupSourceProviding,
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
@@ -98,6 +98,14 @@ public enum BackupError: Error, Equatable, Sendable {
         /// finishing. A programming error, named so it does not
         /// masquerade as a corrupt archive.
         case archiveWriterFinished
+        /// The local backup state exists but could not be read or
+        /// decoded. Never treated as "no backup configured": the next
+        /// save would overwrite the pinned terms and the pending
+        /// operations with an empty file.
+        case stateUnreadable
+        /// The operator's upload grant does not describe the snapshot we
+        /// are holding. Refused before any byte is sent.
+        case invalidUploadGrant
     }
 }
 
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -63,9 +63,19 @@ public actor BackupRepository {
         let snapshot = try await composer.compose(
             keyMaterial: keyMaterial,
             acceptedTermsId: acceptedTermsId,
-            supersedes: state.snapshots.last.map {
-                try? SnapshotReference(digest: $0.digest, sealedByteSize: $0.sealedByteSize)
-            } ?? nil,
+            supersedes: try state.snapshots.last.map {
+                // `try?` here would turn an unreadable ancestor into a
+                // snapshot claiming to supersede nothing, quietly
+                // breaking the chain. The store refuses to *read*
+                // corrupt state for the same reason; it should not be
+                // written past here either.
+                do {
+                    return try SnapshotReference(
+                        digest: $0.digest, sealedByteSize: $0.sealedByteSize)
+                } catch {
+                    throw BackupError.localFailure(reason: .stateUnreadable)
+                }
+            },
             now: now
         )
         state.lastAttemptAt = now
@@ -88,7 +98,7 @@ public actor BackupRepository {
             snapshotReference: try SnapshotReference(
                 digest: pending.digest, sealedByteSize: pending.sealedByteSize),
             sealedBytesURL: url,
-            sealedAt: pending.refusedAt,
+            sealedAt: pending.sealedAt,
             acceptedTermsId: pending.acceptedTermsId,
             supersedes: try pending.supersedesDigest.flatMap { digest in
                 try pending.supersedesByteSize.map {
@@ -135,6 +145,7 @@ public actor BackupRepository {
                     acceptedTermsId: snapshot.acceptedTermsId,
                     supersedesDigest: snapshot.supersedes?.digest,
                     supersedesByteSize: snapshot.supersedes?.sealedByteSize,
+                    sealedAt: snapshot.sealedAt,
                     refusedAt: now
                 )
                 try stateStore.save(state)
@@ -185,6 +196,7 @@ public actor BackupRepository {
                 BackupState.PendingOperation(
                     operationId: snapshot.operationId,
                     digest: snapshot.snapshotReference.digest,
+                    sealedByteSize: snapshot.snapshotReference.sealedByteSize,
                     startedAt: now,
                     acceptedTermsId: snapshot.acceptedTermsId
                 )
@@ -257,12 +269,19 @@ public actor BackupRepository {
         var stillPending: [BackupState.PendingOperation] = []
         for pending in state.pendingOperations {
             if let row = retainedByDigest[pending.digest] {
+                // Our own pin, not the operator's echo. Which terms a
+                // stored snapshot was accepted under is not a fact a
+                // counterparty gets to restate — and an empty pin is not
+                // something this type writes.
+                guard let pinnedTerms = pending.acceptedTermsId ?? state.acceptedTermsId,
+                      !pinnedTerms.isEmpty
+                else {
+                    stillPending.append(pending)
+                    continue
+                }
                 state.record(
                     reference: row.snapshotReference,
-                    // Our own pin, not the operator's echo. Which terms a
-                    // stored snapshot was accepted under is not a fact a
-                    // counterparty gets to restate.
-                    acceptedTermsId: pending.acceptedTermsId ?? state.acceptedTermsId ?? "",
+                    acceptedTermsId: pinnedTerms,
                     at: pending.startedAt,
                     status: .retained
                 )
@@ -290,13 +309,31 @@ public actor BackupRepository {
             switch outcome.status {
             case .retained, .alreadyRetained:
                 // Not in the list but claimed retained. Believe the
-                // digest we sent, never the one echoed back.
+                // digest and the size *we* recorded, never the ones
+                // echoed back — and if we cannot rebuild the reference
+                // from our own record, keep the operation pending rather
+                // than throwing.
+                //
+                // Throwing here wedged everything: reconcile ran before
+                // composing, so a single unbuildable reference aborted
+                // `backUp` before it could save, leaving the operation
+                // unresolved and every subsequent run to hit the same
+                // line. A terse — or hostile — operator could brick
+                // backups permanently with a *success* answer.
+                guard
+                    let size = pending.sealedByteSize ?? outcome.snapshotReference?.sealedByteSize,
+                    let reference = try? SnapshotReference(digest: pending.digest, sealedByteSize: size),
+                    let pinnedTerms = pending.acceptedTermsId ?? state.acceptedTermsId,
+                    !pinnedTerms.isEmpty
+                else {
+                    // An empty terms pin is exactly what the rest of this
+                    // type refuses to write.
+                    stillPending.append(pending)
+                    continue
+                }
                 state.record(
-                    reference: try SnapshotReference(
-                        digest: pending.digest,
-                        sealedByteSize: outcome.snapshotReference?.sealedByteSize ?? 0
-                    ),
-                    acceptedTermsId: pending.acceptedTermsId ?? state.acceptedTermsId ?? "",
+                    reference: reference,
+                    acceptedTermsId: pinnedTerms,
                     at: pending.startedAt,
                     status: .retained
                 )
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -1,0 +1,204 @@
+import Foundation
+
+/// Drives one backup: compose, preflight, upload, record.
+///
+/// The state machine is small and its edges are the interesting part —
+/// what it refuses to do is most of the design.
+public actor BackupRepository {
+    private let port: any BackupPort
+    private let composer: BackupComposer
+    private let stateStore: any BackupStateStoring
+    private let keyMaterial: BackupKeyMaterial
+
+    public init(
+        port: any BackupPort,
+        composer: BackupComposer,
+        stateStore: any BackupStateStoring,
+        keyMaterial: BackupKeyMaterial
+    ) {
+        self.port = port
+        self.composer = composer
+        self.stateStore = stateStore
+        self.keyMaterial = keyMaterial
+    }
+
+    /// What a caller sees after asking for a backup.
+    public enum RunResult: Sendable, Equatable {
+        case retained(SnapshotReference)
+        case alreadyRetained(SnapshotReference)
+        /// The operator wants payment. The sealed bytes are kept so the
+        /// retry after purchase sends the *same* snapshot.
+        case paymentRequired(componentId: String, offerIds: [String], pending: SealedSnapshot)
+        /// Terms moved. Uploading would pin something nobody agreed to.
+        case termsChanged(currentTermsId: String?)
+        /// The request went out and we never learned what happened.
+        case unknown(operationId: String)
+    }
+
+    /// Compose a fresh snapshot and try to place it.
+    public func backUp(now: Date = Date()) async throws -> RunResult {
+        var state = try stateStore.load()
+
+        // Reconcile before adding to the pile. An operation whose result
+        // we never learned may well have succeeded, and composing a
+        // second snapshot on top of an unresolved first is how a person
+        // ends up paying to store the same history twice.
+        try await reconcile(&state)
+
+        guard let acceptedTermsId = state.acceptedTermsId else {
+            // Nothing has been consented to. Enrolment is a decision, and
+            // the repository does not make it on the person's behalf.
+            throw BackupError.termsUnavailable
+        }
+
+        let connection = try await port.connect()
+        if connection.acceptedTermsId != acceptedTermsId {
+            // Do not upload, do not re-pin, do not "helpfully" accept.
+            // Fresh consent is a screen, not an inference.
+            state.lastAttemptAt = now
+            try stateStore.save(state)
+            return .termsChanged(currentTermsId: connection.acceptedTermsId)
+        }
+
+        let snapshot = try await composer.compose(
+            keyMaterial: keyMaterial,
+            acceptedTermsId: acceptedTermsId,
+            supersedes: state.snapshots.last.map {
+                try? SnapshotReference(digest: $0.digest, sealedByteSize: $0.sealedByteSize)
+            } ?? nil,
+            now: now
+        )
+        state.lastAttemptAt = now
+        try stateStore.save(state)
+        return try await place(snapshot, state: &state, now: now)
+    }
+
+    /// Retry a snapshot that was refused for payment, byte for byte.
+    ///
+    /// Takes the `SealedSnapshot` the refusal handed back rather than
+    /// composing again: a fresh composition mints a new salt and a new
+    /// digest, which would defeat both the retry and `already_retained`,
+    /// and would charge the person for a second copy of the same
+    /// history.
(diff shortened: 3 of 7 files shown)


dev-onym (commented):
**REQUEST_CHANGES**

- **Pending record dropped on `submitted`/`accepted`** — `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:141` removes the pending operation *before* the `isRetention` guard at line 142. A `submitted` outcome then returns `.unknown` with no pending record left, so `reconcile` will never re-query it and the uncertainty is silently discarded — the exact invariant the PR body says this record exists to preserve ("silence is not success… only if the uncertainty is durable"). Move the `removeAll` after the guard, keeping the record for non-retention outcomes.
- **Unreadable state file reads as empty** — `Packages/OnymBackup/Sources/OnymBackup/BackupStateStore.swift:61` uses `try? Data(contentsOf:)`, conflating file-missing with file-unreadable (I/O error, or `.completeFileProtection` while the device is locked). An unreadable-but-present file returns `BackupState()`, and a subsequent `save` overwrites the real state — dropping pinned terms and the pending list, the same failure the corrupt-decode branch below deliberately refuses. Distinguish ENOENT from other read errors.
- **Grant arithmetic trusts operator-controlled numbers** — `Packages/OnymBackup/Sources/OnymBackup/URLSessionBackupClient.swift:152`: `index * grant.chunkBytes` traps on overflow for a hostile `chunkBytes` (only `> 0` is checked), crashing the client from untrusted input. Also, a grant whose `chunkCount` is too *small* silently truncates the upload (the `!chunk.isEmpty` guard only catches over-count), failing only at commit — validate `chunkCount == ceil(sealedByteSize / chunkBytes)` against the local file before sending anything.
- **Short download streams succeed silently** — `Packages/OnymBackup/Sources/OnymBackup/URLSessionBackupClient+Transport.swift:146`: `stream` enforces only the upper bound; a body shorter than the reference's `sealedByteSize` returns success with a truncated file. `exportSnapshots` verifies afterward, but `downloadSnapshot` hands the truncated file to the caller unchecked — cheap to assert `total == maxBytes` here.
- **Pending operations can never drain once outcome records expire** — `BackupRepository.swift:167`: `reconcile` keeps a 404/no-record operation pending forever, but the terms fixture itself declares `operationOutcomes: PT6H`, so any response lost longer than that becomes an immortal pending entry re-queried on every run. `queryOutcome`'s comment says "reconciliation moves to the reference," but no `listSnapshots` fallback exists — implement it or bound the retention of pending records.
- Minor: the `.alreadyRetained` path (`BackupRepository.swift:111`) never appends to `state.snapshots`, so the next `backUp`'s `supersedes` points at a stale ancestor after a retry-after-crash resolves this way.

The refusal-first shape, redirect/oversize handling, and the pre-upload pending write are all sound; the tests genuinely pin the interesting edges. The first two bullets contradict the PR's own stated invariants, hence request changes.

rinat-enikeev:
All six in da2a80c, one test each. Two contradicted claims I made in the PR description, and the first is the sharpest kind of miss.

**Pending record dropped on `submitted`.** The removal sat one line above the `isRetention` guard, so "silence is not success, and the only way that stays true is if the uncertainty is durable" held right up until the next statement. Removal now happens only on retention.

**Unreadable state read as empty.** Same failure as the corrupt-decode branch directly below it, reached through a different door — and you're right that complete file protection on a locked device is a realistic way in, not a theoretical one. Absent and unreadable are now distinguished, with `stateUnreadable` as its own reason.

**Pendings could never drain.** Correct, and the fix is better than bounding retention: `reconcile` now asks `listSnapshots` first, because it is the authority on what is *retained* — which is the question actually being asked. A digest in the list resolves as retained; a digest absent from a list that came back resolves as not retained. Only a list we could not fetch leaves things pending, since that is the one case where we genuinely don't know. `queryOutcome` stays as a secondary check for an operator that distinguishes something more specific. A test covers the failed-list path, since resolving uncertainty by failing to look would be worse than the bug.

**Grant arithmetic.** Both halves fixed by validating the grant against the sealed file before any byte moves: `chunkBytes` bounded, and `chunkCount == ceil(size / chunkBytes)`. The under-count case was the one I'd missed entirely — the `!chunk.isEmpty` guard only ever caught over-count.

**Short downloads.** `total == maxBytes` now asserted, and the partial file removed. You're right that `exportSnapshots` re-verifying made this look covered when the path that matters most — restore — was the one without a check.

**`.alreadyRetained` and `supersedes`.** Both retention paths now go through one recorder, which also replaces any earlier record of the same digest so a reconciliation can't double-count what an upload already wrote.

```
BackupAdapterTests   16 passed  (+6)
** TEST SUCCEEDED **
```

dev-onym (commented):
REQUEST_CHANGES

- **Reconcile drops in-flight operations** — `BackupRepository.swift:217-218`: only `.unknown`/`.unreachable` keep a pending record, but `queryOutcome` can legitimately return `.submitted` or `.accepted` (in progress, not yet in the retained list). `place` explicitly preserves the pending record for exactly those statuses (`BackupRepository.swift:151-158`); reconcile then resolves the same operation negatively on the next run, composes a fresh snapshot, and re-creates the double-payment the PR says this design prevents.
- **`try?` on `queryOutcome` conflates "operator says no record" with "couldn't ask"** — `BackupRepository.swift:217`: a transient network error mid-reconcile (after a successful list) silently drops the pending record. The doc comment's rule is "only a list we could not fetch leaves things pending," but a failed *query* also resolves uncertainty it never obtained.
- **Operator's echoed reference is believed verbatim** — `BackupRepository.swift:112` and `URLSessionBackupClient.swift` (preflight `alreadyRetained` branch): `outcome.snapshotReference` is recorded into local state with no check that its digest equals the snapshot we preflighted. A hostile operator can poison the local chain, and the next run's `supersedes` points wherever it chose. Same for `row.acceptedTermsId` recorded from `listSnapshots` in reconcile. Given the PR's own trust posture, refuse a mismatched echo.
- **Payment-pending snapshot isn't durable** — `BackupRepository.swift:101`: the refused `SealedSnapshot` lives only in the returned enum; a purchase flow that survives a process restart (StoreKit routinely does) loses it, and `discard`'s comment says the working directory is swept on the next run — which would delete the very bytes the retry needs, forcing re-composition and a second charge.
- **Overflow trap on a hostile manifest limit** — `URLSessionBackupClient.swift` (`listSnapshots`/`exportSnapshots` cap): `maximumRetainedSnapshots * perSnapshotEntryBytes` traps for a huge declared value — a crash where every other operator-controlled figure gets a refusal. Clamp before multiplying.
- **Error-body read caps retention, not consumption** — `URLSessionBackupClient+Transport.swift:157`: the `where` clause stops *appending* at 8 KiB but keeps draining the stream, so a failed download with an unbounded error body is read to completion. `break` out instead.
- Test quality: `testRedirectsAreRefused` (`BackupAdapterTests.swift:118-135`) catches any error and asserts nothing about where the request went — it passes even if the redirect were followed (the second origin also hits the stub); consider capturing requested URLs and asserting only one origin was contacted.

The state-machine edges the PR body highlights (write-before-upload, same-bytes retry, no re-pin, unreadable-state throw) are all present and tested — the first two bullets are the ones that undermine the stated invariants.

rinat-enikeev:
All seven in 95154b8. The first two undo the double-payment guarantee this PR was written to make, so thank you for catching them.

**In-flight operations resolved negatively.** Right, and the contradiction is internal: `place` deliberately keeps the pending record for `submitted`/`accepted`, then `reconcile` treated exactly those as resolved on the next run — composing a second snapshot on top of one the operator was still working on. Only a definitive answer resolves now: `retained` records it, `rejected` drops it, everything else stays pending.

**`try?` on `queryOutcome`.** Same rule I'd written for the list, not applied to the query. Both failures keep the record now.

**Echoed references believed verbatim.** This is the one I'm most glad you flagged — it's a trust-boundary hole in a PR whose whole argument is about the trust boundary. A mismatched `already_retained` echo is refused outright, and the terms recorded on reconciliation are the ones *we* pinned, now carried on the pending record rather than read back from the operator. Which terms a stored snapshot was accepted under is not a fact a counterparty gets to restate.

**Payment-pending snapshot not durable.** Correct, and the two halves compound: the bytes owed a retry were precisely the bytes `discard`'s sweep would delete. It's written to state now, with `pendingPayment(in:)` to resume after a relaunch — returning `nil` when the sealed bytes are actually gone, so a caller can't be stranded in a purchase flow with no snapshot at the end of it.

**Overflow trap** — clamped. You put it well: every other operator-controlled figure earns a refusal, and this one was earning a crash.

**Error-body `where` clause** — replaced with `break`. It stopped appending but kept draining, so the bound wasn't a bound.

**`testRedirectsAreRefused`.** Fair, and it was worse than weak — it passed whether or not the redirect was followed, because the second origin also hits the stub. It now records hosts and asserts only `backup.example` was contacted.

```
BackupAdapterTests   21 passed  (+5)
** TEST SUCCEEDED **
```

dev-onym (commented):
APPROVE-worthy craft overall, but one wedge bug tips it: **REQUEST_CHANGES**

- **Reconcile can wedge every future backup.** `BackupRepository.swift:295-298` — a `queryOutcome` answer of `.retained` with no echoed `snapshotReference` feeds `sealedByteSize: 0` into `SnapshotReference.init`, which throws (`SnapshotReference.swift:24` requires `> 0`). `reconcile` throws before saving, `backUp()` rethrows before composing, and since the pending op is never resolved, every subsequent run hits the same path. A terse — or deliberately hostile — operator can permanently brick backups with a *success* answer. Fall back to the pending record's own size or keep the op pending instead of throwing.
- **`invalid_entitlement` drops the issuers it exists to carry.** `URLSessionBackupClient+Transport.swift:248` hardcodes `entitlementIssuers: []`; the envelope (line 211) never decodes a top-level `entitlementIssuers` field. The PR body says per-code fields the profile pins are preserved — this one isn't, and the caller can't know which issuer to try next.
- **Silent chain break in `supersedes`.** `BackupRepository.swift:67` — `try?` means a recorded snapshot with an invalid digest/size yields `supersedes: nil` silently, minting a snapshot that claims to supersede nothing. Given the store refuses to read corrupt state precisely to protect this data, swallowing the error here is inconsistent; at least distinguish "no ancestor" from "ancestor unreadable".
- **Partial file left at `destination` on mid-stream I/O error.** `URLSessionBackupClient+Transport.swift:178-193` — the over-long and short paths delete the file, but a thrown `handle.write` or a network error mid-iteration propagates without cleanup. `downloadSnapshot`'s destination is caller-owned, not the swept working directory, so the partial sits there; the doc comment's own reasoning ("a restore is exactly where a silently-short file must not arrive") argues for a `defer`-style cleanup on any throw.
- **Byte-at-a-time reads are a real cost at these caps.** `URLSessionBackupClient+Transport.swift:135,178` — `for try await byte` with per-byte `Data.append` runs up to 32 MB (`maxListResponseBytes`) per list call on-device. Batch via the underlying buffers or `AsyncBytes`' collect patterns.
- Nits: `pendingPayment` reports `sealedAt: pending.refusedAt` (`BackupRepository.swift:96`, mislabels the timestamp); `reconcile` records `acceptedTermsId: … ?? ""` (`BackupRepository.swift:280`), and an empty terms pin is exactly what the rest of this PR refuses to write.

The refusal tests are genuinely good — `testRedirectsAreRefused` asserting *where bytes went* rather than that an error surfaced, and the derived (not fabricated) terms digest, are the right instincts. The reconcile wedge is the only blocker.

rinat-enikeev:
All six in 4c049e9. The wedge was the real one.

**Reconcile wedge.** Exactly as described, and the reachability is what makes it bad: reconcile runs *before* composing, so one unbuildable reference aborted `backUp` before it could save, leaving the operation unresolved for the next run to hit the same line. A permanent brick delivered by a `retained` answer. Reconciliation now rebuilds the reference from the size **we** recorded — pending records carry it for that purpose — and keeps the operation pending rather than throwing when it can't. The test asserts both halves: the operation resolves, and a subsequent run still works.

That pass also removed the `?? ""` terms fallbacks you flagged in the nits. An empty pin isn't just untidy — it would become the thing a later regression check compares against, so reconciliation now keeps the operation pending rather than writing one.

**`invalid_entitlement` issuers.** Decoded now, falling back to the `paymentRequired` block. You're right that it's the field the refusal exists to carry.

**`supersedes` via `try?`.** Agreed on the inconsistency — the store refuses to *read* corrupt state precisely to protect this data, and swallowing the same corruption one layer later defeated it. It throws `stateUnreadable`.

**Partial file on mid-stream error.** Fixed with a `complete` flag and cleanup in `defer`, so any exit that isn't a whole file takes the partial with it. The over-long and short paths cleaning up while a write error didn't was an arbitrary distinction — a caller-owned restore path shouldn't have to distinguish which way the transfer failed.

**Per-byte reads.** Batched through `[UInt8]`. Fair point that 32 MiB of `Data.append` on a phone is a real cost, not a theoretical one.

**Nit: `sealedAt`.** A resumed snapshot is as old as its contents, not as old as the refusal. It carries its own now.

```
BackupAdapterTests   25 passed  (+4)
** TEST SUCCEEDED **
```

dev-onym (commented):
**REQUEST_CHANGES** — the design is strong, but two edges contradict the PR's own double-pay invariant.

- **`backUp` composes anyway when reconcile can't conclude.** `reconcile` swallows a failed `listSnapshots` (`BackupRepository.swift:240`, `guard let retained = try? ... else { return }`) and returns; `backUp` then proceeds to compose and upload a *new* snapshot on top of unresolved pending operations — exactly the "pay to store the same history twice" case the PR body says reconcile-before-compose prevents. Worst path: a healthy operator with an oversized list body trips our own cap (`operatorUnavailable`), reconcile silently gives up, and every run mints a fresh salt/digest the operator retains alongside the lost one. Either refuse to compose while `pendingOperations` is non-empty and unreconciled, or justify the tradeoff in a comment/test.
- **Reconciled retentions corrupt "newest last" chain order.** `BackupState.record` does removeAll-then-append (`BackupStateStore.swift:101`), so a pending op resolved later lands at the *tail* even though it's older. The next `backUp` builds `supersedes` from `state.snapshots.last` (`BackupRepository.swift:67`), so the chain can point at the stale reconciled ancestor instead of the true newest snapshot. Insert by `uploadedAt` or pick the ancestor by timestamp.
- **Actor reentrancy can lose the uncertainty record.** `backUp`/`retry` each load state at entry (`BackupRepository.swift:40`, `:117`) and save across many `await` suspension points. Two interleaved calls (actor reentrancy is real across `port.*` awaits) can save stale copies over each other — e.g. one drops the other's just-written `pendingOperations` entry, which is the record the whole design exists to keep. Serialize runs (an in-flight flag) or reload-before-save.
- **`awaitingPayment` outlives resolution.** When reconcile resolves an operation via the retained list or `queryOutcome` (`BackupRepository.swift:255`, `:305`), it never calls `clearPendingPayment` or removes the sealed bytes, so `pendingPayment(in:)` can report a purchase owed for a snapshot the operator already holds — stranding the caller in the purchase flow the doc comment warns about.
- Minor: `uploadChunkBytes` (`URLSessionBackupClient.swift:44`) is never referenced — dead constant; `discard` is declared `throws` but only `try?`s (`BackupRepository.swift:357`); reconcile-resolved retentions never update `lastSuccessAt`.

Everything else holds up well: redirect refusal is asserted by *where bytes went*, the grant-arithmetic guard is ordered so `Int.max` can't overflow, and the terse-`retained` unwedge fix is correctly tested.

rinat-enikeev:
All five in 92caf9c. Two were the double-pay invariant again, and you're right that the argument I made didn't hold: reconciling first only prevents double storage if *failing* to reconcile also stops us.

**Composing over unresolved work.** Fixed by refusing. `backUp` now returns `.awaitingReconciliation(operationIds:)` while anything is unresolved, so a UI can say it's checking on an earlier backup rather than showing a silent no-op. The path you describe is the one that convinced me it needed to be a refusal rather than a comment — it needs no hostility at all, just a healthy operator whose list body exceeds our own cap, and then every run mints a fresh salt the operator retains beside the copy we lost. A test asserts nothing is even preflighted in that state.

I'll note the tradeoff explicitly since it's a real one: a genuinely stuck pending operation now stops backups rather than quietly duplicating them. That's the safer failure, and it's visible, but it does mean the drain paths have to work — which is why the earlier fixes to those matter more than they looked.

**Chain order.** Good catch, and subtle: `removeAll`-then-append is fine until something resolves out of order. Snapshots are kept in upload-time order now and the ancestor is picked by time rather than position.

**Actor reentrancy.** Real, and it would have eaten precisely the record the design exists to keep. Runs are serialized behind an in-flight flag, with a test driving two concurrent `backUp` calls.

**`awaitingPayment` outliving resolution.** This one is nastier than it reads — it strands someone buying storage for a snapshot the operator already holds, i.e. the exact failure `pendingPayment`'s doc comment warns about, caused by the code that resolves it. Resolution now clears the payment record and the sealed bytes, and records `lastSuccessAt` for a retention found by reconciliation rather than by the run that uploaded it.

**Minors** — dead constant dropped; `discard` no longer claims to throw when it only ever swallows.

```
BackupAdapterTests   29 passed  (+4)
** TEST SUCCEEDED **
```

dev-onym (commented):
APPROVE

- `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:56` — `backUp()` never consults `state.awaitingPayment`. A scheduled run during a pending purchase composes a fresh snapshot, gets refused again, and line 178 overwrites the `PendingPayment` record without discarding the old sealed bytes — the prior file is orphaned on disk and the "retry sends the same bytes" promise silently degrades to "retry sends the newest refusal". Consider returning the existing pending payment (or discarding its bytes on replacement).
- `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:321` — the reconcile list-hit path records `row.snapshotReference`, i.e. the operator's `sealedByteSize`, while the terse-`retained` path at line 360 explicitly prefers `pending.sealedByteSize` ("believe the size *we* recorded"). Same trust decision, two answers — a lied-about size here flows into the next snapshot's `supersedes`. Prefer `pending.sealedByteSize` when present, list row as fallback.
- `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:125` vs `:432` — `pendingPayment(in:)` accepts a working-directory override, but `discardSealedBytes` always resolves against `composer.workingDirectory`. A caller using the override gets a pending snapshot whose bytes `resolve` can never delete. Either drop the override or thread it through.
- `Packages/OnymBackup/Sources/OnymBackup/URLSessionBackupClient.swift:188` — a sealed file shorter than `sealedByteSize` uploads truncated chunks (only emptiness is checked) and fails at commit, after paying for the transfer — the exact failure mode the grant guard's comment says it exists to prevent. Cheap fix: `chunk.count == expected size for index` per chunk, or stat the file against `size` before the loop.
- `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:152` — `retry()` skips both reconciliation and the `connect()` terms check that `backUp()` performs; it relies on the operator's `terms_changed` refusal rather than the local pin. Consistent-enough, but worth a comment saying the omission is deliberate.
- `Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:324` — reconcile force-records `.retained`, ignoring the list row's own `status` (e.g. a row already `pendingErasure`); the recorded chain then says retained for something the operator is erasing. Minor, but `row.status` is already decoded and available.

The refusal architecture is sound and well-tested: redirect refusal asserts on *where bytes went*, the terse-`retained` wedge, corrupt-state-throws, and pending-blocks-composition tests all pin the edges the PR body claims. Nothing above is blocking.

PR #278, opened 2026-08-20
Add OnymBilling: purchases, and the credential a seat actually checks

rinat-enikeev:
Fifth of the device-backup stack, on top of #277 — and the first piece that isn't about backup. `OnymBackup` only ever produces a `paymentRequired` refusal; this is what resolves one. Relays and blob storage will sell through the same machinery.

### The credential is the house format, not a JWT

Canonical JSON with a detached Ed25519 signature, per [WHITEPAPER §17.5](https://github.com/onymchat/onym-system/blob/main/WHITEPAPER.md). Not a stylistic call: the broker, the operator and this client must agree on *bytes*, and there is exactly one implementation of those rules here — `ServiceManifestCanonical`, which every other signed document already uses.

### Verification order is the design

`SeatEntitlementVerifier` checks the signature **before believing any field**, because an unsigned document's `audience` is not evidence of anything. Then:

- **audience** — one operator never receives another's credential;
- **subject** — what stops a captured credential being useful to anyone else, since it's bound to a key only this device holds;
- **window**, then **revocation**.

The trusted issuer set is pinned from the operator's signed manifest and never read from the entitlement — otherwise a document nominates its own authority, which isn't a check.

### The envelope gets its own suite

`x25519-aes-256-gcm-seat-v1` with HKDF salt `onym-seat-entitlement-v1`, rather than reusing the invitation suite. Sharing one would make an entitlement ciphertext and an invitation ciphertext interchangeable at the crypto layer — same key agreement, same derivation context, same sender slot, with only a field name between two unrelated trust relationships. A test feeds the invitation scheme to the seat opener and expects a refusal.

### Purchase ordering — worth review

StoreKit signs → broker validates with the platform and issues → **we verify it ourselves** → we store it → *only then* the transaction is finished.

Finishing last is what makes a broker outage survivable: an unfinished transaction is replayed by `Transaction.updates`, so someone who paid and lost the network gets their credential on the next launch rather than a charge and nothing. `redeem` is shared by the purchase path and the replay path for that reason.

### Free stays first-class

`StoreKitEntitlementProvider` composes with `FreeTierEntitlementProvider` rather than replacing it. A free offer resolves with no store, no network and no credential — a self-hosted operator charging nothing is a first-class case, not a degraded one. An offer with no catalog entry is simply not sellable here; a manifest cannot mint a store product at runtime, and saying so beats a round trip that was never going to succeed.

### Two deliberate omissions

- **`ChannelOffers.json` ships empty.** No store product exists yet, and a placeholder identifier in a committed catalog is exactly the kind of thing that survives into a build. It carries `operatorShareBps` when filled, so the consent surface can *show* a commission rather than merely have one.
- **No `.storekit` file**, for the same reason — it would have to invent a product. StoreKit Testing lands with the first real one, which also means the StoreKit paths here are currently exercised only by compilation. The parts that can be tested without inventing a product are, thoroughly.

### Verification

```
BillingTests   12 passed
** TEST SUCCEEDED **
```

Nothing here is reachable from the app yet — the `makeEntitlements` closure in `OnymIOSApp` still returns the free-tier stub. That swap lands with the wiring PR.


dev-onym (commented):
REQUEST_CHANGES

- **Entitlement wipe on transient read failure** — `Packages/OnymBilling/Sources/OnymBilling/SeatPurchaseFlow.swift:96`: `var stored = (try? store.load()) ?? []` swallows a load error, then `store.save(stored)` overwrites the file with only the new credential. `FileSeatEntitlementStore` uses `.completeFileProtection`, and `redeem` is explicitly the `Transaction.updates` replay path — which can fire at launch before first unlock, when that read *will* fail. One replayed transaction then silently destroys every other purchased credential. This also contradicts the store's own stated posture ("unreadable is an error", SeatEntitlementStore.swift:27). Propagate the error and leave the transaction unfinished for a later replay.
- **Related, same block** — `SeatPurchaseFlow.swift:99`: the dedup's `else { return true }` deletes any stored entry the *current* build can't decode. A version-2 entitlement written by a newer build gets silently dropped by an older one during any redeem. Keep undecodable entries; only remove ones you positively matched.
- **Revocation check is fail-open by default** — `URLSessionBillingBrokerClient.swift:33,68`: `issuerKey` defaults to `nil`, and when nil the epoch's signature is never checked yet the epoch is returned as trusted. The comment ("believing one would let anyone who can answer this URL revoke a person's storage") describes exactly what the default configuration does. Make the key required, or return an explicit unverified type callers can't confuse with a verified epoch.
- **Envelope sender never pinned at the one call site** — `SeatPurchaseFlow.swift:85`: `SeatSealedEnvelope.open` is called without `expectedSender`, so the carefully built sender-authentication path (SeatSealedEnvelope.swift:73–86) is dead code in practice; an unsigned or anyone-signed envelope opens. Mitigated by the inner Ed25519 verification of the entitlement itself, but either wire the broker key through or say in the PR why sealing-without-sender-auth is sufficient.
- Nit — `SeatEntitlement.swift:118`: the doc says quota is "integer or string only", but decode reads `as? Int`, so a string quota silently becomes `nil` (i.e. "subscription, no quota") after a valid signature check. Refuse it as malformed instead.
- Nit — `SeatEntitlement.swift:87` / `URLSessionBillingBrokerClient.swift`: both timestamp decoders reject fractional seconds; fine if the house format forbids them, but worth a one-line test pinning that.

The design itself is sound and well-argued — verification order, domain-separated envelope suite, finish-last transaction ordering, and the free-tier-first composition all check out against the existing `ServiceManifestCanonical`/`SealedEnvelope` conventions, and `Resources/` is already bundled into the app target so the empty catalog lands where `bundled()` looks.

rinat-enikeev:
All six in a42d80a. The first destroys data, and your reachability argument is what makes it urgent rather than theoretical.

**Entitlement wipe.** `redeem` is exactly the `Transaction.updates` replay path, it fires at launch, and the store uses complete file protection — so the read genuinely fails before first unlock, and one replayed transaction silently wipes every purchase. It also contradicted the store's own posture, which refuses an unreadable file *precisely* so this cannot happen; I wrote that rule and then swallowed it one layer up. The load propagates now, leaving the transaction unfinished for a later replay: a delayed credential is recoverable, deleted ones aren't.

**Undecodable entries deleted.** Right, and it's a downgrade hazard specifically — an older build eating a newer build's entitlement during an unrelated redeem. Only positively matched entries are removed.

**Revocation fail-open.** The sharpest framing in your review: the comment described exactly what the default configuration did. `issuerKey` is required now, so there's no configuration in which an epoch is returned unverified.

**Sender never pinned.** Fixed rather than justified — I don't think "sealing is sufficient" holds up, because sealing proves only that the answer was meant for this device, not who sent it. It takes the operator's declared issuer keys now, and an empty expectation must be passed explicitly rather than inherited from a default: a check that can be skipped by omission isn't a check.

**Quota nit — upgraded, because it isn't cosmetic.** Reading a string quota as `nil` means "subscription, no quota" *after* a valid signature check, which is the quietest possible way to give away a consumable's units. Integers and decimal strings are accepted; floats and anything else are malformed.

**Fractional seconds** — test added, pinning that the house form rejects them so a decoder change can't quietly start accepting a shape the signature was never computed over.

```
BillingTests   17 passed  (+5)
** TEST SUCCEEDED **
```

dev-onym (commented):
REQUEST_CHANGES

- **Vacuous regression test** — `Tests/OnymIOSTests/BillingTests.swift:181`: `StubBroker(sealed: Data())` makes `redeem` throw at `SeatSealedEnvelope.open`, long before `store.load()` is ever called. `FailingLoadStore.load()` never runs, so the test guarding the headline data-loss fix ("replayed transaction deletes every credential") would still pass if `try store.load()` regressed to `try? store.load() ?? []`. Seal a real envelope (or stub past it) so the throw actually comes from the failing load.
- **Comment contradicts behavior** — `Packages/OnymBilling/Sources/OnymBilling/SeatPurchaseFlow.swift:63`: claims "if that leaves nothing, the envelope opener refuses," but `SeatSealedEnvelope.open` (`SeatSealedEnvelope.swift:83-87`) treats empty `expectedSenders` as "accept unauthenticated." If every issuer reference is unparseable, `senderKeys` returns `[]` and the envelope opens with no sender check — recovered only downstream by the verifier. Either make `redeem` refuse an empty key set or fix the comment.
- **Test exercises a copy, not the code** — `Tests/OnymIOSTests/BillingTests.swift:202`: `testUndecodableStoredEntriesAreKept` re-implements the `removeAll` predicate inline instead of calling `redeem`, so it can't catch a regression in `SeatPurchaseFlow` itself.
- **No positive envelope test** — the decrypt path of `SeatSealedEnvelope.open` (HKDF salt, sharedInfo, GCM assembly) is never exercised against a matching seal; a wrong constant would surface only at broker integration. A test-side seal helper would cover it and fix the first finding too.
- **Test-order coupling** — `Tests/OnymIOSTests/BillingTests.swift:374`: `FailingLoadStore.didSave` is static and never reset, so a future second use poisons results across tests.
- Minor — `StoreKitEntitlementProvider.swift:62`: `try? store.load()` maps a locked/corrupt store to "no entitlement." Fail-closed and defensible, but it's the same locked-at-launch state `redeem` handles deliberately; worth a comment stating the intent.

Architecture, verification ordering, canonical-bytes handling, and the finish-last purchase flow are all sound; the requested changes are test-integrity fixes plus one comment/behavior mismatch, not design problems.

rinat-enikeev:
All six in 49e4fee. Five were tests that looked like coverage and weren't, which is a worse category than a bug.

**The vacuous regression test.** You're exactly right, and it's the one guarding the data-loss fix from the previous round — `StubBroker(sealed: Data())` threw in the opener long before `store.load()` ran, so it passed for the wrong reason. It now seals a real envelope, reaches the load, and asserts `didAttemptLoad` alongside the absence of a save.

I checked rather than assumed: reinstating `(try? store.load()) ?? []` makes it **fail**, where before it would have passed.

```
Test Case '...testRedeemRefusesToSaveOverAnUnreadableStore' failed (0.627 seconds)
```

**No positive envelope test.** Your suggestion that a seal helper would fix both findings was right — building one for the above gave the decrypt path its first real exercise. Until now a wrong HKDF salt or `sharedInfo` would have surfaced at broker integration and nowhere earlier. There's a negative case too: an envelope sealed by an untrusted broker, addressed correctly, must not open.

**Testing a copy.** `testUndecodableStoredEntriesAreKept` re-implemented `redeem`'s predicate inline, so it couldn't catch a regression in the thing it named. It drives `redeem` now.

**Comment vs behaviour.** Fixed toward the comment rather than the code: `redeem` refuses when the trusted-issuer set parses to nothing. It could never have succeeded — the verifier requires a trusted issuer — but failing there says *why*, instead of opening an unauthenticated envelope and reporting a signature problem two steps later.

**Static test state.** Instance-scoped now; you're right that a second use would have poisoned whichever test ran second.

**`try? store.load()` in the provider.** Kept, with the intent stated: it's deliberately the opposite call from `redeem`'s. Asking *may this offer be selected* can fail closed; `redeem` goes on to **write**, so answering "no entitlements" for a file it couldn't read is how the credentials get destroyed.

```
BillingTests   20 passed  (+3)
** TEST SUCCEEDED **
```

dev-onym (commented):
COMMENT

- `Packages/OnymBilling/Sources/OnymBilling/StoreKitEntitlementProvider.swift:59` — the catalog guard gates **use**, not just sale: a stored, signed, in-date credential stops granting the moment its catalog entry disappears (product delisted, catalog edit in an app update). Sellability and validity are different questions; the PR body only argues the former. A paying user with a valid credential shouldn't lose access to a catalog file change.
- `Packages/OnymBilling/Sources/OnymBilling/SeatPurchaseFlow.swift:82` — the finish-last design is justified entirely by `Transaction.updates` replay, but nothing in the repo listens to it yet (only comments mention it). Fine for this PR, but note the replay path also needs `offerId`/`componentId`, which a raw `Transaction` doesn't carry — the wiring PR will have to reverse-map `productID` through the catalog, and that mapping isn't guaranteed unique (`ChannelOfferCatalog` has no uniqueness check on `productId`).
- `Packages/OnymBilling/Sources/OnymBilling/URLSessionBillingBrokerClient.swift:52` — revocation has no rollback protection: the epoch signature is verified, but `epoch` is never checked against a last-seen value, so anyone who can serve a stale response (CDN cache, on-path with a replayed 200) un-revokes a refunded credential. The field exists; nothing enforces monotonicity. Worth at least a tracking note before revocation is load-bearing.
- `Packages/OnymBilling/Sources/OnymBilling/StoreKitEntitlementProvider.swift:33` — `revokedIds` defaults to `{ [] }`, so revocation is silently off unless the composition root remembers to wire it. Consider making it non-defaulted like `issuerKey` was (the PR itself cites that exact "left a default alone" failure).
- `Packages/OnymBilling/Sources/OnymBilling/SeatPurchaseFlow.swift:110` — `untrustedIssuer(issuer: componentId)` stuffs a component id into an issuer-shaped payload; misleading diagnostic.
- `Packages/OnymBilling/Sources/OnymBilling/SeatSealedEnvelope.swift:87` — the sender signature covers only the ephemeral key bytes, not the scheme/salt, so a broker's signed ephemeral key is transplantable between the invitation and seat suites. Not exploitable today (attacker lacks the ephemeral private key), but signing `scheme || ephemeral` would make the domain separation self-contained. Low priority.

Strengths worth keeping: verification order, fail-closed store handling in `redeem` (and the test that actually reaches `store.load()`), keeping undecodable stored entries, and the distinct envelope suite with a negative test. Tests are unusually well-aimed.

PR #279, opened 2026-08-20
Add OnymBackupUI: the consent surface, and honest copy about scheduling

rinat-enikeev:
Sixth of the device-backup stack, on top of #278. The screen someone reads before their history leaves the phone, the settings surface that follows, and the scheduling policy both describe.

### Scheduling — a decision I made, easy to reverse

You asked what the schedule is; there wasn't one, so this PR picks a shape and puts it in a single type rather than scattering it through views. **v1 is deliberately not a cron**, for two reasons that were already decided:

- **Uploads are foreground-only.** A multi-hundred-megabyte transfer wants a background `URLSession` with a persistent identifier — precisely the long-lived, linkable session state this codebase avoids everywhere else.
- **Snapshots are whole, never incremental.** Every run re-uploads the entire history, which argues for deliberate and occasional over frequent and automatic.

So: explicit **Back Up Now** is the primary path, plus an opportunistic run while the app is open on Wi-Fi and charging, at most daily, with hours of jitter. The jitter isn't battery politeness — upload timing that tracks send activity re-leaks per-group activity through volume and cadence, the same signal the padding and the opaque archive exist to suppress.

If you'd rather have a background task (and accept the session-state tradeoff), or drop the opportunistic path entirely and make it purely manual, it's one type to change.

### The copy has to match

Because uploads are foreground-only, the enrolment screen says: *"It cannot back up in the background, so if you never open the app, nothing is backed up."* Calling this automatic would have been the easiest lie in the product — a test fails if the word "automatically" appears in the schedule copy.

### The enrolment surface

Leads with the three things someone would otherwise learn too late:

1. **what a backup does to everyone else** — a snapshot extends both sides of every conversation, for people who never chose this operator (§7.4, §10.4);
2. **that nobody, including us, can recover it** without the phrase (§11);
3. **when backups actually happen.**

The operator's declared terms follow **in full and unsummarised**. An operator that keeps access logs, or excludes a great deal from erasure, or names four sub-processors, has published that in a signed document — the screen repeats it rather than smoothing it. The excluded erasure scope travels verbatim, since it's the part a person is likeliest to assume away.

The accept button stays disabled until the disclosure has been scrolled through. The button isn't the consent; the reading is.

### Disclosure as data, so §18.10 can be satisfied

`BackupDisclosure` is built as a value rather than assembled inline in a view, so the fixture the spec asks for is possible: consent verified **by fixture rather than by review**. A disclosure is the one part of this seat that degrades silently — a paragraph dropped in a refactor still compiles, still renders, and still looks like a consent screen. The fixture names every required term and fails when one goes missing.

### Status reports uncertainty as uncertainty

`unknown` → "result unknown, checking". `awaitingReconciliation` → "checking an earlier backup", with a note that nothing new uploads until it resolves *so nobody is charged twice*. A stale backup is a visible state rather than something that still reads as on (§7.15).

### Verification

```
BackupDisclosureTests   9 passed
** TEST SUCCEEDED **
```

### Note on naming

`SettingsView` already had `makeBackupFlow` — that's the **recovery phrase** backup, which protects the key. This is `makeDeviceBackupView`, which protects the history. Different things, and the copy distinguishes them.

Restore is not here: it needs `IdentityRepository.restore(mnemonic:)`, which wipes existing identities, so it belongs with the wiring PR where the fresh-device path lives.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -396,6 +396,45 @@ public actor BackupRepository {
         try stateStore.save(state)
     }
 
+    /// What the operator says it holds for this holder.
+    ///
+    /// The operator's list is authoritative about retention; the local
+    /// chain records what *we* did. They can differ — a snapshot
+    /// uploaded from another device, or one whose response we lost — and
+    /// a surface showing "your backups" should show the operator's
+    /// answer rather than our memory of it.
+    public func listSnapshots() async throws -> [RetainedSnapshot] {
+        try await port.listSnapshots()
+    }
+
+    /// Erase, and record the receipt.
+    ///
+    /// The receipt is kept because its `excludedScope` outlives the
+    /// snapshot it describes: what an erasure did not reach is the part
+    /// a person may need to re-read long afterwards.
+    @discardableResult
+    public func erase(scope: ErasureScope) async throws -> ErasureReceipt {
+        let receipt = try await port.eraseSnapshot(scope: scope)
+        var state = try stateStore.load()
+        state.receipts.append(receipt.rawBytes)
+        if case .snapshot(let reference) = scope {
+            state.snapshots.removeAll { $0.digest == reference.digest }
+        } else {
+            state.snapshots.removeAll()
+        }
+        try stateStore.save(state)
+        return receipt
+    }
+
+    /// Export every retained snapshot in portable form.
+    ///
+    /// Never gated on payment here, and it must not be gated at the
+    /// operator either — retention is not leverage over someone's own
+    /// history.
+    public func export(to directory: URL) async throws -> BackupExport {
+        try await port.exportSnapshots(to: directory)
+    }
+
     /// Close out a pending operation: clear any purchase it was waiting
     /// on, drop its sealed bytes, and — when it landed — record that a
     /// backup succeeded.
--- a/Packages/OnymBackupUI/Package.swift
+++ b/Packages/OnymBackupUI/Package.swift
@@ -1,0 +1,25 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
+    name: "OnymBackupUI",
+    platforms: [.iOS("18.0")],
+    products: [
+        .library(name: "OnymBackupUI", targets: ["OnymBackupUI"]),
+    ],
+    dependencies: [
+        .package(path: "../OnymBackup"),
+        .package(path: "../OnymBilling"),
+        .package(path: "../OnymDesign"),
+    ],
+    targets: [
+        .target(
+            name: "OnymBackupUI",
+            dependencies: [
+                "OnymBackup",
+                "OnymBilling",
+                "OnymDesign",
+            ]
+        ),
+    ]
+)
--- a/Packages/OnymBackupUI/Package.swift
+++ b/Packages/OnymBackupUI/Package.swift
@@ -11,6 +11,7 @@ let package = Package(
         .package(path: "../OnymBackup"),
         .package(path: "../OnymBilling"),
         .package(path: "../OnymDesign"),
+        .package(path: "../OnymFoundation"),
     ],
     targets: [
         .target(
@@ -19,6 +20,7 @@ let package = Package(
                 "OnymBackup",
                 "OnymBilling",
                 "OnymDesign",
+                "OnymFoundation",
             ]
         ),
     ]
--- a/Packages/OnymBackupUI/Sources/OnymBackupUI/BackupDisclosure.swift
+++ b/Packages/OnymBackupUI/Sources/OnymBackupUI/BackupDisclosure.swift
@@ -1,0 +1,165 @@
+import Foundation
+import OnymBackup
+
+/// Everything the enrolment surface must say before a single byte
+/// leaves the device.
+///
+/// Built as data rather than assembled inline in a view so a fixture can
+/// assert its content. `UI-Backup.md` §18.10 asks for exactly that: the
+/// consent surface is verified "by fixture rather than by review",
+/// because a disclosure that quietly loses a paragraph in a refactor is
+/// worse than one that was never written.
+public struct BackupDisclosure: Sendable, Equatable {
+    public struct Item: Sendable, Equatable, Identifiable {
+        public let id: String
+        public let label: String
+        public let value: String
+    }
+
+    /// The sentence §7.4 and §10.4 require, and the reason this seat
+    /// treats disclosure as a contract term rather than a courtesy: a
+    /// snapshot extends the life of both sides of every conversation in
+    /// it, for people who never chose this operator.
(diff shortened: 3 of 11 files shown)


dev-onym (commented):
REQUEST_CHANGES

- **Scroll gate is a no-op** — `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupEnrolmentView.swift:93`: the sentinel `Color.clear` sits in a plain (non-lazy) `VStack` inside a `ScrollView`, so `onAppear` fires at initial layout, offscreen included. `scrolledToEnd` becomes true immediately and "Turn On Backup" is enabled without any reading. This is the PR's central consent claim ("the button isn't the consent; the reading is") and it's broken — use `LazyVStack` or a visibility check, and add a test/UI test, since this is exactly the "degrades silently" class §18.10 worries about.
- **Jitter is declared, never applied** — `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupSchedule.swift:33,88`: `maximumJitter` has no reads anywhere; `permitsOpportunisticRun` gates only on Wi-Fi/charging/`minimumInterval`. The PR body sells jitter as a privacy property (decoupling upload timing from send activity), but an opportunistic run fires as soon as conditions are met — deterministic, activity-correlated timing. Either apply the jitter or delete the field and the claim.
- **`refresh()` hides pending reconciliation** — `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift:66-70`: `BackupState.pendingOperations` is ignored, so after relaunch an unresolved upload renders as "On"/idle. `checkingEarlierBackup` only appears transiently after tapping Back Up Now — contradicting the §7.15 "uncertainty as uncertainty" claim on the screen's initial load.
- **Failed erase destroys the last receipt** — `DeviceBackupSettingsFlow.swift:125`: the error path sets `state.lastReceipt = nil`, wiping the receipt of a *previous successful* erasure that the same doc comment says a person may need later. Leave it untouched on failure.
- **Orphaned comment** — `Packages/OnymSettings/Sources/OnymSettings/SettingsView.swift:224-231`: the backup block was inserted between the pre-existing moderation-gating comment and the moderation `if let` it describes; the comment now reads as if it applies to BACKUP. Move it back above line 253.
- Minor — `BackupDisclosure.swift:160`: "at most once a day" is hardcoded while `minimumInterval` is configurable; `scheduleSentence` tracks Wi-Fi/charging but would lie if the interval changes.
- Minor — `BackupEnrolmentView` never renders `disclosure.operatorName` outside the terms list; field may be dead.

The disclosure fixture tests, honest schedule copy, verbatim operator terms, and the `BackupRepository` list/erase/export additions all look solid.

rinat-enikeev:
All seven in a0ab20e. Two were this PR's own claims being false, which is the worst way to be wrong.

**The scroll gate didn't gate.** You're right about the mechanism — `onAppear` on a sentinel in a plain `VStack` fires at initial layout for offscreen children, so `scrolledToEnd` flipped immediately and the button was live before anyone read a word. "The button is not the consent, the reading is" was false in precisely the way §18.10 describes: it compiled, it rendered, and it looked like a consent screen.

Replaced with `onScrollGeometryChange` over a **pure predicate**, which is the part that answers your "add a test" point — a claim about consent that only a human can check is the one most likely to quietly stop being true. Writing it as a function surfaced a second bug in passing: content shorter than the container has already been read, and gating on a scroll that can never happen would have locked the button forever on a large display. That's now its own test.

**Jitter declared, never applied.** Correct, and the consequence is exactly the one the PR body claimed to prevent: a run firing the instant the interval elapses is deterministic, and therefore correlated with whatever the person was doing when the previous one ran. It's applied now, drawn once per flow rather than per check — re-drawing would let frequent polling sample until it found a small value, which is the same as no jitter.

**`refresh()` hiding pending work.** Good catch, and the user-visible consequence is worse than the display bug: the person taps Back Up Now, nothing new uploads (correctly — #277 refuses to compose over unresolved work), and the screen says "On". Unresolved work now outranks idle on load.

**Failed erase destroying the receipt.** Fixed. The receipt describes an erasure that *did* happen, and its `excludedScope` is what someone may need long afterwards — a later failure isn't a reason to destroy that record.

**Cadence hardcoded** — follows `minimumInterval` now, with a test. You put it well: a sentence that starts true beside a configurable value.

**Orphaned comment** — moved back above the moderation block.

**`operatorName` dead** — rendered rather than removed. Who is receiving this history is arguably the first thing the screen should answer, and it wasn't answering it outside a terms row.

```
BackupDisclosureTests   14 passed  (+5)
** TEST SUCCEEDED **
```

dev-onym (commented):
APPROVE

- `Tests/OnymIOSTests/BackupDisclosureTests.swift:36` — the doc comment claims "adding a field to `BackupTerms` without disclosing it fails here", but the assertion is `required.isSubset(of: ids)`: it catches *dropped* items only. A new terms field never enters `required` or `items`, so nothing fails. Either fix the comment or assert set equality plus a `Mirror`-based sweep of `BackupTerms` — the current wording promises a guarantee the fixture doesn't provide.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift:120` — `backUpIfDue` has no caller anywhere, yet the enrolment copy promises "Onym may also back up on its own while the app is open." Until the wiring PR adds a scenePhase/foreground hook, that sentence is the one overpromise on a screen whose whole point is not overpromising. Fine if the wiring PR is imminent; worth a tracking note otherwise.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift:108` — on `.alreadyRunning` the status is set to `.running` and nothing ever observes the in-flight run's completion; the Back Up Now button stays disabled until the view is recreated. Consider falling back to `refresh()` after a beat, or polling the pending state.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupDisclosure.swift:157` — `cadence` rounding can understate frequency: an interval of 2.6h renders "once every 3 hours" while `permitsOpportunisticRun` allows a run every 2.6h. Unreachable with the 24h default, but `Int(hours.rounded(.up))` would keep the sentence a true upper bound for any config.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift:135` (also 125) — `backUpNow` success path calls `refresh()` but not `loadSnapshots()`, so a snapshot retained moments ago doesn't appear in the SNAPSHOTS list until re-entering the screen.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift:143` / `:126` — `String(describing: error)` surfaces raw enum dumps (`localFailure(reason: .stateUnreadable)`) as user-facing copy; map known `BackupError` cases to sentences.

The load-bearing claims check out: the scroll gate is a pure tested predicate with the short-content case handled, `FileBackupStateStore.save` is atomic so the flow's MainActor reads racing the repository actor get old-or-new (never torn), jitter is clamped and actually applied in `permitsOpportunisticRun`, and the "no 'automatically'" copy test passes against the real sentence.

PR #281, opened 2026-08-20
Restore history from a backup

rinat-enikeev:
The reason the rest of the stack exists — everything before it wrote snapshots nobody could read back.

### It corrects a distinction I had wrong

I'd said restore couldn't live in Settings because `IdentityRepository.restore(mnemonic:)` wipes every existing identity. That's true of **identity** restore, and it isn't what this does.

- *Identity* restore replaces the device's keys and wipes existing identities. It stays in onboarding, where nothing is at stake. Not reachable from here.
- *History* restore derives its key from the identity the device **already has**, and writes rows through the app's own stores with `insertOrUpdate`. Nothing is replaced, nothing deleted.

So this is safe on a device that already has its identity, which means it can be reachable now rather than waiting on onboarding surgery — and it's useful to more than someone with a new phone. Someone who lost a chat, or set up a second device, has the same reason to want it.

### Behaviour worth reviewing

- **Three gates, unchanged**: verify the whole reference → decode and validate the entire archive → then write. A failure leaves the device exactly as it was, and the screen says so.
- **An empty list is an ordinary answer, not an error.** A different operator or identity has a different holder key and sees nothing. Telling someone their history is gone when they've picked the wrong operator would be the cruellest bug available here, so the empty state says exactly that.
- **The summary reports what the sink wrote**, names rows this build couldn't read, and states plainly when attachments couldn't be restored.

### ⚠️ This branch turns a pre-existing test red, and it isn't my code

`InviteIntroducerTests.test_rotate_concurrentWithCurrentOrMint_leavesOneSharedKey` fails on this branch and passes on `main`. I traced it rather than assuming:

- My production changes are confined to `OnymBackup`, `OnymBackupUI` and `BackupComposition.swift` — nothing near `InviteIntroducer`.
- Removing **only my new test file** and re-running makes it pass again. So the *presence* of `BackupRestoreTests` in the bundle changes scheduling and exposes a genuine race in `InviteIntroducer.serializingLink`, which the test's own comment anticipates ("unserialized, it can be a link rotate revoked").

I've deliberately **not** fixed it here. It's invite-link revocation logic in a component this PR doesn't touch, and getting a concurrency fix wrong there means either handing out a revoked link or failing to revoke one — that deserves its own PR and its own review, not a drive-by inside a restore change.

Happy to take it as the next PR if you want it; say the word.

### Verification

```
BackupRestoreTests   4 passed
OnymIOSTests         1400 tests, 3 skipped, 1 failure — the InviteIntroducer race above
Debug                ** BUILD SUCCEEDED **
```

### DIFF
--- a/Packages/OnymBackup/Package.swift
+++ b/Packages/OnymBackup/Package.swift
@@ -9,12 +9,14 @@ let package = Package(
     ],
     dependencies: [
         .package(path: "../OnymFoundation"),
+        .package(path: "../OnymTransportBlossom"),
     ],
     targets: [
         .target(
             name: "OnymBackup",
             dependencies: [
                 "OnymFoundation",
+                "OnymTransportBlossom",
             ]
         ),
     ]
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupError.swift
@@ -98,6 +98,15 @@ public enum BackupError: Error, Equatable, Sendable {
         /// finishing. A programming error, named so it does not
         /// masquerade as a corrupt archive.
         case archiveWriterFinished
+        /// A restore failed *after* it had begun writing.
+        ///
+        /// Gate 3 is not atomic: groups, messages and invitations may
+        /// already be in the stores when a later write throws. The
+        /// partial state is benign — every write is `insertOrUpdate`, so
+        /// restoring again converges — but a screen that says "nothing
+        /// was changed" would be lying, and this seat does not get to
+        /// claim more than it knows.
+        case restoreInterrupted
         /// The local backup state exists but could not be read or
         /// decoded. Never treated as "no backup configured": the next
         /// save would overwrite the pinned terms and the pending
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -520,6 +520,15 @@ public actor BackupRepository {
         try await port.listSnapshots()
     }
 
+    /// Fetch a snapshot's sealed bytes.
+    ///
+    /// The caller verifies the reference before treating any byte as
+    /// restorable; the client refuses a short or over-long body, so what
+    /// arrives is either the whole thing or an error.
+    public func download(_ reference: SnapshotReference, to destination: URL) async throws {
+        try await port.downloadSnapshot(reference, to: destination)
+    }
+
     /// Erase, and record the receipt.
     ///
     /// The receipt is kept because its `excludedScope` outlives the
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift
@@ -91,12 +91,24 @@ public actor BackupRestorer {
         // archive's totals would claim to have restored things that are
         // not there — the same overstatement this seat refuses
         // everywhere else.
-        let wroteGroups = try await sink.restore(groups: groups)
-        let wroteMessages = try await sink.restore(messages: messages)
-        let wroteInvitations = try await sink.restore(invitations: invitations)
-        let wroteConsents = try await sink.restore(consents: consents)
-        for blob in blobs {
-            try await sink.restore(blob: blob)
+        // From here on a failure is *not* "nothing happened": the
+        // earlier writes have landed. Reported as `restoreInterrupted`
+        // so the screen can say so, rather than inheriting the
+        // pre-write promise.
+        let wroteGroups: Int
+        let wroteMessages: Int
+        let wroteInvitations: Int
+        let wroteConsents: Int
+        do {
+            wroteGroups = try await sink.restore(groups: groups)
+            wroteMessages = try await sink.restore(messages: messages)
+            wroteInvitations = try await sink.restore(invitations: invitations)
+            wroteConsents = try await sink.restore(consents: consents)
+            for blob in blobs {
+                try await sink.restore(blob: blob)
+            }
+        } catch {
+            throw BackupError.localFailure(reason: .restoreInterrupted)
         }
 
         var skipped: [String: Int] = [:]
--- a/Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift
@@ -14,6 +14,13 @@ import OnymTransportBlossom
 /// Wrapping the client is what makes them real, and it covers every
 /// loader at once — image, voice, and video all take a `BlossomClient`,
 /// so none of them needs to know this exists.
+///
+/// **Nothing prunes this directory yet** (onymchat/onym-ios#283).
+/// Restored ciphertext outlives the chat it belongs to, which is the
+/// wrong default in a product that argues deletion means something. It
+/// is only ever written by an `.includeCiphertext` restore, which is not
+/// selectable yet, and must be fixed before that policy becomes
+/// reachable.
 public struct RestoredBlobClient: BlossomClient {
     private let restored: URL
     private let upstream: any BlossomClient
--- a/Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift
@@ -1,0 +1,50 @@
+import CryptoKit
+import Foundation
+import OnymTransportBlossom
+
+/// Serves restored attachment ciphertext before going to the network.
+///
+/// A snapshot taken with `.includeCiphertext` carries the exact bytes
+/// the blob store served, so a restore can put media back that the blob
+/// operator has since dropped — which is most of the reason to pay for
+/// carrying it. But the media loaders reach for a `BlossomClient`, and
+/// bytes written to a directory none of them reads are bytes nobody ever
+/// sees. Counting those as "restored" would be a lie in the summary.
+///
+/// Wrapping the client is what makes them real, and it covers every
+/// loader at once — image, voice, and video all take a `BlossomClient`,
+/// so none of them needs to know this exists.
+public struct RestoredBlobClient: BlossomClient {
+    private let restored: URL
+    private let upstream: any BlossomClient
+
+    public init(restoredDirectory: URL, upstream: any BlossomClient) {
+        self.restored = restoredDirectory
+        self.upstream = upstream
+    }
+
+    public func download(sha256: String) async throws -> Data {
+        if let local = try? Data(contentsOf: restored.appending(path: sha256)) {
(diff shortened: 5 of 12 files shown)


dev-onym (commented):
REQUEST_CHANGES

- **Path collision between download and decrypt** — `BackupRestoreFlow.swift:71` downloads sealed bytes to `workingDirectory/restore-<digestHex>`, and `BackupRestorer.swift:37` uses the *identical* path for the decrypted plaintext. `BackupOpener.open` (BackupOpener.swift:49–85) then reads the sealed file and `createFile`s the plaintext at the same path while the input handle is open — correctness depends on whether Foundation replaces the inode (rename) or truncates in place; on truncate semantics every restore fails as `incompleteSnapshot`. It evidently passes on your machine, but it's one Foundation implementation detail away from breaking. Rename the download (e.g. `sealed-<digestHex>`).
- **Restored attachments land where nothing reads them** — `BackupComposition.swift:93` points the sink's `blobDirectory` at `BackupSeat.workingDirectory()/blobs`, but the media loaders read the plaintext cache at `Caches/OnymChatImages` or fetch from the Blossom client (`ChatImageLoader.swift:31–33, 109`); neither consults that directory. The sink's own doc ("so the media loaders find it without a round trip") promises wiring that doesn't exist, so an `includeCiphertext` restore reports N attachments restored while every one still renders via network-or-nothing. Either wire a loader fallback to that directory or stop counting these as restored.
- **Raw error dumps in user-facing UI** — `BackupRestoreFlow.swift:63` and `:88` set `failed(message: String(describing: error))`, which `BackupRestoreView.swift:170` renders verbatim. A person mid-restore gets `incompleteSnapshot` or a `URLError` debug description. Map the known `BackupError` cases to sentences, keep the raw string for logs.
- **Test double verifies less than the test names claim** — `BackupRestoreTests.swift` `RecordingSink` records only groups; `testFailedRestoreWritesNothing` would pass even if messages/invitations were written before the failure. Record all record types.
- **The known-red `InviteIntroducer` test** — the analysis is credible and deferring the fix is right, but don't merge a red main: land the follow-up first, or link a tracking issue in this PR and mark the test as an expected failure/skip with that reference so CI stays green.
- Minor: `BackupRestoreView.swift:126` sorts skipped-row strings lexicographically ("10 messages" before "2 groups"); sort the pairs, then format.

The gating design, the empty-state copy, and keeping identity restore unreachable from here are all right — the first two bullets are what block this.

rinat-enikeev:
All six in d23a0f2.

**Path collision.** Fixed — the download is `sealed-<digest>` now. You're right that it was passing on an implementation detail: `BackupOpener` held a read handle on the sealed file while creating the plaintext over the same path, and on truncate semantics every restore would fail as `incompleteSnapshot`. Not a bug I'd have found by testing, since the machine I test on happens to replace the inode.

**Restored attachments landing nowhere.** The sharper half of this is that the sink's own comment promised wiring that didn't exist — "so the media loaders find it" was aspirational.

I took the wire-it option rather than the stop-counting one, because carrying ciphertext is most of the reason to pay for `.includeCiphertext` at all: `RestoredBlobClient` wraps the Blossom client and serves a restored blob from disk before asking the network. One decorator covers image, voice and video, since all three take a `BlossomClient` and none needs to know it exists. Bytes are re-verified against their content address on read — they live on a device now, and a content address that doesn't match its content is the one thing a loader must never decrypt against.

**Raw error dumps.** Mapped to sentences, with an operator's own message preferred over our guess at what its code meant. Test asserts a `URLError` never reaches the screen as `Error Domain=…`.

**Test double.** Correct, and it's the same failure mode you caught twice in #280: the test named a property and checked a quarter of it. `RecordingSink` records every kind, and `wroteNothing` covers all five.

**Sort order.** Pairs sorted, then formatted.

**The red `InviteIntroducer` test.** Agreed on not merging a red main. The fix is #282, which should land first — it makes the shared-link queue FIFO, verified by reinstating the polling loop and watching a repetition harness fail at iteration 30.

Worth being precise about this branch's suite, though: it's green here (1402 tests, 0 failures), but that's **timing luck, not a fix** — my new tests shifted the bundle's scheduling again. That's exactly why #282 exists rather than leaving it to chance.

#282 also documents a second, narrower race it does *not* close, where `currentOrMint` legitimately arrives first and the rotation behind it revokes what the caller was handed. That one needs a product decision about whether a pending rotation should win regardless of arrival order.

### Verification

```
BackupRestoreTests   6 passed  (+2: path collision, failure copy)
OnymIOSTests         1402 tests, 3 skipped, 0 failures
Debug                ** BUILD SUCCEEDED **
```

dev-onym (commented):
REQUEST_CHANGES

**Verified before findings**: the three-gate ordering in `BackupRestorer` is real; the composition's blob directory matches `BackupSeat.restoredBlobDirectory()`; `downloadSnapshot` enforces `maxBytes: reference.sealedByteSize`; no dependency cycle from the new `OnymTransportBlossom` edge; `RestoredBlobClient` re-hashes on read with consistent lowercase hex.

- **"Nothing on this phone was changed" can be false.** Gate 3 is not atomic: `AppBackupSink.restore(consents:)` can throw (`Sources/OnymIOS/BackupSinks.swift:164,168`) and `restore(blob:)` can throw on a disk write (`BackupSinks.swift:188`) *after* groups, messages, and invitations were already written. That error lands in the flow's catch — whose comment claims nothing partial was written (`Packages/OnymBackupUI/Sources/OnymBackupUI/BackupRestoreFlow.swift:91`) — and renders the fallback copy at `BackupRestoreFlow.swift:129,131` plus the hardcoded line at `BackupRestoreView.swift:158`. The partial state itself is benign (`insertOrUpdate` converges on retry), but this PR's own standard is honest copy; say "some items may already have been added — restoring again is safe" for write-phase failures, or make the claim true.
- **Merging this turns the suite red.** The `InviteIntroducer` race analysis is credible and deferring the fix is right, but landing a known failure masks every subsequent regression. Add an `XCTSkip`/`XCTExpectFailure` with a tracking issue in this PR, or land the race fix first.
- **`testDownloadAndPlaintextDoNotShareAPath` doesn't pin its regression** (`Tests/OnymIOSTests/BackupRestoreTests.swift:163-167`): it only proves the restore completed, which also passes with a shared path wherever Foundation replaces the inode — i.e., on the platforms the tests actually run. Assert the sealed file's bytes are intact after `BackupOpener.open`, or that the two paths differ.
- **Restored ciphertext lives forever** (`Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift:26`, `BackupSinks.swift:188`): nothing prunes the `blobs/` directory, so restored attachment ciphertext outlives chat deletion and duplicates the loaders' own caches. Fine as a follow-up, but worth a tracking issue in a privacy-branded app.
- Minor: `DeviceBackupSettingsView.restoreSection` calls `makeRestore()` on every body evaluation, constructing a fresh `BackupRestoreFlow` per render (`Sources/OnymIOS/BackupComposition.swift:113-124`); harmless today because `@State` keeps the first one after push, but a throwaway-allocation trap.

rinat-enikeev:
Rebased on `main`, and all five in ee67052.

**The red suite.** Resolved by sequencing rather than a skip: the `InviteIntroducer` race fix landed on `main` as #282, and this branch is rebased on it. Nothing here is expected-failing.

**"Nothing on this phone was changed" could be false.** You're right, and it's the same standard I applied to `submitted` vs `retained` — promising more than we know. Gate 3 isn't atomic, so a consent or blob write throwing leaves groups, messages and invitations in place. The restorer now distinguishes a pre-write failure from a mid-write one, and the copy for the latter is *"some items may already have been added — restoring again is safe and will finish the job"*, which is both true and actionable given every write is `insertOrUpdate`.

**The path test pinned nothing.** Correct, and my first attempt at fixing it was wrong in exactly the way you'd predicted: I asserted `open(sealedURL: X, to: X)` **throws**, and it doesn't — on this platform Foundation replaces the inode and it succeeds, which is why the collision survived review in the first place.

So it pins the actual damage instead: after a same-path open, the sealed snapshot no longer contains the bytes it was handed — reading it destroyed it. Distinct paths leave it intact. That holds regardless of which semantics the platform has.

**Per-render factory.** Split into `canRestore` and `makeRestore`, so deciding whether to show a row no longer builds a flow and a repository to answer it.

**Restored ciphertext never pruned** — filed as #283 and referenced where the directory is written. Your framing is the right one for this product: "deleted the chat, kept the pictures" is the wrong default even when the bytes are ciphertext. It's only written by an `.includeCiphertext` restore, which isn't selectable yet, so the issue notes it must be fixed before that policy becomes reachable.

### Verification

```
BackupRestoreTests   7 passed  (+ partial-write reporting, real path-hazard pin)
OnymIOSTests         1403 tests, 3 skipped, 0 failures
```

dev-onym (commented):
COMMENT — the restore code itself is sound and I'd approve it, but this branch merges with a red test suite; land the `InviteIntroducer` tracking issue/quarantine first so `main` stays green.

- **CI turns red on merge** (`Tests/OnymIOSTests/InviteIntroducerTests`): the race is credibly pre-existing and well-triaged in the PR body, but merging as-is makes every subsequent PR's suite fail. Quarantine the test or land the fix first, then this can go in unchanged.
- `Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift:110` — the catch swallows the underlying error entirely. `restoreInterrupted` carries no cause, so a mid-write failure is undiagnosable from the screen or any log. Consider `restoreInterrupted(underlying:)` or logging before rethrowing.
- `Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift:102` — the catch also wraps a failure in the *first* write (`restore(groups:)`), reporting "partial" when nothing has landed yet. Overstating in the safe direction is defensible, but it contradicts the PR's own precision standard; checking whether any count landed would be cheap.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupRestoreFlow.swift:95-103` — two contradictory comment blocks: the first ("Nothing partial has been written") is the stale pre-fix claim the second one corrects. Delete the first.
- `Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift:33` — `Data(contentsOf:)` plus a full SHA-256 run synchronously inside an async method; for a restored video this blocks a cooperative-pool thread. Fine for now, worth a `Task.detached` or streaming hash if large media shows up.
- `Packages/OnymBackup/Sources/OnymBackup/RestoredBlobClient.swift` has no test — the hash-mismatch → fall-through-to-upstream branch is exactly the safety property the comment argues for, and it's unpinned.
- Verified the wiring claims hold: `BackupComposition.swift:92` (`workingDirectory/blobs`) matches `BackupSeat.restoredBlobDirectory()`; the sink creates the directory before writing (`BackupSinks.swift:180`); `bound(toServer:)` preserves the wrapper; `sealed-`/`restore-` paths no longer collide and the regression test pins the hazard, not the symptom.

Nice touches worth keeping: the empty state distinguishing "wrong operator" from "history gone", counts sourced from the sink rather than the archive, and the partial-failure honesty in `BackupRestoreView.failed`.

PR #284, opened 2026-08-20
Back up one history to several operators at once

rinat-enikeev:
A person could choose a backup operator; they could not keep a backup with two. Everything below the UI assumed one: a single `state.json`, a single consented manifest, and `rebind(to:)` — which is correct for *switching* operators and exactly wrong for *adding* one, since it discards the previous operator's chain the moment a second is enrolled.

What one operator cannot offer is the reason to want two. An operator can shut down, be seized, lose a region, or decide it no longer wants your business, and none of those is a reason to lose a history.

## The choice worth reviewing

**The snapshot is sealed once per operator, not once.** The archive root comes from the recovery seed alone, so one seal would be openable at every operator — tempting, and wrong. Identical bytes at two operators is a digest that links their two holders, undoing what per-operator key derivation and the profile's ban on convergent keying (`UI-Backup-Object-HTTP.md` §8.2) exist to prevent. So `BackupComposer` splits: `composeArchive` reads the history once, `seal` mints a fresh salt per operator.

The cost is disk — every operator's sealed copy exists at once during a run. That is the deliberate side to pay on: the alternative (seal, upload, seal the next) keeps the plaintext archive, the one seed-readable artefact, on disk for the length of every transfer.

## What changed

- **State is per operator**, in its own file, with the operator in the filename rather than a field inside it. Terms pins, chains, operation ids, entitlements, payments and access keys were already per operator; §14.12 says one operator never receives another's, and the only way that stays true is for none of them to have a shared home. A single-operator install's `state.json` is migrated on first launch — without it, backup reads as off for an operator already enrolled, and the next snapshot supersedes nothing and is paid for twice.
- **`BackupRepository.place(prepared:)`** is where bytes become addressed. Terms digest, chain position and operation id are stamped there because none is a property of the bytes: two operators enrolled at different times supersede different ancestors. `prepare()` lets the fan-out ask before it composes — a snapshot no operator will accept is a whole history read and sealed for nobody. Preconditions are extracted so all three entry points apply the same rules.
- **`BackupFanOut`** runs the order that matters: retry anything already sealed and refused for payment, ask each operator, compose once, seal per operator, destroy the plaintext *before* the first upload, then upload one at a time. One operator failing, refusing payment, publishing new terms or never having been set up does not stop the others.
- **The summary is pessimistic on purpose.** "On" means every enrolled operator is up to date; anything else names the number that are not. A backup surface that rounds up is the failure mode this stack exists to avoid. Restore reads across every operator, labels each snapshot with who holds it, and names operators it could not reach rather than showing a short list that looks complete.
- **Enrolment says what it is doing.** With another operator already set up, the consent surface leads with the fact that this adds a second complete copy — different company, different country, different terms, paid separately, and everyone in those chats has their messages kept in one more place (§14.11). Someone who thinks they are switching finds out before they consent.
- **Media policy is the strictest any enrolled operator was consented under**, because one archive is composed for all of them. The other direction would send an operator attachments the person agreed to give somebody else — a widening of the eligible set by composition order, which §14.10 forbids without an explicit choice.

## Testing

- `OnymIOS` app and the `OnymBackup` / `OnymBackupUI` packages build for the simulator.
- `OnymIOSTests`: 1403 tests, 0 failures. `BackupRestoreTests` updated for the new `sources:` initializer.
- A later serial run hit an unrelated CoreData crash in `VerdictValidatorTests`; reproduced on clean `origin/main` with these changes stashed, so it is pre-existing.

## Known gaps

- No tests for the fan-out, the per-operator state paths, or the multi-operator surfaces yet — they belong in a following PR.
- No erase action on the per-operator screen. The disclosure copy that would have promised one was removed rather than left pointing at a button that does not exist.
- The DEBUG `--backup-base-url` override applies to every operator, so a multi-operator debug run points them all at the same endpoint.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
@@ -46,13 +46,19 @@ public actor BackupComposer {
         supersedes: SnapshotReference? = nil,
         now: Date = Date()
     ) async throws -> SealedSnapshot {
+        // Minted before a byte is written, not between the seal and the
+        // return. The refactor had it second, where a failure would
+        // throw *after* `seal` had produced a full-size file — one that
+        // no state record mentions, so nothing claims it and nothing
+        // ever deletes it.
+        let operationId = try Self.newIdentifier()
         let archive = try await composeArchive(now: now)
         // Before the upload, not after: the plaintext is the one
         // artefact here that is readable without the recovery phrase.
         defer { archive.destroy() }
         let prepared = try seal(archive, archiveRoot: keyMaterial.archiveRoot)
         return prepared.addressed(
-            operationId: try Self.newIdentifier(),
+            operationId: operationId,
             acceptedTermsId: acceptedTermsId,
             supersedes: supersedes
         )
@@ -173,12 +179,30 @@ public actor BackupComposer {
     /// Age-guarded because a manual per-operator run can be sealing into
     /// this directory right now, and a sweep is not worth a race with
     /// it.
+    /// Two ages, because two kinds of file are at stake.
+    ///
+    /// `plaintextAge` covers everything readable without the recovery
+    /// phrase: the archive a compose writes, the scratch it builds it
+    /// in, and `restore-<digest>`, which is the *whole history in the
+    /// clear* and is cleaned today only by an in-process `defer` — a
+    /// force-quit or a jetsam during a several-hundred-megabyte restore
+    /// skips it. Keeping that for a day to avoid racing a concurrent
+    /// compose is the wrong trade; an hour is long enough that no live
+    /// run is touching it.
+    ///
+    /// `ciphertextAge` covers sealed bytes, where the cost of being
+    /// early is a re-upload and the cost of being wrong is deleting
+    /// something a purchase is waiting on.
     public func sweepWorkingDirectory(
         claiming claimed: Set<String>,
-        olderThan age: TimeInterval = 24 * 60 * 60,
+        plaintextAge: TimeInterval = 60 * 60,
+        ciphertextAge: TimeInterval = 24 * 60 * 60,
         now: Date = Date()
     ) {
-        let prefixes = ["pending-", "archive-", "scratch-"]
+        // `sealed-` is the restore path's download: ciphertext, so space
+        // only, and it leaks by the same route as `restore-`.
+        let plaintextPrefixes = ["archive-", "scratch-", "restore-"]
+        let ciphertextPrefixes = ["pending-", "sealed-"]
         guard
             let entries = try? FileManager.default.contentsOfDirectory(
                 at: workingDirectory,
@@ -189,7 +213,15 @@ public actor BackupComposer {
         }
         for url in entries {
             let name = url.lastPathComponent
-            guard prefixes.contains(where: name.hasPrefix), !claimed.contains(name) else { continue }
+            guard !claimed.contains(name) else { continue }
+            let age: TimeInterval
+            if plaintextPrefixes.contains(where: name.hasPrefix) {
+                age = plaintextAge
+            } else if ciphertextPrefixes.contains(where: name.hasPrefix) {
+                age = ciphertextAge
+            } else {
+                continue
+            }
             let modified = (try? url.resourceValues(forKeys: [.contentModificationDateKey]))?
                 .contentModificationDate
             guard let modified, now.timeIntervalSince(modified) > age else { continue }
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
@@ -46,22 +46,35 @@ public actor BackupComposer {
         supersedes: SnapshotReference? = nil,
         now: Date = Date()
     ) async throws -> SealedSnapshot {
+        let archive = try await composeArchive(now: now)
+        // Before the upload, not after: the plaintext is the one
+        // artefact here that is readable without the recovery phrase.
+        defer { archive.destroy() }
+        let prepared = try seal(archive, archiveRoot: keyMaterial.archiveRoot)
+        return prepared.addressed(
+            operationId: try Self.newIdentifier(),
+            acceptedTermsId: acceptedTermsId,
+            supersedes: supersedes
+        )
+    }
+
+    /// Read the whole history into one plaintext archive.
+    ///
+    /// The expensive half, and the half that is the same for every
+    /// operator: someone keeping their history with two operators reads
+    /// it once and seals it twice. The caller owns the result and must
+    /// `destroy()` it — including on the failure path, and before the
+    /// first upload rather than after the last.
+    public func composeArchive(now: Date = Date()) async throws -> BackupPlaintextArchive {
         try FileManager.default.createDirectory(
             at: workingDirectory,
             withIntermediateDirectories: true,
             attributes: [.protectionKey: FileProtectionType.complete]
         )
-        let operationId = try SecureRandom.data(16).map { String(format: "%02x", $0) }.joined()
-        let plaintextURL = workingDirectory.appending(path: "archive-\(operationId)")
-        let scratchURL = workingDirectory.appending(path: "scratch-\(operationId)")
-
-        // The plaintext archive is the one artefact on disk that is
-        // readable without the seed, so it is removed the moment the
-        // seal exists — including on the throwing path.
-        defer {
-            try? FileManager.default.removeItem(at: plaintextURL)
-            try? FileManager.default.removeItem(at: scratchURL)
-        }
+        let runId = try Self.newIdentifier()
+        let plaintextURL = workingDirectory.appending(path: "archive-\(runId)")
+        let scratchURL = workingDirectory.appending(path: "scratch-\(runId)")
+        defer { try? FileManager.default.removeItem(at: scratchURL) }
 
         let writer = try BackupArchiveWriter(scratchURL: scratchURL)
         writer.setIdentityCount(await source.identityCount())
@@ -71,37 +84,61 @@ public actor BackupComposer {
         encoder.outputFormatting = [.sortedKeys, .withoutEscapingSlashes]
         encoder.dateEncodingStrategy = .iso8601
 
-        let groups = try await source.groups()
-        try writer.append(kind: .groups, bytes: try encoder.encode(groups))
+        do {
+            let groups = try await source.groups()
+            try writer.append(kind: .groups, bytes: try encoder.encode(groups))
 
-        var messages: [BackupMessageRecord] = []
-        for group in groups {
-            messages += try await source.messages(
-                groupID: group.id,
-                ownerIdentityID: group.ownerIdentityID
-            )
-        }
-        try writer.append(kind: .messages, bytes: try encoder.encode(messages))
-
-        try writer.append(kind: .invitations, bytes: try encoder.encode(try await source.invitations()))
-        try writer.append(kind: .consents, bytes: try encoder.encode(try await source.consents()))
-
-        if mediaPolicy == .includeCiphertext {
-            let blobs = try await collectBlobs(from: messages)
-            if !blobs.isEmpty {
-                try writer.append(kind: .blobCiphertext, bytes: try encoder.encode(blobs))
+            var messages: [BackupMessageRecord] = []
+            for group in groups {
+                messages += try await source.messages(
+                    groupID: group.id,
+                    ownerIdentityID: group.ownerIdentityID
+                )
             }
+            try writer.append(kind: .messages, bytes: try encoder.encode(messages))
+
+            try writer.append(kind: .invitations, bytes: try encoder.encode(try await source.invitations()))
+            try writer.append(kind: .consents, bytes: try encoder.encode(try await source.consents()))
+
+            if mediaPolicy == .includeCiphertext {
+                let blobs = try await collectBlobs(from: messages)
+                if !blobs.isEmpty {
+                    try writer.append(kind: .blobCiphertext, bytes: try encoder.encode(blobs))
+                }
+            }
+
+            try writer.finish(writingTo: plaintextURL, createdAt: now)
+        } catch {
+            // A half-written plaintext archive is readable without the
+            // seed. It does not survive the throw.
+            try? FileManager.default.removeItem(at: plaintextURL)
+            throw error
         }
+        return BackupPlaintextArchive(url: plaintextURL, createdAt: now)
+    }
 
-        try writer.finish(writingTo: plaintextURL, createdAt: now)
-
-        let sealedURL = workingDirectory.appending(path: "pending-\(operationId)")
+    /// Seal one archive for one operator.
+    ///
+    /// Called once per operator, and each call mints a fresh snapshot
+    /// salt — so the same history sealed for two operators produces two
+    /// unrelated ciphertexts under two unrelated references. Handing both
+    /// operators the same bytes would be cheaper and would hand a pair of
+    /// colluding operators the link between their two holders.
+    ///
+    /// `archiveRoot` is derived from the seed alone and is therefore the
+    /// same at every operator: the snapshot must be openable by whoever
+    /// holds the recovery phrase, not by whoever stored it.
+    public func seal(
+        _ archive: BackupPlaintextArchive,
+        archiveRoot: SymmetricKey
+    ) throws -> PreparedSnapshot {
+        let sealedURL = workingDirectory.appending(path: "pending-\(try Self.newIdentifier())")
         let reference: SnapshotReference
         do {
             reference = try BackupSealer.seal(
-                plaintextURL: plaintextURL,
+                plaintextURL: archive.url,
                 to: sealedURL,
-                archiveRoot: keyMaterial.archiveRoot
+                archiveRoot: archiveRoot
             )
         } catch {
             // A half-written seal is ciphertext, so it is not a
@@ -111,14 +148,15 @@ public actor BackupComposer {
             try? FileManager.default.removeItem(at: sealedURL)
             throw error
         }
-        return SealedSnapshot(
-            operationId: operationId,
+        return PreparedSnapshot(
             snapshotReference: reference,
             sealedBytesURL: sealedURL,
-            sealedAt: now,
-            acceptedTermsId: acceptedTermsId,
-            supersedes: supersedes
+            sealedAt: archive.createdAt
         )
+    }
+
+    static func newIdentifier() throws -> String {
+        try SecureRandom.data(16).map { String(format: "%02x", $0) }.joined()
     }
 
     /// Fetch the ciphertext behind every attachment the messages
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupComposer.swift
@@ -155,6 +155,48 @@ public actor BackupComposer {
         )
     }
 
+    /// Delete working files nobody will claim.
+    ///
+    /// Written because a comment in `BackupRepository` claimed the
+    /// working directory "is swept on the next run" and nothing ever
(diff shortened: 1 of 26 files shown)


dev-onym (commented):
REQUEST_CHANGES

The architecture (compose-once/seal-per-operator, per-operator state, precondition extraction) is sound and matches the profile's isolation rules. But three places contradict the PR's own honesty guarantees:

- **Restore empty state overclaims on partial outage** — `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupRestoreFlow.swift:124` only fails when *all* sources are unreachable; with 1 of 2 unreachable and the reachable one empty, state becomes `.ready([])` and the view renders `empty` (`BackupRestoreView.swift:82`: "No operator you have set up holds anything") — the unreachable footnote lives only in `list()` (`BackupRestoreView.swift:116`), so it never renders on the empty branch. This tells someone their history is gone on the evidence of a network failure — the exact failure mode the PR names.
- **The "pessimistic" summary rounds up after a fan-out** — `place(prepared:)` returning `.termsChanged` or `.operatorChanged` writes nothing to local state, so `DeviceBackupSettingsFlow.refresh()` (DeviceBackupSettingsFlow.swift:139) shows `.idle` and `DeviceBackupVendorsFlow.summary` counts it healthy (`DeviceBackupVendorsFlow.swift:75-80`). The row subtitle surfaces only `.failed` run results (`DeviceBackupVendorsView.swift:130`), so an operator blocked on new terms reads "On · backed up \<date\>" until it happens to go stale, with no route to re-accept. `.ran(...)` blocked results should surface in status/subtitle like failures do.
- **Consent-order comment is wrong, and the bug it hides rebuilds the view mid-backup** — `PinnedConsentStore.accept` appends the fresh record at the end (PinnedConsentStore.swift:102), so re-consenting moves that operator's newest record to the tail; `consentedManifests`' final `.reversed()` (`Sources/OnymIOS/BackupSeat.swift:67`) yields store-position order, not acceptance order. Re-reading terms reshuffles the list and flips RootView's order-sensitive guard (`Sources/OnymIOS/RootView.swift:265`), rebuilding flows under a running backup — the very thing that guard exists to prevent. Sort by a stable key (e.g. first-acceptance date or componentId).

Minor:
- `Packages/OnymBackup/Sources/OnymBackup/BackupFanOut.swift:122` — a corrupt/unreadable state file reads as "not enrolled" and the vendor is silently omitted from outcomes; `DeviceBackupSettingsFlow.refresh()` treats the same condition as `.failed`. Distinguish "no state" from "unreadable state".
- `BackupRepository.swift:120` — `place(prepared:)` repeats the full `connect` + `reconcile` already done in `prepare()`, so a fan-out makes two network round-trips per operator per run. Acceptable, but worth a note or a cheaper re-check.
- `DeviceBackupVendorsView.swift:213` — "\(attention) of your operators needs attention" misconjugates for attention > 1.
- Known gaps (no fan-out/migration tests, shared DEBUG endpoint) are acknowledged; the fan-out ordering and migration edge cases are exactly where tests are needed before this ships.

rinat-enikeev:
Thanks — all three were real. Fixed in a67cd55.

**Restore empty state on partial outage.** Correct, and it was the exact failure mode the PR names. The unreachable note now renders outside both branches rather than inside `list()`, and the empty state says something different when an operator did not answer: "The operators that answered hold nothing for this identity. That is not the whole picture — one of them could not be reached, and what it holds is unknown."

**The summary rounding up.** Also correct, and the in-memory fix alone would not have been enough — so there are two changes:

- A run now records *why* it refused, in `BackupState.lastBlockedReason`, cleared only by getting past the check that set it. `termsChanged` and `operatorChanged` were the two refusals that left no other trace, so the fact died with the return value and every later launch read a device that last backed up successfully. Same rule `pendingOperations` already follows.
- The fan-out feeds each outcome back into that operator's flow (`DeviceBackupSettingsFlow.apply(_:)`), since it drives the repositories directly. `backUpNow` routes through the same method now, so the two mappings cannot drift.

Two consequences worth flagging: the refusal outranks a pending payment in `refresh()` (a purchase made before the terms moved buys nothing until somebody has read them, and "Payment needed" would send them to buy it again), and `summary` counts these as attention rather than "not set up" — uploads have stopped at an operator that *is* set up, and calling that Off reads as a choice the person made.

**Consent order.** Right on both counts: `accept` appends, so store position is not acceptance order, and my comment claimed otherwise. Sorted by componentId — `acceptedAt` on the active record moves for the same reason, and the earliest record is evicted past the history cap, so it is the only key here that does not move. The unused singular `consentedManifest` helper went with it; "most recently consented" is not a question this list can answer any more.

Minor: unreadable state no longer takes the not-enrolled exit in the fan-out (it reports `.failed` for that operator instead of vanishing from the outcomes); pluralisation fixed. The two round trips are documented as deliberate rather than made cheaper — the second is a small request immediately before a multi-hundred-megabyte upload, and it re-checks the terms pin against what the operator publishes *now* rather than what it published before the seal. Carrying a `BackupConnection` forward from `prepare()` would make the upload's precondition a stale one.

On tests: per this repo's convention they land in a following PR rather than this one, and I agree the fan-out ordering and the `state.json` migration are where they need to be pointed. Happy to bring that PR up before this merges if you would rather not ship the sequence untested.

Verified: app and packages build, `OnymIOSTests` 1403/1403 pass.

dev-onym (commented):
**REQUEST_CHANGES**

The multi-operator architecture is sound — per-operator state, per-operator seals with fresh salts, and the pessimistic summary all hold up under inspection. One state-machine bug blocks it:

- **Blocker — re-accepting terms doesn't clear `lastBlockedReason`, locking a single-operator install.** `BackupEnrolmentFlow.accept()` (`Packages/OnymBackupUI/Sources/OnymBackupUI/BackupEnrolmentFlow.swift:93`) writes the new terms pin but leaves `lastBlockedReason = .termsChanged` in state (`rebind` at line 85 is a no-op for the same operator). `refresh()` then returns `.termsChanged` before anything else (`DeviceBackupSettingsFlow.swift:130`), `needsEnrolment` routes straight back to the enrolment screen the person just completed, and with one operator `enrolledCount == 0` hides "Back Up Now" on the vendors screen (`DeviceBackupVendorsView.swift:34`) — the only interactive path to the fan-out run that would commit the cleared reason. `backUpIfDue` is not wired anywhere at app level, so nothing heals it: accepting new terms appears to do nothing, forever. Fix: clear `lastBlockedReason` in `accept()`.

- **`prepare()` clears the blocked reason in memory but never persists it** (`BackupRepository.swift:232`, comment says "the caller commits this" but the `.ready` branch at `BackupRepository.swift:153` discards the state). If compose fails after `prepare` says ready, the stale `termsChanged`/`operatorChanged` survives and the screen keeps reporting a refusal that no longer exists. Adjacent to the blocker; same fix family.

- **Silent narrowing for an `includeCiphertext` operator** (`Sources/OnymIOS/BackupComposition.swift:99`): with one shared composer, an operator the person explicitly consented to *with media* stops receiving attachments the moment a descriptors-only operator is enrolled — including via that operator's own per-screen "Back Up Now". Strictest-wins is the right §14.10 direction, but nothing on that operator's screen or disclosure says its backups no longer contain media. At minimum, surface it.

- **Doc/impl contradiction** (`Sources/OnymIOS/BackupSeat.swift:38`): header says "in the order the consents were accepted"; the function returns them sorted by componentId (deliberately, per the inline comment). Fix the header.

- **Nit — filename-privacy claim overstated** (`BackupVendorStorage.swift:29`): the digest is deterministic and unsalted, so anyone with container access can confirm enrolment with a known operator by hashing candidate componentIds. Fine as obfuscation; the comment shouldn't promise more.

Verified against source (not just the diff): `FileBackupStateStore.load()` does return empty-for-missing and throw-for-unreadable, so the fan-out's skip-vs-fail split is correct; the compose-once/seal-per-operator plaintext lifecycle destroys the archive on all paths including per-vendor seal failure.

rinat-enikeev:
Fixed in 29f8157.

**Blocker — the refusal that could not be resolved.** Correct, and it was introduced by the previous round's own fix: making `lastBlockedReason` durable was the point, and nothing cleared it on the one path that resolves it. Your trace is exactly right, including the part that makes it permanent rather than merely annoying — with one operator, `enrolledCount == 0` hides Back Up Now, so the run that would have cleared the reason is not reachable.

Cleared in `accept()`, and on principle rather than as a workaround: the reason a run refused was that nobody had read these terms, and somebody just has. That *is* the resolution, so it belongs at the point of resolution rather than deferred to whichever later run happens to fire.

**`prepare()` not persisting the clear.** Right — the comment described an intent the `.ready` branch dropped on the floor. It commits the state it was handed now, which matters precisely in the fan-out case you name: `prepare` runs before anything is composed, so a compose failure means there is no later write and a disproved refusal would survive to be reported again.

**Silent narrowing.** Fair, and "at minimum, surface it" is the right bar — the strictest-wins direction is the one I want to keep, so the fix is to stop it being quiet. The affected operator's screen now says: "You set this operator up with attachments included, but another operator you back up to was not. One backup is made for all of them, so attachments are currently left out of every one — including this operator's." The composition root already knows both policies, so it passes `attachmentsWithheld` to that operator's flow.

Not done, and worth naming: there is no way to *change* an operator's media policy from these screens, so the person can read that sentence and have nowhere to act on it beyond re-enrolling. That gap predates this PR (media policy is written once at enrolment) and I would rather fix it with the erase/settings work than bolt a control onto this.

**Doc/impl contradiction.** The stale comment described the bug rather than the fix — corrected to say componentId order, deliberately not acceptance order.

**Filename-privacy nit.** Agreed, and the comment overpromised. Rewritten to say it is obfuscation: deterministic and unsalted, so anyone reaching the container can confirm enrolment with an operator they can name; salting would mean storing the salt beside the files it hides. File protection is what closes that directory.

Thanks for checking `FileBackupStateStore.load()` and the plaintext lifecycle against source rather than the diff — the skip-vs-fail split depends entirely on that distinction holding.

Verified: build clean, `OnymIOSTests` 1403/1403.

dev-onym (commented):
REQUEST_CHANGES — architecture and state model are sound; one consent-surface claim is unverifiable and, by this codebase's own standard, blocking.

**Blocking**
- `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupDisclosure.swift:174` — the added-copy disclosure asserts the new operator is "held by a different company, in a different country, under different terms." Nothing verifies company or jurisdiction; two operators in the same country (or same parent company) make this a false statement on a consent screen — the exact failure mode §14.11 copy exists to prevent. Same phrasing in the `BackupEnrolmentView` headline. Soften to what is known (a second complete copy, separate terms, paid separately) or derive jurisdiction from the pinned manifests.

**Non-blocking**
- `Packages/OnymBackup/Sources/OnymBackup/BackupFanOut.swift:198` — the `.unknown` comment claims "the next run's reconciliation" needs the sealed file; `BackupRepository.reconcile` resolves purely via `listSnapshots`/`queryOutcome` and never re-reads the bytes, and no working-directory sweep exists despite `BackupRepository.swift:717` claiming one. Each `.unknown` strands a full-size sealed snapshot per operator, permanently. Leak is pre-existing on the single-operator path, but the fan-out multiplies it and the comment rationalizes it incorrectly.
- `Packages/OnymBackup/Sources/OnymBackup/BackupFanOut.swift:170` (retry path) — a vendor whose pending payment resolves is excluded from this run's fresh compose, then `refresh()` reads `lastSuccessAt = now` and reports it up to date while it actually holds the older snapshot. "Not re-charging" is the right call, but the summary rounds up here — consider surfacing "older copy retained; new snapshot next run."
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift:96` — doc comment says "hold at least one snapshot"; `enrolledCount` actually counts operators not needing enrolment.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift:141` — `backUpIfDue` has no caller anywhere (also true of the old per-flow variant on main), so the opportunistic path is dead code; worth a tracking issue so "backups only run while the app is open" stays honest.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/BackupRestoreFlow.swift:29` — the flow's class doc comment now attaches to `BackupRestoreSource` instead of `BackupRestoreFlow`.
- `Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsView.swift:99` — `SettingsRow` subtitles default to one line; "name · status · last run: \<error\>" will middle-truncate exactly when the error matters.

Verified clean: legacy migration (decoder matches `FileBackupStateStore`, failed-move and nil-componentId paths correct), per-operator supersedes chains, media-policy narrowing can never widen mid-session (enrolment stores the composer's own policy), `isRunning` serialization between fan-out and per-operator runs, and no stale references to removed APIs.

rinat-enikeev:
Fixed in 7ae2107. Every item taken, and one of them widened — no pushback on any.

**Blocking — the unverifiable claim.** You are right, and the fix is not a softening. Jurisdiction *is* knowable on both sides: the new operator's from the terms just fetched, the others' from `acceptedTermsRaw`, which is the exact bytes each was consented under and is kept for precisely this kind of question. Company ownership is not knowable by anything here, so the copy no longer mentions it.

So the sentence now states what is signed, including when that is the unhelpful answer:

> Both this operator and Foo store in EE, so one authority can reach both copies.

versus, when they differ, which one stores where. Silent when either side's terms did not decode — an unstated jurisdiction is not evidence of a different one. Someone adding a second operator for resilience should find out that both sit under the same authorities before they pay for it.

**The `.unknown` leak.** Confirmed on both counts: `reconcile` never re-opens the bytes, and the sweep the other comment promised does not exist anywhere in the package. So this was a permanent full-size leak per unresolved run per operator. Rather than only fix the comment I added the sweep it was describing — `BackupComposer.sweepWorkingDirectory`, run at the top of a fan-out, sparing the sealed bytes of anything awaiting payment, age-guarded so it cannot race a manual per-operator run. It collects the single-operator path's leftovers too. Both comments now describe what the code does.

**The resumed-payment row.** Agreed the exclusion from the fresh compose is right and the reporting was not. `Outcome` carries `resumedPayment`, and the row says "the backup you paid for went up; the newest changes go next time" rather than letting a fresh local success time imply this operator holds what the others just received.

**`backUpIfDue` dead code.** This one grew. The consequence was not only dead code: the consent surface said "Onym may also back up on its own while the app is open, on Wi-Fi" — a policy description on the one screen whose job is to say what will happen, for a behaviour that cannot occur. Someone reading it would reasonably never tap the button again. Copy is tap-only until something calls it, `testScheduleCopyDoesNotOverpromise` pins the honest claim, and the old `testScheduleSentenceTracksThePolicy` is replaced — it was pinning the wrong property, since tracking a policy is only a virtue when the policy runs. Tracking issue: #286, including what supplying `Conditions` would need and the open question of whether an opportunistic run should fan out to everyone or only the stale.

**`enrolledCount` doc, restore-flow doc placement, subtitle truncation** — all three fixed. The subtitle now gets two lines on the operator rows and three on the per-operator status row, since what renders there is a failure message.

#285 (restore purchases on a new phone) is rebased onto this.

Verified: build clean, `OnymIOSTests` 1403/1403.

rinat-enikeev:
Max-effort review results addressed in 0edc769 — 20 of the 21 findings fixed, one filed as #287. All but two were introduced by this PR.

**#1, the ratchet.** Right, and it is the one I would have shipped without noticing. `accept` deactivates records for the same componentId, so consenting to a second operator used to be a switch and is now an add — with nothing anywhere that removes one. Every operator a person ever chose stayed active, kept receiving a sealed copy on every run, and kept charging for it.

`PinnedConsentStore.withdraw(componentId:)` is the missing half, and **Stop Backing Up Here** is the route: *Erase My Backup and Stop* or *Stop Without Erasing*, as two choices rather than a fallback, because erasure is a request to a counterparty that can refuse and someone who wants to stop paying today should not be held there by an operator that will not answer. A failed erasure changes nothing and says so. Receipts survive the clear.

**#2, the unguarded window.** Confirmed exactly as described, including the `.alreadyRunning` swallow at the end. `prepare()` now holds a reservation through to `place(prepared:)`, `cancelReservation()` releases it on every path that never places, and the fan-out marks participating operators as running so the per-operator Back Up Now is not live during the seal.

**#4/#5, the sweep.** Both right, and #4 is the worse one: `restore-<digest>` is the entire history in the clear, cleaned only by an in-process `defer`, and my prefix list did not include it. Swept now, with a one-hour guard for plaintext against the day for ciphertext — the guard exists to avoid racing a live run, and the plaintext is the file whose stated invariant is that it does not outlive the run that made it.

**#11, the disclosure.** All three sub-points. It now fires only for an operator that holds nothing yet (re-enrolment no longer greets a months-old copy with "this adds a second copy"), counts instead of saying "a second", and normalises case and whitespace before comparing jurisdictions — and where the strings still differ it states both lists and says plainly that whether they are the same place is not something this app can check. "Stores your backup somewhere else" was a reassurance drawn from two free-text fields, which is exactly what that sentence must never be.

**#7, #8, #9, #13, #14, #15, #3, #6** and the gap sweep's six: all fixed as described. Three of them were the same mistake — `try?` on a store that throws deliberately — in the media-policy vote, the sweep's claim list, and the disclosure. Worth naming as a pattern: `FileBackupStateStore.load()` throwing rather than returning an empty state is load-bearing, and every `try?` on it converts a read failure into a confident wrong answer.

**#10** (the `@State` pin) was PLAUSIBLE and I treated it as real: the flows are owned by the composition root, so they are held as plain `let`s now. `@Observable` needs `@State` only for ownership, not for observation.

**#12 is the one I did not fix: #287.** The finding is correct — `mediaPolicy` has a fixed point at `.descriptorsOnly`, so attachment backup is unreachable and `attachmentsWithheld` is permanently false. It is pre-existing rather than introduced here, and wiring it needs decisions this PR should not make quietly: whether the choice is per operator (not expressible while one archive serves all of them) or per device, how the size cost is disclosed, and how widening it later avoids happening without an explicit choice. I left the narrowing logic in place rather than deleting it — it is correct the day a chooser exists. Say the word if you would rather it came out until then.

#285 is rebased on this.

Verified: build clean, `OnymIOSTests` 1403/1403 on both branches.

rinat-enikeev:
## Review: Approve

Read the full diff independently rather than taking the commit messages' self-assessment at face value, and re-checked each invariant this PR claims to hold against the actual code.

**Verified holding:**
- Fresh random salt per operator (`BackupComposer.seal()` called once per operator in `BackupFanOut.backUpAll()`) — no cross-operator digest linkage.
- Per-operator state files, componentId-hashed filenames, legacy single-operator migration fails safe rather than guessing.
- Plaintext archive destroyed before the first upload, not after all uploads.
- Pessimistic summary — `.on` is unreachable while anything is running/needs attention/not set up; no rounding up.
- Stop Backing Up Here — both erase-and-stop and stop-without-erasing are present and reachable.
- **Reservation window genuinely spans prepare→place** (not just the initial check): `isReserved` is set synchronously inside `prepare()` right after `commit()` with no intervening `await`, and released via `defer` in `place(prepared:)` on every exit path including throws; `cancelReservation()` covers the fan-out's other failure paths. This was the exact class of bug the 5th review round fixed, and it holds up under a second look.
- Strictest media policy across enrolled operators — vote only from operators with an accepted terms id, unreadable state votes conservatively rather than being skipped.
- Fail-closed on unreadable state in all three places that matter (fan-out vendor load, plaintext sweep claim list, media-policy vote).
- Consent list sorted by componentId, not acceptance order — avoids the reorder-mid-run hazard called out in the commit message.

**Two non-blocking notes, not asking for changes before merge:**
1. `BackupFanOut`'s reservation/cancellation logic and the media-policy vote have no new unit tests in this diff — given this is the part of the PR most likely to regress silently later (and the part that took 5 rounds to get right), a fast-follow test would be worth it. The PR description already flags fan-out/per-operator tests as a known gap for a following PR, so this isn't new information, just a vote to prioritize it.
2. In `BackupComposition.swift`, if `BackupKeys.material(...)` fails for any single consented manifest while building per-operator stacks, the whole multi-operator screen returns `nil`, including operators whose material derivation would have succeeded independently. Worth confirming this can only fail for "no recovery phrase" (identity-wide) and not per-component-id, since the code's comment assumes the former.

No blocking correctness or security defects found.

PR #285, opened 2026-08-20
Restore a purchase, and a backup, on a phone that never made either

rinat-enikeev:
Stacked on #284.

A person whose phone is gone restores their identity from the recovery phrase, opens Device Backup, and meets two dead ends.

## The purchase they already made was invisible

`Transaction.updates` replays only transactions that were never *finished*, and a purchase completed on the old phone was finished there. On the new one it exists solely in `Transaction.currentEntitlement` — which nothing was asking. `StoreKitPurchaseCoordinator` has had `currentTransaction(for:)` and `sync()` since #278, with no caller anywhere in the app. So a paid operator answered `payment_required` and the only route offered was to buy the same storage a second time.

`SeatPurchaseFlow.restorePurchases(componentId:)` sweeps the catalog's products for that operator, redeems what the store confirms, and stores the credential. Everything the broker needs is derivable from the recovery phrase — the subject is the seat access key for this operator, and the credential is sealed to the agreement key derived beside it — so none of it depends on the device that made the purchase.

Offers already covered by a credential that verifies *today* are skipped rather than re-redeemed. That also handles a lapsed subscription since renewed: the stale credential fails verification, so the offer is not counted as held and the sweep replaces it.

It runs two ways, and the difference matters:

- **Quiet, when the screen appears.** The person who needs this does not know to ask — nothing has been uploaded yet, so no operator has refused anything, and the refusal that would have mentioned a purchase has not happened. It announces nothing unless it has news.
- **A Restore Purchases row they can tap**, which asks StoreKit to sync first. Only the tapped path syncs: `AppStore.sync()` prompts for an App Store password, and a screen that demands one on appear is a screen people learn to back out of.

## Restore was hidden behind enrolment

The row lived inside the enrolled gate, so someone who had consented to an operator and set nothing up saw "Off" and no way to read the backup they came for. Reading one needs the recovery phrase and the operator's name, not a terms pin — `listSnapshots` and `download` never consult local state. Asking someone to agree to *store* a backup before they may read what they already have is an obligation taken on for nothing.

## Also

One factory now serves both the transaction observer and the sweep. A `SeatPurchaseFlow` pins the issuer key it will verify a broker's answer against, and that key comes from the operator's own signed manifest — there is no app-wide broker client, and a second copy of that rule is a second place for it to drift.

## Testing

- App and packages build; `OnymIOSTests` 1403/1403 pass.
- Not exercised end to end against a sandbox Apple ID on a second device — that path needs a real StoreKit environment, and it is the one I would most want covered before shipping.

## Known gaps

- Nothing in the recovery phrase records *which* operator held the backups. On a new phone the person must remember, or recognise it in the discovery list; with two operators set up, remembering either one recovers the history. Worth a line on the enrolment screen telling people to write it down — not in this PR.
- Tests for the sweep (already-held short-circuit, lapsed-then-renewed, broker failure reporting) belong in the tests PR that follows #284.

### DIFF
--- a/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift
+++ b/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupSettingsFlow.swift
@@ -32,6 +32,12 @@ public final class DeviceBackupSettingsFlow {
         case failed(message: String)
     }
 
+    /// Whether this operator is waiting on a purchase.
+    public var isPaymentRequired: Bool {
+        if case .paymentRequired = state.status { return true }
+        return false
+    }
+
     /// Whether the person needs to go through enrolment.
     ///
     /// `.off` alone was not enough, and the gap was the whole point of
@@ -294,6 +300,17 @@ public final class DeviceBackupSettingsFlow {
         case .alreadyRunning:
             break
         }
+    }
+
+    /// A credential arrived for an operator that is waiting on one.
+    ///
+    /// The status stays `paymentRequired`, because it is derived from a
+    /// snapshot still sitting unsent — a credential does not place those
+    /// bytes, only a run does. Saying so beats a screen that keeps
+    /// reading "Payment needed" with no explanation after someone has
+    /// just restored the purchase it was asking for.
+    public func noteRestoredPurchase() {
+        lastRunNote = "Purchase restored — tap Back Up Now to finish this backup."
     }
 
     /// Report that a run threw before it could reach a result.
--- a/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift
+++ b/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift
@@ -68,8 +68,16 @@ public final class DeviceBackupVendorsFlow {
         /// `heldUnknown` means this phone could not check what it
         /// already holds — different from finding nothing, and not
         /// something to report as an answer about the App Store.
+        /// `syncFailed` means the account-wide sync did not happen or
+        /// did not work, so nothing here is an answer about what the
+        /// App Store holds.
         case finished(
-            restored: Int, held: Int, checked: Int, heldUnknown: Bool, failures: [String])
+            restored: Int,
+            held: Int,
+            checked: Int,
+            heldUnknown: Bool,
+            syncFailed: Bool,
+            failures: [String])
     }
 
     private let fanOut: BackupFanOut
@@ -85,9 +93,12 @@ public final class DeviceBackupVendorsFlow {
     /// as "the store has nothing".
     private let restorePurchasesForOperator:
         (@Sendable (String) async -> SeatPurchaseFlow.RestoreResult?)?
-    /// Asks StoreKit to sync. Account-wide, and prompts for an App Store
-    /// password, so it happens once per tap and never per operator.
-    private let syncPurchasesWithStore: (@Sendable () async -> Void)?
+    /// Asks StoreKit to sync. Account-wide, and prompts for an App
+    /// Store password, so it happens once per tap and never per
+    /// operator — and it reports whether it worked, because a sweep
+    /// that follows a failed sync must not go on to say the App Store
+    /// has nothing.
+    private let syncPurchasesWithStore: (@Sendable () async -> Bool)?
     /// Guards the sweep itself, rather than what the screen is showing.
     /// The quiet sweep displays nothing, so a status-based guard did not
     /// hold it — and a tap during one started a second concurrent sweep,
@@ -100,7 +111,7 @@ public final class DeviceBackupVendorsFlow {
         schedule: BackupSchedule = .default,
         restorePurchasesForOperator:
             (@Sendable (String) async -> SeatPurchaseFlow.RestoreResult?)? = nil,
-        syncPurchasesWithStore: (@Sendable () async -> Void)? = nil
+        syncPurchasesWithStore: (@Sendable () async -> Bool)? = nil
     ) {
         self.vendors = vendors
         self.fanOut = fanOut
@@ -257,13 +268,27 @@ public final class DeviceBackupVendorsFlow {
         // prompts for a password: doing it per operator asked a person
         // with two operators to authenticate twice for one tap, the
         // second time seconds later and out of context.
-        if syncWithStore { await syncPurchasesWithStore?() }
+        // A sync that did not happen is not a sync that found nothing.
+        // Skipping it silently — or swallowing its failure — and then
+        // reporting "the App Store has no purchase for this identity"
+        // is a definitive answer to a question nobody managed to ask,
+        // told to the one person who most needs it to be right: someone
+        // on a replacement phone.
+        var syncFailed = false
+        if syncWithStore {
+            syncFailed = await syncPurchasesWithStore?() == false
+        }
 
         var restored = 0
         var held = 0
         var checked = 0
         var heldUnknown = false
         var failures: [String] = []
+        /// The operators a credential was actually recovered for. A
+        /// total is not enough: with two paid operators, one restore and
+        /// one failure, a total would tell the operator that got nothing
+        /// that its purchase had been restored.
+        var restoredFor: Set<String> = []
         for vendor in vendors {
             guard let result = await restore(vendor.id) else {
                 // Not checked: no pinned issuer, or nothing this build
@@ -274,6 +299,7 @@ public final class DeviceBackupVendorsFlow {
             }
             checked += 1
             restored += result.restored.count
+            if !result.restored.isEmpty { restoredFor.insert(vendor.id) }
             held += result.alreadyHeld.count
             // Named by operator: with several set up, "something went
             // wrong" without saying where is not something a person can
@@ -309,6 +335,7 @@ public final class DeviceBackupVendorsFlow {
                 held: held,
                 checked: checked,
                 heldUnknown: heldUnknown,
+                syncFailed: syncFailed,
                 failures: failures)
         } else {
             purchaseRestore = .idle
@@ -319,8 +346,9 @@ public final class DeviceBackupVendorsFlow {
         // is derived from a snapshot still awaiting payment, and only a
         // run that actually places those bytes resolves it.
         refresh()
-        if restored > 0 {
-            for vendor in vendors where vendor.flow.isPaymentRequired {
+        if !restoredFor.isEmpty {
+            for vendor in vendors
+            where restoredFor.contains(vendor.id) && vendor.flow.isPaymentRequired {
                 vendor.flow.noteRestoredPurchase()
             }
             // A credential is what lets a paid operator answer at all,
--- a/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift
+++ b/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift
@@ -62,7 +62,10 @@ public final class DeviceBackupVendorsFlow {
         /// separate because "nothing to restore" and "already set up"
         /// are different answers, and only one of them means something
         /// went wrong.
-        case finished(restored: Int, held: Int, failures: [String])
+        /// `checked` is how many operators were actually asked about.
+        /// Zero means nothing was, which is not a fact about the App
+        /// Store.
+        case finished(restored: Int, held: Int, checked: Int, failures: [String])
     }
 
     private let fanOut: BackupFanOut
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
d

rinat-enikeev (commented):
Reviewed the diff and the surrounding code, not a build — the `Transaction.updates` vs `currentEntitlement` diagnosis is right and is the actual bug, and the fix is in the right shape. Two things ship visibly wrong today, though.

## The Restore Purchases row always appears, and always says nothing to restore

`canRestorePurchases` is `restorePurchasesForOperator != nil` (`DeviceBackupVendorsFlow.swift:90`). In the app, `makeSeatPurchaseFlow` is a closure literal — never nil — so `purchaseRestorer` always returns non-nil and the row always renders.

And `Resources/ChannelOffers.json` is `[]`. So `catalog.offers(forComponentId:)` is empty for every operator, the sweep iterates nothing, and `RestoreResult` comes back all-empty. Every tap: an `AppStore.sync()` password prompt, then *"The App Store has no purchase for this identity to restore."*

The PR body says the row is absent when "this build cannot sell anything — a free operator, or a build with no store products", but neither condition is checked. The gate wants to be on the catalog — at least one vendor with a non-empty `offers(forComponentId:)`.

Worth noting the copy makes it worse rather than merely useless: someone who genuinely paid on another phone is told *the App Store* has no purchase, when the truth is that this build has no product mapping. That sends them to the wrong place to look.

## Swift error dumps are rendered as user-facing copy

`SeatPurchaseFlow.swift:143` stores `String(describing: error)`; the flow prefixes the vendor name; `DeviceBackupVendorsView.swift:139` returns it as the row subtitle.

`BillingError` doesn't conform to `LocalizedError`, so what a person reads is:

```
Onym Backup: brokerRejected(code: "invalid_transaction", message: Optional("…"))
Onym Backup: expired(expiresAt: 2026-08-20 10:00:00 +0000)
```

The comment above it — "The operator's own words, not a count" — has the right instinct and the wrong implementation. `brokerRejected` already carries `message`, which *is* the operator's own words; that's the thing to surface, with a mapped fallback for the rest.

## The quiet sweep isn't covered by its own re-entrancy guard

`DeviceBackupVendorsFlow.swift:127-128` — `.running` is set only when `announcingNothingFound` is true, which is the tapped path. The quiet sweep never sets it, so:

- quiet sweep in flight, person taps → guard passes, both run → two broker round trips per offer
- quiet sweep finishes second → line 148 writes `.idle` over the tapped run's `.running`, re-enabling the button mid-run

Setting `.running` unconditionally and keeping *announcing* as the conditional part would fix both, since the distinction being drawn is about what to say, not about what is happening.

## Smaller

- **`heldOfferIds` failure inverts the short-circuit silently** (`SeatPurchaseFlow.swift:112`). `(try? …) ?? []` means a throwing `seatSubject` or `store.load()` reads as "nothing held", so the sweep re-redeems every offer. That's the safe direction, but the doc-comment promises a short-circuit that has quietly stopped existing, and it costs a broker round trip per offer. "Couldn't tell" and "nothing held" are worth distinguishing.
- **`failures.first` over a `Dictionary`.** The flow maps `result.failures.values`, whose order is unspecified — with two failing offers the subtitle picks arbitrarily and can differ between runs. Sorting by offer id makes it repeatable.
- **`notPurchased` is populated and never read.** No caller consumes it and `isEmpty` ignores it. Either surface it or drop it.
- **`hasSweptPurchases` is per-flow-instance** and `makeDeviceBackupView()` builds a fresh flow per navigation, so "once per screen" is really once per *visit*. Cheap either way once a credential is held — but the doc-comment claims the stronger thing.

## A question about moving restore outside the enrolled gate

The reasoning is right that reading a backup needs no local terms pin. But a paid operator refuses `download` outside grace, so someone who consented and never paid can tap Restore From Backup and be refused by the operator. The new summary copy — "you can still restore an earlier backup below" — is a promise the operator may not keep.

Moving the row is still correct. The copy may want to be less certain, or the restore screen may want to surface the payment refusal in terms of the Restore Purchases row sitting right above it.

## What's right

Sync only on the tapped path is the correct call and the justification is the real one — a screen that demands a password on appear is a screen people learn to back out of. One factory for both callers removes a genuine drift risk in the issuer pinning. And skipping offers whose credential verifies *today* rather than merely exists is a nice way to get lapsed-then-renewed for free.

## On the testing note

The deferral of tests to the follow-up PR is reasonable for a stack, but the first two findings are exactly what a sweep test would have caught — an empty-catalog fixture for one, a subtitle assertion for the other — and both are visible to a person on the first build. Those two feel like they belong here rather than in the tests PR.

rinat-enikeev:
All eight fixed in 1411edd. The three high ones were mine and share a root cause — treating a StoreKit sweep as if it were cheap, local and single-threaded.

**1 — `AppStore.sync()` per operator.** Correct, and the two-operator case is the premise of this stack, so it would have hit the people the feature is for. The sync is hoisted out of the loop and runs once per tap; the per-operator closure only redeems now.

**2 — the quiet sweep took no lock.** Right that the guard was on the *displayed* state, which the quiet sweep deliberately never sets. There is a sweep-level flag now, separate from what the screen shows, so a tap during the on-appear sweep no longer starts a second one.

**3 — concurrent redeems.** Confirmed, including that the lost credential is unrecoverable because its transaction was finished when it was redeemed. `SeatEntitlementStoring.mutate` now takes a process-wide lock around the whole read-modify-write — the same shape `PinnedConsentStore` uses, for the same reason. You are right that the shape predates this PR; adding a redeemer that fires when a screen opens is what turns it from rare into ordinary, so it belongs here.

**4 — cancellation as failure.** Both halves fixed: cancellation is not recorded as a failure, and `hasSweptPurchases` is set only by a run that finished, so leaving the screen early no longer disables the automatic attempt for the life of the flow.

**5 — skipped vs checked.** The result carries a `checked` count. With nothing checked the row says "No operator here sells storage through this app" rather than making a positive claim about the App Store nobody asked.

**6 — `canRestorePurchases` never off.** Right, and both doc comments described behaviour the wiring did not implement. Gated on the catalog actually having offers for a consented operator, so a free-operator-only setup or a build with no `ChannelOffers.json` shows no row.

**7 — the comment claiming `refresh()` clears "Payment needed".** It cannot: that state is derived from a snapshot still sitting unsent, and only a run places those bytes. Dropped the claim, and the operator's row now says what actually remains — "Purchase restored — tap Back Up Now to finish this backup."

**8 — the sweep blocked the list.** Moved to its own `.task` so it runs beside `loadSnapshots` rather than in front of it, and it reloads the list itself in the one case where a recovered credential changes the answer.

Still the case, and worth repeating from the PR body: the cross-device redemption is not exercised against a sandbox Apple ID on a second device. That is the path I would most want confirmed before this ships.

Verified: build clean, `OnymIOSTests` 1403/1403.

rinat-enikeev (commented):
## Verdict: reject (request changes)

Re-reviewed at `1411edd3`, fetched and confirmed against the PR head. Eight findings fixed, and the three you took as your own were the right ones to prioritise — the per-operator `AppStore.sync()` in particular would have hit exactly the two-operator case this stack exists for. The `sellsOffers` gate is the correct fix for the always-present row: I initially misread it, because `canRestorePurchases` still reads `restorePurchasesForOperator != nil`, and the gate now lives in the composer. It's right.

One blocking defect remains, and three minors. It looks missed rather than dismissed — it wasn't in the set of eight.

## Blocking: a Swift enum dump is still the user-facing failure copy

The path is intact end to end at `1411edd3`:

```
SeatPurchaseFlow.swift:143   failures[offerId] = String(describing: error)
DeviceBackupVendorsFlow:276  failures += result.failures.values.map { "\(vendor.displayName): \($0)" }
DeviceBackupVendorsView:157  return first        // ← the row subtitle
```

`BillingError` still conforms only to `Error, Equatable, Sendable` — no `LocalizedError` — so what a person reads on the Device Backup screen is:

```
Onym Backup: brokerRejected(code: "invalid_transaction", message: Optional("…"))
Onym Backup: expired(expiresAt: 2026-08-20 10:00:00 +0000)
```

The comment directly above it says *"The operator's own words, not a count. A failure here is the difference between someone getting what they paid for and buying it twice."* That's the correct instinct and it's the argument for fixing this, not for shipping it — the sentence explains why this particular string matters more than most, and it's the one string in the flow rendered as a debugger would print it.

`brokerRejected` already carries `message`, which genuinely *is* the operator's own words. That's the thing to surface, with a mapped fallback for the other cases. It's a small change and it's the last thing between this and an approval from me.

## Minors, non-blocking

- **`failures.first` over a `Dictionary`** (`Flow:276`). `.values` has unspecified order, so with two failing offers the subtitle picks arbitrarily and can differ between runs. Sorting by offer id makes it repeatable — worth doing while you're in the same expression for the cancellation filter.
- **`heldOfferIds` failure reads as "nothing held"** (`SeatPurchaseFlow:112`, `(try? …) ?? []`). A throwing `seatSubject` or `store.load()` silently removes the short-circuit the doc-comment promises and re-redeems every offer. Safe direction, but "couldn't tell" and "nothing held" deserve to be distinguishable.
- **`notPurchased` is populated and never read.** No caller consumes it, and `isEmpty` ignores it — now more visibly odd next to the `checked` count you added for finding 5, which is the thing callers actually needed.

## On the sandbox gap

Agreed with your own framing, and I'm not blocking on it: cross-device redemption against a sandbox Apple ID on a second device is the path I'd most want confirmed before this ships. Worth saying that it's also the path that would have caught the copy defect above — the first time anyone sees that string is a redemption that fails on a device that didn't make the purchase.

Fix the error copy and I'll approve.

rinat-enikeev:
Fixed in 0440bac. You were right that it was missed rather than dismissed — I read the `failures` plumbing as carrying an operator's words and never checked what `String(describing:)` does to a `BillingError` on the way there.

**The blocking one.** `BillingError` conforms to `LocalizedError` now, and the sweep goes through it. `brokerRejected` prefers the broker's own `message` — as you say, it is the one case carrying words written for a person — and falls back to naming the code rather than swallowing it, since the code is what support will ask for. `purchaseFailed` distinguishes the two StoreKit outcomes that are not failures at all: cancelled, and waiting for approval. Anything that is not a `LocalizedError` gets a sentence rather than its type name.

**Failure ordering.** Sorted by offer id before the row picks one. Right that `Dictionary.values` made the subtitle arbitrary and unstable between runs of the same state.

**`heldOfferIds` throwing.** The result carries `heldUnknown`, and the row says "Could not check what this phone already has — try again". The sweep still re-checks everything, which is the safe direction, but "already set up" and "could not tell" are different answers and only one of them is a reason to stop looking.

**`notPurchased`.** Gone. Nothing read it, `isEmpty` ignored it, and `checked` is what callers actually needed.

**One thing your finding turned up that was outside this PR.** Grepping for the same pattern found `String(describing:)` reaching user-facing copy in three more places on the base branch — the backup status row twice, and the consent surface when an operator's terms cannot be fetched. Same defect, worse audience: `rejected(code: "quota_exceeded", message: Optional("…"))` shown to somebody trying to find out whether their history is safe. Fixed in e60dfac on #284, with `BackupError: LocalizedError` and a single `BackupCopy.describe` so a surface cannot forget to ask. `BackupRestoreFlow.describe` stays — its wording is restore-specific by design.

This branch is rebased on that.

On the sandbox gap: agreed, and your point that it is the same path that would have caught this copy is the strongest argument for doing it that anyone has made. The first person to see that string would have been someone whose redemption failed on a device that did not make the purchase — the worst possible moment to be shown a Swift enum.

Verified: build clean, `OnymIOSTests` 1403/1403 on both branches.

rinat-enikeev:
## Verdict: **REJECT - request changes**

I reviewed head `0440bac` against base `e60dfac`. The full `OnymIOSTests` suite passes locally: **1,403 tests executed, 3 skipped, 0 failures**. The earlier error-reporting changes are an improvement, but two correctness blockers remain in the cross-device, multi-operator path this PR introduces.

1. **The explicit StoreKit sync can be silently skipped or fail, followed by a false "no purchase" result.** The restore row is enabled when *any* stack sells offers ([BackupComposition.swift L216-L219](https://github.com/onymchat/onym-ios/blob/0440bac30566a4b70fb6797f04f9f10deb584e88/Sources/OnymIOS/BackupComposition.swift#L216-L219)), but the account-wide sync is routed through the lexicographically first operator. If that first operator is free or has no pinned issuer, `makePurchaseFlow` returns `nil` ([OnymIOSApp.swift L892-L904](https://github.com/onymchat/onym-ios/blob/0440bac30566a4b70fb6797f04f9f10deb584e88/Sources/OnymIOS/OnymIOSApp.swift#L892-L904)) and the sync becomes a no-op. Even when a flow exists, `try?` discards an `AppStore.sync()` failure ([BackupComposition.swift L396-L404](https://github.com/onymchat/onym-ios/blob/0440bac30566a4b70fb6797f04f9f10deb584e88/Sources/OnymIOS/BackupComposition.swift#L396-L404)). The subsequent sweep can therefore tell a new-device user that the App Store has no purchase even though the account was never successfully synced. Please decouple account-wide sync from any operator and propagate its outcome so a failed/skipped sync cannot produce a definitive no-purchase message.

2. **A purchase restored for one operator is attributed to every payment-required operator.** The loop aggregates only a total restored count ([DeviceBackupVendorsFlow.swift L262-L290](https://github.com/onymchat/onym-ios/blob/0440bac30566a4b70fb6797f04f9f10deb584e88/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift#L262-L290)); when that total is nonzero, every payment-required vendor receives `noteRestoredPurchase()` ([L316-L328](https://github.com/onymchat/onym-ios/blob/0440bac30566a4b70fb6797f04f9f10deb584e88/Packages/OnymBackupUI/Sources/OnymBackupUI/DeviceBackupVendorsFlow.swift#L316-L328)). With paid operators A and B, a successful restore for A plus no entitlement or a redemption failure for B makes B claim "Purchase restored." Please retain the component IDs that actually restored credentials and update only those vendor flows.

The existing suite has no test exercising either mixed free/paid sync routing or partial multi-operator restore attribution. I would add focused tests for both while fixing them.

rinat-enikeev:
Both fixed in 428411b, with the two tests you asked for.

**1 — the sync routed through an operator.** You are right, and the free-operator case makes it worse than a swallowed error: `makePurchaseFlow` returns `nil` exactly when the operator pins no entitlement issuer, so a free operator sorting first turned the sync into a silent no-op — and then the sweep reported a definitive "the App Store has no purchase for this identity" on the strength of it. I routed it through a `SeatPurchaseFlow` because that is where `syncWithStore()` happened to live, which was the wrong reason: `AppStore.sync()` needs no seat, no broker and no issuer.

It is built from a bare `StoreKitPurchaseCoordinator` now, so no operator's absence can skip it, and it returns whether it worked. The outcome is propagated: a failed sync does not stop the sweep — `currentEntitlement` still answers from what the device knows — but it does stop the sweep concluding anything, and the row reads "Could not reach the App Store — try again".

**2 — attribution.** Also right, and it is the worst possible row to be wrong on: with A restored and B failed, B told the person their purchase had come across while B is still the one owed a payment. The componentIds that actually restored are retained, and only those flows are noted.

**Tests.** `BackupPurchaseRestoreTests`, two cases, both pinning the cross-operator property rather than the mechanism:

- a restore for A leaves B's note untouched (this fails against the previous commit — B took the note from the total);
- a failed sync reaches the surface that renders the no-purchase claim.

Both were invisible in a single-operator build, which is the honest reason neither appeared until now, and is the argument for the tests being cross-operator by construction rather than parameterised over one.

Suite is 1405 tests, 0 failures. The sandbox-on-a-second-device gap is unchanged and still the thing I would most want exercised before this ships — it is also the only way either of these would have surfaced without your reading.

rinat-enikeev:
## Re-review verdict: **REJECT - request changes**

Reviewed current head `428411b` against base `e60dfac`.

The two prior blockers are fixed: StoreKit sync is now account-wide with its outcome propagated, and restored credentials are attributed only to the operator that received them. The two new regression tests pass. After regenerating the Xcode project, the complete suite also passes: **1,405 tests executed, 3 skipped, 0 failures**.

One blocking false-negative remains:

1. **An unverified current StoreKit transaction is still reported as no purchase.** `currentTransaction(for:)` converts every verification error to `nil` via `try?` ([StoreKitPurchaseCoordinator.swift L49-L55](https://github.com/onymchat/onym-ios/blob/428411b85b4a684b65341e2cb4d41ca3e86193b1/Packages/OnymBilling/Sources/OnymBilling/StoreKitPurchaseCoordinator.swift#L49-L55)), even though `jws(from:)` deliberately throws for `.unverified` ([L73-L83](https://github.com/onymchat/onym-ios/blob/428411b85b4a684b65341e2cb4d41ca3e86193b1/Packages/OnymBilling/Sources/OnymBilling/StoreKitPurchaseCoordinator.swift#L73-L83)). The restore sweep then treats that `nil` exactly like a product that was never bought ([SeatPurchaseFlow.swift L154-L161](https://github.com/onymchat/onym-ios/blob/428411b85b4a684b65341e2cb4d41ca3e86193b1/Packages/OnymBilling/Sources/OnymBilling/SeatPurchaseFlow.swift#L154-L161)). After a successful account sync, the UI can therefore state that the App Store has no purchase when StoreKit actually returned one but its verification failed. That recreates the duplicate-purchase risk this flow is designed to prevent and contradicts the method contract that failures are reported rather than swallowed.

Please preserve the distinction: make `currentTransaction` throwing (with `nil` reserved for no entitlement), catch the verification error per offer in `restorePurchases`, and add it to `failures` so the UI asks the user to retry instead of concluding there was no purchase.

rinat-enikeev:
Fixed in cfa5b65, exactly as prescribed.

`currentTransaction(for:)` is throwing, with `nil` reserved for the one thing it means — the store has no entitlement for this product. `restorePurchases` catches per offer and records the verification error in `failures`, so the row asks the person to try again instead of concluding there was no purchase.

The `try?` was undoing the throw on the line directly above it: `jws(from:)` refuses `.unverified` deliberately, and I collapsed that back into `nil` in the method whose caller spells `nil` as "never bought".

One addition beyond the letter of the finding. `transactionUnverified` is now its own `BillingError` case rather than `purchaseFailed(reason: "unverified: …")`. On the purchase path those two read alike; on the restore path they are opposites — a purchase that could not be read versus no purchase at all — and since these strings now reach a person directly, the sentence has to come down on the right side of that. It reads "The App Store's record of this purchase did not verify on this phone. Try again."

Two more tests, bringing this file to four: an unverifiable transaction survives to the surface as a failure rather than an absence, and the error sentences that reach a person say what happened — the broker's own words where it supplied them, its code where it did not.

Suite: 1,407 tests, 3 skipped, 0 failures.

Worth naming, since you have now found it three times in a row: sync-not-attempted, held-check-failed, transaction-unverified. Each was an unknown collapsed into a confident negative, each on the same screen, and each would have cost the same person the same second purchase. I have re-checked the remaining `try?`s in `OnymBilling` — the two left are in credential *filters* (`heldOfferIds` and `StoreKitEntitlementProvider`), where a credential that fails verification is genuinely unusable rather than merely unknown, and the outer failure of the held-check is already reported as `heldUnknown`. If you read either of those differently I would rather hear it now than find it as a fourth.

rinat-enikeev:
## Re-review verdict: **APPROVE**

Reviewed current head `cfa5b65` against base `e60dfac`.

The remaining blocker is resolved correctly:

- `currentTransaction(for:)` now reserves `nil` for genuine absence and throws on verification failure.
- `restorePurchases` catches that failure per offer and carries it to the UI instead of producing a false no-purchase result.
- The dedicated `transactionUnverified` copy is clear and does not expose implementation details.
- The earlier account-wide sync and per-operator attribution fixes remain intact.

Validation after pulling the branch and regenerating the Xcode project:

- `BackupPurchaseRestoreTests`: **4 passed**
- Full `OnymIOSTests`: **1,407 executed, 3 skipped, 0 failures**
- `git diff --check`: clean
- PR head remains mergeable and unchanged at `cfa5b65`

I found no remaining merge blocker. The untested sandbox-Apple-ID restore on a genuinely separate device remains worthwhile release validation, but it does not block this code review verdict.

PR #289, opened 2026-08-20
A way to reach the backup seat at all

rinat-enikeev:
## What was unreachable

The device-backup seat has been complete since #280 and reachable in
none of the PRs since. The chain breaks in one place:

- `BackupSeat.consentedManifests` reads pinned consent records whose
  `seat` is `storage.backup`.
- The only writer of a pinned consent record is
  `ModuleConsentAcceptance.accept`.
- The only caller of that is `ModuleConsentFlow`, which is only ever
  built inside a `DiscoveryModulePicker`.
- There were exactly three pickers — relayers, Nostr relays, Blossom
  servers — and none of them is for `storage.backup`.

So `consentedManifests` always returned empty,
`BackupSeatComposer.makeDeviceBackupView()` always resolved to `nil`,
and the Settings BACKUP section — gated on that view alone — never
appeared. The surface a person would consent from was gated on their
having already consented. Everything downstream (sealing, upload,
fan-out, restore, erase, purchases) worked and had no way to be
started.

This adds the fourth picker and the screen that reaches it.

## What is here

- **`BackupCatalogConsent`** — the picker's `apply` step, extracted the
  way `BlossomCatalogConsent` was so the one piece with behavior is
  unit-tested.
- **`BackupOperatorSettingsFlow` / `BackupOperatorSettingsView`**
  (OnymSettings) — the catalog half of the transport pickers, with the
  configured-list half removed.
- **`SettingsView`** — the BACKUP section now raises on *either*
  factory, and carries a Backup Operators row beside Device Backup.
- **`BackupConsentSignal`** — a consent raises it, `RootView`
  re-resolves.
- **`SeatManifestVocabulary`** — an explicit `storage.backup` entry.
- Tests for each, in the established style.

## The four judgement calls

**Apply writes nothing, and says so.** The transport pickers apply by
copying an endpoint into a repository, because their repository is the
operational source of truth for the seat. Backup has none: the pinned
record *is* the enrolment, read directly by `consentedManifests`, and
`BackupSeatComposer` builds one stack per record it finds. A second
copy anywhere else would be a second place for "who holds my history"
to drift from what was agreed — the Device Backup screen's Stop
Backing Up withdraws the consent and nothing else, which is right only
while consent is the record.

What apply does instead is **refuse**. It runs
`BackupOperatorManifest(manifest:)` — the same parse
`consentedManifests` runs over every pinned record — and translates
`BackupError.invalidEndpoint` into `ModuleApplyError.noUsableEndpoint`,
which the consent flow renders with its honest two-part message: your
consent was recorded, the service was not added. A backup manifest
with no read-write https endpoint is exactly as unusable as a Nostr
manifest with no `wss`, and without this check it would earn a
"Service added" screen followed by a Settings section that never
appears. Other refusals (unimplemented profile, missing declared
terms) are rethrown as themselves rather than dressed up as a missing
endpoint that is in fact fine.

**Entitlements pin from the manifest under review.** The shared
`makeEntitlements` resolves trusted issuers through
`BackupSeat.entitlementIssuers`, i.e. from the active consent record —
correct everywhere except on the one surface where that record does
not exist yet. On a first consent it answers with no issuers, so every
stored credential fails verification, no paid offer resolves as
entitled, and `canAccept` refuses a paid-only manifest outright.
Somebody restoring onto a new phone — who has already paid, and whose
credential is derivable from the recovery phrase — would find Accept
permanently disabled with nothing on screen to explain it. So
`makeBackupEntitlements` reads the issuers from the digest-pinned,
signature-verified bytes the person is being asked to agree to, via
`BackupOperatorManifest` rather than a second hand-rolled field read.
Any *other* component still falls back to its pinned record: a
manifest may only speak for itself, and a credential still never
nominates its own authority.

**The section, and making a fresh consent show up.** The BACKUP
section now appears when either factory is wired, so the picker is
reachable from the state everybody starts in. `RootView` previously
re-resolved on two events — the Settings tab being selected, and the
moderation consent cover closing. Backup consent is given from a sheet
pushed over the Settings tab, so neither fires, and the person who has
just agreed to back up their history would return to a Settings screen
with no Device Backup row on it. `BackupConsentSignal` (same shape and
same reason as `OnboardingRestartController`) is raised by the apply
step *after* the manifest is known usable, and `RootView` observes it.
The resolution itself is untouched: it still compares the consented
operators and rebuilds only when they changed, so re-reading an
operator's terms does not throw away a running backup.

**Vocabulary: explicit, no aliases.** `storage.backup` is written into
`acceptedManifestSeats` even though it accepts precisely what the
strict fallback would. The table is the readable list of every seat
this app will consent to, and a seat absent from it reads as one
nobody considered rather than one deliberately left on the default. No
aliases: the other three earn theirs from manifests really published
under an older spelling; this seat has never had one, and an invented
alias would be a second accepted spelling only a mislabeled catalog
entry could use.

## Verification

`xcodebuild build` and `xcodebuild test -only-testing:OnymIOSTests` on
an iPhone 17 simulator: **1421 tests, 3 skipped, 0 failures**.

### DIFF
--- a/Packages/OnymSettings/Sources/OnymSettings/BackupOperatorSettingsFlow.swift
+++ b/Packages/OnymSettings/Sources/OnymSettings/BackupOperatorSettingsFlow.swift
@@ -1,0 +1,81 @@
+import Foundation
+import Observation
+import OnymDiscovery
+import OnymFoundation
+
+/// `@Observable @MainActor` view-model for Settings → Backup
+/// Operators: the discovery catalog's `storage.backup` entries, and
+/// nothing else.
+///
+/// Thinner than the transport flows on purpose. `RelayerSettingsFlow`
+/// and friends pair the catalog section with a configured list they
+/// can add to, remove from and reorder, because their repository is
+/// the operational source of truth for the seat. The backup seat's
+/// source of truth is the pinned consent record itself — read straight
+/// out of the consent store by `BackupSeat.consentedManifests` — so
+/// there is no configuration for this screen to edit and no custom-URL
+/// field to offer: an operator is adopted by consenting to its signed
+/// manifest and dropped by withdrawing that consent, which is the
+/// Device Backup screen's "Stop backing up".
+///
+/// What is left is the catalog plumbing, and that is shared with the
+/// other pickers (`DiscoveryCatalogState`) rather than copied: the
+/// entries stream, so the section populates when the boot-time
+/// discovery refresh lands while the screen is open, and the memoized
+/// per-component consent lookups, so a row costs a dictionary hit
+/// instead of a full consent-store decode per render.
+@MainActor
+@Observable
+public final class BackupOperatorSettingsFlow {
+    /// Catalog-section state (entries + memoized consent lookups),
+    /// shared plumbing with the relayer/Nostr/Blossom flows.
+    private let catalog: DiscoveryCatalogState
+
+    /// Discovery-sourced `storage.backup` entries. Empty when the app
+    /// runs without discovery, or when no provider this person trusts
+    /// lists a backup operator — the screen says so rather than
+    /// pretending to be loading forever.
+    public var catalogEntries: [AttributedCatalogEntry] { catalog.entries }
+
+    /// Optional discovery seam, `nil` in builds without the discovery
+    /// stack (the UI-test harness) — the same seam the transport flows
+    /// expose, and the reason this screen degrades to an explanation
+    /// instead of a blank list.
+    let discovery: DiscoveryModulePicker?
+
+    public init(discovery: DiscoveryModulePicker?) {
+        self.discovery = discovery
+        self.catalog = DiscoveryCatalogState(discovery: discovery)
+    }
+
+    /// Begin draining discovery catalog updates. Idempotent.
+    public func start() {
+        catalog.start()
+    }
+
+    /// Paired with `start()`: the drain retains the flow and the stream
+    /// never finishes on its own, so without this the continuation
+    /// would accumulate on every visit.
+    public func stop() {
+        catalog.stop()
+    }
+
+    /// Re-read the discovery aggregate once (consent-sheet dismiss — a
+    /// fresh consent changes a row's badge with no catalog change to
+    /// push through the stream).
+    public func refreshCatalog() {
+        catalog.refresh()
+    }
+
+    /// The active pinned consent for a catalog entry's component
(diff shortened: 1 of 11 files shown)


dev-onym (commented):
**APPROVE** — the unreachability analysis checks out end-to-end; wiring, refusal path, and tests are sound. Two non-blocking follow-ups.

- **Verified the core claims against the code.** `ModuleConsentFlow` does pin the record before `apply` and renders `noUsableEndpoint` with the two-part message (`Packages/OnymDiscovery/Sources/OnymDiscovery/ModuleConsentFlow.swift:373-384`); `BackupError` has exactly the cases the tests assert; `AttributedCatalogEntry` is `Identifiable` so `.sheet(item:)` is valid; the retry-after-apply-failure guard (`ModuleConsentFlow.swift:317`) composes correctly with a write-nothing apply.
- **Follow-up 1 (real, edge-case): same-operator re-consent never reaches the running stack.** `resolveDeviceBackupView` skips the rebuild when component IDs are unchanged (`Sources/OnymIOS/RootView.swift:274-277`), and `consentedBackupComponentIds` maps to `componentId` only (`Sources/OnymIOS/OnymIOSApp.swift:1496`). This PR introduces the first path that changes a backup manifest's bytes under the same component ID — TERMS CHANGED → re-accept from Backup Operators — and after it the signal fires, the comparison sees no change, and uploads keep going to the *old* manifest's endpoint under the old terms until app restart. Comparing manifest hashes instead of component IDs would close it.
- **Follow-up 2 (minor, consent hygiene): a refused consent is unwithdrawable.** When `BackupCatalogConsent.apply` refuses (`Sources/OnymIOS/BackupCatalogConsent.swift:73`), the record stays pinned and active; the catalog row badges it CONSENTED, but no Device Backup stack is built for it, so the only withdrawal surface ("Stop Backing Up") never exists for that operator. Since consent *is* the enrolment for this seat, an active record with no withdrawal path deserves either a withdraw affordance on the catalog row or an unpin on refusal.
- **Nit:** `BackupOperatorSettingsView.swift:56` renders an empty sheet if `flow.discovery` is nil — unreachable in practice (no entries → no selection), fine to leave.
- Vocabulary entry, entitlement pinning from the manifest under review, the either-factory section gate, and the counter-not-flag signal are all correct as argued; tests cover each judgement call in the established style. Could not build here (Linux worktree); relying on the stated 1421-test run.

PR #290, opened 2026-08-20
Verify the terms signature that actually covers the bytes

rinat-enikeev:
Enrolment against a correct operator always failed with **"The operator's terms could not be verified"**, and Status stayed off. The operator was serving a valid document; the client was checking the wrong signature against the wrong message.

## The two signatures cover different bytes

- The document's **embedded `signature`** covers the canonical bytes with `termsId` and `signature` removed. It is what the pinned `termsId` is a digest of.
- The detached **`.sig` sibling** covers the **exact served bytes**, which are *not* canonical — they still carry both of those fields.

`connect()` fetched the detached `.sig`, passed it in as `signature`, and `verifyTermsSignature` checked it against the canonical message. That cannot verify for any honest operator.

Measured against the deployed `backup.onym.app`:

```
detached sig over RAW served bytes : VERIFIES
detached sig over CANONICAL bytes  : FAILS      ← what this code did
embedded sig over CANONICAL bytes  : VERIFIES   ← what it ignored
raw == canonical bytes?  False
```

## The fix

`BackupTerms` now carries `embeddedSignature` and `detachedSignature`, named for what they cover, and **reads the embedded one out of the document** rather than accepting it from a caller — so a call site cannot swap them again. That parameter confusion is the whole bug.

Both are verified, each against the bytes it signs. The detached one stays optional on the wire, but a served `.sig` that does not verify is refused rather than ignored: that is two things the same operator published disagreeing with each other, which is worse than an absent sibling.

## Tests

Four, including one that asserts the detached signature does **not** verify over canonical bytes — so nobody re-introduces this by concluding the two signatures are interchangeable. The fixture also asserts `raw != canonical`, so it fails loudly if it ever stops distinguishing the two messages.

**I ran only `BackupTermsSignatureTests` (4 passed), not the full suite.** The change touches `BackupTerms.decode`'s signature and one call site, so a wider run is worth doing before merge — `ConsentedTermsEnforcementTests` in particular constructs terms fixtures.

Note `./generate-xcodeproj.sh` must be re-run after pulling this, since it adds a source file and the `.xcodeproj` is gitignored.

## Not an operator bug

`backup.onym.app` is serving correctly: detached `.sig` verifies over the served bytes, the embedded signature verifies over canonical bytes, and `termsId` recomputes to its own content address. Nothing to change there or in the discovery catalog.

Worth noting onym-backup's own commit adding those routes predicted this exactly — *"the embedded signature covers the canonical document minus itself, so a verifier checking the detached one the same way would reject a valid document."*

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift
@@ -38,7 +38,17 @@ public struct BackupTerms: Sendable, Equatable {
     public let metadataRetention: MetadataRetention
 
     public let rawBytes: Data
-    public let signature: Data?
+    /// The document's own `signature` field, which covers the canonical
+    /// bytes with `termsId` and `signature` removed.
+    ///
+    /// Read from the document rather than accepted from a caller. The
+    /// detached `.sig` sibling served beside it signs *different bytes*
+    /// — the exact served ones — so the two are not interchangeable, and
+    /// checking one against the other's message rejects a document that
+    /// is perfectly valid.
+    public let embeddedSignature: Data?
+    /// The detached `.sig` sibling, which covers `rawBytes` verbatim.
+    public let detachedSignature: Data?
 
     public struct Retention: Sendable, Equatable {
         /// `measurable` or `best-effort`, verbatim.
@@ -114,7 +124,7 @@ extension BackupTerms {
     /// a partial value: terms are a precondition for enrolment
     /// (`UI-Backup.md` §13), and half-read terms are worse than none —
     /// they would be pinned and later compared against.
-    public static func decode(raw: Data, signature: Data? = nil) throws -> BackupTerms {
+    public static func decode(raw: Data, detachedSignature: Data? = nil) throws -> BackupTerms {
         guard let object = try? JSONSerialization.jsonObject(with: raw) as? [String: Any] else {
             throw BackupError.termsUnavailable
         }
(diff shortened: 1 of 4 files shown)


dev-onym (commented):
No Swift toolchain is available in this environment, so I verified by reading; the author ran the 4 new tests and CI should cover the rest.

**APPROVE**

The fix is correct and well-anchored:
- The security model holds: `termsId` is recomputed locally from canonical bytes (`Packages/OnymBackup/Sources/OnymBackup/BackupTerms.swift:173`, `:239-247`), so digest pin + embedded-signature-over-canonical binds content to the manifest's operator key; the detached check over `rawBytes` (`URLSessionBackupClient+Transport.swift:336-339`) is a genuine independent guarantee, as claimed.
- All other `decode(raw:)` call sites use the default parameter (BackupAdapterTests, BackupDisclosureTests, BackupComposition.swift:363), so the signature rename breaks nothing else.
- Reading `embeddedSignature` out of the document (`BackupTerms.swift:215-216`) does structurally prevent the swap from recurring, and the negative test (`BackupTermsSignatureTests.swift:96-101`) locks the distinction in.

Minor findings, none blocking:
- `URLSessionBackupClient.swift:130` — `Data(base64Encoded:)` is strict; a `.sig` served as base64 with a trailing newline (very common) decodes to nil, falls back to the raw text bytes, fails detached verification, and is now a *hard* `termsUnavailable` rather than tolerated absence. Trim whitespace before decoding.
- `URLSessionBackupClient+Transport.swift:336-340` — the new refuse-on-bad-detached-sig behavior has no test; the four tests cover the happy path and the swap, not the new strictness.
- `BackupTermsSignatureTests.swift:110-118` — `testAnUnsignedDocumentIsRefused` only asserts `embeddedSignature == nil`; it never exercises the refusal in `verifyTermsSignature`, so the name overstates what it proves.
- `BackupTerms.swift:215-216` — a present-but-malformed base64 `signature` field silently maps to nil while every other malformed field throws; same end state via `verifyTermsSignature`, but it's the one field where decode doesn't fail loudly.
- Full suite unrun (author flagged it; `ConsentedTermsEnforcementTests` builds terms fixtures) — run CI before merge.

PR #291, opened 2026-08-20
Keep a rolling window, so a full operator is not a wedged holder

rinat-enikeev:
A holder's third backup fails with `quota_exceeded` — "the operator has no room left for this backup" — and so does every backup after it. The deployed operator declares `maximumRetainedSnapshots: 2`, the client has never removed anything, and retention fills on the third run. From then on the newest thing the operator holds is the oldest backup that person ever made, for the life of the install.

`supersedes` does not help and was never going to. It is a recorded pointer and nothing more: nothing in the profile obliges an operator to erase what a newer snapshot supersedes, and the operator does not. Raising the declared limit moves the wall; it does not remove it.

**The client keeps the window, and the operator keeps its property.** `BackupRepository.makeRoom` runs before the preflight: it counts what the operator says it is holding, and if one more would exceed the declared limit it erases exactly enough of the oldest to fit — one snapshot, in the ordinary case. Each erasure goes out as an ordinary §11 erase and comes back as a signed receipt this device keeps. Doing it here rather than at the operator is the whole point: erasure stays holder-initiated, and no code path at the operator deletes a snapshot the holder did not ask it to delete. `BackupError.quotaExceeded` says as much in its own doc comment, and this is what makes that stay true.

**Counted, not caught.** `quota_exceeded` also covers the byte limit and the concurrent-grant limit, so treating it as "prune and retry" would erase somebody's backup to fix a problem that deleting backups does not solve. Counting first, against `GET /v1/snapshots` rather than the local state file, only erases when a slot is what is actually missing — and the local file is not the authority on what the operator holds: it misses what another device put there and keeps what the operator has since expired.

**Before the upload, not after the commit.** This opens a window: if the upload then fails, the holder is one snapshot down and has gained nothing. Two things make it the better side. What is erased is never their most recent backup — the newest retained row, and whatever the new snapshot supersedes, are both excluded, and it prunes not at all rather than partially — so throughout that window they still hold a newer copy of the same history at the same operator. And the alternative, pruning to `limit - 1` after a successful commit, permanently spends a slot: an operator declaring 2 would hold 1, one declaring 5 would hold 4, paid every day to avoid a transient loss on the runs where an upload fails. It would also still need this code for anyone already at the limit today, which is everyone this fixes.

A consequence worth naming: an operator declaring a limit of **1** is never pruned by this at all, because its only snapshot is also its newest. Such a holder still sees `quota_exceeded` and can erase deliberately. That is a worse experience than a rolling window and a better one than this device silently leaving them with nothing while an upload is in flight.

**An absent limit means no pruning.** `maximumRetainedSnapshots` is optional on the manifest, and absent is absent. Inventing a figure — a local default, or the terms' `snapshotsRetained` prose, which is a description and not a bound — would have this device erasing backups to satisfy a rule nobody published. Filling up an operator that never said it was full is the better failure: the refusal is at least the operator's own.

The limit rides out of `evaluatePreconditions` on the manifest the operator check already fetched, so this costs one `GET /v1/snapshots` per upload and no second `connect`. It is per operator throughout — one operator's limit says nothing about another's, and the fan-out runs each repository against its own.

Five new tests in `BackupAdapterTests`: at the limit, exactly one erasure and the upload lands; below the limit, nothing is touched; an operator declaring no limit is never pruned; a failed upload after pruning still leaves the most recent backup at the operator; and a limit of one erases nothing. `ScriptedPort` now performs erasure the way the operator does — the row stops being listed and a receipt comes back — so a test cannot claim room was made without the list getting shorter.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -84,12 +84,14 @@ public actor BackupRepository {
 
         var state: BackupState
         let acceptedTermsId: String
+        let retentionLimit: Int?
         switch try await evaluatePreconditions(now: now) {
         case .blocked(let result):
             return result
-        case .ready(let ready, let terms):
+        case .ready(let ready, let terms, let limit):
             state = ready
             acceptedTermsId = terms
+            retentionLimit = limit
         }
         let enrolled = state.componentId
 
@@ -101,7 +103,7 @@ public actor BackupRepository {
         )
         state.lastAttemptAt = now
         try commit(state, expecting: enrolled)
-        return try await place(snapshot, state: &state, now: now)
+        return try await place(snapshot, state: &state, retentionLimit: retentionLimit, now: now)
     }
 
     /// Place bytes sealed elsewhere in this run at *this* operator.
@@ -135,12 +137,14 @@ public actor BackupRepository {
 
         var state: BackupState
         let acceptedTermsId: String
+        let retentionLimit: Int?
         switch try await evaluatePreconditions(now: now) {
         case .blocked(let result):
             return result
-        case .ready(let ready, let terms):
+        case .ready(let ready, let terms, let limit):
             state = ready
             acceptedTermsId = terms
+            retentionLimit = limit
         }
         let enrolled = state.componentId
 
@@ -151,7 +155,7 @@ public actor BackupRepository {
         )
         state.lastAttemptAt = now
         try commit(state, expecting: enrolled)
-        return try await place(snapshot, state: &state, now: now)
+        return try await place(snapshot, state: &state, retentionLimit: retentionLimit, now: now)
     }
 
     /// Ask whether a run may proceed, without composing anything.
@@ -169,7 +173,7 @@ public actor BackupRepository {
         switch try await evaluatePreconditions(now: now) {
         case .blocked(let result):
             return .blocked(result)
-        case .ready(let state, _):
+        case .ready(let state, _, _):
             // Committed here, not left to the caller. A fan-out asks
             // every operator before it composes anything, and if
             // composing then fails there is no later write — the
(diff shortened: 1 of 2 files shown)


dev-onym (commented):
Review complete. I read the full `BackupRepository.swift`, the supporting types (`RetainedSnapshot`, `ErasureScope.wireValue`, manifest parsing), the new tests, and the `ScriptedPort` changes; checked the fan-out/retry call paths and existing tests for interaction with the new pruning (none trigger it — default stub limit 8, small retained lists). No Swift toolchain in this environment, so tests were verified by inspection, not execution.

**Verdict: APPROVE**

- `makeRoom` logic verified sound: surplus arithmetic, oldest-first sort with digest tie-break, newest+supersedes exclusions, all-or-nothing guard, and the limit-of-1 / no-limit / already-held / failed-list early-outs all behave as the PR body claims (`Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift:662-763`).
- Receipt can be silently dropped: if `commit` returns `false` (re-enrolment raced the run), the erasure has already executed at the operator but the receipt — "the only durable record that this happened" per the adjacent comment — is discarded, and the loop keeps erasing further rows whose receipts are also lost (`BackupRepository.swift:760-761`). Suggest `guard try commit(...) else { return }` so a failed commit stops further erasure; the existing `erase(scope:)` shares the pattern but there the person at least asked for the erase.
- The supersedes exclusion is effectively untested: in every new test the superseded row is also the newest held row, which `dropLast()` already removes — deleting the filter at `BackupRepository.swift:734-736` would pass the whole suite. A multi-device-shaped test (supersedes pointing at a non-newest row) would pin it.
- Minor coverage gaps: no test for the failed-list → no-pruning path (`setListFails` already exists in the stub), and none for `surplus > 1` / operator-lowered-limit refusing to partially prune (`BackupRepository.swift:740`).
- `try? await port.listSnapshots()` also swallows `CancellationError`, proceeding to preflight on a cancelled task (`BackupRepository.swift:683`) — harmless in practice since the next `await` throws, but worth knowing.
- When pending operations exist, a run now lists twice (reconcile at `BackupRepository.swift:784`, then `makeRoom` at `683`); the PR's "one GET per upload" holds only for the clean path. Not worth restructuring.
- Refactor of `operatorOrTermsMismatch` → `checkOperatorAndTerms` preserves behavior on all three call sites, including the mismatch bookkeeping in `evaluatePreconditions` (`BackupRepository.swift:245-262`); the dropped `now` parameter was unused.

PR #292, opened 2026-08-20
Name an identity after its recovery phrase, not after a dice roll

rinat-enikeev:
Restoring a backup after importing a recovery phrase produced a device
that held every row and showed none of them. QA saw the restore summary
report "1 chats, 50 messages" and the chat list stay empty across an app
restart, with the identity renamed to "Identity".

## What was wrong

`IdentityID` was a random UUID, minted by `IdentityID()` in
`IdentityRepository.addLocked` on **every** creation — importing an
existing mnemonic included. The entropy never touched the ID.

So importing a phrase minted identity B while the archive's groups and
messages carried identity A. Everything downstream is owner-scoped —
`MessageStore.list(groupID:ownerIDString:)`, `ChatGroup.ownerIdentityID`,
the chat surfaces — so nothing matched. The "Identity" name is the same
cause showing through `fallbackName(forNewSlot:)`; it was the only
visible tell that anything was wrong. (This PR does not fix the name.
That is archive-carried data and a separate change; the ID is what
unblocks the chat list.)

## The fix

The entropy is the one thing an import actually carries across devices,
so the ID now comes from it:

```
HKDF-SHA256(
    ikm:  BIP39 entropy,
    salt: "app.onym.bip39",
    info: "identity-id-v1",
    len:  16
)
```

with the version and variant bits stamped into the result.

**Domain separation is the point of the label.** This value is not
secret — it lands in a keychain service name and in plaintext SwiftData
rows — while the same entropy roots the Nostr and BLS secret keys.
HKDF's `info` is exactly the mechanism that keeps those worlds apart:
the ID is computationally unrelated to any key derived from the same
seed, so possessing it reveals nothing about them and it can never be
mistaken for, or fed in as, key material. The `v1` is there so a future
change of scheme is a new label rather than a silent reinterpretation of
the same bytes. The salt matches `Bip39.deriveNostrKey` /
`deriveBlsKey` so the whole seed hierarchy has one root salt and the
labels alone do the separating.

It takes the raw entropy rather than the PBKDF2 seed the secret keys
come from — both reproduce, but the entropy is one step further from the
key material and far cheaper for another platform to match.

**Both branches derive**, generated and imported. Today's new identity
is the one someone re-imports from a phrase in a year; randomising only
the generate path would have left the bug perfectly intact for every
identity that does not exist yet.

**The output is stamped v4** even though it is derived. `IdentityID`
round-trips through `UUID(uuidString:)` for keychain service names
(`app.onym.ios.identity.<uuidString>`) and persisted `ChatGroup` rows,
so it has to be a well-formed UUID or it fails to parse a long way from
here. RFC 9562's v8 ("custom") is the more literally honest nibble, and
it was rejected: it would make every derived ID visually distinguishable
from a legacy random one, printing "this identity is seed-derived" into
every service name and every stored row. Nothing downstream needs to
know which way an ID was minted.

Importing a phrase the device already holds now resolves to an ID that
is already loaded, and returns that identity instead of appending a
second `orderedIDs` entry that would double every broadcast summary.
Under random IDs it produced two indistinguishable identities with
identical keys.

`IdentityID()` stays available — tests mint hundreds of throwaway owner
IDs that never touch a mnemonic — but `addLocked` was its only
production call site, and its doc comment now says so, so a second one
has to argue for itself.

## No migration — read this before you test

**This is deliberate, and it has a consequence.** Identities that
already exist on a device keep the random UUID they were persisted
under. Nothing is rewritten. **A backup taken by one of those identities
still will not restore after a re-import.** Only identities created or
imported after this change are restorable.

If you test with pre-existing data you will see the old empty chat list
and conclude the fix does not work. Mint or import an identity on this
build first.

`test_load_preexistingIdentity_keepsItsPersistedRandomID` asserts the
no-migration property rather than leaving it assumed — a well-meaning
"fix up the ID on load" would silently orphan every row already keyed to
the old one.

## Why a seed-derived ID is safe here

A value derived from the seed would be a stable cross-service
correlator if it ever left the device. Audited independently: it does
not.

- **No wire payload carries it.** `ChatMessagePayload`,
  `ChatReceiptPayload`, `GroupInvitationPayload`,
  `GroupInviteOfferPayload`, `JoinRequestPayload`,
  `MemberAnnouncementPayload`, `GroupStateRefreshRequest`,
  `SealedEnvelope`, and the on-chain SEP types contain no identity
  field.
- **No invite link or QR carries it.** The deeplink payload in
  `IntroCapability` is exactly `intro_pub`, `group_id`, `group_name`.
  `InviteIntroducer.mint` takes an owner but uses it only for a keychain
  row and an in-process actor dictionary key.
- **`IdentityID` is not referenced at all** in OnymTransport,
  OnymTransportNostr, OnymTransportBlossom, OnymDiscovery, OnymChain,
  OnymBilling, OnymModeration, OnymRecovery, or OnymFoundation.
- **No telemetry**: there is no analytics or crash-reporting SDK, and no
  log site prints it.
- **The one network path is inside the seal.** `BackupArchiveRecords`
  carries `ownerIdentityID`, but that plaintext is AES-GCM sealed under
  a random per-snapshot salt before upload; preflight sends only a
  digest, size, and terms ID. The operator sees opaque, non-convergent
  ciphertext.

The `ownerIdentityID` fields in `BackupArchiveRecords` are the single
place a future change — a plaintext manifest, a preflight index, an
unsealed export — would turn this into a leak. Worth a note on whoever
touches that next.

## Android is not ready for this

Cross-platform restore (profile §18.18) needs both platforms to derive
the *same* ID from the same phrase. onym-android currently cannot:

- `IdentityId` is `IdentityId(UUID.randomUUID().toString())` —
  `modules/identity/.../IdentityId.kt:27`, its only mint site.
- Re-minted on every restore: `IdentityRepository.kt:303` (`restore`),
  `:356` (`add`), `:643` (`generateNewLocked`) — all
  `IdentityId.new()`, all independent of the entropy sitting on the
  line above.
- Worse, and independent of this PR: Android's restore sink writes
  archived `ownerIdentityId` values back verbatim
  (`app/.../BackupSinks.kt:65,97`) with no remap, while
  `GroupRepository.snapshots` filters by the active owner — so Android
  already orphans restored chats on its own platform, not just across
  platforms.

Android needs the same HKDF, the same salt, and the same
`identity-id-v1` label. `test_derivedFromEntropy_matchesFixture` pins
the value it must reproduce: the canonical BIP39 phrase must derive
`59B0B267-F746-46B7-AD6C-31863606156F`. Note that the existing
cross-platform fixture comment already records that the `chat.onym.*`
rebrand diverged the salts from onym-android, so key-derivation interop
is separately broken today.

No changes were made to onym-android.

## Tests

Added to `IdentityIDTests`: same entropy gives the same ID; different
entropy gives different IDs; a one-bit entropy flip changes the ID; the
derived value is a well-formed v4/RFC-4122 UUID that survives the
`uuidString` round trip; and a derivation fixture pinning the label and
salt against the canonical BIP39 vector.

Added to `IdentityRepositoryTests`: re-importing a phrase lands on the
same `IdentityID` across a wipe; two phrases do not collide; a
generated identity is reachable by its own phrase; importing the same
phrase twice does not duplicate the identity; and the no-migration
property above.

```
xcodebuild test -project OnymIOS.xcodeproj -scheme OnymIOS \
  -destination "platform=iOS Simulator,name=iPhone 17,OS=latest" \
  -only-testing:OnymIOSTests
```

`** TEST SUCCEEDED **` — 1431 tests, 3 skipped, 0 failures.

### DIFF
--- a/Packages/OnymIdentity/Sources/OnymIdentity/IdentityID.swift
+++ b/Packages/OnymIdentity/Sources/OnymIdentity/IdentityID.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 
 /// Stable opaque handle for one of the user's identities.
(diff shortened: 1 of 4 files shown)


dev-onym (commented):
APPROVE

Verified independently, not just read:

- **Fixture is correct** — I recomputed HKDF-SHA256(zero-entropy, salt `app.onym.bip39`, info `identity-id-v1`, 16 bytes) with the v4/variant stamping in Python and got exactly `59B0B267-F746-46B7-AD6C-31863606156F` (`Tests/OnymIOSTests/IdentityIDTests.swift:135`).
- **Domain separation holds** — `nostr-secp256k1-v1` / `bls12-381-v1` labels at `Packages/OnymFoundation/Sources/OnymFoundation/Bip39.swift:2199,2210` are distinct, and the keys take the PBKDF2 seed while the ID takes raw entropy, so IKMs differ too.
- **Audit claims hold** — no remaining production `IdentityID()` call sites and zero `IdentityID` references in OnymTransport*/OnymDiscovery/OnymChain (grep-verified).
- **Cache check is sound** — every `addLocked` caller runs `ensureLoaded()` first (`IdentityRepository.swift:70,93,168`), so the `cache[id] != nil` dedup at `IdentityRepository.swift:785` can't miss a loaded identity.

Minor notes, none blocking:

- `IdentityID.swift:80` — `SymmetricKey(data: entropy)` copies the entropy outside the `resetBytes` zeroization at `IdentityRepository.swift:783`; unavoidable with CryptoKit and consistent with `Bip39.deriveNostrKey`, but the zeroization is best-effort, not airtight.
- `IdentityRepository.swift:785` — the dedup early-return doesn't cover quarantined identities (distinct service prefix, `IdentityKeychainStore.swift:64`, never loaded into cache). Re-importing a quarantined phrase writes a fresh active item under the same derived ID while the quarantined copy lingers; harmless, and arguably the intended recovery path.
- `IdentityRepository.swift:169-175` — `restore` of the currently-held phrase still wipes and re-mints the snapshot, resetting a custom name to the fallback. Pre-existing semantic, not introduced here, but the new same-ID behavior makes it slightly more visible.
- The no-migration stance and its QA consequence are clearly documented and pinned by `test_load_preexistingIdentity_keepsItsPersistedRandomID`; Android divergence is honestly scoped out.

ux-onym (commented):
APPROVE

- `IdentityID.swift:78-94` — HKDF-SHA256(entropy, salt `app.onym.bip39`, info `identity-id-v1`, 16 bytes) with version/variant bits stamped is correct and matches the salt convention used by `Bip39.deriveNostrKey`/`deriveBlsKey` (`Bip39.swift:2198,2209`) and `deriveSeedScopedKey` (`IdentityRepository.swift:311`); the `info` label doesn't collide with any existing label (`stellar-ed25519-v1`, `x25519-key-agreement-v1`, `local-storage-v1`, `nostr-secp256k1-v1`, `bls12-381-v1`).
- `IdentityRepository.swift:784-785` — dedup check (`if cache[id] != nil { return id }`) sits after both the generate and import branches, so both paths get the same-identity-collapses-to-one-slot behavior described in the PR; `restore` (`:169-174`) clears the cache before calling `addLocked`, so it never hits this path spuriously.
- Confirmed via `IdentityID(` grep that every other call site in the repo uses the string-parsing initializer (`IdentityID(String)`/`IdentityID?`), not the random-UUID one — `addLocked` (`IdentityRepository.swift:784`) is indeed the sole production caller of `IdentityID()`, matching the doc comment's claim (`IdentityID.swift:17-26`).
- `add`/`select`/`bootstrap` (`IdentityRepository.swift:69-113`) correctly leave `currentID` untouched when `addLocked` returns an already-cached id and there was already a current identity — no accidental identity switch on a duplicate import.
- Test fixture (`IdentityIDTests.swift`) pins the canonical BIP39 test vector to a fixed UUID for Android parity; reasonable given the PR explicitly flags Android as not yet compatible.

No correctness, security, or test-coverage issues found — the no-migration behavior, dedup-on-reimport, and domain-separation reasoning are sound and well-tested.

PR #293, opened 2026-08-20
Show a restored chat to the person who restored it, and count what landed

rinat-enikeev:
QA restored a backup, read "1 chats, 3 messages", and watched the chat
list stay empty. The chat appeared later — but only once they created a
new one — and the thread it opened onto was still empty. The rows were
on disk and correct; #292 fixed that. What was broken is that nothing
which renders them knew.

## What was wrong

`AppBackupSink` writes through `GroupStore` and `MessageStore`. That is
deliberate: it is what re-encrypts every restored row under *this*
device's at-rest key rather than importing another device's database
file. The cost is that it bypasses the repositories that own the caches
the UI subscribes to.

- `GroupRepository` serves `filteredCache()` to every `snapshots`
  subscriber and calls `refreshFromStore()` only from its own mutators.
  A restore is not one of them. Creating a chat is — which is exactly
  why the restored chat surfaced then, and not before.
- `MessageRepository` caches per `(group, owner)` thread and yields
  `cached[key] ?? []` on subscribe. A thread whose cache was populated
  before the restore keeps serving what it held then, and since nothing
  else in the app writes those rows, it stays empty for good.

Two more things were wrong in the same family, and they are why this
took a QA pass to see at all.

**The counts were of calls made, not rows stored.** Both sinks did
`await store.insertOrUpdate(…)` and then `written += 1`, discarding a
`@discardableResult Bool` that meant precisely the thing being counted.
`SwiftDataGroupStore.insertOrUpdate` returns `false` having persisted
nothing when `try? Self.encode(group)` gives way. So "1 chats, 3
messages" read identically whether the rows had landed or not — the
direct contradiction of `BackupRestorer`'s own comment about counts
coming from the sink so the summary cannot "claim to have restored
things that are not there".

**"Already here" and "could not be read" were the same number.**
`restore(consents:)` returned only *newly added* records, `BackupRestorer`
computed `skipped = held - written`, and `BackupRestoreView` renders any
shortfall as "could not be read by this version of Onym". Every restore
carries consents the device already holds — you cannot reach an
operator's snapshot without having consented to that operator — so a
perfectly ordinary restore told someone four of their own consents were
unreadable.

### Making a restore visible

`BackupRestoreFlow` gains a `didRestore` hook, supplied by
`BackupSeatComposer`, that reloads `GroupRepository` and
`MessageRepository`.

**Not in the sink, and not in the restorer.** Teaching `AppBackupSink`
about repositories turns a thin store adapter into something that knows
about view state. Teaching `BackupRestorer` puts UI cache invalidation
inside the crypto and I/O it exists to do. The composition root is the
one place that already holds both the repositories and the restorer.

**Inside the flow rather than at the call site**, because the ordering
*is* the requirement: the refresh runs before `state = .restored`, so the
summary and the list behind it are read in the same glance and agree. The
flow is the one object that knows both that the write finished and that
the summary is about to be presented, so it can enforce that rather than
ask every caller to remember it. The parameter has no default value for
the same reason — a caller with nothing to refresh should have to write
`{}` and mean it.

`MessageRepository.reload()` is new, in the shape of the existing
`GroupRepository.reload()`. It refreshes **every** cached thread rather
than ones a caller believes were touched: threads never opened this
session are not in the cache at all and read the store on first
subscribe, so the cached ones are the entire exposure — and there are
only ever as many as the person has opened since launch. Narrowing it
would mean the restore handing over a list of written keys, which is
wrong the moment one record is skipped, and whose failure mode is a chat
that silently stays empty.

### Counting what landed

`BackupSinkOutcome` replaces the sinks' bare `Int`, carrying `landed`
and `unreadable` separately, and `skipped` in the summary is now what
the sink *said* was unreadable rather than what is left after
subtracting.

Honouring the return value alone would have moved the same conflation
one layer down: `GroupStore.insertOrUpdate` answers `false` for a row
updated in place **and** for one it could not encode, so counting
`false` as a failure would report every second restore — and every
restore onto a phone that has since received some of the same rows — as
unreadable. The sink reads the store's keys once before writing;
`false` for a key that was already present is the update branch, which
persisted, and `false` for one that was not is the encode guard, which
did not. `MessageStore` already returns `.inserted` / `.updated` /
`.failed`, so messages need no pre-read.

Invitations had the identical ambiguity — `InvitationStore.save` returns
`false` for a dedup hit and for a payload it could not encrypt — and are
handled the same way. Consents count every record that decoded, since
all of them are on the device afterwards, and reserve `unreadable` for
bytes that are not a consent record at all.

Widening `GroupStore.insertOrUpdate` into an outcome enum the way
`MessageStore` has one is the better shape and was **not** taken here:
it reaches into every caller of a hot path for the sake of the one
caller that is cold.

## Tests

In `Tests/OnymIOSTests/BackupRestoreTests.swift`, in the file's
established iterator-over-`snapshots` style:

- a restored group reaches a `snapshots` subscriber with no other
  mutation, subscribing *before* the restore so the cache is primed
  empty first;
- restored messages reach a thread cached empty before the restore;
- a sink whose store refuses the write counts nothing as landed;
- a consent already on the device is not reported as unreadable, and one
  that will not decode still is;
- a second restore of the same rows still counts them as landed.

The two visibility tests were checked against a build with the refresh
removed, and fail there — they are not vacuous.

`xcodebuild test -project OnymIOS.xcodeproj -scheme OnymIOS -destination
"platform=iOS Simulator,name=iPhone 17,OS=latest"
-only-testing:OnymIOSTests` — 1446 tests, 0 failures, 3 pre-existing
skips.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRestorer.swift
@@ -95,10 +95,10 @@ public actor BackupRestorer {
         // earlier writes have landed. Reported as `restoreInterrupted`
         // so the screen can say so, rather than inheriting the
         // pre-write promise.
-        let wroteGroups: Int
-        let wroteMessages: Int
-        let wroteInvitations: Int
-        let wroteConsents: Int
+        let wroteGroups: BackupSinkOutcome
+        let wroteMessages: BackupSinkOutcome
+        let wroteInvitations: BackupSinkOutcome
+        let wroteConsents: BackupSinkOutcome
         do {
             wroteGroups = try await sink.restore(groups: groups)
             wroteMessages = try await sink.restore(messages: messages)
@@ -111,14 +111,31 @@ public actor BackupRestorer {
             throw BackupError.localFailure(reason: .restoreInterrupted)
         }
 
+        // What could not be read is what the sink *said* could not be
+        // read, not what is left over after subtracting what it wrote.
+        //
+        // The subtraction was `held - written`, and it was wrong for
+        // every row that was already on the device. A restore's writes
+        // are `insertOrUpdate` precisely so that restoring twice, or
+        // onto a phone that has since received some of the same
+        // messages, converges rather than duplicating — and a sink that
+        // reported only *new* rows made every one of those convergences
+        // look like a parse failure. The consent list made it
+        // unmissable: you cannot reach an operator's snapshot without
+        // having consented to that operator, so the consents in the
+        // archive are always already held, and the screen said all of
+        // them "could not be read by this version of Onym".
+        //
+        // Only the sink knows which of the two happened, so only the
+        // sink may say. Arithmetic here cannot recover the distinction.
         var skipped: [String: Int] = [:]
-        for (name, held, written) in [
-            ("groups", groups.count, wroteGroups),
-            ("messages", messages.count, wroteMessages),
-            ("invitations", invitations.count, wroteInvitations),
-            ("consents", consents.count, wroteConsents),
-        ] where held > written {
-            skipped[name] = held - written
+        for (name, outcome) in [
+            ("groups", wroteGroups),
+            ("messages", wroteMessages),
+            ("invitations", wroteInvitations),
+            ("consents", wroteConsents),
+        ] where outcome.unreadable > 0 {
+            skipped[name] = outcome.unreadable
         }
 
         // Only a snapshot that meant to carry blobs can be missing any.
@@ -135,10 +152,10 @@ public actor BackupRestorer {
         }
 
         return BackupRestoreSummary(
-            groups: wroteGroups,
-            messages: wroteMessages,
-            invitations: wroteInvitations,
-            consents: wroteConsents,
+            groups: wroteGroups.landed,
+            messages: wroteMessages.landed,
+            invitations: wroteInvitations.landed,
+            consents: wroteConsents.landed,
             blobs: blobs.count,
             skipped: skipped,
             unresolvedBlobs: unresolved
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupSource.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupSource.swift
@@ -64,18 +64,54 @@ public enum BackupBlobAvailability: Sendable, Equatable {
     case gone
 }
 
+/// What one kind of row did on its way into local state.
+///
+/// Two numbers rather than one, because a single count cannot tell the
+/// difference between the two reasons a row is not *new* on this device,
+/// and the screen says something very different about each:
+///
+/// - `landed` — the row is on the device now. Freshly inserted, updated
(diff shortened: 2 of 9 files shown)


dev-onym (commented):
**COMMENT** — the fix is correct and well-placed; one gap in the same family it closes, plus two smaller notes.

- `Sources/OnymIOS/BackupComposition.swift:307` — `didRestore` reloads groups and messages but not invitations, and `AppBackupSink.restore(invitations:)` writes straight through `invitationStore`. `IncomingInvitationsRepository` has exactly the shape the PR indicts: a full `cached` roster refreshed only from its own mutators (`Packages/OnymInbox/.../IncomingInvitationsRepository.swift:20,83`), and `subscribe` re-reads only when `cached.isEmpty` — so a device with at least one existing invitation serves the pre-restore roster to live subscribers *and* to the next presentation of that screen, until relaunch. `incomingInvitations` is a sibling property of `groupRepository` at `Sources/OnymIOS/OnymIOSApp.swift:39`, so this is one more line in the composer and one more field on `BackupSeatComposer`. Worth doing here rather than as the next QA report.
- `Packages/OnymGroup/Sources/OnymGroup/GroupStore.swift:15` — the sink's whole inference now rests on `false` meaning *either* update *or* encode-failure, but the protocol documents only "true on insert, false on update". `SwiftDataGroupStore.swift:59` does return `false` from the encode guard, so the behaviour is right today; the contract the sink depends on should be written down where the next implementer of `GroupStore` will read it. Same for `InvitationStore.swift:54`, whose doc likewise omits the encrypt-failure `false` that `SwiftDataInvitationStore.swift:96` returns.
- `Sources/OnymIOS/BackupSinks.swift:68` — `present` comes from `groupStore.list()`, which is `compactMap(Self.decode)` (`SwiftDataGroupStore.swift:54`): a row on disk this build cannot *decrypt* is absent from `present`, so `insertOrUpdate` updates it in place, answers `false`, and the sink calls it unreadable — reporting "could not be read" for the one row the restore just made readable again. Narrow (needs a row encrypted under a key this install lost), and the inverse direction of the bug being fixed, so noting rather than blocking.

Smaller things, take or leave:

- `Packages/OnymChatsCore/.../MessageRepository.swift:177` — `refresh(_:)` already ends in `notifyChange()` (line 283), so `reload()` fires the coarse signal once per cached thread plus once more; the chat list recomputes every row each time. The trailing call is still needed for the zero-cached-threads case, so a `notifyChange()`-free inner refresh (or hoisting the notify) is the cleanup, not deleting the line.
- `Sources/OnymIOS/BackupSinks.swift:68,187` — both pre-reads run even for an empty record array, decrypting every group and invitation payload on the device for nothing. A `guard !groups.isEmpty` pays for itself on snapshots that carry no rows of that kind.
- The `didRestore`-has-no-default argument holds up, and the two visibility tests do subscribe before the restore, so they genuinely pin the ordering (`MessageRepository` caches `[]` as populated — `cached[key] == nil`, line 274 — which is what makes the thread test non-vacuous).

PR #294, opened 2026-08-20
Let the stores say which of the three things happened

rinat-enikeev:
Follow-up to #293. Reads as one with it.

#293 fixed a restore that counted calls made rather than rows stored,
and hit an obstacle on the way: `GroupStore.insertOrUpdate` answered
`Bool`, and `false` meant a row updated in place *or* a row it could not
encode and did not write. Those are opposite facts. Counting `false` as
a failure reports every restore onto a device that already holds rows as
unreadable — the very bug #293 was fixing — and counting it as a success
reports a store that refused the write as restored.

It worked around it in the sink, reading the store's keys once before
writing so `false` for a key already present is the update branch and
`false` for one absent is the encode guard. Correct, and tested. But it
reconstructs from the outside what the store knew from the inside, and
the next caller to read that `Bool` at face value finds the bug again.
`InvitationStore.save` had the same shape: `false` for a dedup hit and
`false` for a payload it could not encrypt.

`MessageStore` has answered `.inserted` / `.updated` / `.failed` since it
was written, which is exactly why the message path in #293 needed no
pre-read. The other two now do the same, and the pre-read is gone —
kept alongside it would be worse than either, two mechanisms answering
one question with nothing to say which is authoritative.

## Three enums, not one

All three modules hang off `OnymFoundation`, so a shared type was
available; the reason against it is not mechanical. The stores agree on
the *shape* of the answer, not on the middle case. A group `.updated`
was overwritten in place — the row now holds new values.
`InvitationSaveOutcome.duplicate` is the opposite: the store declined
the write so the original `receivedAt` and `status` survive, which is
the entire point of the dedup. One shared type has to pick one of those
words and then be wrong about the other store every time someone reads
it, which is the class of confusion being ended here.

What they do share is the part that mattered: "nothing was persisted"
can never again be spelled the same way as "persisted, just not as a new
row".

## The hot-path objection did not survive the audit

#293 declined this change on the grounds that it reaches into every
caller of a hot path for the benefit of one cold caller. Counting them:

| seam | production callers | what they do with the value |
| --- | --- | --- |
| `GroupStore.insertOrUpdate` | `GroupRepository`, `AppBackupSink` | pass through / count |
| `InvitationStore.save` | `IncomingInvitationsRepository`, `AppBackupSink` | branch on it / count |
| `GroupRepository.insert` | 9 sites across Inbox, Group, ChatsCore | **all nine discard it** |

Not one caller depends on `true`-means-insert. The single caller that
branches on the value at all is
`IncomingInvitationsRepository.recordIncoming`, which skipped its
snapshot refresh on `false`; that is now `guard outcome == .saved`, and
it is more correct than it was, because `.failed` used to take the same
path by accident rather than on purpose. There is no behavioural risk to
handle, only a mechanical migration.

Both repositories widen rather than flatten back to `Bool`, following
`MessageRepository.insert`, which already passes `MessageInsertOutcome`
through. Narrowing at that layer would move the ambiguity up one floor
and leave it for whoever next writes `if inserted`.

`GroupRepository.insert` refreshes unconditionally, including after
`.failed` — the snapshot is re-read from the store rather than patched,
so re-yielding what the store actually holds is right in all three
cases.

## Tests

New: each store's three outcomes distinguished by a caller, with an
explicit assertion that the persisted and the not-persisted answers are
not equal; a group updated in place counted as landed by the sink and
not duplicated in the store; a second restore of the same invitations
still counting them as landed; an invitation store that refuses the
write counted as unreadable.

#293's assertions are untouched. One line in its test file had to
change: `RefusingGroupStore.insertOrUpdate` spells its refusal `.failed`
instead of `false`, because the protocol it conforms to no longer has a
`Bool` to return. That is the change making the fake unable to mean
"updated" by accident.

`xcodebuild test -only-testing:OnymIOSTests` — 1451 tests, 3 skipped,
0 failures.

### DIFF
--- a/Packages/OnymGroup/Sources/OnymGroup/GroupRepository.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/GroupRepository.swift
@@ -30,11 +30,24 @@ public actor GroupRepository {
     /// `GroupStore.insertOrUpdate`). Any subsequent insert with the
     /// same id overwrites the row in place — the chain-anchor flow
     /// uses this to flip `isPublishedOnChain` and bump the commitment.
+    ///
+    /// Passes the store's `GroupInsertOutcome` straight through, the
+    /// way `MessageRepository.insert` passes `MessageInsertOutcome`.
+    /// Narrowing it back to `Bool` here would put the ambiguity the
+    /// store just shed one layer up, where the next caller to write
+    /// `if inserted { … }` finds it again — and every caller in this
+    /// repo discards the value anyway, so nothing is paying for the
+    /// wider type.
+    ///
+    /// The refresh is unconditional, including after `.failed`: the
+    /// snapshot is re-read from the store rather than patched, so
+    /// re-yielding what the store actually holds is correct in all
+    /// three cases and costs a list on a path that just did a write.
     @discardableResult
-    public func insert(_ group: ChatGroup) async -> Bool {
-        let inserted = await store.insertOrUpdate(group)
+    public func insert(_ group: ChatGroup) async -> GroupInsertOutcome {
+        let outcome = await store.insertOrUpdate(group)
         await refreshFromStore()
-        return inserted
+        return outcome
     }
 
     /// Mark a group as anchored on chain. The commitment, when
--- a/Packages/OnymGroup/Sources/OnymGroup/GroupStore.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/GroupStore.swift
@@ -1,5 +1,31 @@
 import Foundation
 
+/// Result of `GroupStore.insertOrUpdate`. Mirrors
+/// `MessageInsertOutcome` case for case, and deliberately is not the
+/// same type: the three stores agree on the *shape* of the answer, not
+/// on what the middle case means. A group `.updated` was overwritten in
+/// place — the row on disk now holds the new field values.
+/// `InvitationSaveOutcome.duplicate` is the opposite, a write the store
+/// declined so the original row survives untouched. A shared enum would
+/// have to pick one of those two words and be wrong about the other
+/// store every time someone read it, which is the class of confusion
+/// this type exists to end. (All three modules do hang off
+/// `OnymFoundation`, so a common type was available and was not the
+/// thing standing in the way.)
+///
+/// What every one of them *does* share is that the last case is
+/// distinct: "nothing was persisted" can never again be spelled the
+/// same way as "persisted, just not as a new row".
+public enum GroupInsertOutcome: Sendable, Equatable {
+    /// New row persisted.
+    case inserted
+    /// A row with the same `(id, owner)` already existed and its
+    /// sensitive + mutable fields were overwritten in place.
+    case updated
+    /// Nothing was persisted — encode gave way.
+    case failed
+}
+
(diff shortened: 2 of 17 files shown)


dev-onym (commented):
**APPROVE** — the migration is complete and behavior-preserving; the three-outcome case is well argued, and the removal of the pre-read (rather than keeping both) is the right call.

Verified independently:
- All 10 `GroupStore`/`InvitationStore` conformances are migrated (2 production, 8 test); no production caller depended on `true`-means-insert, so the "no behavioural risk" claim holds.
- `IncomingInvitationsRepository.recordIncoming:57` — `guard outcome == .saved` preserves the old `guard inserted` behaviour exactly (`.failed` skipped the refresh before too, and should: nothing was written).
- `GroupRepository.insert:48` unconditional refresh is safe — `refreshFromStore` re-reads, so `.failed` re-yields the truth.
- The removed `key(_:_:)` helper's `(id, owner)` concern is genuinely subsumed: `SwiftDataGroupStore` matches on the composite predicate, so a second identity's arrival returns `.inserted` (pinned by `SwiftDataGroupStoreTests:202-208`).

Findings:

- `Packages/OnymGroup/Sources/OnymGroup/SwiftDataGroupStore.swift:87,92` and `Packages/OnymPersistence/Sources/OnymPersistence/SwiftDataInvitationStore.swift:104` — the persisting branches are `try? context.save()` and then return `.updated`/`.inserted`/`.saved` unconditionally. If the save throws (disk full, migration conflict) the store reports persisted having written nothing, which is exactly the guarantee `GroupStore.swift:26-28` claims the enum makes. `SwiftDataMessageStore.insertOrUpdate:196-215` — the model this PR follows — does the opposite: `do/catch` with `context.rollback()` / `context.delete(encoded)` and `return .failed`. Not a regression (the old `Bool` had the same hole, and the sink's pre-read would have counted it as landed too), but the sink now rests entirely on the store's word, and this is the one path where the word can be wrong. Worth closing here or in an immediate follow-up, since matching MessageStore is a ~6-line change and leaves the enum's promise actually true.
- `Packages/OnymPersistence/Sources/OnymPersistence/InvitationStore.swift:63-65` — `.duplicate` is documented as "a row with this `id` already existed", and the store dedups on `id` alone, unlike the group and message stores' `(id, owner)`. So for a fan-out archive holding the same invitation id under two identities, the second restore counts as landed while that owner has no row. Behaviour is unchanged (the deleted pre-read keyed on `\.id` too), but the doc comment is now the authoritative statement of the contract — say "already existed, for any owner" so the next reader of `.duplicate` doesn't assume the composite key the sibling stores use.
- `SwiftDataInvitationStore.swift:94` — if `fetchCount` throws, `try?` falls through to the insert and returns `.saved`, which can mint a duplicate row and report it as new. Pre-existing, but the same `try?`-swallows-and-lies shape as the first finding.
- `Tests/OnymIOSTests/BackupRestoreTests.swift:436-475` — the two store-contract tests largely restate `SwiftDataGroupStoreTests:137-139` and `SwiftDataInvitationStoreTests:41-48`, which this PR already updated. The `XCTAssertNotEqual` framing is new and worth keeping; consider moving them next to the stores they specify rather than into the backup-restore file. Nit only.

PR #295, opened 2026-08-21
A chat you are waiting to be let into is a chat, not an inbox

rinat-enikeev:
The invitee side had a waiting room nobody visits. A pushed offer landed in `PendingInvitesStore` and was reachable only behind an envelope in the Chats toolbar; a group approved but not yet verifiable against chain state sat in `PendingVerificationStore`, deliberately hidden from the list and surfaced on that same sheet. Both are the same thing to the person waiting — a chat that hasn't opened yet — and neither was where they were looking for it.

So the sheet goes, and the wait becomes a row in the chats list. Tapping it opens a thread-shaped screen: the founder's invitation message, an Accept for an offer nobody has answered, and otherwise what is being waited on. When the founder approves and the invitation materializes the group, the pending row is consumed and the real thread takes its place — opening on the "You joined X" notice the dispatcher already mints. Nothing here writes that notice; this screen just stops existing.

**A pending chat is not a `ChatGroup`.** It carries no group secret, no roster and no epoch, and it must never reach `GroupRepository`: `currentGroups()` is the "do I already hold this group?" oracle for both `JoinRequestApprover` and `materializeGroup`, so a placeholder there would suppress the real materialization and take the join notice with it. The two kinds of row are merged in the view instead, where each fact keeps its owner.

**Keyed by (group, owner), and durable.** The old store keyed on the inbound Nostr event id and lived in memory, which was right when the offer was the only thing in it: a replaceable event is re-delivered on every launch. It stops being right once the row *is* the surface. A re-delivered offer must not come back asking to be accepted a second time, and — with the link/QR half landing in the next PR — a join the user asked for is replayed by nothing, so the row would be the only evidence they ever asked. `SwiftDataPendingChatStore` follows the `PersistedIntroRequest` posture: plain columns for what we filter and sort on, `StorageEncryption` for names, aliases and keys.

`GroupStateVerifier` and `PendingVerificationStore` are untouched. `PendingChatsFlow` overlays their statuses onto the matching row rather than mirroring them, so each state keeps one owner. A verification with no row of its own still gets one: that is the stale-invitation case the sheet existed to show, and without it the group would be stuck forever with no screen left to surface it.

Tests land in the last PR of the stack.


dev-onym (commented):
REQUEST_CHANGES — the architecture and the persistence split are right, but the pending thread dead-ends and a re-delivered offer can resurrect a row for a chat you're already in.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/PendingChatThreadView.swift:31` — when the row is consumed while this screen is open (the approval-lands case the PR body describes), the user is left on an indefinite `ProgressView` and has to tap Back themselves; nothing navigates to the real thread. `PendingChatsFlow.materialized` / `materializedGroupID(for:)` (`PendingChatsFlow.swift:96,187`) were built exactly for this and are read by nobody in this PR — so the "the real thread takes its place, opening on the 'You joined X' notice" claim isn't implemented, only enabled. Either consume it here (replace the route with the group id) or drop the dead map until the PR that uses it.
- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:287` — `recordOffer` writes unconditionally, with no "do I already hold this group?" check. Relays replay the inbox on every reconnect (the dispatcher says so at line ~470), and the only sweep is `PendingChatRepository.consumeForMaterializedGroups`, driven solely by `GroupRepository` emissions. So an offer replayed *after* the group materialized inserts a row that nothing clears until the next group mutation or relaunch — and it is now durable and rendered in the chats list next to the real chat, showing "Alice invited you" with a live Accept that re-sends a join request. Gate the record on `groupRepository.currentGroups()`, or sweep on insert.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatRepository.swift:73` + `SwiftDataPendingChatStore.swift:103` — rows are keyed `(group, owner)` but `deleteForGroups(hexes:)` matches on hex alone, while `groupIDHexes` comes from the identity-filtered `GroupRepository` stream. Identity A joining group X therefore deletes identity B's pending row for group X. Pass the owner too, or intersect against `cached` filtered by owner.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChat.swift:105` — `PendingChatWriteOutcome`'s stated reason for existing is that callers must distinguish `.alreadyPresent` from `.failed`, but the only caller (`recordOffer`) discards it, so an encryption/store failure drops the offer with no row, no retry and no log. Same shape makes `NoopPendingChatRecorder`'s `.failed` indistinguishable from a real disk failure.
- `Sources/OnymIOS/AppDependencies.swift:46` and `Sources/OnymIOS/OnymIOSApp.swift:730` — comments still describe the deleted surface ("Backs the Chats toolbar 'Invitations' badge + modal", "'Invitations' surface"). Nit, but these are the doc comments a reader lands on first.
- No tests for `SwiftDataPendingChatStore` round-tripping, the `.alreadyPresent` no-clobber invariant, or `PendingChatsFlow.rebuild`'s verification-wins overlay. Noted as deferred to the last PR of the stack; the dedup invariant is the one I'd most want landed with the code that depends on it.

rinat-enikeev:
All six addressed in `63be71c`.

**Replayed offer for a group you're already in** — real bug, and worse than described: the row is durable, renders beside the real chat, and its Accept would ship a join request for a group the user is in. Gated in `recordOffer`, scoped to the invited identity — another identity on this device being in the group says nothing about whether this one was invited. The comment there now names why the store's dedup can't cover it: after materialization there is no row left to collapse onto.

**Cross-identity delete** — real bug, and the same composite-key mistake `PersistedGroup` documents one layer down. `deleteForGroups(hexes:)` is now `deleteForIDs(_:)` and `consumeForMaterializedGroups` takes `(groupIDHex, owner)` pairs, matching on the same id `insert` dedupes on. Covered by tests in #297 that fail against the old matching.

**Dead `materialized` map** — moved to #296, where the screen that reads it lives. You were right that #295 shipped it enabled but unimplemented.

**`PendingChatWriteOutcome` discarded** — added `.notRecorded` so the no-op recorder can't read as a disk that gave out, and rewrote the doc to say where each case is consumed (the deeplink caller in #296 distinguishes all three; that PR now also acts on `.alreadyPresent`). The dispatcher still ignores it, and now says why: a pushed offer is a retained event replayed on every reconnect, so the next delivery *is* the retry — and there is nothing it may log, since an invitation is precisely the activity this app does not record.

**Stale comments** in `AppDependencies` and `OnymIOSApp` — fixed.

**Tests** — deliberately in #297, per the convention this repo uses (implementation PRs, then one test PR). The dedup invariant you'd most want landed here is `test_replayedOffer_doesNotResetAnAcceptedRow` plus the new `test_deleteForIDs_leavesTheOtherIdentitysRowForTheSameGroup`; happy to hoist them into this PR instead if you'd rather they travel with the code.

1487 unit tests green; `MultiIdentityChatUITests` green end to end.

dev-onym (commented):
**REQUEST_CHANGES** — solid architecture (pending rows never touching `GroupRepository` is the right call, and the durability rationale holds up), but the happy path dead-ends on screen.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/PendingChatThreadView.swift:29-33` — **the central flow dead-ends.** The stack has no path binding (`RootView.swift:308` is a plain `NavigationStack`) and nothing calls `dismiss()`, so when the founder approves while this screen is open the pushed route survives, `flow.row(id:)` goes nil and the user stares at a bare `ProgressView` forever. The PR body says "the real thread takes its place — opening on the 'You joined X' notice"; nothing implements that. At minimum `@Environment(\.dismiss)` on row-disappearance; ideally a bound path that swaps `PendingChatRoute` → group id.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatStore.swift:10-13` — never-overwrite is right for `.requested`, but it also drops a **re-invite carrying a fresh `introPublicKey`**. Since "Generate new link" revokes the old key (`OnymGroup/IntroKeyStore.swift:43-45`), Accept then seals to a dead key, `JoinRequestSender` reports `.sent`, and the row sits in `.waiting` forever with no error and no way to refresh. Consider updating `introPublicKey`/`groupName` when the existing row is `.offered` or `.failed`.
- `Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift:176-186` — `decode` returns nil if the *cosmetic* `encryptedInviterAlias` won't decrypt, silently deleting from view the row that this PR argues is "the only evidence they ever asked". Alias and group name should degrade to empty/nil, not drop the row.
- `Packages/OnymInbox/Sources/OnymInbox/PersistedPendingChat.swift:34` vs `:43` — `groupIDHex` is stored plaintext while `encryptedGroupID` holds the same value encrypted; the `id` doc on the same type argues encrypting a group id "would buy nothing". Zero confidentiality gain, one more fatal decrypt path. Derive the bytes from the hex.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:226,231` — `markFailed` persists an already-localized sentence, freezing the language on disk. This contradicts the `Row.name` rationale ("the view supplies the placeholder, so history isn't frozen into one language"). Persist a reason code; localize at render.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift:86,414` — the pitch now gates on `flow.items` (via `rows`), not `flow.groups`. `items` lags `recompute()`'s awaits (`ChatsFlow.swift:103-128`), so a first group landing renders the empty-state CTA for a frame or two instead of an empty list. `flow.groups.isEmpty && pendingChatsFlow.rows.isEmpty` avoids the flash.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:293-296` — the verification overlay wins unconditionally, so a row still in `.offered` that acquires a verification entry loses its Accept button with no way back.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:55` — `State.isRetryable` is public and unused; both views hand-roll the same switch.

Verified clean: no dangling `PendingInvites*` references, all new/reused xcstrings keys present in both locales (including the `%@`-interpolated ones), `retryVerification` wiring, and `GroupStateVerifier.resolve` still clears materialized entries so no duplicate row appears next to a real chat.

dev-onym (commented):
COMMENT — architecture is sound (no placeholder in `GroupRepository`, verification statuses overlaid not mirrored, `(group, owner)` keying); a few real gaps worth addressing before the stack lands.

- **Dismiss doesn't stick.** `ChatsView.swift:468` → `PendingChatRepository.remove` deletes the row, but a pushed offer is a retained event replayed on every reconnect and `IncomingMessageDispatcher.recordOffer` guards only on `alreadyJoined` — so `store.insert` returns `.inserted` and the dismissed row reappears in the main chats list. This was tolerable when the surface was a modal behind an envelope; now that the row *is* the chats list and the store is durable, swipe-to-dismiss is effectively a no-op across launches. Needs a tombstone (a `dismissed` status kept on disk, or a dismissed-key set) — otherwise the swipe action is a lie.

- **Contradictory comment, and the churn it denies is real.** `PendingChatRepository.swift:35-39`: line 35 calls `refreshOffer` on `.alreadyPresent`, and lines 36-38 then assert "No refresh on `.alreadyPresent` … would otherwise re-yield the whole list to every subscriber for no reason." `refreshOffer` ends in `refreshFromStore()`, so every replayed offer does a disk write + full `list()` + publish to all subscribers + `PendingChatsFlow.rebuild()`. Delete the stale comment, and consider skipping the write when `introPublicKey` and the descriptive fields are unchanged — that's the common case (plain replay), not a re-invite.

- **`refreshOffer` can silently erase fields.** `SwiftDataPendingChatStore.swift:111-112`: `try?` around `.map(encrypt)` collapses "encryption failed" into "no value", wiping `groupName` / `invitationMessage` on a row that had them — while `insert` treats the same failure as `.failed`. Encrypt into locals first and bail if either throws, as the lines above already do for key/alias.

- **Silent saves.** `SwiftDataPendingChatStore.swift:91,113,122,134,143` all use `try? context.save()`. For `setStatus` this diverges disk from view: `list()` reads the same context so the UI shows `.requested`, but after relaunch the row is back to `.offered` and asks the user to accept a request that already went out. Worth at least returning success from `setStatus` so `markRequested` doesn't claim a state it didn't persist.

- **Transient duplicate row.** `ChatsView.swift:426` merges `flow.items` with `pendingChatsFlow.rows`; both are driven by independent `groupRepository.snapshots` subscribers, so on materialization the real chat can render alongside its own pending row until `consumeForMaterialized` completes. Filtering pending rows whose `groupIDHex` is already in `flow.groups` would make it deterministic.

- Minor: `PendingChatThreadView.swift:15` says this screen "is replaced by the real thread" — it pops to the list; the user has to tap in. `PendingChatRepository.swift:126` `currentChats()` is unused until the deeplink PR (fine, just noting the untested `cached.isEmpty` path). Localization coverage checked — all new `String(localized:)` keys are present in `Localizable.xcstrings`, and no dangling `PendingInvite*` references remain.

PR #296, opened 2026-08-21
Tapping the link is the intent, so stop asking for it twice

rinat-enikeev:
Stacked on #295.

The join sheet asked a person who had just tapped an invite link to type a display name, tap Send, and then keep the sheet open while a spinner said "Waiting for the host to approve…". The name was pre-filled with the active identity's alias, the Send tap added nothing the link tap had not already said, and the wait behind it was in memory: a force-quit erased every trace that they had asked, with the sheet's own copy having told them to stay put.

So the sheet goes. A tapped link — and a scanned QR, which now takes the identical path — ships the request under the identity's own name and lands the user in the chat it created, waiting until the founder lets them in. Re-tapping a link the user already acted on is deliberately not a second request: an existing row is returned as it stands, and the Retry inside it is how you send again. A link into a chat they are already in opens that chat.

**The identity comes from the repository, not the flow.** A cold-start deeplink is handled before any tab's `.task` has populated `IdentitiesFlow`, so reading the UI's copy answered "no identity" for exactly the case that matters most: the link that launched the app.

**The path is hoisted into `ChatsNavigation`.** The link is handled at the app root, above the tab bar and above the navigation stack, while what it produces is a row inside the chats list — without a shared path the user taps a link in Messages, comes back, and finds that nothing visibly happened.

The UI test drops its form-filling and waits on the pending thread instead, which is a stronger assertion than the old one: reaching that state means the request actually left the device, where previously the test typed a name and dismissed a modal. It passes end to end — link tap → pending chat → founder approves in-thread → the real chat opens → messages and read receipts round-trip.

One trap worth naming, since it cost a run: SwiftUI pushes a container's `accessibilityIdentifier` down onto every descendant, so the id on the state stack silently renamed the very label the test queried. Leaves carry their own ids now and nothing above them competes.


dev-onym (commented):
REQUEST_CHANGES — the cold-start argument that justifies the `currentIdentityID` fix applies equally to the display name, which still reads the UI flow.

- **The name is still read from the flow, so a cold-start link joins anonymously.** `join()` → `send(chat)` reads `displayLabel()` (`Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:307`), which is wired to `identitiesFlow.identities.first { $0.id == identitiesFlow.currentID }?.name ?? ""` (`Sources/OnymIOS/OnymIOSApp.swift:791`). That is the exact read the PR body condemns — "reading the UI's copy answered 'no identity' for exactly the case that matters most" — and only `currentIdentityID` was moved off it (`OnymIOSApp.swift:799`). A link that launches the app fires `.onOpenURL` before the root `.task` finishes `bootstrap()` + `identitiesFlow.start()`, so the request ships with `joinerDisplayLabel: ""` and the founder sees an unnamed asker, not "the identity's own name." `IdentityRepository.currentIdentityName()` (`Packages/OnymIdentity/Sources/OnymIdentity/IdentityRepository.swift:226`) is the repository-side twin of the getter already used here — inject that the same way.
- **A tapped link does nothing when a pushed offer for the same group already exists.** `.alreadyPresent` returns `.waiting(rowID:)` without sending (`PendingChatsFlow.swift:255`). If the dispatcher already recorded the row as `.offered` (same `groupIDHex:owner` id), the user lands on a thread whose only content is "Accepting sends a request…" + an **Accept** button (`PendingChatThreadView.swift:122-135`) — asked for the intent a second time, which is the thing this PR removes. `JoinOutcome.waiting`'s doc ("the request is on its way") is also false in that branch. Fire `send` when the existing row is `.offered`; keep the no-op for `.requested`.
- **The link push never selects the Chats tab.** `handleCapability` only appends to `chatsNavigation.path` (`OnymIOSApp.swift:1997`), while `selectedTab` is private `@State` in `RootView` (`Sources/OnymIOS/RootView.swift:31`). Tap a link while the app was last on Settings and you come back to Settings with the pending chat pushed on an off-screen stack — "nothing visibly happened", the failure the hoisted path was meant to prevent. Hoist tab selection alongside the path, or have `ChatsNavigation` carry a "wants focus" signal `RootView` observes.
- **`join()` is untested, and 273 lines of join-state-machine tests went away with `JoinFlowTests`.** No test file references `PendingChatsFlow` at all. All four outcomes are cheap to cover against the in-memory doubles: no identity → `.failed`, already-a-member → `.alreadyJoined`, fresh → `.inserted` + one `submitJoin` call, second call → `.waiting` with no second submit. The UI test is a good end-to-end assertion but won't catch the label regression above.
- Minor: `.onChange(of: flow.materializedGroupID(for: rowID))` (`PendingChatThreadView.swift:49`) has no `initial: true`, so a route pushed after the group already materialized shows a `ProgressView` with no exit but Back — one word to close.
- Nit: the new key `"Couldn't use that invite"` uses a straight apostrophe while its neighbour `"Couldn’t save this invite on your device."` uses `’` (`Resources/Localizable.xcstrings:313`, `:330`) — the rest of the file is curly, and the literal is the lookup key.
- Nit: the alert `Binding(get:set:)` boilerplate is duplicated verbatim in `ChatsView.swift:218` and `OnymIOSApp.swift:1975`; `.alert(item:)` over an `Identifiable` reason, or one small extension, collapses both.

rinat-enikeev:
All seven addressed in `7f15ffb`.

**The name still read from the flow** — you're right, and the argument I wrote against that read applied to the line right next to it. `displayLabel` now asks `IdentityRepository.currentIdentityName()`, the repository-side twin you pointed at. A cold-start link was shipping `joinerDisplayLabel: ""`, so the founder was asked to admit an unnamed stranger.

**A tapped link did nothing over an existing offer** — fixed as suggested: send when the existing row is `.offered`, no-op when it is `.requested`. Landing someone on a screen asking them to Accept is the second ask this PR exists to remove. `JoinOutcome.waiting`'s doc no longer claims the request is always on its way.

**The push never selected the Chats tab** — `ChatsNavigation` carries a `focusRequests` counter and `RootView` watches it. A counter rather than a flag so two links are two requests and neither needs clearing by whoever handled it. `replaceTopWithChat` deliberately doesn't bump it: that only happens while the person is already looking at the screen being replaced.

**`initial: true`** — added.

**Apostrophe** — the literal is the lookup key, so a straight one is a silently missing translation. Both the key and the two call sites are curly now.

**Duplicated alert bridge** — extracted as `View.reasonAlert(_:reason:)` in `OnymDesign`; both call sites use it.

**`join()` untested** — the four outcomes are covered in #297 (`test_join_*`), which is where this stack keeps its tests. #297 also gained `test_join_onAnUnansweredOffer_sendsInsteadOfAskingAgain` for the branch you found; it fails against the version in this PR before the fix. You're right that the UI test wouldn't have caught the empty-label regression — nothing asserts on `joinerDisplayLabel` end to end, and the accept-path test asserting `label == "Bob"` only covers the flow's side of it.

dev-onym (commented):
REQUEST_CHANGES — the join-on-tap direction is right, but the shared path has an unguarded top-of-stack assumption, a lost request has no recovery, and the logic that replaced 273 lines of tested state machine has no unit tests.

- **`replaceTopWithChat` pops whatever is on top, not the caller** — `ChatsNavigation.swift:58`. `PendingChatThreadView`'s `onChange` fires while the view is still alive but *buried*: tap link A (path `[pendingA]`), then tap a second link that resolves to `.alreadyJoined` or another pending (`openChat`/`openPending` append at `ChatsNavigation.swift:34,40`, reachable any time from the app root). Path is now `[pendingA, X]`; when A materializes, `removeLast()` deletes X — the screen the user is actually reading — and pushes A's group. Two rows materializing in one snapshot compounds it. Guard the replacement on the top element actually being `PendingChatRoute(id: rowID)`.
- **A request that is sent but never answered is a dead end** — `PendingChatThreadView.swift:141` renders `.waiting` with no action, `State.isRetryable` excludes `.waiting`, and `join()` deliberately skips re-sending for a `.requested` row (`PendingChatsFlow.swift:270`). So a revoked link or a dropped request leaves the row waiting forever: re-tapping the link is a no-op, and the only escape is swipe-deleting the row and re-tapping. `JoinFlow`'s 90s `.unanswered` + "Try again" (deleted, `JoinFlow.swift:141-155`) was the one thing covering this; nothing replaced it. A staleness threshold on `receivedAt` that flips `.waiting` → retryable would restore it.
- **`materialized` is recorded from `rows`, which can be empty when the group snapshot arrives** — `PendingChatsFlow.swift:170-181`. This is the exact ordering `consumeForMaterialized` documents and defends against by reading through the store (`PendingChatRepository.swift:73-77`); the new loop has no such guard, so the row gets consumed with no mapping recorded and `PendingChatThreadView`'s fallback `ProgressView` (`:37`) becomes an exit-less spinner. Keying off the raw `pending` snapshot — or recording `groupIDHex → groupIDHex` and resolving lazily — closes it.
- **No unit coverage for the new join path** — `PendingChatsFlowTests` doesn't exist anywhere, and this PR deletes `JoinFlowTests.swift` (273 lines). Untested and non-obvious: `.alreadyJoined` owner matching, the `.alreadyPresent` + `.offered` auto-send, `.failed`/`.notRecorded` → alert, and `materializedGroupID`. The UI test is end-to-end only and can't reach the branches.
- Nit: `JoinRequestSender.swift:18` still references "the JoinScreen UI (PR-7)", now deleted.

Verified as correct: `PendingChat.id` is deterministic (`groupIDHex:owner`), so the `.alreadyPresent` row id the caller navigates to is the right one; `currentGroups()`/`currentSelectedID()`/`currentIdentityName()` all read through, so the cold-start fix the PR describes genuinely holds; verification-only rows can't spuriously materialize, since a deferred group isn't in `GroupRepository` yet; and the accessibility-identifier move to leaves is consistent with no remaining references to the old container ids.

rinat-enikeev:
`13fbe5e`. Three accepted, two pushed back on — details below.

**`replaceTopWithChat` popping the wrong screen — accepted, real.** A buried `PendingChatThreadView` is alive and still watching its row, so `[pendingA, X]` + A materializing deleted X. One note on the suggested fix: the guard isn't expressible against `NavigationPath`, which exposes `count`/`isEmpty`/`removeLast()` and nothing that reads an element. So the path is a typed `[ChatsRoute]` now and the swap runs only when the top is the pending screen that asked for it. That also collapses the two `navigationDestination`s into one route type, retiring the "both ids are `String`s" hazard the pair carried.

**`materialized` recorded from `rows` — accepted, fixed differently.** No need to key off `pending` or resolve lazily: a pending row's id *is* `<groupIDHex>:<owner>`, so the mapping is computed straight from the group snapshot and has no ordering dependency at all. I'd downgrade the severity though — reaching that fallback means being *on* the pending thread, which means `rows` was populated, and the screen has a Back button. It now also dismisses itself when the row ends with no mapping, so the spinner isn't a resting state either way.

**Stale `JoinScreen` doc — fixed.**

**Pushback 1: the dead-end wait — finding accepted, fix changed.** A staleness threshold on `receivedAt` re-introduces the timer the durable row makes awkward: what is stale when the app was closed for a week, and what does the row say at hour 23? Instead the `.waiting` state simply carries "Ask again", always. Asking twice is safe on its own terms — `JoinRequestApprover` collapses repeats by `(joiner, group)` (`JoinRequestApprover.swift:843`) and a decline stays declined — so there's no reason to make someone wait out a timer before offering the one action that can help. The verification case routes to the verifier instead of the founder, since past the approval the founder has already done their part.

**Pushback 2: "no unit coverage for the new join path".** Tests live in #297 by this repo's convention, and it is stacked directly on this branch: `.alreadyJoined` owner matching, the `.alreadyPresent` + `.offered` auto-send, `.waiting` with no second submit, and `materializedGroupID` are all covered there, plus the "Ask again" routing added this round. The genuinely uncovered branch is `.failed` → alert, which is view code in `OnymIOSApp`. Worth noting the "273 lines deleted" framing overstates it: most of `JoinFlowTests` exercised a sheet state machine (`.sending` → `.awaitingApproval` → `.unanswered`) that no longer exists, not behaviour that moved.

1492 unit tests green; `MultiIdentityChatUITests` green end to end.

dev-onym (commented):
COMMENT — the design is sound and the cold-start identity fix is the right call; two behavioural gaps and a coverage gap are worth addressing before merge.

- `PendingChatsFlow.swift:198` — `materialized` is only ever added to, never pruned, while `groupRepository.snapshots` re-emits after `delete(id:ownerID:)` (`GroupRepository.swift:76`). Delete a chat, then tap a fresh invite to the same group: `join()` sees no group in `currentGroups()`, records a pending row, and `PendingChatThreadView.swift:59`'s `initial: true` reads the stale mapping and immediately `replaceTopWithChat` into a group that no longer exists. Rebuilding the dict from each snapshot (`materialized = Dictionary(uniqueKeysWithValues: groups.map { ("\($0.id):\($0.ownerIdentityID.rawValue.uuidString)", $0.id) })`) still satisfies the "kept after the row is gone" requirement — the group is in the snapshot, only the row is gone — and self-heals.
- `PendingChatsFlow.swift:280-289` — `record()` on `.alreadyPresent` calls `refreshOffer`, whose own comment says the row's old intro key is now "a dead address — reported as sent, never heard, waiting forever." But `join()` only re-sends when `status == .offered`, so the founder-regenerated-link case (`.requested` row + new key just written) returns `.waiting` and sends nothing. Tapping the new link is the user's only signal that the old request is dead; requiring them to find Retry inside the thread defeats it. Re-send when `refreshOffer` actually changed `introPublicKey`.
- No unit coverage for any of the new logic: `JoinFlowTests.swift` (273 lines) is deleted and nothing replaces it. `join()`'s four branches (no identity / already-joined / inserted / already-present-and-offered) and `ChatsNavigation.replaceTopWithChat`'s `path.last` guard — the one the file's own comment says protects against "reliably, whenever two links are in flight" — are asserted only indirectly by one UI test. `replaceTopWithChat` in particular is pure logic and trivial to test.
- `ChatsNavigation.swift:41` — `openChat` appends unconditionally, so re-tapping a link for the chat currently on screen pushes a second identical `.chat` route. A `guard path.last != .chat(groupID: groupID)` matches the care already taken in `replaceTopWithChat`.
- `PendingChatThreadView` still carries `.accessibilityIdentifier("pending_chat.thread")` on the root container while the PR body documents that container ids propagate down onto descendants. It doesn't appear to break the leaf ids the test queries, but it's the same trap the PR just paid for — worth confirming it's load-bearing or dropping it.

Verified as fine: no dangling `JoinFlow`/`JoinView`/`makeJoinFlow`/`join.*` identifier references remain; the `alreadyJoined` check correctly filters `currentGroups()` by owner (that cache is the unfiltered roster); `materialized` keys include the owner uuid so identity switches can't cross-fire; new xcstrings entries have both en and ru.

PR #297, opened 2026-08-21
Tests for the waiting room now that it is a chat row

rinat-enikeev:
Stacked on #296.

Three files, one per layer, and each one holds a claim the previous PRs made in prose.

`PendingChatStoreTests` — the key is `(group, owner)`, so a relay replaying an offer collapses onto the row that exists and a chat the user already accepted does not come back asking to be accepted again; two identities on one device keep their own rows for the same group; and rows survive a reopen, which is the entire reason the store is on disk rather than in memory. `reopened()` exists for that last one: a cache that answers correctly right after writing proves nothing about durability.

`PendingChatRepositoryTests` — the per-identity filter, including the cold-start case where no identity is selected yet and the honest answer is nothing rather than someone else's invitations. Also that a group snapshot which matches no pending row touches nothing: every group emission runs through that path, and re-yielding the list each time would repaint the chats list on every unrelated change.

`PendingChatsFlowTests` — where the three sources meet. A verification overlays the stored status rather than adding a second row, because past the founder's approval the wait has moved on from them; a verification with no offer under it still gets a row, undismissable, since that is the stale-invitation case with nowhere else left to show; `chainSettling` offers no Retry; and a link tapped twice is not two requests.

`PendingVerificationStore.record` became public for the overlay tests. The state it holds is rendered a module away, so a test of that seam either gets to write one, or has to stand up the whole verifier to assert on how a row reads.

Full suite: 1474 unit tests green, plus `MultiIdentityChatUITests`, `DeleteChatUITests` and `SearchUITests`.

### DIFF
--- a/Packages/OnymInbox/Sources/OnymInbox/PendingVerificationStore.swift
+++ b/Packages/OnymInbox/Sources/OnymInbox/PendingVerificationStore.swift
@@ -64,10 +64,7 @@ public actor PendingVerificationStore {
 
     /// Idempotent on `groupIDHex`. A re-deferred snapshot (re-delivery)
     /// keeps the existing entry/status rather than resetting it.
-    /// Public because the state it holds is rendered a module away, by
-    /// `PendingChatsFlow` — a test of that overlay has to be able to put
-    /// a verification in without standing up the whole verifier.
-    public func record(_ entry: PendingGroupVerification) {
+    func record(_ entry: PendingGroupVerification) {
         guard !all.contains(where: { $0.groupIDHex == entry.groupIDHex }) else { return }
         all.append(entry)
         publish()
--- a/Packages/OnymInbox/Sources/OnymInbox/PendingVerificationStore.swift
+++ b/Packages/OnymInbox/Sources/OnymInbox/PendingVerificationStore.swift
@@ -64,7 +64,10 @@ public actor PendingVerificationStore {
 
     /// Idempotent on `groupIDHex`. A re-deferred snapshot (re-delivery)
     /// keeps the existing entry/status rather than resetting it.
-    func record(_ entry: PendingGroupVerification) {
+    /// Public because the state it holds is rendered a module away, by
+    /// `PendingChatsFlow` — a test of that overlay has to be able to put
+    /// a verification in without standing up the whole verifier.
+    public func record(_ entry: PendingGroupVerification) {
         guard !all.contains(where: { $0.groupIDHex == entry.groupIDHex }) else { return }
         all.append(entry)
         publish()
--- a/Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift
+++ b/Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift
@@ -51,11 +51,15 @@ public actor SwiftDataPendingChatStore: PendingChatStore {
         return SwiftDataPendingChatStore(container: container)
     }
 
-    /// A second store over the same container — a relaunch, modelled.
-    /// Durability is the reason this store exists at all, so a test has
-    /// to be able to close the context and read the rows back through a
-    /// fresh one rather than trust the cache it just wrote.
-    public func reopened() -> SwiftDataPendingChatStore {
+    /// A second store over the same container, reading through a fresh
+    /// `ModelContext`.
+    ///
+    /// Not a relaunch: the container here is the same object and, for
+    /// `inMemory()`, never touches disk. What it does show is that a
+    /// write reached the container's store rather than sitting in the
+    /// context that made it — the half of durability a test can check
+    /// without a device.
+    func reopened() -> SwiftDataPendingChatStore {
         SwiftDataPendingChatStore(container: container)
     }
 
--- a/Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift
+++ b/Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift
@@ -51,6 +51,14 @@ public actor SwiftDataPendingChatStore: PendingChatStore {
         return SwiftDataPendingChatStore(container: container)
     }
 
+    /// A second store over the same container — a relaunch, modelled.
+    /// Durability is the reason this store exists at all, so a test has
+    /// to be able to close the context and read the rows back through a
+    /// fresh one rather than trust the cache it just wrote.
+    public func reopened() -> SwiftDataPendingChatStore {
+        SwiftDataPendingChatStore(container: container)
+    }
+
     private init(container: ModelContainer) {
         self.container = container
         self.context = ModelContext(container)
--- a/Tests/OnymIOSTests/PendingChatRepositoryTests.swift
+++ b/Tests/OnymIOSTests/PendingChatRepositoryTests.swift
@@ -77,7 +77,7 @@ final class PendingChatRepositoryTests: XCTestCase {
         XCTAssertEqual(snapshot.first?.status, .failed(reason: "no relay connection"))
     }
 
-    func test_consumeForMaterializedGroups_dropsTheWaitThatIsOver() async throws {
+    func test_consumeForMaterialized_dropsTheWaitThatIsOver() async throws {
         let repository = PendingChatRepository(store: InMemoryPendingChatStore())
         await repository.setCurrentIdentity(alice)
         let landed = makeChat(group: 0x11, owner: alice)
@@ -85,28 +85,53 @@ final class PendingChatRepositoryTests: XCTestCase {
         await repository.record(landed)
         await repository.record(waiting)
 
-        await repository.consumeForMaterializedGroups([landed.groupIDHex])
+        await repository.consumeForMaterialized([(landed.groupIDHex, alice)])
 
         let snapshot = try await firstSnapshot(from: repository)
         XCTAssertEqual(snapshot.map(\.groupIDHex), [waiting.groupIDHex])
     }
 
-    func test_consumeForMaterializedGroups_ignoresGroupsWithNoPendingRow() async throws {
+    func test_consumeForMaterialized_touchesNothingWhenNoRowMatches() async throws {
         // Every group snapshot flows through here on every emission, and
-        // almost none of them match. Touching the store for those would
-        // re-yield the whole list to every subscriber on each repaint.
-        let repository = PendingChatRepository(store: InMemoryPendingChatStore())
+        // almost none of them match. Counting the writes rather than the
+        // surviving rows is the point: a row count of 1 also passes for
+        // an implementation that deleted nothing but re-yielded the whole
+        // list to every subscriber on each repaint.
+        let store = CountingPendingChatStore()
+        let repository = PendingChatRepository(store: store)
         await repository.setCurrentIdentity(alice)
-        let chat = makeChat(group: 0x11, owner: alice)
-        await repository.record(chat)
+        await repository.record(makeChat(group: 0x11, owner: alice))
+        await store.resetCounts()
 
-        await repository.consumeForMaterializedGroups(["ffff"])
+        await repository.consumeForMaterialized([("ffff", alice)])
 
+        let deletes = await store.deleteForIDsCalls
+        let lists = await store.listCalls
+        XCTAssertEqual(deletes, 0, "no row matched, so nothing may be written")
+        XCTAssertEqual(lists, 0, "and the list must not be re-read to say so")
         let snapshot = try await firstSnapshot(from: repository)
         XCTAssertEqual(snapshot.count, 1)
     }
 
-    func test_consumeForMaterializedGroups_worksAgainstAColdCache() async throws {
+    func test_consumeForMaterialized_leavesTheOtherIdentitysRowAlone() async throws {
+        // The group stream is filtered to the selected identity, so the
+        // pairs arriving here name one owner. Matching on the group
+        // alone deleted the other identity's row for the same chat.
+        let store = InMemoryPendingChatStore()
+        let repository = PendingChatRepository(store: store)
+        let mine = makeChat(group: 0x11, owner: alice)
+        let theirs = makeChat(group: 0x11, owner: bob)
+        await repository.setCurrentIdentity(alice)
+        await repository.record(mine)
+        await repository.record(theirs)
+
+        await repository.consumeForMaterialized([(mine.groupIDHex, alice)])
+
+        let rows = await store.list()
+        XCTAssertEqual(rows.map(\.ownerIdentityID), [bob])
+    }
+
+    func test_consumeForMaterialized_worksAgainstAColdCache() async throws {
         // Launch order isn't guaranteed: the group watcher starts from
         // the flow while the startup `reload()` runs from another task,
         // so the first emission can arrive before anything has been
@@ -118,7 +143,7 @@ final class PendingChatRepositoryTests: XCTestCase {
 
         let repository = PendingChatRepository(store: store)
         await repository.setCurrentIdentity(alice)
-        await repository.consumeForMaterializedGroups([chat.groupIDHex])
+        await repository.consumeForMaterialized([(chat.groupIDHex, alice)])
 
         let snapshot = try await firstSnapshot(from: repository)
         XCTAssertTrue(snapshot.isEmpty, "the row must go even if the cache was never read")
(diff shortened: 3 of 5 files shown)


rinat-enikeev:
@dev-onym review the whole PR stack

dev-onym:
## Review — the whole stack (#295 → #296 → #297)

Read all three end to end. The shape is right: `PendingChat` is deliberately not a `ChatGroup`, the `(group, owner)` key is the correct dedupe unit, and moving the wait into the chats list removes two surfaces (`PendingInvitesView`, `JoinView`) that nobody could find. `PendingChatStore` / `PendingChatRepository` mirror `GroupStore` / `GroupRepository` closely enough that the seams read as one idea rather than three. No dangling references to the removed types outside `EXTRACTION-NOTES.md`.

Fixed in `cc556cd` (the two things that were unambiguous and mechanical):

- **Two new user-facing strings were missing from `Localizable.xcstrings`** — `"Couldn't use that invite"` (the alert title in `OnymIOSApp`) and `"Couldn’t save this invite on your device."` (`PendingChatsFlow.join`). `LocalizationCatalogTests` only checks that keys *in* the catalog have both `en` and `ru`; it never checks that every `String(localized:)` has a key, so CI stayed green while a `ru` user saw English on the one screen the whole stack is about. Added with translations.
- Stale comment in `ChatsView` still saying a scan and a tap "reach `JoinView` identically" — `JoinView` was deleted in #296.

Four things I did **not** touch, because each is a design call rather than a slip:

1. **`PendingChatThreadView` becomes a permanent spinner when the wait ends** — `PendingChatThreadView.swift:25`. The type comment says "the pending row is consumed and this screen is replaced by the real thread", but nothing pops or replaces it: `ChatsNavigation.path` is never touched on materialization, so `flow.row(id:)` goes nil and the branch at line 31 renders a bare `ProgressView` forever. This is the *success* path for the headline flow — tap link → land in the pending thread → founder approves → indefinite spinner, and the user has to guess that Back is the way into their new chat. Cheapest fix is probably for the flow to publish "row `X` became group `Y`" and have `ChatsNavigation` swap the top of the path.

2. **A failed scan-to-join fails silently** — `ChatsView.swift:195`. The deeplink path surfaces `.failed` through the `joinLinkError` alert in `OnymIOSApp`; the scanner path sets `flow.lastError`, and the only place `lastError` is rendered is the banner inside `PendingChatThreadView` — a screen the user isn't on, and by definition can't reach, because the join is what failed. So a scan that can't be recorded looks exactly like one that worked. The two entry points were deliberately unified in `handleScannerDismiss`; the error surface is the half that didn't come with them.

3. **`navigationDestination` is registered only when the list is non-empty** — both destinations hang off `groupList` (`ChatsView.swift:459`, `:483`), and the `rows.isEmpty` branch renders `emptyState` instead. The push from `ChatsNavigation` now comes from *outside* the view, so a first-ever user with zero chats tapping an invite link is appending a `PendingChatRoute` to a stack that may not have a destination for it at that moment. It probably survives — the row insert and the path append coalesce into one render pass — but it survives by scheduling luck rather than by construction. Hoisting the two `.navigationDestination` modifiers onto the outer `Group` costs nothing and removes the question.

4. **`consumeForMaterializedGroups` can no-op against a cold cache at launch** — `PendingChatRepository.swift:60` computes `live` from `cached`, and in `OnymIOSApp` the group watcher starts in `pendingChatsFlow.start()` while `pendingChatRepository.reload()` runs from a different `.task`. If a group emission lands before the reload, `matched` is empty, nothing is deleted, and the stale row then sits in the list until the next unrelated group mutation. Narrow (it needs a row that outlived its own materialization), and reading `store.list()` when `cached` is empty would close it.

On #297 itself: the three files cover the claims the prose made, and `reopened()` earning its keep is the right instinct. Two gaps worth a line each — nothing covers `deleteForGroups` / `consumeForMaterializedGroups` against a cold cache (#4 above is exactly the case with no test), and `PendingChatsFlowTests.Harness.init` fires `setCurrentIdentity` in an un-awaited `Task`, which `test_join_whenAlreadyAMember_opensTheChatInstead` then races: it asserts on `groupRepository.currentGroups()` without first waiting for the identity filter to land. The polling helpers hide this everywhere else; that one test reads the result synchronously.

rinat-enikeev:
All four addressed, each on the PR whose code it is. Your `cc556cd` moved down to #296 (cherry-picked, authorship intact) since the strings it adds are used by that PR's code rather than by the tests.

**1 — permanent spinner on the success path (#296, `353a9e2`).** Fixed the way you suggested: the flow records `row → group` *before* the row is consumed (#295, `57e75a9`), the pending screen reports it upward, and `ChatsNavigation.replaceTopWithChat` swaps the top of the path. Replace rather than push — the pending screen has nothing left to show once its row is gone, and leaving it underneath puts a dead screen behind the Back button of a brand-new chat. The view reports rather than navigates; the path belongs to `ChatsNavigation`.

**2 — silent scan failure (#296, `353a9e2`).** The scanner path gets an alert where the scan happened, with the wording the deeplink path already uses. You were right about why it was invisible: `flow.lastError` renders in the banner inside the pending thread, which is exactly the screen a failed join can never reach.

**3 — `navigationDestination` registered only on the populated list (#295, `57e75a9`).** Both hoisted onto the outer view. Correct by scheduling luck isn't correct.

**4 — cold-cache no-op (#295, `57e75a9`).** Reads through to the store when `cached` is empty, behind an early return for an empty input so the common emission still costs nothing. Covered by a new test that asserts the row leaves the *store*, not just the cache.

**On #297:** `Harness` construction is an async factory now, so the identity filter is in place before any test looks — `test_join_whenAlreadyAMember_opensTheChatInstead` was deciding on a repository that had never been told whose rows it holds. Added the two missing tests: the cold-cache sweep, and the `row → group` mapping surviving the row it describes.

One thing worth recording: the first cut of that factory shipped an infinite recursion (`make` calling itself), which XCTest reports as a 50-second silence and a runner relaunch per test rather than as a failure. Caught by running the suite, not by the build.

Verification: 1476 unit tests, 0 failures; `MultiIdentityChatUITests` green end to end — link tap → pending chat → founder approves in-thread → the real chat with its "You joined" notice → messages and read receipts.

dev-onym (commented):
COMMENT — solid, well-targeted tests; findings are non-blocking (unnecessary API widening, a few assertions that don't test what their comments claim).

- `Packages/OnymInbox/Sources/OnymInbox/PendingVerificationStore.swift:70` — `record` didn't need to become `public`. `PendingChatsFlowTests.swift:330` constructs `PendingGroupVerification` through its **internal** memberwise init, so that file already depends on `@testable import OnymInbox` for internal access; `record` would have been reachable unchanged. Four other package modules are `@testable`-imported elsewhere in this suite, so this is established practice. Revert to internal and drop the "public because…" comment.
- `Packages/OnymInbox/Sources/OnymInbox/SwiftDataPendingChatStore.swift:58` — same argument for `reopened()`; also the doc overstates what the test can show. `inMemory()` builds an `isStoredInMemoryOnly` container, so a second `ModelContext` proves the write reached the container's store rather than the context's cache — not disk durability — and nothing "closes the context" (the original stays alive). Worth rewording; `test_rowsSurviveAReopen` (`PendingChatStoreTests.swift:110`) is still a real assertion, just a narrower one than the comment claims.
- `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:94` — the stated claim ("touches nothing / no re-yield") is not asserted. `XCTAssertEqual(snapshot.count, 1)` also passes if the implementation dropped the `matched.isEmpty` guard and re-yielded the whole list to every subscriber. Needs a spy store counting `deleteForGroups`/`list` calls, or a subscriber that counts yields.
- `PendingChatRepository.remove(id:)` and `PendingChatsFlow.dismiss(_:)` are untested at every layer — the only user-initiated delete in the feature. `isDismissable` is asserted on both row kinds, but the swipe it gates is never exercised.
- `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:140` — `currentChats()` is only read with a warm cache, so the `cached.isEmpty → refreshFromStore()` branch (`PendingChatRepository.swift:101`) is uncovered. That's the same cold-cache hole `test_..._worksAgainstAColdCache` was written for one method up, and it's the cold-start deeplink path.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:47` — the hold/release gate proves less than it looks: `isSending` is set synchronously inside `send()` before `submitJoin` is ever awaited, so `waitFor { isSending == true }` can pass before the spy is entered, and `release()` then makes the gate a no-op. The debounce is genuinely proven by `sendingIDs`, but wait on `sender.calls.count == 1` before releasing if you want the in-flight state to be the thing under test.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:126` — asserting `isRetryable == false` tests the enum, not the flow. Call `flow.retry(chat.id)` and assert `verifierRetries.values.isEmpty` and `sender.calls.isEmpty`; that's the behavior the comment describes.
- Uncovered branches in the overlay/send paths: `.verifying` and `.chainNotConfigured` status mappings (`PendingChatsFlow.swift:394-402`), `.noIdentityLoaded` from the sender → "Sign in first." on the row, and the malformed-capability `lastError` branch (`PendingChatsFlow.swift:302`).
- `test_inMemoryStore_matchesTheOnDiskContract` (`PendingChatStoreTests.swift:139`) hand-checks three of six protocol methods; `delete(id:)`, `deleteOwner`, and newest-first ordering diverge silently. A shared helper run against both stores would make "matches the contract" true.

rinat-enikeev:
All addressed in `e15de95`.

**Both seams reverted to internal.** You're right on both counts — the suite already reaches internals via `@testable import OnymInbox` (the same file builds `PendingGroupVerification` through its internal memberwise init), so neither `record` nor `reopened()` needed widening. `reopened()`'s doc also overstated things: `inMemory()` never touches disk and the original context stays alive, so what it shows is that the write reached the container's store rather than the context that made it. The doc and the test name say that now.

**Three assertions that didn't test their own claims:**
- "touches nothing" counts calls through a spy store, not surviving rows.
- the debounce gate waits on the spy being entered before releasing — `isSending` is set synchronously, so it could unblock a gate nobody had reached.
- `chainSettling` taps Retry and asserts neither verifier nor sender hears anything, instead of asserting the enum.

**New coverage:** `remove`/`dismiss`, `currentChats` through a cold cache, `.verifying` and `.chainNotConfigured` overlays, `.noIdentityLoaded` landing as a retryable row, and the malformed-invite branch that must not reach the sender. Both store conformers now run one shared contract exercise over all six methods, so "matches the on-disk contract" is true rather than three-sixths true.

Also here: the two rules the code learned from your other reviews — a sweep must leave the other identity's row alone, and a link onto an unanswered offer must send. I mutated both implementations back and confirmed the new tests fail (3 failures), so they're load-bearing.

One process note: the earlier run of these files was passing against a stale `.xcodeproj` generated on a branch where the test files didn't exist — the target silently compiled 1441 tests instead of 1476. Regenerating is now part of the loop after touching test files.

1487 unit tests green; `MultiIdentityChatUITests` green end to end.

dev-onym:
I verified the comment against the tree rather than assuming. **There is nothing to commit — the comment reports work that is already in the branch.**

Comment 5372523825 (by @rinat-enikeev) opens with "All addressed in `e15de95`", and `e15de95` is HEAD of `test/pending-chats` and already on `origin`. Working tree is clean. Every claim in it checks out against the current files:

| Claim | Verified |
|---|---|
| `record` / `reopened()` back to internal | `PendingVerificationStore.swift:67`, `SwiftDataPendingChatStore.swift:62` — both plain `func` |
| `reopened()` doc no longer overstates | doc says "a write reached the container's store rather than sitting in the context that made it" |
| test name matches | `test_rowsSurviveAFreshContext` with the matching "not a relaunch" comment |
| "touches nothing" counts rows, not calls | `CountingPendingChatStore` at `PendingChatRepositoryTests.swift:238`, used by `test_consumeForMaterialized_touchesNothingWhenNoRowMatches:94` |
| debounce waits on the spy, not `isSending` | `PendingChatsFlowTests.swift:62` — `waitForAsync { await harness.sender.calls.count == 1 }` |
| `chainSettling` taps Retry instead of asserting the enum | `PendingChatsFlowTests.swift:197-203` |
| new coverage | `remove`/`dismiss`, `currentChats` cold cache, `.verifying`, `.chainNotConfigured`, `.noIdentityLoaded`, malformed-invite — all present |
| shared contract over both conformers | `assertHonoursContract` driven from `test_swiftDataStore_honoursTheContract` and `test_inMemoryStore_honoursTheContract` |
| the two rules from other reviews | `test_deleteForIDs_leavesTheOtherIdentitysRowForTheSameGroup`, `test_consumeForMaterialized_leavesTheOtherIdentitysRowAlone`, `test_join_onAnUnansweredOffer_sendsInsteadOfAskingAgain` |

Two notes on what I could not do here:

- **The process note about the stale `.xcodeproj` needs no commit.** The project is generated by `generate-xcodeproj.sh` from `project.yml` and `.xcodeproj` is untracked, so "regenerate after touching test files" is a local-loop habit, not a repo artifact. If you want it enforced rather than remembered, the place for it is a `.github` CI step or a pre-test script — say the word and I'll add that as an actual change.
- **I could not re-run the suite to confirm the 1487 green.** This worktree is Linux; there is no `xcodebuild`. The 1487/`MultiIdentityChatUITests` figures are yours from the comment, unverified by me.

I did not create an empty or no-op commit to satisfy the "commit with message …" instruction — that would put a message claiming to address the comment on top of a tree the comment already describes. If you did intend further work beyond `e15de95` (something the comment doesn't spell out), tell me what and I'll do it under that message.

dev-onym (commented):
COMMENT — solid, well-motivated tests that read the production code correctly; two races and a stale PR claim to fix.

- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:270` — race. `materialized[row.id]` is written *before* `await repository.consumeForMaterialized(...)` (`PendingChatsFlow.swift:180-188`), so when the `waitFor` predicate first sees `materializedGroupID != nil`, the delete + re-yield hasn't necessarily landed. `XCTAssertTrue(rows.isEmpty)` on the next statement can fail. Make it a second `try await waitFor { harness.flow.rows.isEmpty }`.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:196-201` — the negative assertion is race-prone and the comment overclaims. `retry` for `.chainSettling` hits `break` synchronously today, but if it regressed to `Task { await retryVerification(hex) }` the spawned task may not have run by the time `await verifierRetries.values` reads, so the test could still pass. Either poll for a short grace period, or drive a *positive* signal through the flow first (e.g. tap retry on a `.chainUnreachable` row, wait for that recorded hex, then assert the settling hex never appears).
- PR body claims "`PendingVerificationStore.record` became public for the overlay tests" — it's still internal (`PendingVerificationStore.swift:67`) and the diff touches that file not at all. `@testable import` already covers it. Drop the paragraph.
- `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:23,56` — names promise re-yield to subscribers, but `firstSnapshot` opens a *fresh* stream every call, so both only assert `subscribe`'s initial `yield(filtered())`. `setCurrentIdentity`'s and the mutators' `publishFiltered()` to *existing* continuations (`PendingChatRepository.swift:94,139`) is untested — these tests pass if you delete those calls. Worth one test that holds a stream open across a mutation and an identity switch, given the file header states "every mutation re-yields" as the contract.
- Untested and cheap with the spy already in hand: `record` returning `.alreadyPresent` must *not* re-yield (`PendingChatRepository.swift:29-32`) — the exact mirror of `test_consumeForMaterialized_touchesNothingWhenNoRowMatches`, and the replayed-offer repaint it prevents is the flow's most frequent event.
- `SwiftDataPendingChatStore.swift:54-63` — `reopened()` is production surface existing only for tests. Its doc comment is honest about what it does and doesn't prove, which is good, but it would sit better as an `extension SwiftDataPendingChatStore` in the test target if `container` visibility allows.
- `SpyJoinSender` (`PendingChatsFlowTests.swift:~500`) — a second `send` while `held` overwrites `gate`, orphaning the first continuation and hanging the test. Only one held send exists today; a `precondition(gate == nil)` would make the misuse loud instead of a 2s timeout.
- Minor: no test calls `flow.stop()`, so three stream-draining tasks per test outlive it; `test_noIdentityLoaded_...` asserts on `String(localized:)` copy, which couples it to wording the `State` enum comment says should be free to change.

dev-onym (commented):
COMMENT — solid, well-targeted tests; four things worth tightening before merge, none blocking.

- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:335` — race. The group watcher sets `materialized[rowID]` *before* awaiting `consumeForMaterialized`, so the `waitFor` on `materializedGroupID` can return while the delete/re-publish round-trip is still in flight and `rows` is not yet empty. `test_theRowDisappearsWhenTheGroupLands` already owns that claim — drop the trailing assertion here, or wait on it separately.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:302-317` — the test can't fail. It waits on `rows.count == 1`, which the pending snapshot alone satisfies, then asserts `.offered` — the same value the row has *before* the verification is folded in. If the overlay-wins-over-offers bug regressed, this passes whenever the verification stream drains a beat late. Needs a signal that the verification was actually applied (e.g. a second verification for an uncovered group whose synthesised row you can wait on, then assert this row).
- `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:23,56` — names promise re-yield ("reyields", "isVisibleToSubscribers") but `firstSnapshot` (:247) subscribes fresh each time and `subscribe` yields from `cached`. Deleting every `publishFiltered()` call from the mutators and from `setCurrentIdentity` leaves both green. Either hold one stream across the mutation and read two elements, or rename to match what's checked — the identity-switch re-yield is covered nowhere else.
- `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:271` — `CountingPendingChatStore` is a 60-line hand copy of `InMemoryPendingChatStore` that never runs through `assertHonoursContract`, and already diverges (`deleteForIDs` counts an empty set the real one returns early on). A decorator wrapping `InMemoryPendingChatStore` gets the counters without the drift.

Smaller:
- `PendingChatRepositoryTests.swift:175` — `var reinvite = makeChat(...)` is discarded on the next line; `let reinvite = PendingChat(...)` says it directly.
- `PendingChatsFlowTests.swift:529,544` — `waitFor` calls `XCTFail` and returns without throwing, so a timeout cascades into downstream assertions instead of stopping at the real failure. `throws` is already on the signature.
- `firstSnapshot` (:247) has no timeout — a regression that stops yielding on subscribe hangs the suite rather than failing it.
- PR body says `PendingVerificationStore.record` became public; it's still internal (`PendingVerificationStore.swift:67`) and the diff doesn't touch that file. `@testable` is carrying it — description is stale.
- `SwiftDataPendingChatStore.reopened()` is production surface that exists only for one test. Internal keeps it contained, and the doc comment is honest about what it does and doesn't prove, so this is fine as-is — `#if DEBUG` if you'd rather it not ship.
- Adjacent, not yours: `PendingChatRepository.swift:36-38` still carries a "No refresh on `.alreadyPresent`" comment directly contradicting line 35. `test_record_onARepeatOffer_takesTheNewerReplyChannel` now pins the real behaviour — good time to delete the stale comment.

PR #299, opened 2026-08-21
Arriving is not consenting: a link resolves to a screen, not a request

rinat-enikeev:
Fixes a HIGH from security review, raised against the Android twin and true of both apps.

**The finding.** The URL types this app registers are exported. Any app on the device can open `onym://join?c=…`, and so can a page in a browser. Since #296, receiving a decoded capability immediately recorded a row and sent a signed join request — carrying this identity's display name and its stable inbox, BLS and signing public keys — to whoever minted the invite key. The person's first sight of it was a new row in their chats list. That was my error: "tapping the link is the intent" holds for a person tapping a link, not for the channel that delivers it, which anything can speak into.

**The fix.** A capability now resolves to a screen. `prepareJoin` records nothing and sends nothing; it answers *already in* (open the chat), *already asked* (open the wait), or *ask them* — the confirmation screen. That screen names the group, the inviter when there is one, the invite key the request would be sealed to, and in as many words what Send discloses. `confirmJoin` is the only path to a send, and only a person can reach it. The scan path resolves identically: a scan is deliberate, but it discloses the same things, so it gets the same review.

Accept on a pushed offer goes through the same screen rather than sending on tap — one path, so what someone is shown before their name and keys leave the device doesn't depend on which door the invitation came through.

**And the name is theirs to choose — without having to type it.** The field arrives pre-filled with the identity's own alias, read through the identity *repository* rather than the identities flow, because a link that launches the app resolves before any tab has populated that flow and the field would otherwise open empty. So joining stays one tap: Send, and nothing else. The name rides with this join only: one identity can be "Sam" to a book club and "S." to a tenants' group without being two identities. It is stored on the row (encrypted, like the rest of the person-supplied text), so "Ask again" re-introduces the same person rather than whoever the identity has since been renamed to — a founder is deciding partly on that name, and a rename between the first request and the second would make it arrive from a stranger.

`test_aTappedLink_recordsNothingAndSendsNothing` is the rule as a test: it asserts the sender was never called *and* that no row was written. Both halves matter — a row left behind is a disclosure of its own to anyone who later looks at the device.

Verification: 1497 unit tests green; `MultiIdentityChatUITests` green end to end. It asserts the name field arrives holding `"Bob"` and then taps Send without typing — the two halves of the rule: a link no longer sends on arrival, and confirming it costs one tap rather than typing your own name.

The same fix is still owed on Android (onymchat/onym-android#240), where the finding was originally raised.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift
@@ -84,6 +84,10 @@ public struct ChatsView: View {
     /// inside the pending thread, which is precisely the screen a failed
     /// join never reaches.
     @State private var scanJoinError: String?
+    /// A scanned invite awaiting confirmation. A scan is a deliberate
+    /// act, but what it discloses is the same, so it gets the same
+    /// review as a link before anything is sent.
+    @State private var scanConfirmation: PendingChatsFlow.JoinConfirmation?
     /// The chat awaiting a swipe-to-delete confirmation, if any.
     @State private var pendingDelete: ChatListItem?
 
@@ -194,6 +198,24 @@ public struct ChatsView: View {
                 onCancel: { showScanner = false }
             )
         }
+        .sheet(item: $scanConfirmation) { confirmation in
+            JoinConfirmView(
+                confirmation: confirmation,
+                onSend: { label in
+                    let flow = pendingChatsFlow
+                    let navigation = navigation
+                    scanConfirmation = nil
+                    Task { @MainActor in
+                        if case .waiting(let rowID) = await flow.confirmJoin(
+                            confirmation, label: label
+                        ) {
+                            navigation.openPending(rowID: rowID)
+                        }
+                    }
+                },
+                onCancel: { scanConfirmation = nil }
+            )
+        }
         .reasonAlert("Couldn\u{2019}t use that invite", reason: $scanJoinError)
         .alert("Not an Onym invite", isPresented: $scanRejected) {
             Button("OK", role: .cancel) {}
@@ -214,11 +236,13 @@ public struct ChatsView: View {
             let flow = pendingChatsFlow
             let navigation = navigation
             Task { @MainActor in
-                switch await flow.join(capability: cap) {
+                switch await flow.prepareJoin(capability: cap) {
                 case .alreadyJoined(let groupIDHex):
                     navigation.openChat(groupID: groupIDHex)
                 case .waiting(let rowID):
                     navigation.openPending(rowID: rowID)
+                case .confirm(let confirmation):
+                    scanConfirmation = confirmation
                 case .failed(let reason):
                     scanJoinError = reason
                 }
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatsView.swift
@@ -186,8 +186,10 @@ public struct ChatsView: View {
             QRCodeScannerView(
                 onScanned: { raw in
                     // Same allowlist + decode the deeplink path uses, so
-                    // a scanned link and a tapped link reach `join`
-                    // identically. A non-invite QR yields nil → reject.
+                    // a scanned link and a tapped link resolve through
+                    // `prepareJoin` identically — a scan is deliberate,
+                    // but it discloses the same things, so it gets the
+                    // same review. A non-invite QR yields nil → reject.
                     if let cap = DeeplinkCapture.introCapability(fromString: raw) {
                         scannedCapability = cap
                     } else {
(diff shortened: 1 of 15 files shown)


dev-onym (commented):
REQUEST_CHANGES — the security fix is right and well-tested; two real defects fall out of it.

- **`PendingChatsFlow.swift:436` — "Ask again" is a dead button after upgrade.** `retry` now bails when `joinerLabel == nil`, but every row already on disk from a pre-#299 install has no label (`encryptedJoinerLabel` is new and nil). A row sitting at `.requested`/`.failed` through the update shows "Ask again" (`PendingChatThreadView.swift:186`, `isRetryable == true`) and tapping it does nothing — no spinner, no error, no send. Those are exactly the rows whose request died in a relay, i.e. the ones that need it. Fall back to `await displayLabel()` when the label is absent (and attach it), or hide/disable the button when there is nothing to repeat. The `guard`'s premise — "a row with no label has sent nothing yet" — is only true for rows created by this version.
- **`JoinConfirmView.swift:186` — the two new catalog keys can never resolve.** `row(_ key: String, …)` passes a `String` to `Text`, which picks the non-localizing `StringProtocol` overload, so the `group` / `invite key` entries added at `Resources/Localizable.xcstrings:477,494` are dead and both labels render English under `ru`. Take `LocalizedStringKey` (or `String(localized:)` at the call sites). Worth noting `test_localizable_xcstrings_everyKey_hasEnAndRu` can't catch this direction — the key exists, it's the lookup that doesn't happen.
- **`ChatsView.swift:209` and `PendingChatThreadView.swift:70` — a failed confirm is silent.** Both swallow `confirmJoin`'s result except `.waiting`; the scan path already has `scanJoinError` wired for `prepareJoin` failures but not for confirm failures, and the thread ignores the return entirely. Someone taps Send, the sheet closes, and a `.failed`/`Couldn't save this invite` outcome is indistinguishable from success. `OnymIOSApp.swift:2025` gets this right — match it.
- **`ChatsView.swift:231` and `:189` — stale comments contradict the change.** "sends the request and opens the chat that is on its way" and "reach `join` identically" both describe the removed behaviour (and `join` no longer exists). Given how much this codebase leans on comments as the record of *why*, these are the ones a future reader will trust.

Smaller things: `nameFocused` (`JoinConfirmView.swift:36`) is bound but never set — either focus the field on appear or drop it; the new `.sheet` at `PendingChatThreadView.swift:70` was inserted between the "The other way a row ends" comment and the `.onChange` it explains; `confirmJoin`'s `.offer` branch (`PendingChatsFlow.swift:363`) re-fetches the row but drops `prepareAccept`'s `status == .offered` check — harmless today only because `send` debounces.

Not blocking, and worth saying: `test_aTappedLink_recordsNothingAndSendsNothing` asserting both the empty sender *and* the empty store is the right shape for this rule, and routing pushed-offer Accept through the same screen rather than leaving two disclosure paths is the part that makes the fix hold up.

dev-onym (commented):
COMMENT — the security fix is sound (no send path bypasses `confirmJoin`, and the record-nothing rule is tested); two label-persistence defects and one pre-existing overwrite are worth fixing before merge.

**Findings**

- `Packages/OnymInbox/Sources/OnymInbox/PendingChatStore.swift:80` — `InMemoryPendingChatStore.refreshOffer` rebuilds `PendingChat` without `joinerLabel`, so it silently resets it to `nil`. `record()` calls `refreshOffer` on every `.alreadyPresent` (PendingChatRepository.swift:35), so a re-invite arriving for a row that already asked erases the name it asked under; the next "Ask again" then falls back to the identity's *current* alias — exactly the stranger-rename the PR argues against. `SwiftDataPendingChatStore.refreshOffer:105` preserves the field (it mutates four columns), so the fallback store and the real store now disagree. Same omission in the test double at `Tests/OnymIOSTests/PendingChatRepositoryTests.swift:301`, which is why no test catches it. Add `joinerLabel: existing.joinerLabel` in both.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:401` — the `.link`/`.alreadyPresent` branch calls `attachJoinerLabel` *before* checking `existing.status == .offered`. If the row moved to `.requested`/`.failed` while the sheet was open, the stored label is overwritten with a name that was never sent, and no send happens — so a later `retry` introduces a different person. The `.offer` branch guards first (line 372); make the link branch symmetric (check status, then attach).
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:394` — pre-existing but re-touched: the link-built chat carries `inviterAlias: ""` / `invitationMessage: nil`, and `record`'s `.alreadyPresent` path feeds those to `refreshOffer`. Confirming a link against a pushed offer therefore wipes "Alice invited you" and her message from the row the confirmation screen just showed, and replaces the row's intro key with the link's — which downgrades a re-invited row to a revoked key. `test_aLinkOnAnUnansweredOffer_confirmsAgainstTheOffersDetails:521` asserts the alias on the confirmation but nothing about the row after Send.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift:196` — `canSend` is false when `suggestedLabel` is empty, which happens for real when the identity has no name (`displayLabel` returns `""`, OnymIOSApp.swift:805). The screen then shows a dimmed Send with nothing saying a name is required. A hint or a placeholder-derived fallback would close it.
- `Sources/OnymIOS/OnymIOSApp.swift:806` — `_ = try? await repository.currentIdentities()` swallows a keychain load failure, so `currentIdentityID` answers `nil` and the link reports "Sign in first." for a device that is signed in. Worth distinguishing, or at least a comment noting the wording is knowingly approximate.

**Verified as claimed:** `prepareJoin`/`prepareAccept` write nothing and never reach `send`; `send` is private with only `confirmJoin` and `retry` as callers; all four catalog keys (`Send join request`, `(Unnamed)`, `group`, `invite key`) exist with `en`+`ru`; the new optional `encryptedJoinerLabel` is a lightweight SwiftData migration and won't trip `PersistentStoreOpener`.

PR #300, opened 2026-08-21
Rules a joiner can read, and a signature that says they agreed

rinat-enikeev:
An invite carried a group's name but not its terms. A founder admitting a stranger could prove *who* was asking and never *what they had agreed to*. This is the wire and the crypto for the second half; the screens follow in the PRs stacked on this one.

**Why this can't reuse the envelope's signature.** `IdentityRepository.sealInvitation` signs the *ephemeral public key* — not the payload beneath it. Whoever holds the envelope can produce a different plaintext for the same signature, so it is proof of delivery and never of content. That is the same defect class as #290, and it means agreement needs a signature of its own rather than a reinterpretation of an existing one.

**The statement.**

```
"onym-group-rules-v1" ‖ group_id ‖ SHA256(rules) ‖ joiner_sending_pub
```

Domain-separated, so it can't be replayed as one of the other things this Ed25519 key signs (moderation mandates, report disclosures). Bound to `group_id`, so an acceptance collected for a permissive group can't be shown as agreement to a stricter group whose text happens to match. Bound to the joiner's key, so a signature can't be lifted from one request and re-attributed. No timestamp: a joiner-supplied one is unverifiable, and the founder's receipt time is already theirs.

Because it is signed with `joinerSendingPublicKey` — which is already announced to every member — **any member can verify any other member's agreement**, not just the founder who admitted them.

**The link carries the text, not a hash.** There is no channel to fetch rules from before joining, and someone asked to agree to a hash they cannot read has not agreed to anything. So `IntroCapability` gains `rules`: public and cleartext, exactly as `groupName` is already documented to be, and capped at 500 characters. The cap is the QR code's, and it is measured rather than estimated — 500 CJK characters plus a 30-character group name comes to a 2258-byte link against a 2331-byte ceiling at correction level M. `GroupRules.maxLength` carries the numbers and the thin case.

An over-long `rules` on decode is rejected, not truncated: signing text that ends mid-sentence, and walking someone into a group whose rules they were never shown, are both worse failures than a link that doesn't work.

**`JoinRequestPayload`** carries the hash beside the signature. The inviter knows their own rules, but the hash is what separates "agreed to an older version" from "signed something that doesn't verify" — only one of those is worth asking a person to redo. Both fields are optional, absent together, and rejected if only one arrives.

**The sender signs what was shown**, taking the text as an argument rather than reading it back off the capability. Those agree for a link, but a pushed invitation keeps its rules on the stored offer, and reaching for the capability's copy would sign text that was never on screen. A signing failure fails the send: nothing goes out unsigned behind someone who was just told that Send means agreement.

Nothing calls this with rules yet — that arrives with the screens.

Cross-platform: the JSON keys (`rules`, `rules_hash`, `rules_signature`) are additions to a wire shape documented as byte-for-byte with Android. Unknown keys are ignored on decode, so old builds degrade to "signed nothing" rather than breaking. The same work is owed on onym-android.

Verification: 1497 existing unit tests green. Tests for this land in the final PR of the stack.

### DIFF
--- a/Packages/OnymGroup/Sources/OnymGroup/GroupRules.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/GroupRules.swift
@@ -1,0 +1,148 @@
+import Foundation
+import CryptoKit
+
+/// The rules a founder sets for a group, and the joiner's signed
+/// agreement to them.
+///
+/// ## Why a signature at all
+///
+/// A founder admitting a stranger is deciding on two things: who is
+/// asking, and what they have agreed to. The first was already
+/// provable — the request carries the joiner's long-term keys. The
+/// second was not, and could not be inferred from the envelope: the
+/// sealed envelope's Ed25519 signature covers the *ephemeral public
+/// key* only (`IdentityRepository.sealInvitation`), not the payload
+/// under it. A founder holding such an envelope can produce a
+/// different plaintext for the same signature, so it proves nothing to
+/// a third party about what the joiner said.
+///
+/// So agreement gets its own detached signature, over bytes that name
+/// what is being agreed to and by whom.
+///
+/// ## The statement
+///
+///     "onym-group-rules-v1" ‖ group_id (32) ‖ SHA256(rules) (32)
+///                           ‖ joiner_sending_pub (32)
+///
+/// Every component after the domain string is fixed-length, so the
+/// concatenation is unambiguous without length prefixes.
+///
+///  - The **domain string** keeps this signature from being replayable
+///    as any other signature this identity produces — the same Ed25519
+///    key signs moderation mandates and report disclosures.
+///  - **`group_id`** binds the agreement to one group. Without it, an
+///    acceptance collected for a permissive group could be shown as
+///    agreement to a stricter group's rules that happen to be equal.
+///  - **`SHA256(rules)`** is what makes it an agreement to *these*
+///    rules rather than to the idea of rules. Changing a comma
+///    invalidates it.
+///  - **`joiner_sending_pub`** names the signer inside the signed
+///    bytes, so a signature lifted from one request cannot be
+///    re-attributed to another joiner.
+///
+/// Deliberately **not** signed: a timestamp. A joiner-supplied one is
+/// unverifiable, and the only trustworthy time is when the founder's
+/// device received the request — which the founder records itself.
+///
+/// ## Keeping the proof
+///
+/// A signature is only evidence if the bytes it covers can be produced
+/// again. The verifier therefore stores the exact rules text alongside
+/// the signature; a stored hash with no text proves that *something*
+/// was agreed to and never what. This is the same discipline the
+/// moderation layer follows for mandates, where the signed manifest
+/// bytes are retained rather than re-derived.
+public enum GroupRules {
+    /// The longest rules text an invite may carry.
+    ///
+    /// The binding constraint is the QR code, not the screen. An invite
+    /// link is `base64(JSON)` in a URL, rendered at correction level M
+    /// (`SettingsQRCode`), whose byte-mode ceiling is 2331 bytes.
+    /// Measured end to end, with both 32-byte keys and a 30-character
+    /// group name in the JSON:
+    ///
+    ///     500 ASCII characters     →   924-byte link
+    ///     500 Cyrillic characters  →  1591-byte link
+    ///     500 CJK characters       →  2258-byte link
+    ///
+    /// So 500 fits even when every character costs three UTF-8 bytes,
+    /// which is the case that decides the cap. A longer one would
+    /// produce links that encode fine and then fail to scan — the worst
+    /// way to find out.
+    ///
+    /// The headroom in that last row is thin, and it is shared with the
+    /// group name, which has no cap of its own: a 300-character name
+    /// alongside 500 characters of CJK rules would push the link past
+    /// the ceiling. The failure is soft — `CIQRCodeGenerator` returns
+    /// no image and the link stays copyable — but it is why this
+    /// number should not be raised without re-measuring the pair.
+    public static let maxLength = 500
+
+    /// Domain separator. Versioned: a change to the statement's shape
+    /// must not let old signatures verify under the new reading.
+    static let domain = "onym-group-rules-v1"
+
+    /// The one canonical form, applied by *both* sides before hashing.
+    ///
+    /// Ends-only trimming, nothing else: the founder wrote this text
+    /// and a joiner is agreeing to it, so collapsing inner whitespace
+    /// or normalising case would mean signing something other than what
+    /// was displayed. Trailing newlines from a text field are the one
+    /// difference that carries no meaning, and letting them through
+    /// would make agreement fail on an invisible character.
+    public static func canonical(_ rules: String) -> String {
+        rules.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    /// `nil` for rules that are absent or blank — a group with no rules
+    /// asks for no agreement, and an empty string must not become a
+    /// thing to sign.
+    public static func normalized(_ rules: String?) -> String? {
+        guard let rules else { return nil }
+        let canonical = canonical(rules)
+        return canonical.isEmpty ? nil : canonical
+    }
+
+    /// SHA-256 over the canonical text's UTF-8. The identity of a set
+    /// of rules, and what the founder compares against.
+    public static func hash(_ rules: String) -> Data {
+        Data(SHA256.hash(data: Data(canonical(rules).utf8)))
+    }
+
+    /// The exact bytes both sides sign and verify. See the type doc.
+    public static func statement(
+        groupID: Data,
+        rulesHash: Data,
+        joinerSendingPublicKey: Data
+    ) -> Data {
+        var bytes = Data(domain.utf8)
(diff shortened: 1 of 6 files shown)


dev-onym (commented):
REQUEST_CHANGES — the crypto shape is sound, but the cap and the canonicalization don't hold up across platforms.

- `GroupRules.swift:79` — `maxLength` is measured in `String.count` (grapheme clusters) but justified in bytes. 500 ZWJ emoji clusters (👨‍👩‍👧‍👦 = 25 UTF-8 bytes) is a legal 500-"character" rules text that produces a ~12.5 KB payload and a ~17 KB link — 7× the 2331-byte ceiling the cap exists to defend. The measurement table only covers ≤3 bytes/char. Cap the UTF-8 byte count (keep a character cap for the UI if you like).
- `IntroCapability.swift:157` — same unit, but now on a wire documented byte-for-byte with Android. Kotlin's `String.length` counts UTF-16 units, so any non-BMP character makes the two platforms disagree: 500 emoji is `count == 500` (iOS accepts) and `length == 1000` (Android rejects). Because decode *rejects* rather than truncates, that asymmetry is an unusable link, not a cosmetic difference. A byte cap is also the only unit both platforms compute identically.
- `GroupRules.swift:94` — `.whitespacesAndNewlines` includes NBSP (U+00A0) and the Unicode space separators; Kotlin's `trim()`/`Character.isWhitespace` does **not** trim NBSP. Rules text with a leading or trailing NBSP (paste from a web page, common) hashes differently on the two platforms, so a genuine agreement fails `isAgreement` — and per the docs elsewhere in this PR the founder reads that as "signed something that doesn't verify", i.e. forgery, rather than a stale version. Pin the exact codepoint set in the doc comment, since Android has to reproduce it exactly.
- `JoinRequestSender.swift:84` — `signWithStellarKey(_:)` signs with whatever identity is *currently selected* at call time, while the payload announces `active.stellarPublicKey` captured at line 55. An identity switch across those awaits yields a signature that verifies against no announced key — silently, and again in the "forgery" bucket. `signWithStellarKey(_:matchingPublicKeyHex:)` (`IdentityRepository.swift:344`) exists for exactly this binding; the existing `sealInvitation` has the same gap but fails loudly instead.
- `JoinRequestSender.swift:53` — `agreedRules: String? = nil` makes "forgot to pass it" the default. A caller that omits it after showing the rules screen sends no agreement and the founder sees a joiner who declined. Given the parameter's whole purpose is that it can't be derived from `capability`, drop the default, or return a failure when `capability.rules != nil && agreedRules == nil`.
- `GroupRules.swift:113` — the unambiguity argument rests on all three components being fixed-length, but `statement` length-checks none of them; `isAgreement` gets the key checked by CryptoKit for free and `groupID` not at all. Cheap `precondition`s (or a throwing initializer) would keep a future caller from silently producing an ambiguous preimage.
- No tests. Deferring to the last PR of a stack means these bytes can land alone if the stack lands partially, and `statement()` is precisely the thing Android must match — a golden-vector test (fixed key, fixed rules, expected signature/hash hex) belongs in this PR, alongside the over-long-`rules` decode rejection and the hash/signature pairing rule.

dev-onym (commented):
Verified the golden vectors independently with an RFC 8032 implementation: hash `4405…6473`, pubkey `ea4a…d22c`, 115-byte statement, and the "foreign" signature `d1b3…810e` all reproduce exactly. The construction (domain sep + group binding + signer binding, no timestamp) is sound, and pinning `trimmedCodepoints` instead of `.whitespacesAndNewlines` is the right call.

**COMMENT** — crypto is correct; the byte budget, one doc claim, and the negative-test coverage need attention.

- `GroupRules.swift:96` — the `maxLength` doc says "Latin text hits neither before the other," but 500 Latin chars = 500 bytes, a third of `maxBytes`. For Latin the char cap binds far first; the two are not commensurate. `maxLength` is also unused in this PR.
- `GroupRules.swift:84-89` + `SettingsQRCode.swift:33` — the admitted overrun is worse than documented. The 2331-byte ceiling is version-40/level-M capacity for an *unoccluded* code, but `SettingsQRCode` overlays a badge across ~22%×22% of the symbol (~5% of modules, contiguous, so per-block recovery is the real limit). A max-length link may be unscannable while still generating an image — a failure mode with no signal at all, unlike the nil-image case the doc describes. Either lower `maxBytes` or gate on measured scannability.
- `IntroCapability.swift:167` vs `:149` — `rules` is canonicalized and capped; `groupName` is neither trimmed nor capped. That asymmetry is exactly what makes the pair overrun possible. Capping the *sum* at encode time would close it in one place.
- `JoinRequestSender.swift:87-92` — nothing checks `agreedRules` against `capability.rules` when both are present. A UI wiring bug signs text the founder cannot reproduce, and the founder sees a signature that fails to verify — the "forged vs. stale" confusion `rulesHash` exists to prevent, in the one case the hash can't distinguish. Cheap guard, same shape as the nil check above it.
- `JoinRequestSender.swift:88,90` — `GroupRules.normalized(agreedRules)` runs twice (three canonicalizations counting `hash`); hoist to a `let`.
- `GroupRulesVectorTests.swift` — only positive paths. `isAgreement` has no test for a wrong `groupID`, a wrong `joinerSendingPublicKey`, tampered rules text, or a 63/65-byte signature; the sender's "invite carries rules, `agreedRules` nil" guard is untested. Those bindings are the entire security claim, and PR #290's lesson is that the untested half is where the defect lives.
- `GroupRules.swift:112-124` — the cross-platform note covers only the Foundation-trims-more direction. Kotlin's `String.trim()` also strips U+001C–U+001F, which `trimmedCodepoints` does not, so Android must not use bare `trim()` either. Worth one sentence, since the doc is the Android spec.
- `JoinRequestApprover` / `PersistedIntroRequest` are untouched: `rules_hash`/`rules_signature` decode and are then dropped — nothing verifies or retains them. Expected for a stacked PR, but the `GroupRules` type doc states the verifier *does* store the text, which reads as describing today.
- PR body is stale against the code: it describes a 500-character cap and a 2258/2331-byte measurement; `aaaf81e` moved to 1500 UTF-8 bytes and a 2273-byte link. Update before merge so the rationale matches the constant.
- Test doc claim at `GroupRulesVectorTests.swift:83` that CryptoKit Ed25519 signing is randomized is dubious (it reproduces RFC 8032 deterministic output here) — harmless, since verify-not-reproduce is the right contract regardless.

dev-onym (commented):
**COMMENT** — the statement construction and vectors are sound (I re-derived all three golden values independently); the findings are doc/scope-level, plus one real gap in the safety net.

**Verified independently** (pure-Python RFC 8032 + SHA-256, not CryptoKit):
- `GroupRulesVectorTests.swift:33` hash `440518f5…` ✓; `:41` pubkey from seed `0x07*32` = `ea4a6c63…` ✓; `:69` statement length 115 ✓; `:87` the "foreign" signature verifies and is exactly the deterministic RFC 8032 signature ✓. `Curve25519.Signing` matches `IdentityRepository.stellarSigningPrivateKey` (`IdentityRepository.swift:918`), and `stellarPublicKey` is always the 32-byte raw representation, so `statement`'s preconditions can't fire from the sender path.
- `GroupRules.swift:110` — the trim set does match `CharacterSet.whitespacesAndNewlines` (Zs ∪ 0009–000D ∪ 0085 ∪ 2028/2029), and Java's `Character.isWhitespace` really does exclude U+00A0/2007/202F, so spelling it out is warranted.

**Findings**
- `GroupRules.swift:90` — references `GroupRulesWireTests`, which doesn't exist; the only test file added is `GroupRulesVectorTests`, and nothing in it pins the name+rules boundary the sentence claims is pinned. Either add that test or drop the claim.
- `GroupRules.swift:81-89` — the cap doesn't actually guarantee the property it exists for: 1500 bytes of rules plus an uncapped `groupName` (`IntroCapability.swift:83`) overruns 2331 bytes, and `SettingsQRCode.generate()` returns nil (`SettingsQRCode.swift:44`) → the invite screen loses its QR for a link the mint side happily accepted. Since the budget is shared, the check belongs on the encoded link length (or `groupName` needs its own cap), not on `rules` alone.
- `JoinRequestSender.swift:88` — the "wiring mistake" guard only fires for link invites. `GroupInviteOfferPayload.introCapability()` (`GroupInviteOfferPayload.swift:123`) has no `rules` field, so for a pushed invitation `capability.rules` is always nil and a forgotten `agreedRules` sends silently — exactly the failure the parameter's doc comment says it's guarding, in the one case the comment cites as the reason for the parameter's existence.
- PR body vs. code: the body describes a 500-character cap and a 2258/2331-byte measurement; the code ships `maxBytes = 1500` with a 2273-byte measurement. The body reads as pre-`aaaf81e` prose — worth updating, since the cap's justification is the reviewable part.
- `IntroCapability.swift:149-164` — the cap is applied *after* trimming, so a hostile link can carry megabytes of NBSP in `rules` (or anything at all in `group_name`) and decode fine. Low severity given `fromLink` returns nil on nothing here, but it means the decode-side rejection isn't a bound on link size.
- Scope note, not a defect: `GroupRules.isAgreement` and both new payload fields have no consumer — the approver never verifies or persists them, so the type doc's "the verifier therefore stores the exact rules text" is a promise the stack hasn't kept yet. `maxLength` (`GroupRules.swift:98`) is likewise unused. Fine if the follow-up PRs land together.

PR #301, opened 2026-08-21
Tell the founder what the joiner agreed to, and keep the proof

rinat-enikeev:
Stacked on #300. The signature arrives there; this is the side that reads it and remembers it.

**The verdict is decided once**, in `JoinRequestApprover` at decrypt time, where the group's own rules are already in hand — a view is the wrong place to be verifying signatures.

Five outcomes rather than a `Bool`, because the cases a `Bool` collapses are exactly the ones a founder is deciding between:

| | what it means | what to do about it |
|---|---|---|
| `notRequired` | the group has no rules | nothing to show |
| `agreed` | verified against the group's current text | — |
| `agreedToOtherRules` | valid signature over a different wording | they accepted an earlier version |
| `notSigned` | rules exist, no agreement came | an older build; ask again |
| `invalid` | a signature that doesn't verify | the only one that should give pause |

**Verification runs against our copy of the rules, never against the hash the joiner sent.** Taking their hash as the thing to check would let a joiner choose the text their own signature is checked against. The hash they send only ever explains a failure.

**None of the five blocks Accept.** Rejecting outright would silently exclude every joiner on a build that predates rules — today, all of Android. The founder is shown the verdict *above* the buttons, not under them, and decides.

**The proof is kept on the member, not the request.** The request is consumed at approval; the question ("did they agree?") outlives it by the whole life of the membership. It rides in `MemberProfile`, so it is announced with the rest of the profile and any member can check any other member's agreement against the `sendingPubkey` they already hold — the founder who admitted them is not a required witness.

Two smaller calls worth flagging for review:

- Malformed agreement bytes on the wire are **dropped, not thrown on**. "We can't show they agreed" is what nil already means, and rejecting the whole profile over it would take a member's inbox and verification keys down with it — breaking the chat for a field that only ever adds evidence.
- The agreement is recorded **even when it said nothing**. The founder saw the verdict and approved anyway; keeping the bytes is what lets that be re-examined. Dropping them makes "approved someone who signed nothing" and "approved someone whose signature was wrong" identical after the fact.

`OnymTokens.amber` is new — caution, for the two states a founder should read before deciding, where `red` would claim the decision is already wrong.

Verification: 1497 existing unit tests green. Tests land in the final PR of the stack.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatJoinRequestCell.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatJoinRequestCell.swift
@@ -1,4 +1,5 @@
 import UIKit
+import SwiftUI
 import OnymDesign
 import OnymGroup
 
@@ -16,6 +17,11 @@ struct ChatJoinRequestDisplay: Equatable {
     /// self-asserted by the joiner and nothing stops them typing someone
     /// else's name.
     let fingerprint: String
+    /// Whether this joiner agreed to the group's rules, already
+    /// decided by `JoinRequestApprover` against the group's own copy of
+    /// the text. The cell renders it and nothing more — a view is the
+    /// wrong place to be verifying signatures.
+    let agreement: JoinRequestApprover.RulesAgreement
     /// A call is in flight: spinner on, both buttons disabled.
     let isInFlight: Bool
     /// Last failure for this request, if any.
@@ -42,6 +48,7 @@ final class ChatJoinRequestCell: UITableViewCell {
     private let card = UIView()
     private let titleLabel = UILabel()
     private let fingerprintLabel = UILabel()
+    private let agreementLabel = UILabel()
     private let errorLabel = UILabel()
     private let hintLabel = UILabel()
     private let declineButton = UIButton(type: .system)
@@ -78,6 +85,10 @@ final class ChatJoinRequestCell: UITableViewCell {
         fingerprintLabel.adjustsFontForContentSizeCategory = true
         fingerprintLabel.textColor = UIColor(OnymTokens.text3)
         fingerprintLabel.numberOfLines = 1
+
+        agreementLabel.font = .preferredFont(forTextStyle: .caption1)
+        agreementLabel.adjustsFontForContentSizeCategory = true
+        agreementLabel.numberOfLines = 0
 
         errorLabel.font = .preferredFont(forTextStyle: .caption1)
         errorLabel.adjustsFontForContentSizeCategory = true
@@ -117,6 +128,7 @@ final class ChatJoinRequestCell: UITableViewCell {
         stack.spacing = 8
         stack.addArrangedSubview(titleLabel)
         stack.addArrangedSubview(fingerprintLabel)
+        stack.addArrangedSubview(agreementLabel)
         stack.addArrangedSubview(buttonRow)
         stack.addArrangedSubview(hintLabel)
         stack.addArrangedSubview(errorLabel)
@@ -168,6 +180,14 @@ final class ChatJoinRequestCell: UITableViewCell {
         titleLabel.text = String(localized: "\(display.alias) wants to join")
         fingerprintLabel.text = String(localized: "inbox \(display.fingerprint)")
 
+        // Above the buttons on purpose: it is one of the two things
+        // this decision turns on, and a founder who has already tapped
+        // Accept is not helped by finding out underneath it.
+        let agreement = Self.agreementText(display.agreement)
+        agreementLabel.text = agreement?.text
+        agreementLabel.textColor = agreement.map { UIColor($0.color) }
+        agreementLabel.isHidden = agreement == nil
+
         var accept = Self.buttonConfiguration(
             title: display.isInFlight
                 ? String(localized: "Anchoring on chain…")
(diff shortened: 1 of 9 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the design is right, but two of the five verdicts don't mean what they say, and the proof doesn't propagate as the PR body claims.

- `JoinRequestApprover.swift:1004` — `agreedToOtherRules` is asserted without verifying anything. The only thing that distinguishes it from `invalid` is that the joiner's *self-supplied* hash differs from ours. 64 random bytes plus a random 32-byte hash therefore render as amber "Signed a different version of the rules" — a claim nothing supports — while the honest failure mode gets red. An attacker picks their own verdict by not echoing our hash, and it's the softer one. Either fold both into `invalid`, or rename/reword to something the code can back ("signed rules we don't have — can't be checked") and keep the color at `red`/neutral, not reassuring amber.
- `IncomingMessageDispatcher.swift:864` (and `:767`) — "any member can check any other member's agreement" doesn't hold. Announcements carry `AnnouncedMember`, which has no rules fields, and the receiver rebuilds `MemberProfile(alias:inboxPublicKey:sendingPubkey:)`, defaulting the evidence to nil. Only the admitting founder — and later joiners who receive the whole map via `GroupInvitationPayload.memberProfiles` — ever hold it; existing members never learn it for anyone admitted after them. Either add the two fields to `AnnouncedMember` (`decodeIfPresent`, same drop-if-wrong-sized rule) or drop the claim from the PR body.
- `MemberProfile.swift:33-48` — `GroupRules`' own doc (`GroupRules.swift:47-54`) says the verifier stores the exact rules *text* alongside the signature, because a hash with no text proves something was agreed and never what. This stores hash + signature and points at the live `ChatGroup.invitationMessage`, which the founder can edit at any time; after one edit the retained agreement is unattributable to any text on the device. Worth retaining the agreed text (or an append-only rules history) before this ships.
- `OnymBrand.swift:27` — `amber` light variant `#FF9500` on the card background is ~2.1:1; as caption-sized body text it fails WCAG AA, and it carries the two verdicts a founder is most expected to actually read. Darken the light value (~`#B25E00` reaches 4.5:1); the dark variant is fine.
- `ChatJoinRequestCell.swift:188` — `textColor = nil` in the `.notRequired` path relies on `UILabel.textColor` being an IUO that resets to default; harmless since the label is hidden, but set it inside the non-nil branch instead. Also worth an `accessibilityIdentifier` here like the buttons have, since the stack's tests land later and this line is the thing under test.
- `JoinRequestApprover.swift:119-121` — defaulting `rulesAgreement` to `.notRequired` in the memberwise init means a call site that forgets it silently reads as "group has no rules" rather than failing to compile. Only tests and this one construction site exist; drop the defaults on the two agreement params while that's still cheap.
- `MemberProfile.swift:108` — the paired nil-out is correct, but the memberwise init accepts any-sized bytes, so `recordJoiner` can persist a malformed local pair that decode would have rejected. `JoinRequestPayload` already validates upstream, so this is belt-and-braces, not a live bug.

dev-onym (commented):
REQUEST_CHANGES — the design and the five-way verdict are right; two wire/persistence issues need fixing first.

- **`rules_text` is unbounded on the wire, and it fans out.** `MemberAnnouncementPayload.AnnouncedMember.init(from:)` (`Packages/OnymGroup/Sources/OnymGroup/MemberAnnouncementPayload.swift:183`) and `MemberProfile.init(from:)` (`Packages/OnymGroup/Sources/OnymGroup/MemberProfile.swift:145`) accept a rules string of any length — `GroupRules.fits`/`maxBytes` is enforced only on the invite side (`IntroCapability.swift:123`). Because `MemberProfile` is snapshotted into `GroupInvitationPayload.memberProfiles` (`GroupInvitationPayload.swift:61`), one authorized-but-hostile announcement with a multi-MB `rules_text` is persisted *and* re-shipped to every future joiner. `paired(hash:signature:)` already exists as the "wrong shape → no agreement" gate; the text belongs in it (`MemberProfile.swift:98`, `MemberAnnouncementPayload.swift:157`). Even benign traffic amplifies: identical text × N members ≈ 150 KB per invitation at 100 members.
- **The re-join path silently erases the proof.** `recordJoiner` (`JoinRequestApprover.swift:697`) and the dispatcher (`IncomingMessageDispatcher.swift:863`) replace the whole `MemberProfile`. A reinstalled member re-joining from a pre-rules build — the case `recordJoiner`'s own comment says the path exists for — arrives with nil bytes and overwrites a retained agreement with nothing. That contradicts this PR's stated reason for keeping bytes even when they said nothing. Preserve the existing triple when the incoming one is nil.
- **Stale doc ref:** `JoinRequestApprover.swift:116` still points at `agreedToOtherRules`; the case is `unknownRules` (`:62`). The PR body's table has the same stale name — worth fixing there too, since the rename is the security-relevant part of the change.
- **Contrast:** the `amber` note is verified — `#B25E00` on white is 4.67:1, dark 8.9:1. But the same caption line renders `.agreed` in `OnymTokens.green`, which is ~3.1:1 on `surface2` light (`ChatJoinRequestCell.swift:236`, `card.backgroundColor = surface2` at `:72`). The reasoning that darkened amber applies unchanged to the green on that line.
- **Minor / no action needed:** the persisted `rulesHash` is the joiner's claimed hash while `rulesText` is our copy, so for `.invalid`/`.unknownRules` the stored pair is internally inconsistent and `agreedToRules(groupID:)` (`MemberProfile.swift:112`) ignores `rulesHash` entirely. A later reader holding only the profile therefore can't tell "signed an earlier wording" from "forged" — the distinction this PR argues is the whole point of five cases. Persisting the verdict alongside the bytes would close that; the hash alone doesn't.

dev-onym (commented):
COMMENT — the verdict logic and "verify against our copy, never their hash" are right; the wire boundary for `rules_text` is the gap.

- `MemberAnnouncementPayload.swift:182` / `MemberProfile.swift:153` — `rules_text` is decoded with only `GroupRules.normalized`, no length check. Every other rules ingress caps it (`IntroCapability.swift:122` rejects over `GroupRules.maxBytes`), so this is the one path where a peer's announcement persists an arbitrarily large string into the group row — then re-encoded on every subsequent group write and re-fanned in the next announcement. Suggest rejecting or dropping-to-nil past `maxBytes` at both decode boundaries.
- `MemberProfile.swift:102` — `paired` keeps hash+signature when `rulesText` is nil, producing exactly the "hash with no text" record `GroupRules`' own doc (`GroupRules.swift:49-53`) argues is not evidence: `agreedToRules` can never return true for it, and nothing distinguishes it from a real agreement in storage. Text belongs in the pairing rule, or the doc should say why this one is kept.
- `IncomingMessageDispatcher.swift:842` — `if group.memberProfiles[key] != nil { return }` fires before the profile write, so on the re-join path (which `JoinRequestApprover.swift:447` calls out as the point of `recordJoiner`) the founder overwrites their local profile with the new agreement while every other member keeps the old one. The evidence the PR wants "any member can check" diverges by device precisely in the case the approver treats as expected.
- `JoinRequestApprover.swift:116` — doc references `agreedToOtherRules`; the case is now `unknownRules`.
- `JoinRequestApprover.swift:695`+ — `recordJoiner` stores the joiner's `rulesHash` beside *our* `rulesText`, and `agreedToRules` (`MemberProfile.swift:113`) ignores the hash entirely. In the `unknownRules` case the stored pair is internally inconsistent by design; worth one line on `agreedToRules` saying the hash is a claim, not part of the check, so a later reader doesn't add a comparison that breaks it.
- Merge order: alone, this commit reads `group.invitationMessage` as the rules while nothing yet mints links carrying them (`InviteIntroducer.swift:73` gains `rules` only in `29783bb`), so every request into a group with a welcome message renders amber "Didn't sign the group rules". Fine if the stack lands in order — not fine if this one is cherry-picked.
- Test debt is real but as declared: nothing pins the five-way branch in `agreement(for:rules:)` or the drop-malformed rule in `paired`, and those two are the parts of this PR most likely to be quietly regressed. Worth ensuring the final stack PR covers the `unknownRules`-vs-`invalid` split specifically, since it's the case the earlier naming got wrong.

PR #302, opened 2026-08-21
Group rules, read on the way in and signed on the way out

rinat-enikeev:
Stacked on #301. The screens, and the rename.

The field's own placeholder already read *"Add a greeting, group rules, or policy…"*. This makes it the rules, gives joiners something to agree to, and carries the agreement back.

**The founder's side.** "Invitation message" → "Group rules", and the footnote now says what that costs the person reading it: anyone joining reads these and signs that they agree. The field is clamped to 500 characters rather than validated at submit — the cap is the invite QR code's, and letting someone type past it into a field that will refuse to mint a link at the end is the worse failure. The counter appears at the last hundred characters and not before: one sitting at 500 on an empty field reads as a demand for 500 characters.

**The link.** `ShareInviteFlow` and the create-time invitee mint both pass the group's *current* rules — including down the `currentOrMint` path that reuses a live intro key. The link is a pointer to the group, and a reused key handed out with rules from whenever it was minted would have joiners signing text the founder had already replaced.

**The joiner's side.** The confirmation screen grows a rules card: the full text, never truncated behind a "more", because Send signs this exact string and a signature over folded-away words is not much of an agreement. The tick under it turns reading into agreeing, and it gates Send — **but only for a group that has rules**.

That last point is the one design call worth arguing with. A group that asks nothing keeps the one-tap join the pre-filled name bought; a tick standing for nothing is friction that teaches people to tick without reading. With rules, the button says what it does: *"Agree and send request"*.

**The rules signed are the ones the screen showed**, threaded through as a value rather than re-read from the capability at send time. A link's copy wins over a stored offer's when both exist, since the person read the invitation they actually opened. And the text is kept on the pending row, because "Ask again" months later has to re-sign the same words — the capability is long gone by then.

Twelve strings, en and ru.

Verification: 1497 existing unit tests green; the whole stack builds. Tests for the feature land in the PR after this one.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift
@@ -42,8 +42,15 @@ public struct JoinConfirmView: View {
             ScrollView {
                 VStack(spacing: 18) {
                     header
+                    // Same field, two names: since the rename, a
+                    // group's invitation message *is* its rules, so an
+                    // offer whose text became the rules card must not
+                    // also draw it as a greeting. Only a message that
+                    // says something different earns its own card.
                     if let message = confirmation.invitationMessage,
-                       !message.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+                       !message.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty,
+                       message.trimmingCharacters(in: .whitespacesAndNewlines)
+                           != confirmation.rules {
                         invitationCard(message)
                     }
                     if let rules = confirmation.rules {
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift
@@ -33,6 +33,9 @@ public struct JoinConfirmView: View {
 
     @State private var label: String = ""
     @State private var isSending = false
+    /// Unticked until the person ticks it. Only ever gates Send for a
+    /// group that has rules — see `canSend`.
+    @State private var agreedToRules = false
 
     public var body: some View {
         NavigationStack {
@@ -42,6 +45,9 @@ public struct JoinConfirmView: View {
                     if let message = confirmation.invitationMessage,
                        !message.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
                         invitationCard(message)
+                    }
+                    if let rules = confirmation.rules {
+                        rulesCard(rules)
                     }
                     nameField
                     disclosure
@@ -106,6 +112,48 @@ public struct JoinConfirmView: View {
             .accessibilityIdentifier("join_confirm.invitation_message")
     }
 
+    /// The rules, and the tick that turns reading them into agreeing.
+    ///
+    /// Send signs this exact text, so it is shown in full rather than
+    /// truncated behind a "more": a signature over words that were
+    /// folded away is not much of an agreement. The text is
+    /// founder-supplied and untrusted, hence rendered plain — no
+    /// markdown, no links to follow.
+    private func rulesCard(_ rules: String) -> some View {
+        VStack(alignment: .leading, spacing: 10) {
+            Text("Group rules")
+                .font(.system(size: 12, weight: .semibold))
+                .foregroundStyle(OnymTokens.text3)
+            Text(rules)
+                .font(.system(size: 14))
+                .foregroundStyle(OnymTokens.text)
+                .frame(maxWidth: .infinity, alignment: .leading)
+                .textSelection(.enabled)
+                .accessibilityIdentifier("join_confirm.rules_text")
+            Divider().overlay(OnymTokens.hairline)
+            Toggle(isOn: $agreedToRules) {
+                Text("I agree to these rules")
+                    .font(.system(size: 13, weight: .medium))
+                    .foregroundStyle(OnymTokens.text)
+            }
+            .tint(OnymAccent.blue.color)
+            .accessibilityIdentifier("join_confirm.agree_toggle")
+            // Said plainly, because it is the part that outlives the
+            // tap: the founder keeps this, and can show it.
+            Text("Your agreement is signed with this identity\u{2019}s key and sent with your request.")
+                .font(.system(size: 11))
+                .foregroundStyle(OnymTokens.text2)
+        }
+        .padding(14)
+        .background(OnymTokens.surface2)
+        .clipShape(RoundedRectangle(cornerRadius: 12))
+        .overlay(
+            RoundedRectangle(cornerRadius: 12)
+                .stroke(OnymTokens.hairline, lineWidth: 1)
+        )
+        .accessibilityIdentifier("join_confirm.rules")
+    }
+
     /// Pre-filled, never blank: the identity's own name is the answer
     /// almost every time, so joining is Send and nothing else. The field
     /// is here for the times it isn't — a name you'd rather this room
@@ -148,6 +196,17 @@ public struct JoinConfirmView: View {
                     .font(.system(size: 12))
                     .foregroundStyle(OnymTokens.text2)
             }
+            if confirmation.rules != nil {
+                Label {
+                    Text("Your signed agreement to the rules goes with it.")
+                        .font(.system(size: 12))
+                        .foregroundStyle(OnymTokens.text2)
+                } icon: {
+                    Image(systemName: "signature")
+                        .font(.system(size: 12))
+                        .foregroundStyle(OnymTokens.text2)
+                }
+            }
             row("group", value: shortHex(confirmation.groupIDHex))
             row("invite key", value: shortHex(hex(confirmation.introPublicKey)))
         }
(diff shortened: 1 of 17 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the rules plumbing has two holes where a joiner ends up signing nothing, or signing text they were never shown.

- **`rotate()` drops the rules entirely** — `ShareInviteFlow.swift:131` calls `introducer.rotate(...)` with no `rules:`, and `InviteIntroducer.swift:174` has no such parameter, so it mints via `mint(...)` at `InviteIntroducer.swift:192` with `rules: nil`. After "Generate new link", every joiner on that link sees no rules card, gets the one-tap join, and lands on the founder's screen as `.notSigned` ("Didn't sign the group rules", `JoinRequestApprover.swift:974-991`) — the same failure the PR closes for `currentOrMint`.
- **`.alreadyPresent` signs the row's text, not the screen's** — the screen showed `capability.rules ?? existing?.invitationMessage` (`PendingChatsFlow.swift:323`), but `confirmJoin` falls through to `send(existing, ...)` (`:426-427`), and `send` reads `chat.invitationMessage` (`:515`). A pushed offer whose stored text differs from the link's rules → signature over words nobody read; offer text `nil` + link rules present → tick required, nothing signed. `:412`'s `invitationMessage: capability.rules` only lands on `.inserted`, so "Ask again" re-signs the stale text too.
- **Rules aren't normalized at the confirmation boundary** — `:323` and `:362` pass raw `chat.invitationMessage`. `CreateGroupFlow` never trims for whitespace-only input, so `"   "` yields non-nil `rules` → an empty rules card, a mandatory tick, "Agree and send request" — then `JoinRequestSender.swift:81` normalizes to nil and sends no signature. Wrap both in `GroupRules.normalized(...)`.
- **The same string renders twice on the offer path** — `:358` and `:362` set `invitationMessage` and `rules` to one value, and `JoinConfirmView.swift:45-51` draws both cards. Every pushed invitation (and every link row re-opened via Ask again, `:412`) shows its text as an invitation card and again as the rules card.
- **The 500 cap is only in the create form** — `CreateGroupFlow.swift:49`. The old doc said "any length", so a group created by an earlier build with a longer message now throws out of `IntroCapability.init` (`IntroCapability.swift:122`) inside `currentOrMint`, and `ShareInviteFlow.swift:105` turns it into `.failed(reason: "shape(rules: …)")` — that group can no longer produce a share link at all. `CreateGroupInteractor.swift:747` has the same shape for creation. Clamp or omit the link's copy instead of failing the share.

Smaller things:

- `Я согласен` (`Localizable.xcstrings`, "I agree to these rules") is masculine-gendered; `Я принимаю эти правила` reads for everyone.
- The clamp sits in the flow as `didSet` while the name field clamps in the view via `onChange` (`CreateGroupView.swift:229`) — worth matching, and worth confirming `@Observable` still tracks a property with an observer so `rulesRemaining` updates live (`CreateGroupFlow.swift:49-62`).

The design call you flagged — gating Send only when rules exist — reads right to me: a tick over an empty card is worse than no tick. The `currentOrMint` fix and threading the text as a value rather than re-reading the capability are both the correct instinct; findings 1 and 2 are the two places that instinct isn't yet enforced end to end.

dev-onym (commented):
**COMMENT** — the happy paths are correct and the "sign what was shown" threading holds up; findings are a defeated backstop, a redundant write, and one founder-facing blind spot.

- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:534` — `send` rebuilds the capability *without* `rules`, so `capability.rules` is always nil by the time it reaches `JoinRequestSender.swift:89`. That guard ("invite carried rules, send has none → fail loudly") is therefore inert on the only production path, which is exactly the wiring bug it was added to catch. Concrete: a row whose `invitationMessage` didn't persist (`SwiftDataPendingChatStore.swift:124` swallows encryption failures with `try?`) makes `retry` (`:497`) send `rules: nil`, and the founder reads that as "Didn't sign the group rules" (`ChatJoinRequestCell.swift:245`). Passing `rules` into the init would fix it but throws for an over-cap legacy offer — so either accept the raw text here for the guard's sake, or move the assertion into `PendingChatsFlow.send`.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:441-453` — redundant. `repository.record(chat)` already calls `refreshOffer(chat)` on `.alreadyPresent` (`PendingChatRepository.swift:35`) with the same `invitationMessage: confirmation.rules`, so the reconstructed `shown` writes identical values and `send(shown, …)` is byte-identical to `send(existing, …)` (only id/keys/groupName are read). The comment's premise ("the row is brought up to what was actually read") is satisfied before line 433 runs — and `existing.inviterAlias` read there has already been blanked to `""` by that first refresh, which the comment doesn't reflect. Drop the block or drop the refresh in `record`.
- `Packages/OnymGroup/Sources/OnymGroup/ChatGroup.swift:149` — the over-cap fallback is silent in both directions. A legacy/imported group (`BackupSinks.swift:97` writes `invitationMessage` unclamped; only `CreateGroupView.swift:267` clamps) mints links with no rules, so every link joiner arrives as "Didn't sign the group rules" while the founder is never told their rules are too long to travel. `ShareInviteFlow.swift:93` is the place to surface it — one line next to the link ("These rules are too long for a link; joiners from a link won't be asked to agree") turns an unexplained accusation into a fact.
- `Packages/OnymGroup/Sources/OnymGroup/CreateGroupInteractor.swift:757` — hand-inlines `linkableRules`. Same two calls, same tradeoff, two copies; a `GroupRules.linkable(_:)` both call keeps them from drifting.
- `Packages/OnymGroup/Sources/OnymGroup/CreateGroupView.swift:289` — `Text("\(flow.rulesRemaining)")` interpolates into a `LocalizedStringKey`, registering a bare `"%lld"` key that isn't in the catalog. `Text(verbatim:)` says what's meant; the `accessibilityLabel` at `:292` is correctly keyed.
- Verification: no Swift toolchain here, so the 1497-green claim is unverified locally. Deferring *all* tests is the wrong call for this particular PR — `prepareJoin`'s link-beats-offer precedence (`:323`) and the `.alreadyPresent` branch decide which bytes get signed, and both are one-liners to pin. The `SpyJoinSender.rules` plumbing is already in place for it.

dev-onym (commented):
**COMMENT** — the threading is sound and the "sign what was shown" invariant holds on every path I traced; two gaps worth closing before merge.

- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:534` — `send` rebuilds the capability with `groupName` only, so `capability.rules` is always nil by the time it reaches the sender. That makes `JoinRequestSender.swift:89` ("fail here, where it is still a bug report") unreachable from the *only* production path, and the app's `submitJoin` (`Sources/OnymIOS/OnymIOSApp.swift:788`) passes that same capability. A future caller that loses `rules` now sends unsigned and silently reaches the founder as a decline — exactly what that guard exists to prevent. Note the fix isn't just `rules: rules`: a legacy offer's `invitationMessage` can exceed `maxBytes` and `IntroCapability.init` throws on it, turning the offer path into "This invite is malformed". Thread a `rulesRequired: Bool` alongside instead.
- `Packages/OnymGroup/Sources/OnymGroup/ChatGroup.swift:150` — omitting over-long legacy rules keeps "Share invite" working, and omitting beats truncating. But the approver still reads the group's *full* `invitationMessage` (`JoinRequestApprover.swift:1016`), so every link joiner on such a group lands as `.notSigned` → "Didn't sign the group rules" — indistinguishable from refusal, and the same failure `InviteIntroducer.swift:196` argues against for `rotate`. `ShareInviteFlow` knows when it dropped them; it should say so to the founder rather than mint quietly.
- `Resources/Localizable.xcstrings` (`%lld characters left`) — plain `stringUnit`, no plural variations. Renders "1 characters left" in en, and ru needs one/few/many. Only an accessibility label, but it's the one new string with a number in it.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/JoinConfirmView.swift:44` — the dedupe trims with `.whitespacesAndNewlines` and compares against `confirmation.rules`, which was canonicalised with `GroupRules.trimmedCodepoints`. `GroupRules.swift:122` pins that set precisely so a Foundation revision can't drift it; comparing across the two sets re-couples them, and a drift shows the same text twice. `GroupRules.normalized(message) != confirmation.rules` is the same line and immune.
- `Packages/OnymGroup/Sources/OnymGroup/CreateGroupInteractor.swift:757` — reimplements `linkableRules` inline (`normalized` + `fits`). Same rule, two places to change; call the property.
- `Packages/OnymGroup/Sources/OnymGroup/CreateGroupFlow.swift:47` — the clamp lives only in the view's `onChange`, so any non-view write to `invitationMessage` overshoots and the interactor then drops the rules from the link with no counter to warn anyone. The `didSet`/`@Observable` reasoning is fair; applying `clampedRules` at the interactor boundary would close it without that.
- Coverage: `canSend`'s tick gate, `ShareInviteFlow` passing `linkableRules`, and `currentOrMint` refreshing rules on a reused key are all untested here — the three claims the PR body leans on hardest. Fine if the follow-up PR covers them; the `currentOrMint` one is the one I'd most want pinned, since it's a live-key path that silently reuses stored state.

PR #303, opened 2026-08-21
Tests for the rules, and for the ways the signature must fail

rinat-enikeev:
Stacked on #302. Forty tests for the whole stack.

The negative cases carry the weight here. A signature that verifies is one property; a signature that *keeps* verifying after the group, the signer, or a comma has changed would make the whole thing decorative. So each binding in the statement gets a test that breaks it:

- edit the rules → stops verifying
- move the acceptance to another group with identical text → stops verifying
- re-attribute it to a different joiner → stops verifying
- hand it garbage (a zero signature, a two-byte key) → answers "no" rather than throwing

**Across versions**, in the directions that only matter later: a link minted before rules existed still opens, a request from an older build decodes with no agreement, and an over-long `rules` is rejected at the **decode** boundary — the mint-time cap is not a defence, because a hostile link never passes through it.

**Two QR tests, and the second pins a limit rather than a guarantee.** 500 CJK characters of rules with a Latin group name fits a level-M code; the same rules with a 30-character CJK name comes to 2338 bytes against a 2331-byte ceiling and does not. The group name has no cap of its own, so the pair is what a future change has to re-measure. I'd rather record that as a measured boundary than have someone rediscover it as a QR code that silently won't render. `GroupRules.maxLength`'s comment carries the same numbers.

**The five verdicts** get a test each, including the two that are easy to conflate:

- a signature *stolen* from another joiner fails as `.invalid` — the hash matches, so it is not a stale version
- a joiner who signs their own terms and sends an honest hash of them reads as `.agreedToOtherRules`

That last one is the test for the rule that verification runs against *our* copy of the rules. Check the hash they sent, and a joiner picks the text their own signature is graded against.

**Screen to send**: a link's rules are what get signed, a pushed offer signs the text it arrived with, a group without rules signs nothing, and "Ask again" re-signs the same words from the row rather than from a capability that is long gone.

**The clamp** is checked to agree with `IntroCapability`'s cap — one character apart and every full-length rules set would fail at share time.

Verification: 1537 unit tests green (40 new).

### DIFF
--- a/Tests/OnymIOSTests/CreateGroupFlowTests.swift
+++ b/Tests/OnymIOSTests/CreateGroupFlowTests.swift
@@ -420,6 +420,22 @@ final class CreateGroupFlowTests: XCTestCase {
 
     // MARK: - Helpers
 
+    // MARK: - Group rules
+
+    func test_clampedRulesStillMintALink() async throws {
+        // The clamp and `IntroCapability`'s cap have to agree: one
+        // apart and every full-length rules set would fail at share
+        // time, after the founder had already typed it.
+        let overshoot = String(repeating: "x", count: GroupRules.maxBytes + 50)
+        XCTAssertNoThrow(
+            try IntroCapability(
+                introPublicKey: Data(repeating: 0x44, count: 32),
+                groupId: Data(repeating: 0x11, count: 32),
+                rules: CreateGroupFlow.clampedRules(overshoot)
+            )
+        )
+    }
+
     private func makeFlow() async -> CreateGroupFlow {
         let env = await CreateGroupFlowTestEnv.make()
         return CreateGroupFlow(interactor: env.interactor)
--- a/Tests/OnymIOSTests/GroupRulesTests.swift
+++ b/Tests/OnymIOSTests/GroupRulesTests.swift
@@ -1,0 +1,175 @@
+import XCTest
+import CryptoKit
+@testable import OnymIOS
+import OnymGroup
+
+/// The agreement itself: what the signature covers, and what it must
+/// refuse to cover.
+///
+/// The negative cases carry the weight here. A signature that verifies
+/// is one property; a signature that keeps verifying after the group,
+/// the signer, or a comma has changed would make the whole thing
+/// decorative.
+final class GroupRulesTests: XCTestCase {
+
+    private let groupID = Data(repeating: 0x11, count: 32)
+    private let otherGroupID = Data(repeating: 0x22, count: 32)
+    private let rules = "Be kind. No links. Ask before adding anyone."
+
+    // MARK: - Canonicalization
+
+    func test_canonical_trimsTheEndsAndNothingElse() {
+        // Inner shape is the founder's, and a joiner is agreeing to it:
+        // collapsing runs or normalising case would mean signing
+        // something other than what was displayed.
+        let messy = "  Rule one.\n\n  Rule  two.   \n"
+        XCTAssertEqual(GroupRules.canonical(messy), "Rule one.\n\n  Rule  two.")
+    }
+
+    func test_normalized_treatsBlankAsNoRules() {
+        // A group asking nothing must not acquire an empty string to
+        // sign — that would make every joiner tick a box about nothing.
+        XCTAssertNil(GroupRules.normalized(nil))
+        XCTAssertNil(GroupRules.normalized(""))
+        XCTAssertNil(GroupRules.normalized("   \n\t "))
+        XCTAssertEqual(GroupRules.normalized("  hi  "), "hi")
+    }
+
+    func test_hash_ignoresTheWhitespaceCanonicalizationRemoves() {
+        // Both sides hash through `canonical`, so a trailing newline
+        // out of a text field can't make agreement fail on a character
+        // nobody can see.
+        XCTAssertEqual(GroupRules.hash(rules), GroupRules.hash("\n\(rules)  "))
+    }
+
+    func test_hash_isSensitiveToTheTextItself() {
+        XCTAssertNotEqual(GroupRules.hash(rules), GroupRules.hash(rules + "."))
+    }
+
+    // MARK: - The statement
+
+    func test_statement_isDomainSeparated() {
+        let statement = GroupRules.statement(
+            groupID: groupID,
+            rulesHash: GroupRules.hash(rules),
+            joinerSendingPublicKey: Data(repeating: 0xEE, count: 32)
+        )
+        XCTAssertTrue(
+            statement.starts(with: Data("onym-group-rules-v1".utf8)),
+            "without the domain prefix this signature is replayable as any other this key makes"
+        )
+        // Domain + three fixed-length fields, which is what makes the
+        // concatenation unambiguous without length prefixes.
+        XCTAssertEqual(statement.count, "onym-group-rules-v1".utf8.count + 32 + 32 + 32)
+    }
+
+    // MARK: - Verification
+
+    func test_agreement_verifiesForTheJoinerWhoSignedTheseRules() {
+        let joiner = Curve25519.Signing.PrivateKey()
+        XCTAssertTrue(
+            GroupRules.isAgreement(
+                signature: sign(rules, as: joiner, group: groupID),
+                rules: rules,
+                groupID: groupID,
+                joinerSendingPublicKey: joiner.publicKey.rawRepresentation
+            )
+        )
+    }
+
+    func test_agreement_survivesTheTrimmingThatCanonicalizationDoes() {
+        // The founder's stored copy and the copy that travelled in a
+        // link can differ by a trailing newline and still be the same
+        // rules.
+        let joiner = Curve25519.Signing.PrivateKey()
+        XCTAssertTrue(
+            GroupRules.isAgreement(
+                signature: sign("\(rules)\n", as: joiner, group: groupID),
+                rules: rules,
+                groupID: groupID,
+                joinerSendingPublicKey: joiner.publicKey.rawRepresentation
+            )
+        )
+    }
+
+    func test_agreement_failsWhenTheRulesChanged() {
+        // The point of hashing the text: editing a rule invalidates
+        // every agreement made to the old wording, rather than silently
+        // extending them to words nobody accepted.
+        let joiner = Curve25519.Signing.PrivateKey()
+        XCTAssertFalse(
+            GroupRules.isAgreement(
+                signature: sign(rules, as: joiner, group: groupID),
+                rules: "\(rules) And no photos.",
+                groupID: groupID,
+                joinerSendingPublicKey: joiner.publicKey.rawRepresentation
+            )
+        )
+    }
+
+    func test_agreement_doesNotTransferToAnotherGroupWithTheSameRules() {
+        // Two groups can adopt identical text. Without `group_id` in the
+        // statement, an acceptance collected by the permissive one would
+        // be presentable as agreement to the strict one.
+        let joiner = Curve25519.Signing.PrivateKey()
+        XCTAssertFalse(
+            GroupRules.isAgreement(
+                signature: sign(rules, as: joiner, group: groupID),
+                rules: rules,
+                groupID: otherGroupID,
+                joinerSendingPublicKey: joiner.publicKey.rawRepresentation
+            )
+        )
+    }
+
+    func test_agreement_cannotBeReattributedToAnotherJoiner() {
+        // The signer is named inside the signed bytes, so a signature
+        // lifted from one request can't be presented as someone else's.
+        let signer = Curve25519.Signing.PrivateKey()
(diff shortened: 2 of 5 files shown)


dev-onym (commented):
## COMMENT

Read the tested code (`GroupRules`, `IntroCapability`, `JoinRequestPayload`, `JoinRequestApprover.agreement`, `PendingChatsFlow`) and re-derived the QR numbers independently. The suite is well-aimed — the negative cases genuinely break each binding, `.invalid` vs `.agreedToOtherRules` is pinned at the right seam, and the arithmetic in the doc comment is correct: JSON is `1645 + name_bytes`, so a 30-char CJK name → 1735 B → 2314 base64url chars → **2338** with the 24-char prefix, and a Latin name → 2234. 40 new test funcs, matching the claim.

- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:568` — "a link's rules are what get signed" is only pinned on the no-existing-row path, and the uncovered path is broken. Pushed offer with rules is still `.offered`, user opens a link for the same group minted by a pre-rules build: `PendingChatsFlow.swift:323` shows the offer's rules on the confirmation screen, then `confirmJoin` → `record` → `.alreadyPresent` → `PendingChatRepository.swift:35` → `refreshOffer` unconditionally overwrites `invitationMessage` with `capability.rules` (nil). `send` at `PendingChatsFlow.swift:515` then signs nothing, and the founder reads `.notSigned` for a joiner who was shown rules and tapped Send. The defect is in #302, but this is exactly the property this PR claims to guard — worth a test here either way.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:105` — `qrCapacityAtLevelM = 2331` is a private literal with no tie to `SettingsQRCode.swift:43` (`correctionLevel = "M"`). Change the renderer to L or H and both QR tests keep asserting against a stale ceiling while silently meaning nothing. Reading the level off the renderer, or at least naming `SettingsQRCode.swift:43` in the comment, makes the pair actually re-measure together.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:107` — "fits a QR code" is asserted as bytes ≤ capacity, but `SettingsQRCode.swift:30-34` overlays a centred mark covering ~22% of the side (~5% of modules at version 40). That eats a third of level M's ~15% recovery budget, so ≤2331 is necessary, not sufficient. Fine as a proxy, but the comment claiming the property rather than the arithmetic overstates it slightly.
- `Tests/OnymIOSTests/JoinRequestAgreementTests.swift` — `agreement(for:rules:)` builds the statement from joiner-supplied `payload.groupId`; the only thing stopping a joiner from presenting a signature bound to a group they control is the `payload.groupId == entry.groupId` check at `JoinRequestApprover.swift:957`, which nothing tests. `test_agreement_doesNotTransferToAnotherGroupWithTheSameRules` covers the primitive, not this seam.
- `Tests/OnymIOSTests/CreateGroupFlowTests.swift:375` — the three new tests land *after* `// MARK: - Helpers`, so that marker now heads a block of tests. Move the new `// MARK: - Group rules` section above it.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:59` — `XCTAssertNil(json?.index(forKey: "rules"))` also passes when `json` is nil, i.e. if the cast fails. `XCTAssertFalse(try XCTUnwrap(json).keys.contains("rules"))` can't pass for the wrong reason.

dev-onym (commented):
**REQUEST_CHANGES** — strong negative-case coverage overall, but three assertions can't fail under a wrong implementation, and the one place a real joiner signature is produced has no test.

- `Tests/OnymIOSTests/CreateGroupFlowTests.swift:389` — non-discriminating. 500×`則` is 1500 bytes *and* 500 characters, so both terms of `rulesRemaining`'s `min` are 0; the assertion (and the "out of bytes long before out of characters" message) passes equally under `max`. With `maxLength=500`/`maxBytes=1500`, 3-byte scripts hit both caps at the same point — only 4-byte scalars separate them (e.g. 400 emoji → chars 100, bytes −100).
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:251` — tautological: a Swift `String` cannot hold a partial scalar, so `clamped == String(clamped.unicodeScalars.map(Character.init))` is true for any value `clampedRules` could return. The real hazard is cutting a grapheme cluster (ZWJ family), which `prefix`/`removeLast` by `Character` also never does. Either assert on a ZWJ input or drop the line.
- `Tests/OnymIOSTests/CreateGroupFlowTests.swift:396` — ASCII overshoot never reaches the byte clamp: 1550 chars trims to 500 chars = 500 bytes, so `clampedRules`' `while utf8.count > maxBytes` loop never runs. The claim the test is named for (clamp output at exactly `maxBytes` still mints) needs multi-byte input, e.g. 600×`則` or the emoji string.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestSender.swift:90` — untested, and it's the producer side. Every flow test goes through `SpyJoinSender`, so nothing checks that the emitted `JoinRequestPayload` verifies under `GroupRules.isAgreement`, nor the guard that refuses to send when the invite carried rules and the send carries none. That's the one path a joiner's real signature comes out of.
- `Packages/OnymInbox/Sources/OnymInbox/PendingChatsFlow.swift:323` — the `capability.rules ?? existing?.invitationMessage` fallback branch is uncovered. Tests cover link-wins-over-offer and offer-alone, but not a legacy link (no rules) landing on a row that has them — which is where a joiner signs text the link never carried.
- Redundant with `GroupRulesVectorTests` already on the stack: `GroupRulesWireTests.swift:72` duplicates `GroupRulesVectorTests.swift:156` verbatim, `GroupRulesWireTests.swift:170` duplicates `:179` verbatim (same name and body), and `GroupRulesWireTests.swift:92` duplicates the over-long-on-the-wire case. Consolidate; ~5 of the "40 new tests" are re-runs.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:678` — looks up the row via a second `harness.makeChat().id` rather than the recorded chat's. It works only because `PendingChat.id` is `groupIDHex:ownerUUID` and the harness owner is fixed; a stray `nil` owner would make `prepareAccept` return `nil` and the test pass vacuously. Use the recorded chat.
- PR body describes the honest-hash case as reading `.agreedToOtherRules`; the enum case is `.unknownRules` (renamed in `JoinRequestApprover.swift:62`) and the test asserts that. Body is stale, tests are right.

Not blocking, and worth keeping as-is: the paired QR tests pinning a measured boundary rather than a guarantee, and `.invalid` vs `.unknownRules` being separated by whether the sender echoed our hash — that pair is the sharpest thing in the PR.

dev-onym (commented):
COMMENT — the negative-case coverage is genuinely strong; a few tests assert less than their comments claim, and three duplicate pre-existing ones.

- `Tests/OnymIOSTests/CreateGroupFlowTests.swift:389` — the byte-cap assertion doesn't discriminate. `maxBytes/3 = 500` CJK chars is *also* exactly `maxLength`, so both terms of `rulesRemaining`'s `min()` are 0 and the message "out of bytes long before out of characters" is false. Only 4-byte scalars separate the caps: 375 × `\u{1F600}` leaves 125 characters but 0 bytes — that input actually pins which cap is reported.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:72,92,170` — `test_rulesAtTheByteCap_areAccepted`, `test_overLongRulesArrivingOnTheWire_areRejectedToo`, and `test_halfAnAgreement_isRejected` are near-verbatim copies (same payloads, same comments) of `GroupRulesVectorTests.swift:156,165,179`, added in `aaaf81e` earlier in this stack. Drop them here or drop them there; two copies drifting apart is worse than one.
- `Tests/OnymIOSTests/PendingChatsFlowTests.swift:678` — vacuous on lookup failure. `prepareAccept(rowID:)` returns an optional and the test only asserts `confirmation?.rules == nil`, so a row that isn't found passes identically to a row with no rules. It also mints a throwaway `makeChat()` purely for its id (which only matches because `PendingChat.id` is `groupIDHex:owner`). `XCTAssertNotNil(confirmation)` first, or unwrap.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:242` — the "never cut a scalar" assertion can't fail. `String(clamped.unicodeScalars.map(Character.init))` re-forms grapheme clusters, so it equals `clamped` for every input, and `removeLast()` removes whole Characters anyway. The byte/char bounds above it are the real assertions; `GroupRulesVectorTests.swift:139` already covers the ZWJ case this was reaching for.
- `Tests/OnymIOSTests/JoinRequestAgreementTests.swift:167` — "the agreement rides the announcement" hand-rebuilds the `MemberProfile` rather than exercising the code that does it (`IncomingMessageDispatcher.swift:863`). If that mapping dropped `rulesText`, this test still passes. Worth noting `MemberProfile.agreedToRules` currently has no production caller, so this path is test-only end to end.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:42` — `XCTAssertNil(json?.index(forKey: "rules"))` also passes when the `as? [String: Any]` cast fails. `XCTUnwrap` the dictionary first.
- `Tests/OnymIOSTests/GroupRulesWireTests.swift:121` — pinning the QR overrun as a measured boundary is the right call, but the failure is softer than "soft": `SettingsQRCode.swift:25` renders a blank grey square with the Onym badge and no explanation, and `ShareInviteView.swift:167` shows no fallback text. A follow-up to cap the group name (or tell the founder why the code is missing) seems worth filing.
- Not verified here: the 1537-green claim — this worktree can't build the iOS targets.

dev-onym (commented):
**COMMENT** — solid negative-case coverage; one assertion looks broken, one test is tautological, and three tests duplicate `GroupRulesVectorTests` from earlier in the stack.

- `Tests/OnymIOSTests/GroupRulesWireTests.swift:48` — `XCTAssertNil(json?.index(forKey: "rules"))` is a double optional (`Index??`). When `json` decodes, the outer level is `.some`, so `XCTAssertNil` receives a non-nil `Any?` regardless of whether the key is there — the test asserts the opposite of its name (and the repo's own idiom is `XCTAssertFalse(json.keys.contains(...))` at `EnforcementBackendClientTests.swift:121`). Worth confirming against a run; either it's failing or it's vacuous. Fix: `let json = try XCTUnwrap(... as? [String: Any]); XCTAssertFalse(json.keys.contains("rules"))`.
- `Tests/OnymIOSTests/CreateGroupFlowTests.swift:425` — `test_clampedRulesStillMintALink` can't fail for the reason its comment gives. `overshoot` is 1550 ASCII chars, so `clampedRules` (`CreateGroupFlow.swift:65`) truncates on the 500-*character* branch to 500 bytes; the byte cap is never approached. And since the `while clamped.utf8.count > maxBytes` loop guarantees the result fits, `IntroCapability` can never throw for any input. To actually pin the clamp↔cap agreement, use 600 CJK characters and assert `GroupRules.clampedRules(...).utf8.count == GroupRules.maxBytes` — that's the boundary where a one-off between the two caps bites.
- `GroupRulesWireTests.swift:71`, `:91`, `:169` duplicate `GroupRulesVectorTests.swift:156`, `:165`, `:179` — `test_rulesAtTheByteCap_areAccepted` is byte-identical, the over-long-on-the-wire payload is identical, and `test_halfAnAgreement_isRejected` shares even its name and comment. Delete one copy (the vector file's, since these aren't cross-platform vectors) rather than leaving two places to update.
- `GroupRulesWireTests.swift:105` — `qrCapacityAtLevelM = 2331` is a second, disconnected copy of `SettingsQRCode.swift:43`'s `correctionLevel = "M"`. Change the level to "Q" and both QR tests stay green while every full-length link stops rendering — exactly the silent failure the pair exists to prevent. Generating with `CIFilter.qrCodeGenerator()` in the test and asserting `outputImage != nil` / `== nil` pins the property instead of the arithmetic; at minimum, cross-reference the constant so a level change trips.
- `GroupRulesWireTests.swift:120` — the pinned overrun asserts `> 2331`, so it fails the day someone caps the group name or lowers `maxBytes`, i.e. when the limit is *fixed*. That is defensible as a forcing function, but the comment should say the fix is expected to update this test, not just re-measure it.
- `JoinRequestAgreementTests.swift:150` — `test_theRetainedTextIsWhatMakesTheStoredSignatureCheckable` verifies retained text against rules that still match the group's live copy, so it doesn't exercise the claim it's named for. Passing a `rulesText` that differs from a (now-edited) group rules value is the case that shows the proof outlives the edit.
- `CreateGroupFlowTests.swift:421` — the new `// MARK: - Group rules` is inserted *after* `// MARK: - Helpers`, orphaning `makeFlow()` under it. Move the test above the helpers mark.
- Nit: `test_clampedRulesStillMintALink` is `async throws` with no `await` and no throwing call; `GroupRulesTests.swift:4` uses plain `import OnymGroup` while the sibling files use `@testable`. Also the PR body still says `.agreedToOtherRules` where the code and tests say `.unknownRules`.

PR #306, opened 2026-08-22
A member's standing on the rules, and a proof that leaves the device

rinat-enikeev:
The agreement has been recorded since #301 and read by nobody: the founder saw a verdict once, at approval, and the bytes then sat on the member profile. This is the model for showing it to everyone and handing it to someone outside the group. Screens are in the PR above; nothing renders this yet.

**`GroupRulesStanding`** answers "where does this member stand" — the same question `JoinRequestApprover.RulesAgreement` answers for a request that hasn't been accepted, asked again for someone already in. A separate type because the answers differ: a request can't have been written by the founder, and a member can't be "an old client we could ask to upgrade". What's left is `author`, `signed`, `signedEarlierVersion`, `didNotSign`, `doesNotVerify`.

`author` earns its case. Founders don't sign their own terms, and rendering the person who wrote the rules as "didn't sign" would read as a failure rather than as the shape of the thing.

Every case is derived by **re-verifying**, never by reading a flag. The signature is checked against the text stored beside it each time the question is asked, because a stored boolean is a claim about a check somebody once ran.

**`GroupRulesProof`** is the export. What makes it a proof is not this app's say-so: it carries the rules text, the member's Ed25519 key and their signature, so a reader who trusts none of this can recompute the statement and check it with any Ed25519 implementation. The `_readme` lines spell out that procedure *inside the file*, because a proof whose verification steps live in a codebase the reader doesn't have is a screenshot with extra steps.

Three deliberate choices:

- The `rules` it carries are the **signed** ones, not whatever the group says today. Those differ exactly when the wording changed after this member joined, and `matches_current_rules` names the divergence rather than hiding it.
- Signature and key are **omitted entirely** unless the standing is proven. Bytes that don't verify, in a document called a proof, invite a reader to assume someone already checked.
- **No group secret, no roster, no inbox keys.** This is meant to be showable to an outsider — a moderator, a committee — without also handing them the ability to read the group or find the people in it.

Verified rather than asserted: an exported file was checked against an independent RFC 8032 implementation, following only the instructions in its own `_readme`. It verifies, and stops verifying when one character of the rules is altered.

Tests land in the last PR of the stack.

### DIFF
--- a/Packages/OnymGroup/Sources/OnymGroup/ChatGroup.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/ChatGroup.swift
@@ -186,7 +186,7 @@ public struct ChatGroup: Identifiable, Equatable, Sendable {
     /// `groupType` themselves.
     public func isAdmin(blsPublicKey: Data) -> Bool {
         guard let storedAdminHex = adminPubkeyHex?.lowercased() else { return false }
-        let candidate = blsPublicKey.map { String(format: "%02x", $0) }.joined().lowercased()
+        let candidate = blsPublicKey.hexString.lowercased()
         return candidate == storedAdminHex
     }
 }
--- a/Packages/OnymGroup/Sources/OnymGroup/DataHex.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/DataHex.swift
@@ -1,0 +1,10 @@
+import Foundation
+
+extension Data {
+    /// Lowercase hex. One implementation, because there were four
+    /// hand-rolled `%02x` loops in this package and a fifth was about
+    /// to be written.
+    var hexString: String {
+        map { String(format: "%02x", $0) }.joined()
+    }
+}
--- a/Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift
@@ -45,18 +45,26 @@ public struct GroupRulesProof: Equatable, Sendable {
     public let currentRules: String?
     public let standing: GroupRulesStanding
 
-    public init(
-        group: ChatGroup,
-        member: MemberProfile,
-        blsHex: String
-    ) {
-        let standing = group.rulesStanding(of: member, blsHex: blsHex)
-        self.groupIDHex = group.id
+    /// `nil` when no member is stored under `blsHex` — the same reason
+    /// `rulesStanding(ofMemberWith:)` takes the key alone: a proof
+    /// about a profile that isn't in this group's roster is a document
+    /// nobody should be able to produce by accident.
+    public init?(group: ChatGroup, blsHex: String) {
+        guard let member = group.memberProfiles[blsHex],
+              let standing = group.rulesStanding(ofMemberWith: blsHex)
+        else { return nil }
+        // Re-hexed from the bytes signing actually used, rather than
+        // echoing `group.id`. `bytes(fromHex:)` is lenient, and the
+        // `_readme` tells the reader this field is 32 bytes — a
+        // malformed id would otherwise ship instructions that cannot be
+        // followed.
+        self.groupIDHex = group.groupIDData.hexString
         self.groupName = group.name
         self.memberAlias = member.alias
         self.memberBlsHex = blsHex
         self.standing = standing
-        self.currentRules = GroupRules.normalized(group.invitationMessage)
+        let currentRules = GroupRules.normalized(group.invitationMessage)
+        self.currentRules = currentRules
         // Only carried where they mean something. Shipping a signature
         // that doesn't verify, or one there is no text for, would put
         // bytes in a document called a proof that prove nothing.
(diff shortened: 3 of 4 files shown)


dev-onym (commented):
REQUEST_CHANGES — the model is well-shaped, but two claims the document makes about itself aren't true of the bytes it ships.

- `GroupRulesStanding.swift:59` — a stored signature with no text collapses into `.didNotSign`, and the export then states "Joined before this group had rules, or from an app version that predates them" (`GroupRulesProof.swift:207`). That's reachable and false: `MemberProfile.swift:153` keeps `rulesSignature` whenever `rulesHash` is present but drops `rulesText` when the wire omits it or sends blank, so an announced/invited profile can hold 64 bytes with nothing to check them against. This is exactly the "bytes exist and are odd" case the type doc says it keeps distinct (`GroupRulesStanding.swift:33`) — it belongs in `.doesNotVerify` or its own case, not in the reassuring one. Same forgery-picks-its-own-verdict argument as `JoinRequestApprover.RulesAgreement.unknownRules` (JoinRequestApprover.swift:51).
- `GroupRulesProof.swift:186-197` — `_readme` says what the signature proves and never what it doesn't. The statement covers `group_id ‖ SHA256(rules) ‖ sending_public_key`; it does not cover `alias` or `bls_public_key`, yet `member` presents both as the identity (`:167`). A moderator following only these instructions concludes "the member with this BLS key agreed" — a pairing the writer of the file chose, and `MemberProfile` documents `alias` as self-asserted and never load-bearing. One readme line naming that boundary is the difference between a proof and a proof-shaped file.
- `GroupRulesProof.swift:45` — `currentRules` is documented "for the reader to compare" but is never encoded; only the `matches_current_rules` bool reaches the file (`:177`). A reader told the wording diverged has nothing to diff against. Either encode it or drop the promise from the comment.
- `GroupRulesProof.swift:177` — `matchesCurrentRules` re-derives what `standing` already decided (`.signed` ⇒ true, `.signedEarlierVersion` ⇒ false). Two readings of one fact, in a type whose thesis is single-sourced derivation; take it from `standing`.
- `GroupRulesStanding.swift:203`/`:57` — `.noRules` and `.author` both discard a verifiable signature the member may actually hold. `invitationMessage` is `var`, so a founder clearing the rules turns every existing agreement into "nothing was asked of anyone" and drops signature, key and text from the export. Checking the stored bytes before short-circuiting on group state would keep the honest answer.
- `GroupRulesStanding.swift:53` — `rulesStanding(of:blsHex:)` doesn't tie `member` to `blsHex`; the admin's hex with another member's profile returns `.author`. Since `memberProfiles` is keyed by that hex, taking the hex alone and looking the profile up makes the mismatch inexpressible — the argument `GroupRules.isAgreement` already makes for taking text over a hash.
- `GroupRulesProof.swift:78` — `locale: .current` makes the "find it again six months later" filename device-dependent (Turkish dotless-i); pass `nil` or a POSIX locale. Also worth noting `isLetter` passes CJK through, so the diacritic folding doesn't get you an ASCII stem.
- `GroupRulesProof.swift:190` — readme asserts `group.id` is 32 bytes, but `ChatGroup.bytes(fromHex:)` is lenient and nothing here checks for 64 hex chars; a malformed id exports instructions that can't be followed.
- `GroupRulesProof.swift:214` — third hand-rolled `%02x` loop in the package (`ChatGroup.isAdmin` has another); worth one shared `Data.hexString`.

Also: the crypto reads correctly against `GroupRules` — domain is 19 bytes as the readme claims, `rulesText` is pre-normalized so `SHA-256(rules.text)` equals `GroupRules.hash(text)`, and re-verifying via `agreedToRules` rather than trusting `rulesHash` is the right call. No secrets, roster, or inbox keys leak into the export, as claimed. I couldn't compile-check (no Swift toolchain here), and this PR adds no tests by design.

dev-onym (commented):
COMMENT — the crypto and derivation logic are sound; the issues are in the exported document's honesty and one unbounded input.

- `GroupRulesProof.swift:217` — the `_readme`'s first line, "Proof that this member agreed to this group's rules," is emitted unconditionally, including for `didNotSign`, `doesNotVerify`, `noRules` and `author` documents where `member.signed` is `false` and there is no signature. Combined with `suggestedFileName`'s `onym-rules-proof-` prefix (`:110`), the file asserts at the top exactly the thing the PR argues a proof must not assert. Make line 0 conditional on `standing.isProven`, or phrase it as "Record of this member's standing on this group's rules" and let the verify steps carry the proof claim.
- `GroupRulesProof.swift:207` — when the founder *clears* the rules after a member signed, standing is `.signedEarlierVersion`, so `matches_current_rules: false` ships with `current_text` absent (because `currentRules` is nil). The readme at `:227` promises `current_text` carries what the group says now, so a reader can't distinguish "group has no rules today" from "field omitted." Emit an explicit signal for the empty case (e.g. `current_text: ""` or a `current_rules_present: false`), or extend the readme to name it.
- `GroupRulesProof.swift:95` — `memberAlias` is attacker-supplied and uncapped: `MemberAnnouncementPayload.AnnouncedMember.alias` decodes any-length `String` (`MemberAnnouncementPayload.swift:169`), and `groupName` is only capped at compose time (`CreateGroupView.swift:229`) while `ChatGroup.name` stays mutable via rename. A remote member with a 10 KB alias produces a filename that exceeds `NAME_MAX`, so export fails with `ENAMETOOLONG` on a document about that member. Truncate the stem (say 60 chars) after collapsing runs. Path separators and dots are already neutralized by the ASCII-alnum filter — that part is right.
- `GroupRulesProof.swift:82` — for `didNotSign` / `doesNotVerify` the document omits `rules` entirely, so a recipient can't see what the member is alleged not to have agreed to, nor that the group has rules at all. The note text is the only signal. Consider carrying the group's current rules with `signed: false` for these cases too — the argument that unverified *bytes* mislead doesn't extend to the rules *text*, which no one is asked to trust.
- `DataHex.swift:1` — the stated rationale ("four hand-rolled `%02x` loops in this package") isn't delivered: only `ChatGroup.isAdmin` was migrated. `IntroKeyEntry.swift:68`, `IntroInboxPump.swift:65`, `CreateGroupInteractor.swift:869` and `ShareInviteView.swift:125` still hand-roll it. Either convert them or trim the comment to what the change does.
- `GroupRulesStanding.swift:79` — the lookup is a raw dictionary hit on `blsHex` while `:105` lowercases both sides for the admin compare. An uppercase-hex caller gets `nil` from `rulesStanding` and `nil` from `GroupRulesProof.init?` rather than a standing. Existing callers all pass lowercase keys, so this is latent, not live — but given the doc's "make the wrong pairing unsayable" framing, lowercasing at the boundary (and storing the normalized form in `memberBlsHex`) costs one line.

Verified against the code: statement layout, the 19-byte domain length in the readme, and `GroupRules.hash`'s canonicalization all match what the `_readme` instructs a third party to recompute — `rulesText` is stored pre-normalized, so the trim is a no-op for the reader. I could not compile-check (no Swift toolchain in this environment).

PR #307, opened 2026-08-22
Show the rules to the people living under them, and who signed

rinat-enikeev:
Stacked on the model PR. This is the screen work, plus one defect the screens exposed.

The rules were only ever visible to someone deciding whether to join. A member who agreed months ago had no way back to the words they agreed to: the confirmation screen is gone by then, and the text lived in an invitation nobody keeps. The group screen shows them now, under their own name rather than as "INVITATION".

Each member carries their standing beside their name — signed, signed an earlier version, didn't sign, wrote them, or a signature that doesn't check out. The wording tracks `ChatJoinRequestCell`'s deliberately: a member's standing and a request's verdict are the same fact at two moments, and two vocabularies would read as two things.

Tapping a member opens what they signed and the bytes behind it. The screen is arranged around the export, because the interesting question about an agreement is nearly always asked somewhere else — to a moderator, a committee, someone deciding whether a person can be held to something.

**The defect, found by running it rather than reading it.** On Bob's own device, Bob was marked as never having signed the rules he had just agreed to. The invitation ships the directory "as known" and leaves the joiner's own row to the receiver, which backfills it from the active identity — and a backfill can reconstruct an alias and keys but not a signature. The joiner keeps no copy of what they sent, so the admitting device holds the only record. The invitation carries their row now, and the receiver's overwrite becomes a merge: alias and keys stay locally asserted (that overwrite exists so a sender can't mislabel us to ourselves), the agreement comes from the wire. Trusting it costs nothing — it's re-verified against that device's own sending key on every read.

That is the same shape as the propagation gap found in review on #301, in the one direction that review didn't reach.

Smaller decisions worth a look:

- The file is written **on appear, not on tap** — `ShareLink` needs a URL up front, and a share sheet waiting on a file write opens empty.
- It goes to the **temporary directory**. The copy that matters is the one the person sends; leaving proofs about other people lying around this device is a disclosure nobody asked for.
- The mark **replaces** the BLS prefix rather than crowding in beside it. The prefix stays for a group with no rules, which is also the only case where a row is inert — a chevron onto a sheet that would say "nothing to show" is an invitation to a dead end.
- Where the signed wording differs from the group's current rules, the **screen** says so as well as the file.

Twenty-one strings, en and ru. Screenshots of the members screen and the proof sheet are in the thread below.


dev-onym (commented):
**REQUEST_CHANGES** — the defect fix is right and well-argued; two things in the screens need another pass, plus one propagation gap the fix leaves open.

- `Packages/OnymInbox/Sources/OnymInbox/GroupStateVerifier.swift:339` — the admin's refresh reply builds a `GroupInvitationPayload` with **no `invitationMessage`**. A joiner whose invitation deferred with `.staleNeedsRefresh` materializes from *that* payload, so they land in a group whose rules are nil: no "GROUP RULES" section, `rulesStanding` returns `.noRules` for every member, no row is tappable — and their own agreement, which this PR just went to the trouble of shipping in `memberProfiles`, renders as "this group has no rules". Same shape as the gap you fixed, in the third direction. One-line fix: add `invitationMessage: group.invitationMessage`.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:305` — the mark *replaces* the BLS prefix, so in any group with rules no row shows a fingerprint, and the sheet doesn't show `memberBlsHex` either (title is the alias; only truncated sending key + signature). Two members self-asserting the same alias become indistinguishable everywhere in the roster. This contradicts the file's own contract at `ChatMembersView.swift:14-16` ("Always show the BLS-pubkey hex prefix… the fingerprint is the load-bearing identifier") and `MemberProfile.swift:15-18`, neither of which the diff updates. `ChatJoinRequestCell` keeps `inbox <fingerprint>` beside its verdict for exactly this reason. Second line, or fingerprint in the sheet.
- No tests, on a wire-behavior change. `JoinRequestApproverTests.swift` and `IncomingMessageDispatcherTests.swift` both exist and #303 set the precedent. Worth pinning: (a) approve ships the joiner's own row with `rulesHash`/`rulesSignature`/`rulesText`; (b) the merge at `IncomingMessageDispatcher.swift:447` keeps the wire agreement while alias/inbox/sending stay local; (c) Bob's standing is `.signed` post-materialize. `rulesStanding` and `GroupRulesProof` have no coverage at all today.
- `ChatMembersView.swift:430` — `rows(for:)` runs an Ed25519 verify per member inside `body`, so every state change re-verifies the whole roster (including each keystroke in the rename alert). Fine at 5 members, less so at 100; the value is a pure function of the group snapshot and could be computed once per group change.
- `MemberRulesProofView.swift:190` — the temp file is written and never removed. "Leaving proofs about other people lying around this device is a disclosure nobody asked for" is the stated goal, and `tmp` is only swept at the OS's discretion; a `try? FileManager.default.removeItem` on dismiss would actually deliver it.
- `ChatMembersView.swift:79` — `.sheet(item:)` yields an empty sheet if `currentGroup` is nil at present-time; and `row.profile` is a snapshot, so a profile update while the sheet is open leaves the proof stale against the group it's rendered beside.
- `ChatMembersView.swift:310` — `members.standing.<id>` is inside a `Button` label, which SwiftUI merges into one accessibility element; that identifier is probably not queryable from a UI test.
- Nit: `ChatJoinRequestCell.swift:239-247` says "Signed the group rules" / "Didn't sign the group rules" / "Their signature on the rules doesn't check out". The new keys drop "group" and "Their". If the two vocabularies are meant to read as one, share the strings rather than adding four near-duplicates for translators to drift apart.
- Nit: for `.author`, `GroupRulesProof` nils `rules`, so the chevron opens a sheet with no rules text and exports `"rules": null` — thin for the one member whose words they are.

dev-onym (commented):
REQUEST_CHANGES — the wire fixes are sound; one SwiftUI presentation issue the codebase already ruled on, plus smaller cleanups.

- **Two `.sheet(item:)` on one view.** `ChatMembersView.swift:79` and `:93` are chained on the same `Group`. This repo already hit that and documented the rule: `GateCheckRequiredView.swift:16` — *"two `.sheet` modifiers live on the same view here, and SwiftUI has a long history of only honoring one of them"* — and consolidated onto one `sheet(item:)` over an enum. Either the proof sheet or the admin's share-invite sheet can silently stop presenting; the second is the harder bug to notice because only admins reach it. Same fix here: one `enum ActiveSheet: Identifiable { case member(MemberRow); case shareInvite(ShareInviteFlow) }`.
- **Orphan string survived the de-duplication.** `Resources/Localizable.xcstrings:7827` still carries `"Wrote the rules"` (en + ru) with no Swift caller — the code uses `"Wrote the group rules"`. The commit says the four near-duplicates were deleted; this one wasn't, so translators still have two of it. That also makes the count 20 live strings, not 21.
- **"WHAT THEY SIGNED" over rules the member wrote.** `MemberRulesProofView.swift:104`: for `.author`, `proof.rules` is the current rules (`GroupRulesProof.swift:82`), so the heading sits directly under "Founders don't sign their own." The exported file dodges this by routing the author through `note`; the screen contradicts itself. Vary the heading on `proof.standing`.
- **Divergence re-derived by string compare.** `MemberRulesProofView.swift:112` uses `proof.rules != proof.currentRules`, while `GroupRulesProof.swift:198` deliberately single-sources the same fact from the standing ("a type whose thesis is single-sourced derivation should not answer the same question twice"). Same answer today; two answers the moment normalization changes. Use `proof.standing == .signedEarlierVersion`.
- **The self-row merge takes `nil` from the wire too.** `IncomingMessageDispatcher.swift:451-453` copies `wire?.rules*` unconditionally, and this path re-runs on every relay replay and on `insert(group)` for a group already on the device. An invitation or refresh reply from an admin that predates this PR carries no self row, so it overwrites a stored agreement with `nil` and the joiner is back to "didn't sign" — the exact defect this PR fixes, reachable by version skew rather than by code path. Fall back to the locally stored profile's rules when `wire` is nil; `groupRepository.currentGroups()` is already read 30 lines below.
- **The joiner can land on a red mark about themselves.** `JoinRequestApprover.swift:408` pairs `req.rulesHash/rulesSignature` with `anchored.invitationMessage`. That matches `recordJoiner` (`:732`) and is consistent, but a founder who edits the rules between request and approval now ships the joiner a row that resolves to `.doesNotVerify` / `.unknownRules` — "Their signature on the rules doesn't check out", about them, on their first look at the group. Consider only attaching the row when the approval verdict was `.agreed`; the admin's own copy can keep the fuller record.
- Minor: `writeFile()` (`MemberRulesProofView.swift:50`, `:205`) is synchronous encode-and-write on the MainActor via `.task`. Tiny JSON today, but it's on the sheet's first frame — `await Task.detached { ... }.value` costs nothing here.
- No tests in this PR; per the stack convention they land separately. The two wire fixes are the ones worth pinning — joiner materializes with `.signed`, and a refresh reply carries the rules — since both are invisible until someone runs it, which is how this defect was found in the first place.

dev-onym (commented):
**REQUEST_CHANGES** — the screens are well-made and the propagation fix is right, but the mark is wrong for two governance types and the receive-side merge has two holes.

- `ChatMembersView.swift:467` + `GroupRulesStanding.swift:105` — in **anarchy and one-on-one groups with rules, every row reads amber "Didn't sign the group rules"**, including the person who wrote them. `.author` is decided solely by `adminPubkeyHex`, which `CreateGroupInteractor` sets to `nil` for both types, and neither type has a signing flow at all (no join request, no approval). The rules field sits above the group-type picker in `CreateGroupView.swift:179`, so this is a default-reachable path, and `MemberRulesProofView`'s explanation — "joined before this group had rules, or from an app version that predates them" — is then simply false. Either suppress the mark where the group has no signing flow, or give that case its own standing.
- `IncomingMessageDispatcher.swift:455` — the stored-agreement floor looks the group up by `groupIDData` **without `ownerIdentityID`**, while every other lookup in this file pairs both (`:516`, `:301`). On a multi-identity device two rows exist for one group, so the floor can read the other identity's row and stamp its `rulesHash`/`rulesSignature`/`rulesText` onto this identity's self profile.
- `IncomingMessageDispatcher.swift:459` — `wire?.rulesSignature != nil ? wire : stored` makes any wire row with 64 bytes in it beat a stored agreement that *verifies*. Re-approval of an old pending request after a rules edit sends `rulesText: nil` with the signature intact (`JoinRequestApprover.swift:402`), downgrading the joiner's own row from `signed` to `unknownRules`. `MemberProfile.agreedToRules(groupID:)` is right there — prefer the wire only when it checks out.
- `MemberRulesProofView.swift:50,215` — `.task` fires once and `writeFile()` guards on `file == nil`, so when the group updates while the sheet is open the screen re-renders from the live group but **the exported file keeps the old bytes**. That's exactly the case `ChatMembersView.swift:105`'s comment exists to handle; `.task(id: proof)` plus dropping the previous file would close it.
- **No tests.** Four behaviour changes cross device boundaries here (invite carries the joiner's row, the merge floor, `textCovering`, the refresh reply carrying rules), and `JoinRequestApproverTests` / `IncomingMessageDispatcherTests` already have the fixtures for all of them. The PR's own framing is that this defect was found by running it rather than reading it — nothing in the diff stops it coming back.
- `Localizable.xcstrings` — the Russian strings gender the member as male where the English is deliberately neutral: "Составил правила группы", "ПРАВИЛА, КОТОРЫЕ **ОН** УСТАНОВИЛ", "Подписал…". Nominal phrasing ("Автор правил группы", "УСТАНОВЛЕННЫЕ ПРАВИЛА") keeps parity with the English.
- Nits: `mark(for:)` lives on `ChatMembersView` (`:381`) but is now shared vocabulary — it belongs beside `GroupRulesStanding` or in a small presentation type; `MemberRulesProofView.swift:261`'s Done button has no accessibility identifier while `:45`'s does; stray double blank line at `ChatMembersView.swift:55`.

Good calls worth keeping: writing on appear rather than on tap, the temp-directory choice, hoisting `rows(for:)` out of the `ForEach`, and `GroupStateVerifier.swift:341` — the refresh reply omitting the rules was a genuine third gap.

dev-onym (commented):
**REQUEST_CHANGES** — the model/UI work is sound and every invite path now carries the rules, but the fixed defect ships without a regression test and the new copy talks about you in the third person.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/GroupRulesMark.swift:34` / `MemberRulesProofView.swift:87` — the whole vocabulary is third-person, borrowed from `ChatJoinRequestCell` where the subject is always someone else. On the row this PR exists for — the joiner looking at their own agreement, rendered right next to `(you)` (`ChatMembersView.swift:322`) — Bob reads "They read these rules and signed them…", and in the failing case "Their signature on the rules doesn't check out" about himself. The sheet title is his own alias. Needs an `isSelf`/second-person variant at least for the sheet's `explanation` and the `doesNotVerify` mark.
- No tests, for either fix. `Tests/OnymIOSTests/` already has `JoinRequestApproverTests`, `IncomingMessageDispatcherTests` and `JoinRequestAgreementTests` sitting on exactly these seams, and both behaviours are cheap to pin: `JoinRequestApprover.textCovering` (`JoinRequestApprover.swift:1063` — rules edited between request and approval ⇒ nil text ⇒ `unknownRules`, not `doesNotVerify`), and the three-way merge at `IncomingMessageDispatcher.swift:460-489` in its two stated skew cases (older admin sends no self row; re-approval sends signature with no text). The merge's failure mode is "a replay silently erases the agreement" — that is a regression test, not a comment.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:346` — the chevron is suppressed only for `.noRules`, but `GroupRulesMark` returns nil for `.notCollected` too (`GroupRulesMark.swift:34`). Such a row gets a chevron onto a sheet with no icon, no headline, and an "Export proof" button for a group that never collected one — the dead end the `.noRules` rule was written to avoid. Unreachable today (only `.tyranny` is produced), so it's cheap to fix now: gate on `GroupRulesMark(row.standing) == nil` rather than on `.noRules`.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:250` — the quadratic fix is right, but the linear cost stays: `rows(for:)` runs one Ed25519 verify plus a SHA-256 per member on the main actor on *every* body render, and this body re-renders off the group stream. A hundred-member group pays ~100 verifies per repaint. Worth deriving standings once per group snapshot (memoized on the group value) rather than per render.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:52` — `onDisappear` removes the file the share sheet was handed. `ShareLink(item: URL)` passes a reference, not a copy; extensions and AirDrop targets that read it lazily can lose the file if the sheet closes first. Deleting on the *next* write (which the code already does, line 240) and letting the OS sweep `tmp` is the safer half of this.
- i18n: the ru changes de-gender the marks ("Правила группы подписаны"), then the new explanations reintroduce slash-forms — "установил(а)", "прочитал(а)…подписал(а)" — on the same screen. Pick one register. Otherwise en+ru are complete: all 21 new keys resolve, no missing or ru-less entries.
- Nit: `Resources/Localizable.xcstrings` drops "Add a greeting, group rules, or policy…" outright; other dead keys in the file keep `"extractionState": "stale"`. Unrelated to this PR and it discards the ru translation.
- Nit: `IncomingMessageDispatcher.swift:456` — "like every other lookup in this file" isn't quite true; the replay shortcut at line 672 matches on `groupIDData` alone.

Verified as correct: all four `GroupInvitationPayload` construction sites now carry `invitationMessage`; the refresh reply also ships the joiner's own row via `memberProfiles`; the merge is owner-scoped; and a hostile admin can't forge a passing agreement — a bad `rulesText` re-verifies to `doesNotVerify`, and `stored` wins when it's the only record that verifies.

dev-onym (commented):
**REQUEST_CHANGES** — the screens and the propagation fix are right; three real defects and a test gap.

- `MemberRulesProofView.swift:247,257` — on a write failure `file` keeps its **previous** value, so the Export button stays live and shares the *old* proof's bytes. That is precisely the failure `.task(id: proof)` was added to prevent ("Whoever tapped Export would have sent the old ones"). Set `file = nil` before/alongside `writeError`.
- `MemberRulesProofView.swift:51-56,252` — the comment claims replacement-on-write covers "a proof about a different member left behind", but `@State file` resets on every sheet presentation. Open member A, dismiss, open member B → A's file is never the `previous` and survives in `tmp` indefinitely. The delete only covers a same-instance proof change. Either fix the claim or track leftovers outside the view (e.g. a per-group subdirectory swept on first write).
- `IncomingMessageDispatcher.swift:462-492` — the merge decides `wire` proves the agreement by verifying against **`wire.sendingPubkey`**, then persists that signature paired with **`selfEntry.value.sendingPubkey`** (`:490`). If those ever differ, an agreement accepted as proven is stored in a form that re-reads as `.doesNotVerify` — red, about yourself, the same class of bug this PR fixes. Compose the merged profile first and call `agreedToRules` on *that*, or gate the wire branch on key equality.
- No tests, in a repo that already has `JoinRequestApproverTests`, `IncomingMessageDispatcherTests` and `JoinRequestAgreementTests`. Untested and non-obvious: `JoinRequestApprover.textCovering` (:1058 — the hash-mismatch → nil behaviour change), the three-way self-row merge above (all three branches, plus the "older admin sends no self row" skew case the comment cites), and `SEPGroupType.collectsRulesAgreements` / `.notCollected`. The merge is the defect fix; it should be pinned.

Smaller:

- `ChatMembersView.swift:473-474` — "Re-derived on every render … not cached" is now false; the value comes from the `standings` cache. Same at `:272-277`, which still says `rows(for:)` "runs an Ed25519 verify per member".
- `ChatMembersView.swift:108-110,456` — `onChange(initial: true)` fires *after* the first body evaluation, so the roster's first frame renders every row as `.noRules`: no marks, no chevrons, then a re-render. The `?? .noRules` fallback also silently mislabels a member who appears between snapshots. A lazily-populated cache (or deriving inside `rows(for:)` keyed on the group) avoids both.
- `ChatMembersView.swift:353,369,380` — `GroupRulesMark(row.standing)` is constructed three times per row; hoist one `let mark` above `content`.
- `IncomingMessageDispatcher.swift:462,542` — two full `currentGroups()` fetches in one materialize path; the first result answers both questions.
- `Localizable.xcstrings` — `"member"`, `"signature"`, `"signing key"` are collision-prone keys in an app-wide catalog shared across 487 strings. Prefix them (`rules_proof.signature`) or add comments.
- The removal of `"Add a greeting, group rules, or policy…"` is unrelated to this PR (I confirmed no Swift reference remains, so it's harmless) — worth a line in the body so it isn't read as an accident.

dev-onym (commented):
REQUEST_CHANGES — the screens are solid and the propagation fix is right; two concrete defects, both narrow but cheap to fix.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:501` — the "neither proves anything" fallback keeps the wire's `rulesText` alongside signature bytes that just failed to verify against the *locally-asserted* `sendingPubkey`. That resolves to `.doesNotVerify` (`GroupRulesStanding.swift:114`): red, "Signature on the rules doesn't check out", on the joiner's own row — the exact failure this PR set out to remove. Reachable whenever the wire row's signature doesn't check against our key (inviter ships arbitrary bytes for our row; or identity key divergence). `JoinRequestApprover.textCovering` already argues the correct handling for this shape — withhold the text, keep hash+signature, land on `.unknownRules`. Apply the same rule here: in the fallback, carry `rulesHash`/`rulesSignature` but drop `rulesText`.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:88` + `MemberRulesProofView.swift:250` — `suggestedFileName` is group name + alias only, and the comment "it keeps its own name" isn't true. Aliases are explicitly non-unique (`MemberProfile` says so, and `bytes` at line 162 exists because two members can pick the same one), and `folding(.diacriticInsensitive)` doesn't transliterate Cyrillic/CJK, so every char maps to `-`: in a Russian-named group with Russian aliases the stem trims to empty and **every member's proof is written to `onym-rules-proof-group-rules.json`**. Since the file is deliberately not deleted, a share extension reading lazily after the user opens a second member gets *that* member's agreement under the first one's name — a disclosure about a third party. Put `memberBlsHex.prefix(12)` in the stem.
- No tests, in a repo that already has `Tests/OnymIOSTests/JoinRequestAgreementTests.swift` covering exactly this logic. Three behavior changes go uncovered: the three-branch self-row merge precedence (including the "wire has signature, no text → prefer stored" case the comment calls out), `textCovering` returning nil on hash mismatch, and `.notCollected`. The merge is the one worth pinning — its correctness argument is entirely in comments right now.

Verified and fine:
- All 24 new/renamed literals exist in `Localizable.xcstrings` with en+ru; the retired `"Their signature…"` and `"Add a greeting…"` entries have no remaining referents (`CreateGroupView.swift:255` now uses the new prompt).
- `GroupInvitationPayload.init` declares `invitationMessage` after `avatar`, so `GroupStateVerifier.swift:340` is in order — and that reply is admin-gated (`:313`), so shipping the rules there doesn't let a peer rewrite them.
- `rulesStanding` asks `.notCollected` after the stored bytes and after `.noRules`, so an old signature under a changed governance type still reads honestly.

Nits, take or leave: `String(localized: "member" / "signature" / "signing key")` are single-word keys in an app-wide catalog — a future button with the same word inherits a lowercase field-label translation; `MemberRulesProofView.swift:162`. And `StandingsMemo` retains the whole `ChatGroup` snapshot, `avatarJPEG` included, for the view's lifetime.

dev-onym (commented):
**REQUEST_CHANGES** — the fix is well-reasoned and the screens are solid, but the admin-side write of the same fact skipped the merge, and none of the new precedence logic is tested.

- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:740` — `recordJoiner` overwrites `rulesText` unconditionally, including with `nil`. The dispatcher (`IncomingMessageDispatcher.swift:486`) explicitly prefers a stored record that verifies over a wire one that doesn't; this write has no such guard. Concrete loss: admin approves J against R1 → stored `(R1, sig)` verifies as `.signed`; seal/send fails so the request survives (approval is deliberately not gated on `alreadyInRoster`, `:495`); admin edits rules to R2; retry → `textCovering` (`:1059`) returns nil → the profile is rewritten with `rulesText: nil`, R1 is gone from the only device that held it, and the member reads `.unknownRules` forever with an export that carries no wording. Same for the fan-out at `:777`, though `applyAnnouncement`'s known-hex bail (`IncomingMessageDispatcher.swift:936`) spares other members. Keep the stored text when `agreedText` is nil and the stored pair still verifies against the same signature — it does, it's the same request.
- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:462` — the four-branch precedence (wire-verifies / stored-verifies / bytes-without-text / stored) is the actual bug fix and carries the version-skew reasoning, and there is no test for any branch. `Tests/OnymIOSTests/JoinRequestAgreementTests.swift` has 14 tests over exactly this surface; the replay case ("erased a stored agreement and put the joiner back to didn't sign") is precisely what silently regresses. Also untested: `textCovering`'s hash-mismatch rule, `.notCollected`, and `suggestedFileName` uniqueness — the last two both landed as review fixes, which is the strongest argument for pinning them.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesStanding.swift:118` — verified: `CreateGroupView.swift:8` confirms 1-on-1 and Anarchy are dimmed, so `.notCollected` is genuinely unreachable and the doc claim holds. `SEPGroupType`'s exhaustive switch is the right call.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:120` — the sheet body re-derives `GroupRulesProof` (an Ed25519 verify) on every re-render of the presenting view while open; `StandingsMemo` doesn't cover this path. Fine at one member, worth a note next to the memo's rationale since the two now disagree about whether that cost matters.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:176` — the export button is absent, not disabled, until the write lands, so the primary CTA pops in a frame or two after the sheet. A disabled-looking placeholder of the same height would stop the layout jumping under a thumb already moving toward it.
- Non-blocking, verified clean: all 24 new literals resolve in `Localizable.xcstrings`; the removed `"Add a greeting, group rules, or policy…"` key has no remaining Swift reference; `GroupRulesMark`'s colors match `ChatJoinRequestCell` exactly, including the WCAG-corrected amber; the `invitationMessage` omission fix in `GroupStateVerifier.swift:349` covers the only other payload builder that was missing it (`CreateGroupInteractor` already passes it on both paths).

dev-onym (commented):
REQUEST_CHANGES — the screen work is solid and the merge reasoning is right, but a wire-controlled string now lands in a filesystem path, and none of the new logic has tests.

- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:120` — the new `-\(memberBlsHex.prefix(12))` suffix is unsanitized wire input, while the stem two lines above is scrubbed to ASCII alnum. `memberProfiles` keys are arbitrary JSON strings: `GroupInvitationPayload.init(from:)` decodes `[String: MemberProfile]` with no key-shape check (`GroupInvitationPayload.swift:141`), and `MemberProfile.init(from:)` validates values only. A profile keyed `"../evil"` renders a row (`ChatMembersView.swift:450`), gets `.didNotSign` in a group with rules, so gets a chevron, and the tap writes `temporaryDirectory.appendingPathComponent("onym-rules-proof-x-../evil.json")` outside `tmp`. Sanitize the suffix the same way as the stem, or reject non-96-hex keys at decode.
- No tests anywhere in the diff, though the repo has the suites for exactly this: `Tests/OnymIOSTests/JoinRequestAgreementTests.swift`, `GroupRulesWireTests.swift`, and `IncomingMessageDispatcherTests.swift:939` (`test_invitation_materializesLocalGroup_withSelfEntry` — the very path being rewritten). The four-branch precedence at `IncomingMessageDispatcher.swift:485-518` is the defect this PR exists to fix and nothing pins it; `textCovering` / `storedTextStillProving` are `static` internal, which reads as testability nobody used. At minimum: wire-verifies-wins, stored-wins-when-wire-has-signature-but-no-text, older-admin-sends-no-self-row-keeps-stored, and two members' filenames differ.
- `IncomingMessageDispatcher.swift:435` — the merge protects only the self row; `profiles` still starts as the wire snapshot wholesale, so every *other* member's `rulesText`/`rulesSignature` is replaced on each replayed invitation. A pre-rules-build admin re-sending an invitation for a group already local drops the whole roster's agreements to "Didn't sign" — the same version-skew shape the self row just got fixed against, one row over. Worth either applying the same "prefer the record that proves something" rule per key, or saying in the comment why the roster is the admin's to overwrite.
- `MemberRulesProofView.swift:283-296` — `.task(id: proof)` cancellation doesn't reach `Task.detached`, and `await …value` on a non-throwing task doesn't throw on cancel, so a superseded write resumes and assigns `file` (and deletes the URL it captured as `previous`). Task B finishing before task A leaves Export pointing at the old proof's bytes — the failure `.task(id:)` was added to prevent. `guard !Task.isCancelled` before the assignment closes it.
- `ChatEmptyStateView.swift:27` — same `invitationMessage`, still headed `INVITATION`. Renaming it on the members screen but not here gives one body of text two names, in the one PR arguing against two vocabularies for one fact.
- `MemberRulesProofView.swift:162` — `String(localized: "member")` is a bare one-word key in a shared catalog; it will collide with the next unrelated use of "member" and give translators no context. Prefix it or use a comment.

dev-onym (commented):
## REQUEST_CHANGES

**Blocking**

- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:263` — the superseded-write guard deletes **by path**, but `suggestedFileName` is derived only from group name + alias + BLS prefix (`GroupRulesProof.swift:100`), so it does *not* vary when the proof changes. The common way `proof` changes while the sheet is open is a rules edit on the *same* member → identical URL. Failure: rules edited, task A cancelled, task B writes and sets `file = F`; A's detached write completes last, sees `Task.isCancelled`, and `removeItem(at: F)` deletes the file B just published. Export then shares a nonexistent URL (share sheet with a dead item). The interleaving the other way leaves A's stale bytes at `F` before deletion. Fix: write into a per-write unique temp subdirectory (`tmp/<uuid>/<suggestedFileName>` — keeps the readable name and makes `previous != written` meaningful), or skip the cancel-path delete when `written` equals the path the live task targets.

**Non-blocking**

- No tests anywhere for this stack: `grep` finds zero references to `suggestedFileName`, `rulesStanding`, `textCovering`, `storedTextStillProving`, or `collectsRulesAgreements` under `Tests/` or any package test target. The three-way agreement precedence in `IncomingMessageDispatcher.swift:511-547` (wire-proves / stored-proves / bytes-without-text) and the approver's `textCovering ?? storedTextStillProving` fallback (`JoinRequestApprover.swift:411`) are exactly the rules that were got wrong twice in review already, and they're pure functions over values — cheap to pin, easy to regress silently.
- `IncomingMessageDispatcher.swift:452` — the preservation loop assigns `profiles[key] = candidate` even when `wire == nil`, so a stored row absent from the admin's roster is re-inserted on every replay. Harmless today (no member-removal path exists), but it makes the local directory permanently additive; gating on `profiles[key] != nil` would keep the fix to the version-skew case it's written for.
- `IncomingMessageDispatcher.swift:604` — `currentGroups()` is queried a second time with the same predicate already evaluated at line 439; `let existing = storedGroup != nil` does it.
- `MemberRulesProofView.swift:2` — `import UniformTypeIdentifiers` is unused (`ShareLink(item: URL)` needs no `UTType`).
- `MemberRulesProofView.swift:171-173` — `"member"`, `"signature"`, `"signing key"` as xcstrings keys are bare generic words; the next feature needing "member" in a different grammatical role inherits this sheet's lowercase «участник». Worth a disambiguating key or a translator comment.

Verified good: all 21 strings present in en+ru; every `GroupInvitationPayload` construction site now carries `invitationMessage` (approver, verifier, both create paths) — the #301-shaped gap is closed in all directions; `SEPGroupType` switch is exhaustive over the 5 real cases and the `.notCollected` "unreachable today" claim holds (`CreateGroupFlow.governance` is pinned to `.tyranny`); the self-row merge verifies the row it *stores* rather than the one that arrived, and the filename sanitizes the wire-supplied key before it reaches a path.

dev-onym (commented):
**REQUEST_CHANGES** — the screens read well and the self-row fix is right, but the non-self merge drops evidence the self branch protects, and none of it is tested.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:453` — the preservation guard is `stored.agreedToRules(...)`, so a stored row that has signature bytes but no text (`unknownRules`) is skipped and replaced wholesale by a wire row from a pre-rules build that carries nothing. The bytes are gone from the DB permanently, and the member flips from "can't be checked" to "Didn't sign the group rules" — a false claim, and exactly the failure this PR was written to remove, one row over. The self branch at `:527` already encodes the correct rule ("neither proves anything, and the wire has bytes → keep the bytes, drop the wording"); the loop needs the same fallback rather than `continue`.
- No test changes anywhere in the diff. `Tests/OnymIOSTests/` has `IncomingMessageDispatcherTests`, `JoinRequestApproverTests`, `MemberProfileTests`, `GroupRulesWireTests` — and grep finds zero references to `GroupRulesStanding`, `rulesStanding`, or `GroupRulesProof` in the whole suite. Untested here: the four-way self-row precedence at `:517-546`, the roster loop, `textCovering` / `storedTextStillProving` in `JoinRequestApprover.swift:1069,1088`, and the filename collision fix. Each is a two-profile fixture; the replay-erases-agreement case in particular is a regression test the next refactor will need.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:242` — `file` isn't cleared when `proof` changes, only on write failure. While the re-write is in flight the screen already shows the new rules/standing (it re-renders from the live group) but `ShareLink` still holds the previous proof's URL, so a tap in that window exports the old bytes. That's the exact thing `.task(id: proof)` is documented at `:230` to prevent. `file = nil` at the top of `writeFile()` costs the disabled label for one write.
- `MemberRulesProofView.swift:288` — `remove(previous)` deletes a directory the user may have handed to a still-open share sheet, which is the same lazy-reader hazard cited at `:50` as the reason not to clean up on dismiss. Pick one rule. Related: with no dismiss-time cleanup, every sheet open leaves a `rules-proof-<uuid>/` holding another member's rules text and signature in `tmp` indefinitely — iOS purges tmp only under pressure, so the "disclosure nobody asked for" argument at `:236` applies to the accumulation too. Sweeping stale `rules-proof-*` siblings on appear would close it.
- `MemberRulesProofView.swift:92,94` and `GroupRulesProof.swift:272,274` — the `.noRules` and `.notCollected` explanations and notes are unreachable: `GroupRulesMark` returns nil for both (`GroupRulesMark.swift:35`), so those rows get no chevron and no sheet. Four strings, two of them newly translated into ru, that nothing can display. Either drop them or make the intended entry point real.
- `MemberRulesProofView.swift:161-164` — `String(localized: "member")`, `"signature"`, `"signing key"` are bare lowercase single words in a 487-key shared catalog with no `comment:`. Russian needs case and gender that the key alone doesn't carry; give them comments or namespaced keys.

dev-onym (commented):
REQUEST_CHANGES — the model-side merge work is sound; the export sweep can delete the file it just wrote.

- **`MemberRulesProofView.swift:49-50` — the sweep races the write it claims to precede.** The two `.task` modifiers start concurrently and each hops to `Task.detached`, so there is no ordering between `createDirectory` (`:286`) and `contentsOfDirectory` (`:334`). Both directories carry the `rules-proof-` prefix, so if the listing observes the fresh one, the sweep deletes this sheet's own export while `file` is published to `ShareLink` — Export then hands out a dead URL. The doc at `:328` ("Runs once on appear, before this sheet writes its own") asserts an ordering the code doesn't establish. Invisible on a clean `tmp` (nothing to sweep on the first-ever presentation), which is why running it wouldn't surface it. Fix: hold the sweep as a `Task` and `await` it at the top of `writeFile()`, or snapshot the pre-existing directory list once and delete only from that list.
- **`MemberRulesProofView.swift:331` — the sweep isn't scoped to this presentation.** Two windows on iPad (or a proof sheet reopened while a share extension still holds the previous URL) means the new sheet's sweep removes a directory whose share sheet is very much alive — the exact hazard `f1f954e` moved to per-write directories to avoid. Snapshotting the list at first appear fixes both this and the above.
- **No tests, in a PR whose headline defect was found by running the app.** `Tests/OnymIOSTests/GroupRules*Tests.swift` already covers this model layer, and the three new pieces of logic are all pure and cheap to pin: `JoinRequestApprover.textCovering` / `storedTextStillProving` (`JoinRequestApprover.swift:1071,1089`), the invitation-merge precedence table (`IncomingMessageDispatcher.swift:445-520` — wire-proves / stored-proves / neither-but-bytes / neither), and `suggestedFileName` uniqueness for two members of a Cyrillic-named group (`GroupRulesProof.swift:100`). The merge in particular is a version-skew regression that will silently return the moment someone reorders those branches.
- **`MemberRulesProofView.swift:74` — the dead end is still reachable, just not by tap.** The sheet re-derives from the live group by design (`ChatMembersView.swift:124`), so a refresh reply that clears `invitationMessage` turns an open `.didNotSign` sheet into `.noRules`: no mark, `explanation` is `""` (`:113`), no rules text, no bytes — a blank card over a live Export button. The chevron gating can't cover a standing that changes after presentation; a `MemberGoneView`-style fallback for a nil mark would.
- Nit — `ChatMembersView.swift:493`: keying the memo on the whole `ChatGroup` means every body evaluation byte-compares `avatarJPEG`, and the memo retains a second copy of it for the screen's life. Still far cheaper than the verifies it replaces, but `(id, invitationMessage, memberProfiles, adminPubkeyHex, groupType)` would be both cheaper and just as correct-by-construction.

Verified as correct: localization is complete for en/ru across all 22 new keys, the removed `"Add a greeting…"` entry has no remaining callers, `section(_:)` takes `LocalizedStringKey` so the renamed heading still translates, `invitationMessage` is the last defaulted parameter so the `GroupStateVerifier` addition is well-formed, and the member-announcement path (`IncomingMessageDispatcher.swift:988`) bails on an existing row rather than overwriting, so the evidence-loss shape doesn't extend there.

dev-onym (commented):
REQUEST_CHANGES — the merge logic is careful and well-argued, but the group's own rules field is left with the exact gap the member rows were just hardened against, and none of it has tests.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:624` — every member row is now protected against a wire record that proves less than the stored one, but `invitationMessage` is still taken from the wire unconditionally. During rollout the live case is an **old admin** answering a `GroupStateRefreshRequest`: pre-fix `GroupStateVerifier` sends `invitationMessage: nil` (that's the bug fixed at `GroupStateVerifier.swift:346`), the receiver stores nil, and every verified agreement on that device flips `.signed → .signedEarlierVersion` (`GroupRulesStanding.swift`, `signedText == current` with `current == nil`), the rules section vanishes, and the proof sheet asserts "The group's rules have changed since." — with `current_text` absent in the export. Same "prefer the record that has something" rule the rows follow: keep the stored rules when the wire omits them.
- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:439` — no tests. `grep` across `Tests/` finds no reference to `rulesStanding`, `GroupRulesProof`, `suggestedFileName`, `textCovering`, `storedTextStillProving`, or `notCollected`. The ~90-line self/roster decision table here is the whole point of the PR, was found by running the app rather than reading it, and `IncomingMessageDispatcherTests.swift` already exists to hang it on. The four branches worth pinning: wire proves; only stored proves; neither proves and wire has bytes (→ `unknownRules`, not `doesNotVerify`); older admin sends no self row at all.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:97` — the traversal fix (`../../../Documents/x` as a roster key) is correct, and it is the one thing in this diff that a test should outlive a refactor for. `suggestedFileName` is pure and public; a three-line assertion costs nothing.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:59` — the file is written on every sheet open regardless of outcome, including the `nothingToShow` branch where there is no Export button to feed. Tapping through twenty members leaves twenty plaintext JSONs (another member's rules text, sending key, signature) in `tmp`, cleaned only by the next sheet's ≥1h sweep — never on dismiss or background. The reasoning for on-appear-not-on-tap is sound for `ShareLink`; gating `writeFile()` on `GroupRulesMark(proof.standing) != nil` costs nothing and matches the disclosure posture the rest of the screen argues for.
- `Resources/Localizable.xcstrings:3281` — `"INVITATION"` is now unreferenced (only `"GROUP RULES"` is used) but was left in place, while `"Add a greeting, group rules, or policy…"` was deleted in the same diff. Pick one.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatEmptyStateView.swift:26` — still decides emptiness with `.whitespacesAndNewlines` while `ChatMembersView.swift:269` now uses `GroupRules.normalized`. The two sets agree today only because `GroupRules.trimmedCodepoints` was written to match Foundation's; the comment there says the point is that Foundation may revise it. Use `GroupRules.normalized` in both.

Verified as correct and not worth changing: the `notCollected` ordering (stored bytes read before `groupType`, so a governance change can't retroactively erase an agreement), the composed-then-re-verified `candidate` in both merge paths, `Task.isCancelled` guarding a superseded detached write, and the per-write UUID directory. Localization is complete — all 21 new strings have en and ru, and no code literal is missing from the catalog.

dev-onym (commented):
REQUEST_CHANGES — the logic is sound and carefully argued, but the merge-precedence rules that carry the fix are shipped with zero test coverage in an area that already has dedicated suites.

- **No tests for any of it.** `JoinRequestApprover.textCovering` / `storedTextStillProving` (`JoinRequestApprover.swift:1069`, `:1083`) are pure static functions, and the receive-side precedence ladder (`IncomingMessageDispatcher.swift:452`, `:499`) is the exact regression class this PR fixes — Bob marked unsigned on his own device. The repo already has `Tests/OnymIOSTests/JoinRequestAgreementTests.swift`, `IncomingMessageDispatcherTests.swift`, `GroupRulesTests.swift`. At minimum pin: joiner's own row survives materialization; a pre-rules invitation replay doesn't downgrade a stored `signed` row; wire-proven beats stored-proven; a wire self-row whose `sendingPubkey` differs lands as `unknownRules`, not `doesNotVerify`; re-approval after a rules edit keeps R1 (`storedTextStillProving`).
- **`suggestedFileName` sanitization is untested** (`GroupRulesProof.swift:100-136`). The comment says a `../../../Documents/x` roster key reached the filesystem and that three levels was verified by hand — a path-traversal fix and a two-members-one-filename collision fix both deserve a test each, since nothing else in the codebase constrains roster-key shape on decode.
- **Exports can outlive the install.** The sweep only runs when *another* proof sheet appears (`MemberRulesProofView.swift:380`), so a user who opens exactly one sheet leaves another member's rules text + signature in plaintext in `tmp` indefinitely. Given the screen's own reasoning about disclosure, also sweep on app launch/scene activation.
- **A presentation type is gating a file write.** `GroupRulesMark(proof.standing) != nil` is the predicate at `MemberRulesProofView.swift:34`, `:290` and `ChatMembersView.swift:373`. Three call sites, one of which decides whether plaintext hits disk — this wants to be `GroupRulesStanding.hasSomethingToShow` in OnymGroup, with `GroupRulesMark` consuming it.
- Nit: `invitation.invitationMessage ?? storedGroup?.invitationMessage` (`IncomingMessageDispatcher.swift:642`) fires only on `nil`. A sender that ships `""` for "no rules" bypasses the fallback and reintroduces the `signed` → `signedEarlierVersion` flip the fallback exists to stop. `GroupRules.normalized(...)` there would make the two spellings agree, as everywhere else in this PR.
- Nit: `exportLabel(enabled: false)` (`:229`) isn't a control — a tap during the write is silently inert, with no disabled trait for VoiceOver. A `ProgressView` overlay or `Button(action: {}).disabled(true)` would say what's happening.
- Nit: `"member"`, `"signature"`, `"signing key"` are context-free keys with no `comment:`; they'll be hard to translate correctly and will collide the moment another screen needs a different rendering of the same word.
- Nit: `standings[key] ?? group.rulesStanding(ofMemberWith: key) ?? .noRules` (`ChatMembersView.swift:457`) — having `StandingsMemo` return a non-optional keyed on the same snapshot would remove a fallback whose failure mode is a silently unmarked row.

dev-onym (commented):
**REQUEST_CHANGES** — the screen work is solid and the defect diagnosis is right, but the headline fix ships with no tests and the new merge double-processes the self row.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:452` — the roster-wide loop iterates `selfEntry.key` too, before the self block reads `storedGroup`. Concrete path: stored self row has `{sig S, text T}` that fails against our sending key (`doesNotVerify`), wire self row comes from a pre-rules admin (`rulesSignature == nil`). The loop rewrites `profiles[selfKey]` to `{S, text: nil}`; the self block then reads that as `wire`, takes the `wire.rulesSignature != nil` branch, and persists `unknownRules` — `T` is gone from disk forever. Without the loop it would have taken `agreement = stored` and kept both. `agreement = stored` and the `stored?.agreedToRules` branch are now partly unreachable. Skipping `selfEntry.key` in the loop (`guard key != selfKey`) makes the two blocks independent and restores the documented "prefer the record that proves something".
- **No tests, anywhere in this diff.** `Tests/OnymIOSTests/IncomingMessageDispatcherTests.swift:939` already covers `test_invitation_materializesLocalGroup_withSelfEntry`, and `JoinRequestAgreementTests.swift` covers the standings — the exact two files this PR changes behaviour in. Untested: the joiner's own row surviving a replay (the defect this PR exists to fix), each of the five merge branches, `textCovering`/`storedTextStillProving` after a rules edit, `GroupStateVerifier`'s new `invitationMessage`, and `invitationMessage` falling back to stored. A regression here is silent — it reads as "didn't sign", not as a crash.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:100` — the path-traversal scrub (`../../../Documents/x` in a roster key) is security-relevant and has no test pinning it. `suggestedFileName` has no coverage at all; a future edit that appends a component raw reintroduces it exactly as the doc says the key did.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:236` — `String(localized: "member")`, `"signature"`, `"signing key"` are single generic lowercase words used as catalog keys with no `comment:`. Any later `String(localized: "signature")` elsewhere in the app silently inherits "подпись", and translators get no context. Add `comment:` at minimum.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesStanding.swift:26` — "Not reachable today" is about the picker, not the type. `CreateGroupInteractor.swift:451` builds `.oneOnOne` groups with `invitationMessage` passed through, and the dispatcher materializes any `groupTypeRaw` off the wire. Worth rewording so nobody deletes the case as dead.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:384` — the sweep only runs from a members screen or a proof sheet. Export once, never return: another member's rules text and signature sit in `tmp` in plaintext until OS pressure, not one hour. A sweep on launch or foreground would close that without touching the "never delete while alive" rule.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:469` — sort has no tie-break, so members sharing an alias (explicitly non-unique) can swap positions when the memo recomputes. Rows are tap targets now; add `blsHex` as the final comparator.

Localization checks out: all 21 new keys are present with both `en` and `ru`, no orphaned removals (`INVITATION` and the old placeholder have no remaining Swift references), and no `en`-only entries in the catalog.

dev-onym (commented):
**REQUEST_CHANGES** — one real evidence-loss bug in the peer merge, and the stack's first PR with no tests.

- **`IncomingMessageDispatcher.swift:498-506` — the loop destroys stored `rulesText` for peers, which is the exact defect 7129450 fixed for the self row.** When a peer's stored row has a signature + text that doesn't verify (`doesNotVerify`) and the wire carries no signature (pre-rules admin build), this branch rewrites the row with `rulesText: nil`. The only copy of the signed wording is gone, permanently, and `doesNotVerify` silently becomes `unknownRules` — a truthful "this doesn't check out" downgraded to "can't be checked". Dropping the text is only justified when the wire announces a *different* `sendingPubkey` (which is why the self row does it — local keys are substituted). Suggest `rulesText: stored.sendingPubkey == wire.sendingPubkey ? stored.rulesText : nil`. Same branch also silently discards a *verified* stored agreement when `wire.rulesSignature != nil` and the recomposed candidate fails (line 498's `continue`) — defensible under key rotation, but undocumented.
- **No tests.** #300–#303 each shipped them; this one changes the invitation wire shape (joiner's own row), adds a merge precedence ladder with six outcomes, a new `GroupRulesStanding` case, filename sanitization defending a stated path-traversal reach, and an age-based tmp sweep. `JoinRequestApproverTests` / `IncomingMessageDispatcherTests` exist and are the natural home. At minimum: `textCovering` returning nil on a rules edit between request and approval, `storedTextStillProving` on double-approval, self-row survival across a replay from a pre-rules admin, and the peer case above.
- `IncomingMessageDispatcher.swift:654` — the `?? storedGroup?.invitationMessage` fallback means rules can never be cleared by omission. The doc admits it; fine for now, but a clearing path will need an explicit "cleared" signal on the wire, not silence. Worth an issue rather than a blocker.
- `GroupRulesMark.swift:44-77` vs `ChatJoinRequestCell.swift:238-247` — the vocabulary is shared via identical string *keys*, so translators see one entry, but there are now two literal sites. `ChatJoinRequestCell` could map its `RulesAgreement` to a `GroupRulesStanding` and read the text from `GroupRulesMark`, which is what the type's own doc argues for.
- `MemberRulesProofView.swift:205-208` — `String(localized: "member")` / `"signature"` / `"signing key"` are bare generic words as catalog keys. Collision-prone and unclear to translate; give them a `comment:` or a namespaced key.
- `GroupRulesProof.swift:110` — `fileSafe(memberBlsHex.prefix(12))` is 48 bits of a *hex* key, but for a non-hex roster key the scrub can collapse it to something short or to `"member"`, reintroducing the collision the prefix exists to prevent. Scrubbing the prefix of an already-scrubbed key, or hashing when the scrub shortens it, would close that.

Verified as fine: `OnymChatsUI`/`OnymGroup` deps, `GroupInvitationPayload.init` arg order at `GroupStateVerifier.swift:349`, `SEPGroupType`'s five cases, no remaining references to the deleted `"INVITATION"` / `"Add a greeting…"` strings or `members.invitation`. No Swift toolchain here, so this is read-only — not compiled.

dev-onym (commented):
**REQUEST_CHANGES** — the merge logic and the screens read well; two things to fix and a coverage gap.

- **`suggestedFileName` bounds the character set but not the length** — `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:101`. `groupName` and `memberAlias` both arrive off the wire unbounded (`JoinRequestPayload.swift:171` decodes `joinerDisplayLabel` with no cap; `GroupInvitationPayload.name` likewise), so a peer with a 400-char alias produces a name past `NAME_MAX` (255). `write(to:)` throws, `file` stays nil, and that member's row is permanently stuck on the disabled Export button + "This device couldn't write the proof file." The function already treats these strings as untrusted for path characters — clamp `named` (e.g. `String(named.prefix(60))`) before appending the key and suffix.
- **The rules section is unreachable on exactly the screen the PR is about, for a group with no profiles** — `ChatMembersView.swift:74` routes to `emptyState` before `list(for:)`, and `list(for:)` is the only caller of `rulesSection` (`:275`). Hoist the rules above the empty/list branch so "a member who agreed months ago" isn't blocked by an unfilled directory.
- **No tests for any of it.** The self/peer precedence merge (`IncomingMessageDispatcher.swift:461-609`) is five interacting branches that the commit log shows being rewritten six times in review; `textCovering`/`storedTextStillProving` (`JoinRequestApprover.swift:1069-1096`), `.notCollected`, and the filename scrubbing are all uncovered, and `GroupRulesProof`/`GroupRulesStanding` have none from the base PR either. `Tests/OnymIOSTests/IncomingMessageDispatcherTests.swift` already has the fixtures. At minimum: wire-proves / stored-proves / neither-proves for the self row, and the peer row's key-mismatch text drop.
- **Roster wipe survives the skew guard it sits next to** — `IncomingMessageDispatcher.swift:466`: `guard let wire = profiles[key]` means an invitation or refresh reply with `memberProfiles == nil` still drops every stored peer row and its agreements. Pre-existing, but it's the same version-skew shape the new comment at `:452-460` claims to have closed, and `invitationMessage` right below got the fallback. Worth either the guard or an explicit note that only *rules-less*, not *roster-less*, senders are handled.
- Nit: `GroupRulesMark.swift:32` makes `:34` dead — the `guard` already excludes `.noRules`/`.notCollected`. Keep one gate.
- Nit: `ChatMembersView.swift:397` puts `members.row.<id>` on the `Group` wrapper rather than the `Button` that is now the tappable element; the identifier may not land on the a11y element that UI tests would drive.
- Nit: `"member"`, `"signature"`, `"signing key"` (`MemberRulesProofView.swift:205-208`) are very generic catalog keys for a shared `Localizable.xcstrings` — a future unrelated "signature" string will collide silently.

dev-onym (commented):
**REQUEST_CHANGES** — the invitation-carries-own-row fix and the `GroupStateVerifier` rules omission are both right, but the merge table is the riskiest code in the stack and ships untested, and one branch destroys evidence irreversibly.

- **No tests, anywhere in 18 commits.** The six-way precedence table at `IncomingMessageDispatcher.swift:461-619`, `JoinRequestApprover.textCovering`/`storedTextStillProving` (`JoinRequestApprover.swift:1069-1093`), `.notCollected`, and the filename scrubbing all land beside existing `IncomingMessageDispatcherTests`, `JoinRequestApproverTests`, `MemberProfileTests`, `GroupRulesTests` — and none of them gained a case. Every branch here is a claim about which record survives a version skew; those are exactly the assertions a reader can't check by eye.
- **A peer's verified wording can be destroyed permanently.** `IncomingMessageDispatcher.swift:513-534`: when `candidate` fails to verify because the wire announces a different `sendingPubkey`, and the wire row carries no signature, the stored hash+signature are kept but `rulesText` is nilled. A member who was `.signed` becomes `.unknownRules` forever — this device held the only copy of those words, and the export now carries none. Keeping `stored.rulesText` yields `.doesNotVerify` (also wrong-looking, but the bytes survive and can still be exported and re-checked). The comment argues the *classification*; it doesn't address that one option is lossy and the other isn't.
- **The owner-scoping fix stopped one path short.** `IncomingMessageDispatcher.swift:439` now scopes the stored-group lookup by `ownerIdentityID`, with a comment about two identities on one device stamping each other's agreements. `applyAnnouncement` (`:1034`), `applyAvatar` (`:1114`) and `applyName` (`:1141`) still take `groups.first(where: { $0.groupIDData == … })`. Pre-existing, but the reasoning now sits 600 lines above them, and an announcement carrying `rulesHash`/`rulesSignature`/`rulesText` (`:1067-1078`) writes agreements into whichever identity's row sorts first.
- **The export gets no data-protection class.** `MemberRulesProofView.swift:332` writes with `.atomic` only. This is another member's rules text and signature in plaintext in `tmp`, deliberately left for up to an hour. `NSFileProtectionCompleteUnlessOpen` would match how carefully the rest of the screen treats the same bytes.
- **Bare single-noun localization keys.** `MemberRulesProofView.swift:205-208` uses `"member"`, `"signing key"`, `"signature"` as catalog keys with no translator comment. In a 485-string shared catalog, `"signature"` is the drift hazard this PR argues against everywhere else — it reads as a UI label here and could be reused as a verb or a heading elsewhere.

Nits:
- `GroupRulesProof.swift:108` — `prefix(60)` can cut mid-run, so the name becomes `…group-name--a1b2c3.json`; trim dashes after clamping.
- `GroupRulesStanding.swift:108-116` — the key drops `avatarJPEG` (≤16 KB) but still retains and deep-compares `memberProfiles`, including every member's `rulesText` (up to 1500 bytes each). That's the bulk of `ChatGroup`, so the comment's "retained a second copy of it" reasoning undersells what's left.
- `MemberRulesProofView.swift:358` — `writeError` is never cleared on a later successful write, so a stale alert can surface over the `nothingToShow` branch.
- `GroupRulesProof.swift:129` — `static func hex(_ data: Data) -> String { Data(data).hexString }` re-wraps a `Data`; `data.hexString` reads the same.

Verified as correct: all four `GroupInvitationPayload` construction sites now carry `invitationMessage`; the removed `"INVITATION"` / `"Add a greeting…"` keys have no remaining call sites; `applyAnnouncement` is insert-only so the merge gap doesn't reach it; the `rows(for:)` sort is now total; `rulesStandingInputs` covers every field `standing(of:)` reads.

dev-onym (commented):
**REQUEST_CHANGES** — the merge logic and the joiner-row fix look right; one real layout regression and a testing gap.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift:80` — moving the rules out of `list(for:)`'s `ScrollView` into the outer `VStack` makes them unscrollable. `rulesSection` (`:322`) sets no `lineLimit`, and `GroupRules.maxBytes` explicitly permits ~1500 bytes — roughly 40 lines at 15pt, taller than the screen. A group with long rules gets them clipped with no way to scroll *and* squeezes `list(for: group)` (`:85`) to near-zero height, so neither the words nor the roster is reachable. That's worse than the old behaviour for exactly the member this screen was written for. Either keep the section inside the scroll view (the "directory hasn't filled in" case can be handled by putting the section above the roster *within* the `ScrollView`, or by scrolling the empty state too), or cap it with `lineLimit` + a "read all" affordance.
- No tests anywhere in the diff. `Tests/OnymIOSTests/` has `JoinRequestApproverTests`, `IncomingMessageDispatcherTests`, `GroupRulesTests` — and the PR's headline defect ("found by running it rather than reading it") is precisely the kind that a regression test pins. Nothing currently covers: the self-row precedence ladder (`IncomingMessageDispatcher.swift:530-600`), the peer-row evidence merge (`:476`), `textCovering` / `storedTextStillProving`, the `invitationMessage` omission fallback, or `.notCollected`. These are five new invariants about which record wins, each justified in prose only; the next edit to that ladder has nothing to fail against. `GroupRulesStanding`/`GroupRulesProof` also arrive from #306 untested.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:124` — the key suffix's stated rationale ("since the export is deliberately left in place… a share extension can pick up the file after a second member's proof has overwritten it") no longer holds: every write now lands in its own `UUID` directory, so two members can't contend for a path. The suffix is still worth keeping for a Files-app copy, but the comment argues for a hazard the directory-per-write already closed — and a future reader will trust it.

Smaller notes, non-blocking:

- `Localizable.xcstrings` — `"member"`, `"signature"`, `"signing key"` are bare single words as catalog keys. They read as labels here, but a different sense elsewhere in the app will silently reuse the same translation. Worth a comment or a qualified key.
- `ChatMembersView.swift:80` shows a "GROUP RULES" section for any group type, including `.oneOnOne`, where `collectsRulesAgreements` is false and nobody can sign — the heading claims more than the group does. Gating the section on `collectsRulesAgreements` would match the row-level gating already in place.
- Verified as behaviour-preserving: `ChatJoinRequestCell.agreementText`'s colours and case mapping through `GroupRulesMark` are identical to the old switch, and the retired `"Their signature…"` / `"INVITATION"` / `"Add a greeting…"` keys have no remaining code references.

dev-onym (commented):
**REQUEST_CHANGES** — reasoning and defect fix are sound; the blocker is that none of it is covered by a test.

- **No tests anywhere for this logic.** `grep` over `Tests/` returns zero hits for `GroupRulesProof`, `GroupRulesStanding`, `suggestedFileName`, or `notCollected`, and this PR adds none — despite `JoinRequestApproverTests`, `IncomingMessageDispatcherTests`, `MemberProfileTests` and `GroupRules*Tests` all existing. The riskiest parts are pure and directly testable:
  - `IncomingMessageDispatcher.swift:461` / `:536` — the whole evidence-precedence table (wire-proves / stored-proves / neither-but-wire-has-bytes / keep-text-iff-same-sending-key, and the self-row `key != selfEntry?.key` skip). Every branch encodes a bug that was already hit once; nothing pins them against the next edit.
  - `JoinRequestApprover.swift:411` — `textCovering ?? storedTextStillProving`, i.e. the R1-then-edit-to-R2 double-approval case the comment says is reachable.
  - `IncomingMessageDispatcher.swift:682` — `normalized(wire) ?? storedGroup?.invitationMessage`, the flip from `signed` to `signedEarlierVersion` on an old admin's refresh reply.
  - The bug in the PR title itself ("Bob marked as never having signed") deserves the regression test that would have caught it: approve → materialize invite → assert self row is `.signed`.
- **`GroupRulesProof.swift:103`** — `suggestedFileName`'s traversal/length hardening (the `../../../Documents/x` key, the 400-char alias, the sub-12-char scrub → SHA fallback) is exactly the kind of string mangling that regresses silently. It's a pure `String → String`; three assertions would cover it.
- **`ChatEmptyStateView.swift:36` and `ChatMembersView.swift:333`** — "GROUP RULES" is applied to `invitationMessage` for *every* group type, but `SEPGroupType.collectsRulesAgreements` (`GroupRulesStanding.swift:186`) says only `.tyranny` treats that field as rules. In a `.oneOnOne` chat the field is a greeting and the thing isn't a group, so the empty thread of a 1:1 now reads "GROUP RULES" over a hello. Worth gating the heading on `collectsRulesAgreements`, since the type already answers the question.
- **`MemberRulesProofView.swift:205`** — `String(localized: "member")`, `"signature"`, `"signing key"` are bare single words in a shared app-wide catalog. No context for a translator and a live collision risk with any future use of the same word; a `comment:` or a namespaced key would cost nothing now and is painful to unpick later.
- **`MemberRulesProofView.swift:397`/`:418`** — the only sweep triggers are a proof sheet opening and a member list appearing, with a 1h floor. Twenty sheets in one session leaves twenty plaintext rules-and-signature directories that survive backgrounding and app exit. The reasoning for not deleting on dismiss is right; a sweep on `scenePhase == .background` would close most of the window without touching a URL anyone might still be reading.

Nothing in the merge precedence or the approver fallback looked wrong on reading — the objection is that "looked wrong on reading" is currently the only check either one has.

dev-onym (commented):
**COMMENT** — the headline fix traces correctly end to end; two symmetry gaps in the precedence logic, and the merge matrix has no test coverage anywhere in the repo.

- `Packages/OnymInbox/Sources/OnymInbox/IncomingMessageDispatcher.swift:577` — the `stored` branch breaks the invariant the `wire` branch at :563 states ("verified as the row that will actually be *stored*, not as the row that arrived"). `stored?.agreedToRules` checks the signature against `stored.sendingPubkey`, but :611 persists that agreement beside `selfEntry.value.sendingPubkey`. If those ever differ, a proven agreement is accepted and re-read as `doesNotVerify` — red, on your own row, the exact failure this PR exists to remove. Not reachable today (a rotated sending key comes with a new BLS key, i.e. a new dict key), but the guard is one `.map` away and the comment already argues for it.
- `Packages/OnymGroup/Sources/OnymGroup/JoinRequestApprover.swift:1083` — same shape, one layer up: `storedTextStillProving` verifies against `existing.sendingPubkey`, then `recordJoiner` stores the returned text beside `req.joinerSendingPublicKey`. A joiner who replays the old signature bytes under a new sending key passes the `rulesSignature ==` guard and lands a row that reads `doesNotVerify` on the founder's device. Compose-then-verify here too.
- No tests in the diff. `grep` finds no test in `Tests/` or `Packages/*/Tests` referencing `rulesStanding`, `GroupRulesProof`, `textCovering`, `storedTextStillProving`, `hasSomethingToShow` or `notCollected` — so the seven-branch self/peer merge, the double-approval fallback and the filename hardening are all unexercised. #303 sets the precedent that tests land in a follow-up PR, so this isn't a blocker, but the follow-up should pin at least: Bob's own row after admission (the reported defect), a pre-rules invitation replaying over a stored agreement, R1→R2 re-approval, and `suggestedFileName` for a `../../../Documents/x` roster key.
- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:397` — the sweep only runs from the members list and the proof sheet, so a user who exports once and never reopens either screen leaves another member's rules text and signature in plaintext in `tmp` until the OS feels pressure. A sweep on app launch or foreground would close that without touching the 1h rule.
- `Packages/OnymGroup/Sources/OnymGroup/GroupInvitationPayload.swift:146` and `MemberProfile.swift:95` — neither decode applies `GroupRules.fits`, while `ChatGroup.linkableRules` (`ChatGroup.swift:149`) does. This PR widens what that costs: unbounded wire text is now rendered on the members screen and written to a temp file per member. Worth a cap at the decode boundary.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesStanding.swift:80` — for a `.notCollected` group the screen still draws a "GROUP RULES" section while no row carries any standing. Fine for tyranny-only today, but if `.oneOnOne` ships, "GROUP RULES" over a two-person dialog's greeting will read wrong; the heading may want to follow `collectsRulesAgreements`.

Verified and fine: all 21 new keys are present in `Localizable.xcstrings` with both `en` and `ru`, and the two removed keys have no remaining referents; `ChatJoinRequestCell.agreementText`'s remap through `GroupRulesMark` preserves every string and colour; the joiner's own row survives the receiver's merge (`wire` recomposed with local keys still verifies, since the joiner signed with that key). No Swift toolchain in this environment, so this is a static review — the layout claims in `ChatMembersView` aren't machine-checked.

PR #308, opened 2026-08-22
Tests for the standing, the proof, and what both devices can see

rinat-enikeev:
Stacked on the screens PR.

**`GroupRulesStandingTests`** covers the standings where a mark would overstate: bytes that don't verify, a signature lifted from another member, and a signature that is good over words the group has since replaced — which stays *proven*, because it is, and says so separately. Plus the founder matched case-insensitively, since `adminPubkeyHex` and the profile key are hex from two sources and a case difference must not demote the author to "didn't sign".

For the export, the load-bearing test **rebuilds the statement from the document's own fields and verifies it**. That is the file's whole claim; asserting anything less would only test the writer against itself. Alongside it: the wording carried is the signed one rather than today's, an unproven standing ships no signature at all, the `_readme` only cites keys the document actually has, two exports of one agreement are byte-identical, and names that reduce to nothing still produce a findable filename.

**`GroupRulesProofUITests`** is a UI test because the claim is cross-device: an agreement is checkable by *any* member, not only the founder who admitted them, and that holds only if the signature survives the invitation payload. Nothing below the UI crosses two devices. It asserts the gate on the way in — Send closed until the rules are agreed to, open after — then the same two standings from Alice's device and from Bob's.

That second half is what failed before the fix in the PR below, which is why it's worth its ~2 minutes of runtime.

1579 unit tests green; the UI test green end to end.

### DIFF
--- a/Tests/OnymIOSTests/GroupRulesStandingTests.swift
+++ b/Tests/OnymIOSTests/GroupRulesStandingTests.swift
@@ -1,0 +1,527 @@
+import XCTest
+import CryptoKit
+@testable import OnymIOS
+@testable import OnymGroup
+import OnymIdentity
+
+/// Where a member stands, and what an export of that standing may say.
+///
+/// The cases that matter are the ones where a mark would overstate:
+/// bytes that don't verify, and bytes that verify over words the group
+/// has since replaced. A screen that called either of those "signed"
+/// would be making a claim this device cannot support.
+final class GroupRulesStandingTests: XCTestCase {
+
+    private let groupID = Data(repeating: 0x1a, count: 32)
+    private let rules = "Be kind. No links."
+    private let adminHex = String(repeating: "ff", count: 48)
+
+    // MARK: - Standing
+
+    func test_aGroupWithNoRules_hasNoStandingToReport() {
+        XCTAssertEqual(standing(of: unsigned(), rules: nil), .noRules)
+    }
+
+    func test_blankRules_areNoRules() {
+        // Whitespace is not a thing anyone can be held to.
+        XCTAssertEqual(standing(of: unsigned(), rules: "   \n "), .noRules)
+    }
+
+    func test_theFounder_wroteThemRatherThanFailingToSign() {
+        // Rendering the author as "didn't sign" would read as a failure
+        // rather than as the shape of the thing.
+        XCTAssertEqual(
+            standing(of: unsigned(alias: "Alice"), key: adminHex, rules: rules),
+            .author
+        )
+    }
+
+    func test_theFounderIsMatchedCaseInsensitively() {
+        // `adminPubkeyHex` and the `memberProfiles` key are both hex
+        // strings from different sources; a case difference must not
+        // demote the founder to "didn't sign".
+        XCTAssertEqual(
+            standing(
+                of: unsigned(alias: "Alice"),
+                key: adminHex,
+                rules: rules,
+                adminHex: adminHex.uppercased()
+            ),
+            .author
+        )
+    }
+
+    func test_aVerifiedSignatureOverTheCurrentRules_isSigned() {
+        XCTAssertEqual(standing(of: signer().profile, rules: rules), .signed)
+    }
+
+    func test_aMemberWithNothingStored_didNotSign() {
+        XCTAssertEqual(standing(of: unsigned(), key: "dd", rules: rules), .didNotSign)
+    }
+
+    func test_rulesChangedSinceTheySigned_saysSoRatherThanFailing() {
+        // The signature is still good; it covers different words. That
+        // is a fact about the group's history, not about the member.
+        let verdict = standing(of: signer().profile, rules: "Be kind. No links. And no photos.")
+        XCTAssertEqual(verdict, .signedEarlierVersion)
+        XCTAssertEqual(verdict?.isProven, true, "an earlier version still checks out")
+    }
+
+    func test_bytesThatDoNotVerify_areNotAMissingSignature() {
+        // Distinct from `didNotSign`: only one of the two is odd.
+        let key = Curve25519.Signing.PrivateKey()
+        let profile = MemberProfile(
+            alias: "Mallory",
+            inboxPublicKey: Data(repeating: 0xAA, count: 32),
+            sendingPubkey: key.publicKey.rawRepresentation,
+            rulesHash: GroupRules.hash(rules),
+            rulesSignature: Data(repeating: 0x00, count: 64),
+            rulesText: rules
+        )
+        let verdict = standing(of: profile, key: "mm", rules: rules)
+        XCTAssertEqual(verdict, .doesNotVerify)
+        XCTAssertEqual(verdict?.isProven, false)
+    }
+
+    func test_aSignatureLiftedFromAnotherMember_doesNotVerifyHere() {
+        // The signer is named inside the signed bytes, so re-attributing
+        // one member's agreement to another fails.
+        let real = signer()
+        let impostor = MemberProfile(
+            alias: "Mallory",
+            inboxPublicKey: Data(repeating: 0xAA, count: 32),
+            sendingPubkey: Curve25519.Signing.PrivateKey().publicKey.rawRepresentation,
+            rulesHash: GroupRules.hash(rules),
+            rulesSignature: real.profile.rulesSignature,
+            rulesText: rules
+        )
+        XCTAssertEqual(standing(of: impostor, key: "mm", rules: rules), .doesNotVerify)
+    }
+
+    // MARK: - Export
+
+    func test_theExportCarriesWhatAnOutsiderNeedsToCheckIt() throws {
+        let signed = signer()
+        let group = makeGroup(rules: rules, members: ["cc": signed.profile])
+        let proof = try XCTUnwrap(GroupRulesProof(group: group, blsHex: "cc"))
+        let json = try object(from: proof)
+
+        XCTAssertEqual((json["group"] as? [String: Any])?["id"] as? String, group.id)
+        let member = try XCTUnwrap(json["member"] as? [String: Any])
+        XCTAssertEqual(member["signed"] as? Bool, true)
+        XCTAssertEqual(
+            member["sending_public_key"] as? String,
+            hex(signed.key.publicKey.rawRepresentation)
+        )
+        XCTAssertEqual(member["signature"] as? String, hex(signed.profile.rulesSignature!))
+        let carried = try XCTUnwrap(json["rules"] as? [String: Any])
+        XCTAssertEqual(carried["text"] as? String, rules)
+        XCTAssertEqual(carried["sha256"] as? String, hex(GroupRules.hash(rules)))
+        XCTAssertEqual(carried["matches_current_rules"] as? Bool, true)
+    }
+
+    func test_theExportedBytesActuallyVerify() throws {
+        try assertExportVerifiesItself(groupRules: rules)
+    }
+
+    func test_theExportVerifiesEvenAfterTheGroupChangedItsRules() throws {
+        // The case the export exists to get right, and the one the
+        // string comparison below can't prove: shipping today's wording
+        // beside an old signature would produce a document that fails
+        // its own instructions. Checking it cryptographically is the
+        // only way to know it doesn't.
+        try assertExportVerifiesItself(groupRules: "Be kind. No links. And no photos.")
+    }
+
+    /// Verifies a proof the way its `_readme` tells a stranger to:
+    /// rebuild the statement out of the document's own fields, and
+    /// check the signature it carries.
+    private func assertExportVerifiesItself(
+        groupRules: String,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) throws {
+        let signed = signer()
+        let proof = try XCTUnwrap(GroupRulesProof(
+            group: makeGroup(rules: groupRules, members: ["cc": signed.profile]),
+            blsHex: "cc"
+        ))
+        let json = try object(from: proof)
+        let member = try XCTUnwrap(json["member"] as? [String: Any])
+        let carried = try XCTUnwrap(json["rules"] as? [String: Any])
+        let groupID = try XCTUnwrap(Data(
+            hexString: try XCTUnwrap((json["group"] as? [String: Any])?["id"] as? String)
+        ))
+        let keyBytes = try XCTUnwrap(Data(
+            hexString: try XCTUnwrap(member["sending_public_key"] as? String)
+        ))
+        let signature = try XCTUnwrap(Data(
+            hexString: try XCTUnwrap(member["signature"] as? String)
+        ))
+        // The byte counts the readme instructs the reader to assume.
+        XCTAssertEqual(groupID.count, 32, "readme says group.id is 32 bytes", file: file, line: line)
+        XCTAssertEqual(keyBytes.count, 32, file: file, line: line)
+        XCTAssertEqual(signature.count, 64, file: file, line: line)
+
+        var statement = Data("onym-group-rules-v1".utf8)
+        statement.append(groupID)
+        statement.append(Data(SHA256.hash(data: Data(
+            (try XCTUnwrap(carried["text"] as? String)).utf8
+        ))))
+        statement.append(keyBytes)
+
+        let key = try Curve25519.Signing.PublicKey(rawRepresentation: keyBytes)
+        XCTAssertTrue(
+            key.isValidSignature(signature, for: statement),
+            "the document must verify by its own instructions", file: file, line: line
+        )
+    }
+
+    func test_theExportCarriesTheSignedWordingNotTodays() throws {
+        // The bytes verify against what was agreed to. Exporting the
+        // group's current text beside the old signature would produce a
+        // document that fails its own instructions.
+        let signed = signer()
+        let group = makeGroup(
+            rules: "Be kind. No links. And no photos.",
+            members: ["cc": signed.profile]
+        )
+        let json = try object(from: try XCTUnwrap(
+            GroupRulesProof(group: group, blsHex: "cc")
+        ))
+        let carried = try XCTUnwrap(json["rules"] as? [String: Any])
+        XCTAssertEqual(carried["text"] as? String, rules, "the wording they signed")
+        XCTAssertEqual(carried["matches_current_rules"] as? Bool, false, "and the divergence is named")
+    }
+
+    func test_anUnprovenStanding_shipsNoBytesToMistakeForProof() throws {
+        // A signature in a document called a proof, which does not
+        // verify, is worse than no signature: it invites a reader to
+        // assume someone already checked.
+        let key = Curve25519.Signing.PrivateKey()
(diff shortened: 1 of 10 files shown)


dev-onym (commented):
I read both new test files against the code under test (`GroupRulesStanding.swift`, `GroupRulesProof.swift`, `MemberProfile.swift`, `ChatMembersView.swift`) and verified the identifiers, labels, page objects, and launch args all exist.

**APPROVE** — the tests match the implementation; findings below are coverage and hygiene, none blocking.

- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:137` — the load-bearing rebuild-and-verify test only runs against the *matching* rules case. The failure it exists to catch (exporting today's wording beside an old signature) is the `signedEarlierVersion` case, and there (`:165`) it's only checked by string equality. Running the same statement rebuild against the changed-rules group would catch that cryptographically rather than by comparison against the writer's own choice.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesStanding.swift:59` — untested branch: signature + hash present but `rulesText` nil. `MemberProfile.init` normalizes blank text to nil while keeping the paired hash/signature (`MemberProfile.swift:95`), so a profile with stored bytes and no text renders as **"Didn't sign"** (amber) rather than **"Signature doesn't check out"** (red). Two very different claims about a member, decided by an unasserted `guard`.
- `GroupRulesStandingTests.swift:179` — the unproven-export test covers only `.doesNotVerify`. The `.author` export is the one every group actually has, is reachable from the UI (`ChatMembersView.swift:331` makes any non-`.noRules` row tappable), and ships with `rules` omitted entirely — a "proof" document for the founder containing no rules text. Worth pinning its note and shape too, along with `.noRules`.
- `GroupRulesStandingTests.swift:152,156,160` — `as! String` inside a file that otherwise uses `XCTUnwrap`. A force cast traps and kills the whole test process instead of failing this one test; the surrounding `try` already covers `XCTUnwrap` here.
- `GroupRulesStandingTests.swift:306` — third private `Data(hexString:)` in this target (`MediaCommitmentFixtureTests.swift:108`, `GroupRulesVectorTests.swift:214`). Reimplementing it rather than using `ChatGroup.bytes(fromHex:)` is right for a verification test, but it belongs in a shared test helper now.
- `GroupRulesStandingTests.swift:137` — consider asserting `groupID.count == 32` and `signature.count == 64`; `_readme` instructs the reader on those byte counts, so the document's own claim about its shape goes unchecked.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:131` — optional-chained `?.tap()` silently no-ops if no row matches "Bob"; the failure then surfaces as "could not open Bob's proof", pointing at the proof screen rather than the missing row. `XCTUnwrap` here names the real cause.
- `GroupRulesProofUITests.swift:146,155,166` — `addIdentity`/`switchIdentity`/`openChat` are verbatim copies of `MultiIdentityChatUITests.swift:251,265,287` (and `openChat` also exists in `SearchUITests.swift:104`). Third copy of the two-identity dance; time to move it into `PageObjects/`.

The gate assertion at `GroupRulesProofUITests.swift:66` is not vacuous — `JoinConfirmScreen.waitReady()` waits on `join_confirm.send` existence first, so `isEnabled` is read on a present element. Cross-device assertion from both sides is the right call for this claim; nothing below the UI covers it.

dev-onym (commented):
COMMENT — the export self-verification test is the right load-bearing test and the cross-device UI test earns its runtime; findings are all test-quality, plus one real coverage gap.

- `Tests/OnymIOSTests/JoinRequestApproverTests.swift:261` — `test_approve_recordsAnAgreementItCouldNotVerify` doesn't test that. `seedEnvironment` only builds `rulesHash`/`rulesSignature` when `joinerSigningKey != nil` (`:798`), and this case passes `nil` — so the request carries *no* agreement and the assertion `.didNotSign` is just the missing-signature path already implied by the other tests. The branch the name and comment describe — `JoinRequestApprover.swift:473` "recorded whatever it said, including when it didn't verify" — has no test. Add a fixture with a well-formed-but-wrong 64-byte signature and assert `.doesNotVerify` plus that `profile.rulesSignature` survived; that is the distinction the comment says must not be lost after the fact.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:64` — the gate assertion is vacuously satisfiable: `XCTAssertFalse(app.buttons["join_confirm.send"].isEnabled)` passes when the button doesn't exist at all (`isEnabled` is `false` for a nonexistent element), so a renamed identifier or an unrendered Send reads as "correctly closed". `waitForExistence` on it first, then assert disabled. Same shape at `:69`: the comment asserts "exactly one switch on this screen" but nothing checks it — if a second toggle appears, `firstMatch` silently taps the wrong one.
- `Tests/OnymIOSUITests/PageObjects/TwoIdentityFlow.swift:5-8` — the doc claim doesn't match the tree. `SearchUITests.swift:104` still has its own private `openChat` verbatim, and `GroupRulesProofUITests` is new in this PR so it never held a copy. Migrate `SearchUITests` (that's the third place the extraction is supposed to be paying for) and reword.
- `Tests/OnymIOSTests/HexTestSupport.swift:6-8` — same overstatement: two copies were removed, `GroupRulesStandingTests` is new. Worth noting the *actual* remaining triplicate in this target is `String.repeated(_:)` (`GroupNamePayloadTests.swift:75`, `SwiftDataGroupStoreTests.swift:306`, `SwiftDataMessageStoreTests.swift:615`) — the two new dispatcher tests use it. The "second implementation on purpose, strict where production is lenient" rationale is right and worth keeping.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:236` — `test_theExportIsStableBetweenRuns` calls `jsonData()` twice on one instance. It does catch a timestamp added inside `document` (recomputed per call), but not per-instance nondeterminism, which is what "two exports of one agreement are byte-identical" claims. Construct two `GroupRulesProof` values from the same group and compare.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:214` — `.unknownRules` gets a standing test but no export test, and it's the case `GroupRulesStanding` documents most carefully (`GroupRulesStanding.swift:33-47`): its `note` is the one that replaced the false "joined before this group had rules". One assertion on that note would pin the fix that motivated the case.
- `:150` — `app.navigationBars.buttons.element(boundBy: 0).tap()` is index-based where the other page objects expose `back()`; fine now, brittle if a leading toolbar item lands on the members screen.

Verified against production: the filename expectations (`maple-garden-bob`, `istanbul-bob`, `group-rules`) all follow from `suggestedFileName`'s fold/lowercase/collapse/trim, and `assertExportVerifiesItself`'s raw `SHA256(text)` is safe against `GroupRules.hash`'s canonicalization only because `MemberProfile.init` normalizes `rulesText` (`MemberProfile.swift:95`) — that's an implicit dependency the export's `_readme` rests on, so it may be worth a line in the test's comment.

dev-onym (commented):
**COMMENT** — the export self-verification test is the right load-bearing choice and the standing matrix is thorough; one test doesn't test what it claims, and two UI assertions can flake.

- `Tests/OnymIOSTests/JoinRequestApproverTests.swift:239` — `test_approve_recordsAnAgreementItCouldNotVerify` records nothing. `seedEnvironment` only builds `rulesHash`/`rulesSignature` under `if let groupRules, let joinerSigningKey` (`:840`), so passing `joinerSigningKey: nil` sends a request with *no* agreement bytes at all. The assertion `.didNotSign` is just the older-client path, already covered. The case the name and comment describe — bytes kept and admitted anyway — is `RulesAgreement.invalid` → `.doesNotVerify`, which needs a signature over *other* text whose hash matches the group's rules (`JoinRequestApprover.swift:1090`). That path is explicitly documented at `JoinRequestApprover.swift:1057` and is untested at the approver level; the unit-level `.doesNotVerify` tests in `GroupRulesStandingTests` construct the profile by hand and don't reach it.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:142` — `app.buttons["rules_proof.export"].exists` with no wait. The button only renders once `@State file` is set by `writeFile()`'s detached encode-and-write (`MemberRulesProofView.swift:166`, `:214`), which is unsynchronized with the `WHAT THEY SIGNED` text appearing. Use `waitForExistence`.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:157` — `rules_proof.done` is tapped, then `members.share_invite_button` is polled for 2s and tapped with no wait for the member sheet to actually disappear. Both go through the same `sheet(item:)` (`ChatMembersView.swift:95`), so a re-present while the first is still dismissing is the classic swallow — the exact failure the comment above it says would go unnoticed. Wait for the sheet to be gone first.
- `Tests/OnymIOSUITests/PageObjects/TwoIdentityFlow.swift:8` — the doc says this existed in three files "before it was extracted", but `SearchUITests.swift:104` still carries its own `openChat` copy verbatim. Either migrate it or drop the claim.
- `Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:61` — `groupIDHex` is deliberately re-hexed from `groupIDData` rather than echoing `group.id`, because `bytes(fromHex:)` is lenient and the `_readme` promises 32 bytes. `test_theExportCarriesWhatAnOutsiderNeedsToCheckIt:127` asserts it equals `group.id` using a well-formed id, so the reason the field exists isn't exercised. A group whose `id` is odd-length or non-hex is the missing case.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:262` — `test_clearingTheRulesDoesNotDeleteWhatPeopleSigned` stops at the standing. Its export is the one document where `matches_current_rules: false` ships with no `current_text` (`GroupRulesProof.swift:207`, `currentRules` is nil), which leaves the reader the readme's compare-against instruction (`:227`) and nothing to compare against. Worth an assertion either way, since it pins intent.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:170` and `:340` — `test_theExportCarriesTheSignedWordingNotTodays` and `test_theDivergedWordingShipsBothVersions` build the identical fixture and both assert `carried["text"] == rules`; they differ only in the second key checked. Fold into one.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:147` — `app.staticTexts["member"].exists` matches the row *label* in `MemberRulesProofView.swift:152`, not the fingerprint beside it, so it passes even if the hex value were empty. Asserting on `rules_proof.bytes` or the BLS prefix is what the comment claims.

I verified the standing matrix, filename derivation (including the `İstanbul 名簿` and empty-stem cases), and the statement-rebuild in `assertExportVerifiesItself` against the production sources — those all line up, and the hex helper consolidation is behavior-preserving.

dev-onym (commented):
APPROVE — thorough, well-aimed tests; the self-verifying export test is the right load-bearing assertion. Findings are minor.

- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:155` — the share-invite check is wrapped in `if shareInvite.waitForExistence(timeout: 2)`, so the one assertion the comment says "would have gone quiet" goes quiet again if the button stops rendering for the admin. Pass the expectation in (`assertMembersScreenShowsStandings(app, from: "Alice", isAdmin: true)`) and `XCTAssertTrue` the button's existence for Alice; 2s is also thin for a freshly-pushed SwiftUI screen.
- `Tests/OnymIOSTests/IncomingMessageDispatcherTests.swift:1560` — `XCTAssertNil(mine.memberProfiles[selfBlsHex]?.rulesSignature)` passes vacuously if the dispatcher stored no self row at all, which is a different (and also wrong) outcome than "row present, no borrowed signature". `try XCTUnwrap(mine.memberProfiles[selfBlsHex])` first, then assert `.rulesSignature` is nil.
- Coverage gap on `GroupRulesProof.note(for:)` (`Packages/OnymGroup/Sources/OnymGroup/GroupRulesProof.swift:266`): the new tests pin only the `.author` and `.noRules` notes. `.unknownRules` is exactly where the false claim lived — `test_aStoredSignatureWithNoTextIsNotAMissingOne` (`GroupRulesStandingTests.swift:236`) pins the standing but nothing pins that its export no longer says "Joined before this group had rules". One export assertion per unproven standing would close it; `.doesNotVerify` currently only asserts `note != nil` (`GroupRulesStandingTests.swift:229`).
- `Tests/OnymIOSTests/HexTestSupport.swift:16` — the doc claims "a non-hex character is nil", but `UInt8(_:radix:)` accepts a leading sign, so `"+f"` decodes to `0x0f`. Harmless for these fixtures; either tighten (reject non-`isHexDigit` chars) or soften the comment, since strictness is this helper's stated reason to exist over `ChatGroup.bytes(fromHex:)`.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:257` — `test_theExportIsStableBetweenRuns` compares two `jsonData()` calls in one process; it catches a timestamp or unsorted keys but not cross-run instability. Name it for what it checks ("twice in a row") or drop the "between runs" claim.
- `Tests/OnymIOSUITests/PageObjects/TwoIdentityFlow.swift:8` — the doc says three copies were extracted, but `SearchUITests.swift:104` still has its own private `openChat`. Fold it in or trim the comment to two.

dev-onym (commented):
COMMENT — solid, well-targeted test suite; a few tests don't assert what their names claim, and one production guarantee is left uncovered.

- `Tests/OnymIOSTests/JoinRequestApproverTests.swift:261` — `test_approve_recordsAnAgreementItCouldNotVerify` doesn't test that. With `joinerSigningKey: nil` (`:267`) the request carries `rulesHash: nil, rulesSignature: nil`, so `MemberProfile.paired` drops nothing — there were never any bytes — and the assertion is just `.didNotSign`, which duplicates the no-agreement case. The behavior `JoinRequestApprover.swift:472-480` explicitly documents ("recorded whatever it said, including when it didn't verify") needs a bogus-but-well-formed signature (sign with a foreign key, or 64 zero bytes) and should assert `profile.rulesSignature != nil` and standing `.doesNotVerify`. As written, nothing in the suite stops a future "only store it if it verifies" change — which would silently collapse "signed nothing" and "signature was wrong".
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:236` — `test_theExportIsStableBetweenRuns` encodes the *same* `GroupRulesProof` twice. It does catch an embedded timestamp (`document` is a computed var), but `Document` has no dictionaries, so key order is already fixed by `CodingKeys`; the `.sortedKeys` half of the claim can't fail here, and the PR-body claim "two exports of one agreement" isn't tested. Build the proof twice from two separately-constructed `ChatGroup`s and compare.
- `Tests/OnymIOSTests/IncomingMessageDispatcherTests.swift:1505` — `XCTAssertNil(mine.memberProfiles[selfBlsHex]?.rulesSignature)` passes vacuously if the self row is never created at all, so a merge that drops the row entirely reads as a pass. Add `XCTAssertNotNil(mine.memberProfiles[selfBlsHex])` first.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:119,123` — the standings are asserted as app-wide static texts, not per row, so they'd pass if the marks landed on each other's rows. You already locate `bobRow` by label at `:141`; asserting `bobRow.label.contains("Signed the group rules")` binds mark to member for the same cost.
- `GroupRulesStandingTests.swift:180` and `:340` are near-duplicates (same diverged group, same `carried["text"] == rules`); `:340` is a strict superset. Fold them.
- `Tests/OnymIOSUITests/PageObjects/TwoIdentityFlow.swift:8` — the doc says the helper existed in `SearchUITests` too, but `SearchUITests` isn't in the diff and doesn't call `TwoIdentityFlow`. Either the third copy was already gone or it's still there; fix the comment to match.
- Nice: `assertExportVerifiesItself` rebuilding the statement from the document's own fields is the right load-bearing test, and it's safe against the canonicalization trap only because `MemberProfile` normalizes `rulesText` (`MemberProfile.swift:95`) — so `SHA256(rules.text)` in the readme matches `GroupRules.hash`. A case with leading/trailing whitespace in the founder's text would pin that coupling.

dev-onym (commented):
**COMMENT** — the crypto assertions are genuinely load-bearing (rebuilding the statement from the document's own fields is the right test); the gaps are in the write path and a few assertions that can pass vacuously.

- `Packages/OnymChatsUI/Sources/OnymChatsUI/MemberRulesProofView.swift:300-403` — no test anywhere for the export **writer**: the per-write UUID directory, the `Task.isCancelled` supersede guard, or the age-based `sweepStaleExports`. That is exactly the logic fixed in the two commits immediately before this one (`8931f8c`, `f1f954e`), and it's the only part of the feature that touches the filesystem. `GroupRulesProofUITests.swift:150` asserts `rules_proof.export` *exists* but never taps it, so a regression that writes nothing, publishes a swept URL, or lets a superseded write win is invisible. Given the PR's framing, this is the coverage I'd most want added — a unit test over a temp dir for write→supersede→sweep.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:59` — `XCTAssertFalse(app.buttons["join_confirm.send"].isEnabled)` passes vacuously: a missing or renamed button reports `isEnabled == false`. The gate is the stated reason this half exists; assert `waitForExistence` first, then `isEnabled`.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:158-168` — `if shareInvite.waitForExistence(timeout: 2)` keys the branch on the presence of the thing under test. The comment says this is the sheet that "would have gone quiet" after the `sheet(item:)` consolidation — but if Alice's button regresses, the block silently skips. `viewer` is already in scope: `if viewer == "Alice" { XCTAssertTrue(...) }`.
- `Tests/OnymIOSUITests/GroupRulesProofUITests.swift:120` — `staticTexts["GROUP RULES"]` and `staticTexts[rules]` are both also rendered by `ChatEmptyStateView.swift:36`, so neither proves the members screen opened. `ChatMembersView.swift:328` already exposes `members.rules`; use it.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift:100-112` — the comment claims the signer is bound inside the signed bytes, but the impostor also gets a fresh `sendingPubkey`, so this only shows Ed25519 fails under a different key. The lift that actually matters — signature *and* `sendingPubkey` copied under a different roster key — returns `.signed` by design (`GroupRulesProof.readme` says so explicitly). Worth a test pinning that as intended, since it's the document's central caveat.
- `Tests/OnymIOSTests/GroupRulesStandingTests.swift` — export `note` is asserted for `.author` and `.noRules`, non-nil only for `.doesNotVerify`, and untested for `.unknownRules` and `.notCollected`. Those strings are the whole content of an unproven document; two more `XCTAssertEqual`s.
- `Tests/OnymIOSTests/IncomingMessageDispatcherTests.swift:1338,1359` — `test_aMemberWhoSignedNothing…` and `test_aGroupTypeWithNoJoinApproval…` dispatch nothing; they seed a group and assert `rulesStanding`. They belong in `GroupRulesStandingTests`, whose `makeGroup` hardcodes `.tyranny` — extending it with a `groupType` parameter puts `.notCollected` where the rest of the standings live.
- `Tests/OnymIOSTests/HexTestSupport.swift` — good call keeping this a second implementation rather than reusing `ChatGroup.bytes(fromHex:)`; the strictness rationale holds up.

No Swift toolchain in this environment, so I verified the assertions against the production sources by hand rather than by running the suite — the "1579 green" claim is unchecked here.

PR #310, opened 2026-08-22
Speak the erase wire the operator actually answers

rinat-enikeev:
## The problem

With five backups retained at an operator declaring `maximumRetainedSnapshots: 5`, the sixth run was refused `quota_exceeded` and Settings showed "1 operator needs attention" — instead of the rolling window erasing the oldest snapshot and the new one landing.

`makeRoom` was doing its job: it listed the operator's snapshots, picked the oldest, and requested a §11 erase. The erase never succeeded, and `makeRoom` deliberately swallows a failed housekeeping erase (a failed erase says nothing about whether the upload can be stored), so nothing surfaced until the operator itself refused the upload.

## Three wire mismatches, all on the erase path

Measured against the deployed operator (`onym-infra/backup/operator`, live at backup.onym.app):

1. **The erase body carried no `operationId`.** The operator requires a client-chosen id — §14.9 reconciliation is keyed on an id the client already knows — and answers 400 without one. Every erase this client ever sent was refused.
2. **The response is an array of receipts; the client decoded a single object.** The operator mints one receipt per distinct pinned `termsId` in scope, so even a single-snapshot erase arrives as a one-element array.
3. **The operator's timestamps carry fractional seconds; the receipt parser refused them.** `acknowledgedAt`/`completionCommittedBy` are stamped `2026-08-22T13:42:25.779692Z` (RFC 3339, `time` crate), and `BackupFormat.date(fromRFC3339:)` was pinned to the fraction-less form.

The unit suite never caught this because `ScriptedPort` fakes the port above the wire; the erase request and response bytes were never exercised against the operator's actual shapes.

## The fix

- `URLSessionBackupClient.eraseSnapshot` sends a client-chosen `operationId` and decodes the receipt array — refusing an empty one, because an erasure that committed nothing is an error, not a receipt.
- `BackupPort.eraseSnapshot` returns `[ErasureReceipt]`; the repository records every receipt's raw bytes (they are the only durable record), and `DeviceBackupSettingsFlow` keeps the first for display.
- `BackupFormat.date(fromRFC3339:)` reads fractional seconds. Reading is wider than writing: emission stays pinned to the fraction-less form the signed bytes use.

## Tests

- `testEraseSendsAnOperationIdAndReadsTheReceiptArray` — pins the request body shape and the array response with fractional stamps, at the transport level through `StubURLProtocol`.
- `testEraseAllKeepsEveryReceipt` — a `scope: "all"` spanning two pinned terms returns both receipts.
- `testAnEmptyReceiptArrayIsRefused`.

### DIFF
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupFormat.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupFormat.swift
@@ -63,8 +63,20 @@ enum BackupFormat {
         return formatter
     }()
 
+    /// Reading is wider than writing: RFC 3339 allows fractional
+    /// seconds, and the operator's stamps carry them, so a reader
+    /// pinned to the emitting form would refuse every receipt the
+    /// operator actually signs.
+    static let fractionalTimestampFormatter: ISO8601DateFormatter = {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        formatter.timeZone = TimeZone(secondsFromGMT: 0)
+        return formatter
+    }()
+
     static func date(fromRFC3339 value: String) -> Date? {
         timestampFormatter.date(from: value)
+            ?? fractionalTimestampFormatter.date(from: value)
     }
 
     static func string(fromDate date: Date) -> String {
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupPort.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupPort.swift
@@ -35,10 +35,12 @@ public protocol BackupPort: Sendable {
     /// restore.
     func downloadSnapshot(_ reference: SnapshotReference, to destination: URL) async throws
 
-    /// Request erasure and return the operator's signed receipt.
-    /// An acknowledged-but-incomplete erasure is `erasureUnconfirmed`,
-    /// carrying the receipt — it is not success.
-    func eraseSnapshot(scope: ErasureScope) async throws -> ErasureReceipt
+    /// Request erasure and return the operator's signed receipts — one
+    /// per distinct pinned `termsId` in scope (§11), so a single
+    /// snapshot yields exactly one and `all` may yield several. Never
+    /// empty. An acknowledged-but-incomplete erasure is
+    /// `erasureUnconfirmed`, carrying the receipt — it is not success.
+    func eraseSnapshot(scope: ErasureScope) async throws -> [ErasureReceipt]
 
     /// The portable container for migration to another operator or to the
     /// person's own storage. Never gated on payment
--- a/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
+++ b/Packages/OnymBackup/Sources/OnymBackup/BackupRepository.swift
@@ -740,9 +740,9 @@ public actor BackupRepository {
         guard surplus <= erasable.count else { return }
 
         for row in erasable.prefix(surplus) {
-            let receipt: ErasureReceipt
+            let receipts: [ErasureReceipt]
             do {
-                receipt = try await port.eraseSnapshot(scope: .snapshot(row.snapshotReference))
+                receipts = try await port.eraseSnapshot(scope: .snapshot(row.snapshotReference))
             } catch {
                 // Stopped, not thrown. A housekeeping erase that failed
                 // says nothing about whether this snapshot can be
@@ -757,7 +757,7 @@ public actor BackupRepository {
             // durable record that this happened, and a batch written
             // after the last erase is a batch lost by a crash in the
             // middle of the first.
-            record(receipt, scope: .snapshot(row.snapshotReference), in: &state)
+            record(receipts, scope: .snapshot(row.snapshotReference), in: &state)
             try commit(state, expecting: state.componentId)
         }
     }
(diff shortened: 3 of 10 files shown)


dev-onym (commented):
REQUEST_CHANGES — the wire fixes are right in shape, but the fractional-timestamp fix stops one function short of the path that actually feeds the rolling window, and the stored receipt bytes are no longer the operator's.

- `URLSessionBackupClient+Transport.swift:15` — `Self.decoder` still uses `dateDecodingStrategy = .iso8601`, which rejects fractional seconds. `listSnapshots` decodes `retainedAt`/`retainedUntil` (`URLSessionBackupClient.swift:244-250`) and `exportSnapshots` decodes `retainedAt` (`:310`) through it. If the operator's `time`-crate clock stamps those the way it stamps `acknowledgedAt`, `listSnapshots` throws `malformed_response`, `makeRoom` never gets a list, and the quota bug in the PR body survives this fix. Point the strategy at `.custom { BackupFormat.date(fromRFC3339:) }` and pin it with a transport test.
- `URLSessionBackupClient.swift:297` — each element is re-serialized with `JSONSerialization(.sortedKeys)`, so `rawBytes` is a client reconstruction: key order, whitespace, number formatting, and unicode escaping all differ from what the operator sent. `BackupRepository.swift:934` writes exactly those bytes as the durable record and `BackupPort.swift:38` calls them "signed receipts" — a detached signature could never be checked against them. Keep the per-element source substrings (or store the whole response body verbatim alongside).
- `URLSessionBackupClient.swift:290-295` — `[]` is refused for every scope. An operator that legitimately holds nothing in scope for `.all` now makes `stopBackingUp(erasingFirst: true)` (`DeviceBackupSettingsFlow.swift:355-366`) report "did not confirm that it erased your backup, so nothing was changed" and refuse to stop. Worth distinguishing nothing-in-scope from malformed; at minimum decide it deliberately.
- `DeviceBackupSettingsFlow.swift:357,393` — `.first` on a multi-receipt `.all` lets the operator choose which `excludedScope` and `completionCommittedBy` the person is shown, by response order. Given the receipt's whole purpose is the excluded scope, hold `[ErasureReceipt]` in state (or select by latest `completionCommittedBy`).
- `URLSessionBackupClient.swift:296-299` — nothing checks that a returned receipt's `scope` matches what was requested; `record` (`BackupRepository.swift:934-940`) then drops the local row by the *requested* digest regardless. Pre-existing, but this path is being rewritten and the file's stated ethos is refusing what the operator asserts.
- `BackupAdapterTests.swift:236-243` — bare `catch {}` passes on any error, including an unrelated transport failure; assert `BackupError.rejected(code: "malformed_receipt")`. Nothing yet pins that a receipt's `rawBytes` equals the operator's bytes, which is why the finding above is invisible to the suite.
- Minor: `BackupFormat.swift:66-79` mirrors the existing `ServiceManifestFormat` pair, which is the right precedent. Note the tolerance is Foundation's — 9-digit/nanosecond stamps or a `+00:00` offset instead of `Z` would still be refused; if the operator can emit either, widen the test.

dev-onym (commented):
**COMMENT** — the three wire fixes look right and well-tested at the transport level; the concerns below are about what happens when a response is partly bad, and about the diagnostic that stayed silent.

- `URLSessionBackupClient.swift:305` — `elements.map { try ErasureReceipt.decode(...) }` throws on the first bad element and discards the receipts already decoded. For `scope: .all` spanning two pinned terms, if receipt #2 has an empty `excludedScope` (or an echoed scope mismatch at :311), the operator has erased everything and this client keeps *zero* receipts and drops no local rows — the same "an erasure that committed nothing is not a receipt" argument used to refuse `[]` at :301 applies harder here. Consider partitioning: record every well-formed receipt's `rawBytes`, then throw.
- `URLSessionBackupClient.swift:282` — the `operationId` is minted inline and never persisted (nothing in `BackupStateStore.swift` holds erase ids, unlike `pendingOperations`). So the comment's §14.9 rationale ("a lost erase response can only be asked about under an id this side already knows") isn't actually achievable yet: a retry after a lost response mints a fresh id. Either persist it or soften the comment to "the operator requires one".
- `BackupRepository.swift:744-752` — the swallow that hid this bug for the whole life of the erase path is unchanged, and nothing records that a housekeeping erase failed. The next erase-wire regression is equally invisible until `quota_exceeded`. A counter/last-error on state (surfaced as the operator-attention reason) would make the class of bug self-reporting rather than requiring another field report.
- `URLSessionBackupClient+Transport.swift:88` — `|| !elements.isEmpty` lets a trailing comma (`[{…},]`) through as a second, empty element, which then fails as `malformed_receipt`. Reachable only via a malformed response, but the guard reads like it was meant to reject it; `[]` is already handled above at :79.
- `BackupFormat.swift:70-79` — the Transport comment (`:17-21`) says the point is not to depend on "a fact about the Foundation build", but `.withFractionalSeconds` on `ISO8601DateFormatter` is exactly such a fact — including whether 9-digit fractions (used in the tests) parse. The upside is that `testTheTimestampReaderAcceptsTheOperatorsClock` now fails loudly instead of the window silently stalling; worth trimming the comment's claim.
- `DeviceBackupSettingsFlow.swift:415` — `receiptToShow` returns `nil` on an empty array, so a port that returned `[]` would silently clear a previously displayed receipt, while the `catch` paths deliberately preserve it. `?? state.lastReceipt` would match that intent. Tie-break by `receiptId` is a good call.
- `ErasureReceipt.decode(raw:)` is still called without a signature, so every wire receipt has `signature == nil` — pre-existing, but it undercuts the "bytes a signature would be checked against" rationale at :288-293. Fine to defer; a note in the comment would keep it honest.
- Tests read well and pin the real shapes. No Swift toolchain in this worktree, so I did not build or run them.

PR #311, opened 2026-08-22
A package that can ask to be woken, and nothing else

rinat-enikeev:
First of three iOS PRs for push notifications, pairing with backend PRs onymchat/onym-push#1–#2. New `Packages/OnymPush`, no app wiring yet — it builds standalone.

## Design recap

The backend is a relay watcher: this device registers its inbox tags (already relay-visible routing handles), the relays to watch, and an encrypted APNs token; the backend wakes it with a content-free "New message" alert when a matching kind-34113 event appears. No group concept anywhere; senders never contact the backend.

## Contents

- `PushRegistrationPayload.swift` — the signed session bytes, mirrored byte-for-byte by `onym-push/apple/src/payload.rs` (length-prefixed, domain-separated; the subscriptions digest covers the wire order exactly, no normalization on either side). Cross-language fixtures are generated from this side in the tests PR.
- `PushTokenEnvelope.swift` — CryptoKit X25519 + HKDF (salt `onym-push-token-v1`) + AES-GCM, so the raw token never sits readable in transit buffers or the backend's request path.
- `PushBackendClient.swift` / `URLSessionPushBackendClient.swift` — the enforcement client's shape: `https://push.onym.app`, ephemeral cookie-less session, https-only in every build, `{error, message}` rejections kept verbatim. Plus an actor stub for tests.
- `PushRegistrationInteractor.swift` — replace-all reconciliation behind `PushSigner`/`PushDeviceAttestationProvider` seams (the `ModerationSigner` pattern): debounced, fingerprinted so unchanged state makes no call, with a 7-day refresh so the backend's 60-day stale sweep never reaps a live device. Registration failures are deliberately quiet per the no-activity-logging stance.
- `PushPreferenceStore.swift` — opt-in, **default OFF**: the tag↔device linkage is the user's trade to make in Settings, never a default.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -229,11 +229,13 @@ jobs:
             -resultBundlePath /tmp/onym-foundation.xcresult
 
       - name: xcodebuild build (OnymPush package)
-        # OnymPush has no package test target (its suites live in
-        # OnymIOSTests, run above) but nothing else compiles the
-        # package on its own, so a signature-payload typo could land
-        # green. Build it standalone here — same `always()` rationale
-        # as the package test steps.
+        # OnymPush's suites live in OnymIOSTests (run above), which
+        # already compiles the package — as a dependency of the app
+        # target. This standalone build proves something the app build
+        # cannot: the package needs no app scaffolding, so an
+        # accidental reference to app-side types fails here instead of
+        # landing green. Same `always()` rationale as the package test
+        # steps.
         if: always()
         working-directory: Packages/OnymPush
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -228,6 +228,21 @@ jobs:
             -derivedDataPath /tmp/onym-foundation-tests \
             -resultBundlePath /tmp/onym-foundation.xcresult
 
+      - name: xcodebuild build (OnymPush package)
+        # OnymPush has no package test target (its suites live in
+        # OnymIOSTests, run above) but nothing else compiles the
+        # package on its own, so a signature-payload typo could land
+        # green. Build it standalone here — same `always()` rationale
+        # as the package test steps.
+        if: always()
+        working-directory: Packages/OnymPush
+        run: |
+          rm -rf /tmp/onym-push-build
+          xcodebuild build \
+            -scheme OnymPush \
+            -destination "platform=iOS Simulator,name=${{ steps.sim.outputs.name }},OS=latest" \
+            -derivedDataPath /tmp/onym-push-build
+
       - name: Upload unit test xcresult
         # `always()` so `report-to-issue` can parse green runs too —
         # not just failures.
--- a/Packages/OnymPush/Package.swift
+++ b/Packages/OnymPush/Package.swift
@@ -1,0 +1,13 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
+    name: "OnymPush",
+    platforms: [.iOS("18.0")],
+    products: [
+        .library(name: "OnymPush", targets: ["OnymPush"]),
+    ],
+    targets: [
+        .target(name: "OnymPush"),
+    ]
+)
--- a/Packages/OnymPush/Sources/OnymPush/PushBackendClient.swift
+++ b/Packages/OnymPush/Sources/OnymPush/PushBackendClient.swift
@@ -1,0 +1,106 @@
+import Foundation
+
+/// A register call as the client presents it. Every field the
+/// signature covers is transmitted (the APNs token inside the
+/// envelope), so the backend can reconstruct the signed bytes and
+/// actually verify — see `SignedPushSessionPayload`.
+public struct PushRegisterRequest: Codable, Sendable, Equatable {
+    /// Fresh `DCDevice` token, or nil when attestation is unavailable
+    /// (simulator, enterprise build). The client never fabricates one.
+    public let deviceToken: Data?
+    public let userKey: String
+    public let timestamp: Date
+    /// User-key signature over `SignedPushSessionPayload.register`.
+    public let signature: Data
+    public let tokenEnvelope: PushTokenEnvelope
+    public let subscriptions: [PushSubscription]
+
+    public init(
+        deviceToken: Data?,
+        userKey: String,
+        timestamp: Date,
+        signature: Data,
+        tokenEnvelope: PushTokenEnvelope,
+        subscriptions: [PushSubscription]
+    ) {
+        self.deviceToken = deviceToken
+        self.userKey = userKey
+        self.timestamp = timestamp
+        self.signature = signature
+        self.tokenEnvelope = tokenEnvelope
+        self.subscriptions = subscriptions
+    }
+}
+
+public struct PushRegisterResponse: Codable, Sendable, Equatable {
+    /// When this registration lapses unless refreshed — the client
+    /// re-registers well before this while push stays enabled.
+    public let expiresAt: Date
+
+    public init(expiresAt: Date) {
+        self.expiresAt = expiresAt
+    }
+}
+
+public struct PushUnregisterRequest: Codable, Sendable, Equatable {
+    public let deviceToken: Data?
+    public let userKey: String
+    public let timestamp: Date
+    /// User-key signature over `SignedPushSessionPayload.unregister`.
+    public let signature: Data
+    public let tokenEnvelope: PushTokenEnvelope
+
+    public init(
+        deviceToken: Data?,
+        userKey: String,
+        timestamp: Date,
+        signature: Data,
+        tokenEnvelope: PushTokenEnvelope
+    ) {
+        self.deviceToken = deviceToken
+        self.userKey = userKey
+        self.timestamp = timestamp
+        self.signature = signature
+        self.tokenEnvelope = tokenEnvelope
+    }
+}
+
+/// The push backend's refusal, kept verbatim: the service's error
+/// vocabulary (`signature_invalid`, `relay_invalid`, `capacity`) is
+/// what the caller branches on.
+public struct PushRejection: Error, Sendable, Equatable {
+    public let statusCode: Int
+    public let rawCode: String
+    public let message: String
+
+    public init(statusCode: Int, rawCode: String, message: String) {
+        self.statusCode = statusCode
+        self.rawCode = rawCode
+        self.message = message
+    }
+}
+
+public enum PushClientError: Error {
+    /// The base URL is not https — never relaxed, in any build.
+    case insecureBaseURL(String)
+    case invalidResponse
+    case malformedResponse(String)
+    case rejected(PushRejection)
+}
+
+/// The interface vendor's push backend — the relay watcher that turns
+/// inbox-tagged events into content-free wakes. Production builds ship
+/// `URLSessionPushBackendClient` against the deployed service; tests
+/// use `StubPushBackendClient`.
+public protocol PushBackendClient: Sendable {
+    /// The X25519 key APNs tokens are encrypted to. Fetched before
+    /// each register call so server-side rotation needs no release.
+    func registrationKey() async throws -> Data
+
+    /// Replace-all upsert of this device's subscription set. Session
+    /// signatures are single-use server-side; a replay is refused 401.
+    func register(_ request: PushRegisterRequest) async throws -> PushRegisterResponse
+
+    /// Idempotent delete of this device's registration.
+    func unregister(_ request: PushUnregisterRequest) async throws
+}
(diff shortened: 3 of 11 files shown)


dev-onym (commented):
REQUEST_CHANGES — solid mirror of the moderation package's shape, but the response decoder and the opt-out path have real holes.

- `URLSessionPushBackendClient.swift:109` — strict `.iso8601` rejects fractional seconds, which is exactly the bug `ModerationJSON.decoder()` exists to avoid ("Foundation's built-in `.iso8601` strategy accepts only the whole-second form, so both shapes are tried"). A Rust backend serializing `chrono::DateTime<Utc>`/`OffsetDateTime` emits `…:00.123456Z` by default, so every `register` would decode-fail *after* succeeding server-side: `malformedResponse` → swallowed at `PushRegistrationInteractor.swift:158` → fingerprint never recorded → a fresh register on every foreground, forever. Use a tolerant custom strategy (or confirm the backend pins whole seconds and say so in the wire notes).
- `PushRegistrationInteractor.swift:70-75` — `pushDisabled()` clears the local record *before* the network call and there is no retry. One offline opt-out leaves the device registered; the user keeps receiving wakes and the tag↔device linkage survives until the 60-day sweep. Persist a pending-unregister (token + intent) and drain it on next launch/foreground; clear the record only on success.
- `project.yml:19-41` + `.github/workflows/release.yml:195-225` — `OnymPush` is in neither the xcodegen package list nor the per-package test steps, so CI never compiles it. "Builds standalone" is currently unverified by anything. Add the `packages:` entry (no target dependency needed yet) or a `Packages/OnymPush` test step.
- `PushRegistrationInteractor.swift:96` — no in-flight guard: `reconcile()` suspends at `registrationKey()`/`sign()`/`register()`, so a second debounced run can enter and read the pre-update fingerprint, issuing two registers with two single-use signatures. Track the running task and coalesce.
- `PushRegistrationInteractor.swift:104,113,145` — `Date()` is called directly; every sibling client injects `clock: @Sendable () -> Date` (`URLSessionModerationAuthorityClient`). Without it the 7-day refresh branch is untestable in the follow-up tests PR.
- `PushBackendClient.swift:83` — `PushClientError` is neither `Sendable` nor `Equatable`, unlike `AuthorityClientError`; and the doc at `:76` says callers branch on `signature_invalid`/`relay_invalid`/`capacity`, but there's no `AuthorityErrorCode`-style enum to branch on — only raw strings.
- `PushRegistrationPayload.swift:78-87` — the digest flattens to `(tag, relay)` pairs, so a `{tag, relays: []}` entry contributes zero bytes and subscription grouping isn't covered. The signature therefore doesn't bind "the wire order exactly" as the comment at `:33` claims — pairs can be regrouped or empty-relay entries added/dropped in flight. Low impact, but either prefix the subscription count and per-subscription relay count, or soften the comment.
- `PushTokenEnvelope.swift:9` — the doc names the scheme `x25519-aes-256-gcm-v1`, but unlike `SealedEnvelope` and `SeatSealedEnvelope` the struct carries no `version`/`scheme` field, so the backend can't refuse a future suite (only the distinct salt separates it from an invitation envelope). Worth adding while the wire is still unshipped.
- `PushBackendClient.swift:38` — `expiresAt` is decoded and discarded; the refresh is a hardcoded 7 days. Either drive `refreshInterval` from it or drop the field.

dev-onym (commented):
**REQUEST_CHANGES** — solid shape and the payload/envelope work reads carefully, but the opt-out path has a reentrancy hole and the test stub's default key looks unusable.

- `PushRegistrationInteractor.swift:85` + `:169` — actor reentrancy across the network awaits: if `pushDisabled()` lands while a `reconcilePass()` is suspended in `client.register`, the disable clears the registration record, sets + drains the pending unregister, and then the resumed pass calls `recordRegistration(...)` and (`:178`) clears a pending token that no longer exists. Net result: the register can reach the backend *after* the unregister, prefs claim "registered" while `isEnabled` is false, and nothing is left pending — so the retry-forever guarantee in the doc comment doesn't hold and the device stays wake-able. Fix: re-read `preferences.isEnabled` after the awaits and, if it flipped, `setPendingUnregister(token:)` instead of recording; and have `pushDisabled()`/`drainPendingUnregister` respect `reconcileInFlight` so two single-use signatures aren't spent concurrently.
- `StubPushBackendClient.swift:16` — `Data(repeating: 0, count: 32)` does decode as a `Curve25519.KeyAgreement.PublicKey`, but agreeing against the all-zero (low-order) point produces an all-zero shared secret, which CryptoKit rejects by throwing rather than returning. So `PushTokenEnvelope.seal` throws, the interactor swallows it at `:181`, and every default-stub test in the tests PR will see zero registrations with no visible failure reason. Default to a freshly generated `Curve25519.KeyAgreement.PrivateKey().publicKey.rawRepresentation` instead, and consider mapping the agreement throw to `SealError.invalidServerKey` so `:47`'s guard isn't the only length-only check.
- `StubPushBackendClient.swift:13` — `unregisterAnswer` has no `set…` method (an actor-isolated `var` can't be assigned cross-actor), so the offline-opt-out retry path — the one the comment at `PushPreferenceStore.swift:71` promises — can't be driven from a test. Add `setUnregisterAnswer(_:)`.
- `PushRegistrationInteractor.swift:141` — `expiryMargin` is a fixed 7 days, equal to `refreshInterval`. If the backend ever returns an `expiresAt` under 7 days out, the short-circuit at `:137` can never be satisfied and every foreground spends a full register + single-use signature. Derive the margin from the returned window (e.g. `min(7d, window/2)`).
- `PushRegistrationInteractor.swift:68` — token rotation registers the new token but never unregisters the previous one, so the old token stays wake-able until the backend's 60-day sweep. Fine if the backend row is keyed by user key/device rather than by APNs token — worth stating that explicitly here, since the replace-all comment implies it.
- `PushBackendClient.swift:10` — synthesized `Codable` uses `encodeIfPresent` for optionals, so a nil `deviceToken` (simulator, enterprise — the common case in testing) omits the key entirely rather than sending `null`. Confirm the Rust struct has `#[serde(default)]` alongside `deny_unknown_fields`, and cover the nil case in the cross-language fixtures.
- `project.yml:244` — `OnymPush` is added to `packages` and to the app target's list at `:44`, but not to `OnymIOSTests`' dependencies; the tests PR can't `import OnymPush` until that line lands.

Nits: `URLSessionPushBackendClient.swift:120` tries whole-seconds before fractional, the reverse of `ModerationJSON` — harmless, but flipping it matches the sibling. The `.iso8601` encoder and `ISO8601DateFormatter.push` do agree on whole-second UTC, so the signed and transmitted timestamps match — good.

dev-onym (commented):
REQUEST_CHANGES — the reconciliation core is careful and well-documented, but the opt-out path cancels the very pass it delegates to.

- **`PushRegistrationInteractor.swift:104` — `pushDisabled()` cancels the Task that is *running* the in-flight pass.** `pendingReconcile` is the same Task that calls `reconcile()` (`:132`), not just a sleeping debounce. So when a disable races an in-flight pass: `cancel()` cancels the holder, `claimPass()` (`:115`) fails, and the drain is delegated to the holder's `runQueuedPasses()` — which now executes inside a cancelled Task, where `client.registrationKey()` / `register` / `unregister` all throw `URLError.cancelled` immediately and get swallowed by the `catch`. The comment at `:110-113` ("A pass holding the slot drains this before it releases it") doesn't hold. Net effect: the user turns push off, nothing reaches the backend, and the device keeps getting woken until the next `appForegrounded()` — exactly the failure mode the persisted-pending design was built to prevent. Fix: keep a separate handle for the debounce sleep and only cancel that, or run the pass body in an unstructured Task so an input-triggered `scheduleReconcile()`/`pushDisabled()` can't abort in-flight HTTP.
- **`PushRegistrationInteractor.swift:130` — same cancellation hazard, milder.** Any `updateToken`/`updateSubscriptions`/`appForegrounded` during a pass aborts an in-flight `register` mid-request. The backend may already have processed it and spent the single-use signature while the client records nothing, so the next pass re-registers unnecessarily.
- **`PushRegistrationInteractor.swift:109` and `:240` — the single-slot `pendingUnregisterToken` can silently drop a queued token.** After a rotation whose drain failed (pending = old token A) and while B's register is in flight, a disable overwrites pending with B; A's unregister is lost and A stays wake-able until the backend's 60-day sweep. Either key pending unregisters as a set, or refuse to overwrite a non-nil slot.
- **`PushRegistrationInteractor.swift:196` — no floor on the backend-granted window.** A `expiresAt` at/behind `registeredAt` gives `expiryMargin == 0` and a never-satisfiable skip condition, so every foreground re-registers (and spends a signature). A minimum window check, or a failure/attempt backoff, would bound it.
- **`project.yml:42` and `.github/workflows/release.yml:230-233` — both rationales are already stale.** `OnymPush` *is* a target dependency (`project.yml:245`, `OnymIOSTests`), so "No target dependency yet" is wrong, and the unit-test step already compiles the package — "nothing else compiles the package on its own" isn't true, making the standalone build step redundant. Keep the step if you want the package to build without the app project, but say that instead.
- Nit, `PushPreferenceStore.swift:9` — this stores raw APNs tokens in plaintext `UserDefaults` (so, in unencrypted backups), which sits oddly beside `PushTokenEnvelope`'s "never sits readable" framing. `UserDefaultsCaseSubmissionStore` carries an explicit why-plaintext-is-fine note (`CaseSubmissionStore.swift:84`); this type should too.
- Nit, `PushRegistrationInteractor.swift:67` — `setOnReconcileFailure` is public production API labelled test-only. Worth noting in the doc comment that a caller could use it to build exactly the activity log the package declines to keep.

Payload/envelope/client layers check out: length-prefixed domain-separated bytes, `.push` formatter matches `JSONEncoder.iso8601` (whole-second UTC) so signed and transmitted timestamps agree, synthesized `encodeIfPresent` really does omit a nil `deviceToken`, and the https gate mirrors `URLSessionEnforcementBackendClient.swift:91`.

dev-onym (commented):
COMMENT — the concurrency design holds up (I traced the disable-mid-register, rotation, and queued-pass paths; the single-slot + persisted-debt scheme is sound), but two rationale comments contradict the diff and one public hook undercuts the privacy stance.

**Fix before merge**

- `.github/workflows/release.yml:230-232` — the comment is false on both counts: this PR adds no OnymPush suites to `OnymIOSTests` (`grep -rn OnymPush OnymIOSTests` is empty; they arrive in the tests PR), and `project.yml:246` adds the package to the **test** target, not the app target — the app target's package list (`project.yml:55-86`) has no `OnymPush`. So "already compiles the package — as a dependency of the app target" names a build that doesn't exist. Same claim repeated at `project.yml:42-44`. The standalone build step itself is right and is currently the *only* thing compiling this package in CI — say that.
- `Packages/OnymPush/Sources/OnymPush/PushRegistrationInteractor.swift:82` — the doc comment correctly identifies `setOnReconcileFailure` as a way for a caller to build the exact failure log the package declines to keep, then ships it `public` anyway with a review-gate comment as the only defence. Make it `internal`: `@testable import OnymPush` works from `OnymIOSTests` for a path-based SwiftPM dependency in Debug, so the seam survives and the foot-gun doesn't. Failing that, `#if DEBUG`.
- `PushRegistrationInteractor.swift:195-197` — "the user's last unhonored instruction" is wrong for the re-enable case: offline opt-out queues token T, user re-enables before connectivity returns, and the pass drains T *first*, then re-registers the same T. Two round trips, two single-use signatures, and a window where a failed register leaves the device unregistered — all to honour an instruction the user has already superseded. Suggest skipping pending entries where `pending == token && preferences.isEnabled` (the replace-all register supersedes them); the rotation drain at :210 still covers the case that matters.

**Worth a look**

- `PushRegistrationInteractor.swift:108` — `pushEnabled()` is a silent no-op unless the caller has already written `preferences.setEnabled(true)`; the interactor reads `isEnabled` (:199, :260) but never writes it, and nothing in the API says so. Similarly, the mid-flight disable re-check at :260 only wins if `setEnabled(false)` lands *before* `await pushDisabled()`. It converges either way (I traced it — the follow-up pass drains the pending entry), but the ordering deserves to be stated on `pushDisabled()` since PR 2 is the caller.
- `PushPreferenceStore.swift:86-99` — `addPendingUnregister`/`removePendingUnregister` are non-atomic read-modify-write on the array. Safe today because every call is interactor-isolated, but the store is `public` and a Settings-side writer would silently drop a debt. Worth a note on the type, or move the pending set behind the actor.
- `PushTokenEnvelope.swift:38` — `SealError` omits `Sendable`, unlike `PushClientError` and `PushRejection`; it crosses actor boundaries from `seal` at :210/:233.
- `URLSessionPushBackendClient.swift:135` — `malformedResponse("push backend response: \(error)")` interpolates a `DecodingError` that can carry response bytes. Harmless while errors are swallowed; it becomes the payload if anyone ever wires the hook above.

Payload/envelope byte layout, https pinning, and the degenerate-expiry cap all read correctly; I couldn't verify the `payload.rs` mirror from this repo.

PR #312, opened 2026-08-22
The app can now ask to be woken, if and only if the user says so

rinat-enikeev:
Second iOS PR, on top of #311. The OnymPush package gets wired into the app; nothing leaves the device until the user flips the Settings toggle.

- **`PushAppDelegate`** — the app's first `UIApplicationDelegate`, existing solely because APNs tokens only arrive through delegate callbacks. It relays them into a replaying multicast `AsyncStream` (replay yielded under the delivery lock, so a stale token can never land after a fresher one) and does nothing else.
- **`PushCoordinator`** — observes `identitiesStream` (tags via the hoisted `InboxTag`), `NostrRelaysRepository.snapshots` (relay set, nil-until-loaded so an empty list is a meaningful "stop watching everything"), and the token stream, feeding the package's reconciling interactor. Relay lists are truncated client-side against the deployed backend's caps (4 per tag, 8 hosts, 4 per host), default relay first. Attestation adapts the same `DCDevice` provider moderation uses. Foreground re-reconciliation hooks the existing `scenePhase` block beside the moderation gate.
- **`DevicePushSigner`** — push registrations are signed by a device-local Ed25519 key minted on first use and kept in the Keychain, used for push and nothing else. The backend verifies and discards the key, so any key authenticates; a per-install random key carries zero identity linkage and never rotates with persona switches — identity keys stay out of the push payload entirely.
- **Settings → Notifications** — optional-factory pattern like moderation/discovery, so harness builds show no row. Toggle default OFF; enabling runs system authorization → APNs registration → backend register, and a denial snaps the switch back and points at system Settings. The copy states the actual trade (transport-honesty style): what leaves the device is inbox codes, an encrypted push token, a per-device signing key linked to no identity, and a device-attestation token; Apple carries a content-free "New message"; nobody sees content, senders, or groups. `NotificationsSettingsFlow` is closure-injected, keeping OnymSettings free of any OnymPush dependency.
- **`InboxTag` hoist** — the same eight-byte SHA-256 formula existed in six private copies (IdentityRepository, SendMessageInteractor, InboxFanoutInteractor, IntroInboxPump, CreateGroupInteractor, ChatReceiptSender); all now delegate to one `OnymFoundation.InboxTag.derive(from:)`. Push registration is the seventh caller and did not want to be the seventh copy.
- **project.yml** — OnymPush package (app + unit-test targets), `aps-environment: development` in the generated entitlements (inert caveat under automatic signing). No `UIBackgroundModes`: pushes are plain alerts.

### DIFF
--- a/OnymIOS.entitlements
+++ b/OnymIOS.entitlements
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
 	<key>com.apple.developer.associated-domains</key>
 	<array>
 		<string>applinks:onym.app</string>
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatReceiptSender.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatReceiptSender.swift
@@ -1,5 +1,6 @@
 import CryptoKit
 import Foundation
+import OnymFoundation
 import OnymTransport
 import OnymIdentity
 
@@ -69,12 +70,8 @@ public struct ChatReceiptSender: ChatReceiptSending {
         return (try? await inboxTransport.send(sealed, to: tag)) != nil
     }
 
-    /// Same derivation as `SendMessageInteractor` / `IntroInboxPump`.
     static func inboxTag(from inboxPublicKey: Data) -> String {
-        var hasher = SHA256()
-        hasher.update(data: Data("sep-inbox-v1".utf8))
-        hasher.update(data: inboxPublicKey)
-        return hasher.finalize().prefix(8).map { String(format: "%02x", $0) }.joined()
+        InboxTag.derive(from: inboxPublicKey)
     }
 }
 
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatReceiptSender.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/ChatReceiptSender.swift
@@ -1,4 +1,3 @@
-import CryptoKit
 import Foundation
 import OnymFoundation
 import OnymTransport
--- a/Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift
+++ b/Packages/OnymChatsCore/Sources/OnymChatsCore/SendMessageInteractor.swift
@@ -1,6 +1,7 @@
 import Foundation
 import CryptoKit
 import OSLog
+import OnymFoundation
 import OnymTransport
 import OnymIdentity
 import OnymTransportBlossom
@@ -1199,16 +1200,8 @@ public actor SendMessageInteractor {
         }
     }
 
-    /// Same derivation as `IdentityRepository.inboxTag(from:)` —
-    /// duplicated here because the repo's helper is private and we
-    /// only need the formula, not the keychain lookup. Matches
-    /// `CreateGroupInteractor.inboxTag(from:)`.
     private static func inboxTag(from inboxPublicKey: Data) -> String {
-        var hasher = SHA256()
-        hasher.update(data: Data("sep-inbox-v1".utf8))
-        hasher.update(data: inboxPublicKey)
-        let hash = hasher.finalize()
-        return hash.prefix(8).map { String(format: "%02x", $0) }.joined()
+        InboxTag.derive(from: inboxPublicKey)
     }
 }
 
--- a/Packages/OnymFoundation/Sources/OnymFoundation/InboxTag.swift
+++ b/Packages/OnymFoundation/Sources/OnymFoundation/InboxTag.swift
@@ -1,0 +1,19 @@
+import CryptoKit
+import Foundation
+
+/// The inbox routing tag: first 8 bytes, lowercase hex, of
+/// `SHA256("sep-inbox-v1" || inboxPublicKey)`.
+///
+/// This is the one formula behind every `inboxTag(from:)` helper —
+/// hoisted here so senders, subscribers, the identity repository, and
+/// the push registration all derive the same tag from the same bytes.
+/// A pure function of the public key: the tag is a *public* routing
+/// handle, already visible to every relay carrying the inbox.
+public enum InboxTag {
+    public static func derive(from inboxPublicKey: Data) -> String {
+        var hasher = SHA256()
+        hasher.update(data: Data("sep-inbox-v1".utf8))
+        hasher.update(data: inboxPublicKey)
+        return hasher.finalize().prefix(8).map { String(format: "%02x", $0) }.joined()
+    }
+}
--- a/Packages/OnymGroup/Sources/OnymGroup/CreateGroupInteractor.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/CreateGroupInteractor.swift
@@ -858,15 +858,8 @@ public struct CreateGroupInteractor: Sendable {
 
     // MARK: - Helpers
 
-    /// Same derivation as `IdentityRepository.inboxTag(from:)` —
-    /// duplicated here because the repo's helper is private and we
-    /// only need the formula, not the keychain lookup.
     private static func inboxTag(from inboxPublicKey: Data) -> String {
-        var hasher = SHA256()
-        hasher.update(data: Data("sep-inbox-v1".utf8))
-        hasher.update(data: inboxPublicKey)
-        let hash = hasher.finalize()
-        return hash.prefix(8).map { String(format: "%02x", $0) }.joined()
+        InboxTag.derive(from: inboxPublicKey)
     }
 
     private static func randomBytes(_ count: Int) throws -> Data {
--- a/Packages/OnymGroup/Sources/OnymGroup/IntroInboxPump.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/IntroInboxPump.swift
@@ -1,5 +1,6 @@
 import CryptoKit
 import Foundation
+import OnymFoundation
 import OnymTransport
 
 /// Sender-side pump for intro inbox subscriptions. Mirrors
@@ -54,15 +55,8 @@ public struct IntroInboxPump: Sendable {
         await subs.applyEmpty()
     }
 
-    /// Mirror of `IdentityRepository.inboxTag(from:)` /
-    /// `InboxFanoutInteractor.inboxTag(from:)`. Pure function of
-    /// the pubkey; safe to recompute.
     public static func inboxTag(from publicKey: Data) -> String {
-        var hasher = SHA256()
-        hasher.update(data: Data("sep-inbox-v1".utf8))
-        hasher.update(data: publicKey)
-        let digest = hasher.finalize()
-        return digest.prefix(8).map { String(format: "%02x", $0) }.joined()
+        InboxTag.derive(from: publicKey)
     }
 }
 
--- a/Packages/OnymGroup/Sources/OnymGroup/IntroInboxPump.swift
+++ b/Packages/OnymGroup/Sources/OnymGroup/IntroInboxPump.swift
@@ -1,4 +1,3 @@
-import CryptoKit
 import Foundation
 import OnymFoundation
 import OnymTransport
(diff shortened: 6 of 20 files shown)


dev-onym (commented):
**REQUEST_CHANGES** — the wiring is clean and the `InboxTag` hoist is correct, but three issues around what actually reaches the push server need addressing.

- `Sources/OnymIOS/PushCoordinator.swift:137` — one registration carries **every** identity's inbox tag, signed by one identity's key. That hands the push server the correlation set the multi-identity design exists to avoid: it learns all personas on this device are the same device. Nothing in the code or the Settings copy acknowledges it. Either register per-identity, or state it in the copy — this is the kind of trade this app documents rather than infers.
- `Sources/OnymIOS/PushCoordinator.swift:138` — the guard covers relays but not tags. `observeRelays()` (:117) fires `apply(relays:)` immediately while `latestTags` is still `[]`, so if `currentIdentities()` (:109 — keychain load + MainActor protected-data hop + possible fresh-install reconcile) exceeds the interactor's 2s debounce, the device registers with an **empty** subscription set, then re-registers when tags land. Empty is meaningful to the interactor ("clear every tag"). Make `latestTags` optional and guard on it.
- `Packages/OnymSettings/Sources/OnymSettings/NotificationsSettingsView.swift:44` says the server "forgets both when you turn this off", but `PushRegistrationInteractor.swift:74` returns without unregistering when `apnsToken == nil`. Launch offline (or with APNs registration failing), then disable: local prefs clear, the backend keeps token + inbox codes until the 60-day sweep. Persist the last token for unregister, or soften the copy.
- `Sources/OnymIOS/PushCoordinator.swift:102` — `appForegrounded()` only re-reconciles; nothing ever calls `getNotificationSettings()`. Revoke notifications in system Settings and the toggle still reads ON while the app keeps re-registering inbox codes for wakes that can never arrive. Check authorization on foreground and disable (or surface) instead.
- `Sources/OnymIOS/PushAppDelegate.swift:26` + `PushCoordinator.swift:82` — `enable()` returns true on authorization alone and the failure callback is a no-op, so the toggle reads ON when registration never happened, with no retry until the next foreground. Consider distinguishing "on" from "on and registered" in `NotificationsSettingsFlow`.
- `Sources/OnymIOS/PushAppDelegate.swift:10` — `pipe` is a `static let` and `deviceTokens` hands out the same stream. If that `.task` is ever cancelled, `onTermination` finishes the continuation permanently and no later token reaches the process. Safe today (it's on the `WindowGroup` root, `OnymIOSApp.swift:1876`) but fragile for a static one-shot.
- No tests. `NotificationsSettingsFlow`'s state machine (denied → snap back, `isWorking` reentrancy guard) is pure and trivially testable, and `project.yml:248` wires OnymPush into the unit-test target with nothing using it.
- `project.yml:184` — `aps-environment: development` is correct under automatic signing, but a manually-signed TestFlight/App Store export ships the development entitlement and gets tokens production APNs rejects. Worth a note next to the existing comment.
- Minor: `NotificationsSettingsFlow.swift:48` doesn't clear `authorizationDenied` on the disable path, so the "blocked in system Settings" footnote can outlive its cause. `import CryptoKit` is now unused in `IntroInboxPump.swift:1` and `InboxFanoutInteractor.swift:1`.
- Minor: `.badge`/`.sound` are requested but no `UNUserNotificationCenter.delegate` is set anywhere, so a push arriving while the app is foregrounded displays nothing.

Verified good: all six former `inboxTag` copies were the identical `SHA256("sep-inbox-v1" ‖ pubkey).prefix(8)` formula, and every affected package already depends on and imports `OnymFoundation`; the `IdentityModerationSigner: PushSigner` conformance is in-module (`Sources/OnymIOS/IdentityModerationSigner.swift`), so no retroactive-conformance warning; both relay paths read the same `cached.endpoints`, so initial and streamed relay sets can't diverge.

dev-onym (commented):
**REQUEST_CHANGES** — the push wiring is careful and the `InboxTag` hoist is clean, but two paths let the app register against the user's expressed OS-level intent.

- **`Sources/OnymIOS/PushCoordinator.swift:86` — cold launch skips the revoke check.** `run()` calls `registerWithAPNs()` on any `preferences.isEnabled`, and the identity/relay observers immediately push subscriptions. The `.denied` reconciliation lives only in `appForegrounded()`, which is driven by `RootView.swift:171`'s `.onChange(of: scenePhase)` — SwiftUI does **not** fire `onChange` for the initial `.active`. So a user who revoked notifications in system Settings has every identity's inbox tag re-registered with the push server on each cold launch, and it stays that way until they background/foreground once. Hoist the `notificationSettings()` `.denied` check into `run()` before `registerWithAPNs()`.
- **`Packages/OnymSettings/Sources/OnymSettings/NotificationsSettingsFlow.swift:69` — the toggle can read ON after the coordinator turned it off.** `refreshRegistrationState()` re-reads only `isRegistrationPending()`; `isEnabled` is mutated solely by `setEnabled`, and `NotificationsSettingsView.swift:63`'s `.onAppear` never resyncs `switchState`. The exact scenario the coordinator handles — leave the app, revoke in system Settings, come back to a still-mounted Notifications screen — leaves the switch claiming ON over a disabled registration. Add an `isEnabled` re-read closure alongside `isRegistrationPending` and resync `switchState` on appear.
- **`Sources/OnymIOS/PushAppDelegate.swift:26` — continuations are never removed.** No `onTermination`, so every subscriber stays in the static array for the process lifetime and its unbounded buffer keeps accumulating tokens after the consumer is gone. Both comparable streams in this repo do clean up (`IdentityRepository.swift:842`, `NostrRelaysRepository.swift:184`). One subscriber today, but a re-created root `.task` (view identity change) leaks one per re-subscription. The header comment argues for replay, which `latestToken` already provides — that doesn't require retaining dead continuations.
- **No tests for any of the new behavior.** `NotificationsSettingsFlow` is pure, `@MainActor`, closure-injected — the denial snap-back, the `isWorking` re-entrancy guard, and the pending state are all cheap to cover, but `Packages/OnymSettings` has no test target at all. `project.yml:251` adds `OnymPush` to the unit-test target while `Packages/OnymPush/Tests/` doesn't exist. At minimum, cover the coordinator's `.denied` → `disable()` path, since that's the privacy-load-bearing branch.
- Minor: `NotificationsSettingsView.swift:44` — "Activating… This finishes on its own" is shown whenever `fingerprint == nil`, including after a permanently failing register (`PushRegistrationInteractor.swift:188` swallows the error). Nothing retries until the next foreground, so the copy over-promises in exactly the case the user would check back on.
- Minor: `PushCoordinator.disable()` leaves `PushAppDelegate.latestToken` set, so a later re-subscribe replays a token for a device that unregistered. Harmless today (`reconcilePass` gates on `preferences.isEnabled`), but it's a stale secret held longer than needed.
- Non-blocking: the `aps-environment: development` caveat at `project.yml:181` is honestly documented, but a comment won't catch it — worth a CI assertion on the archived entitlement rather than trusting that signing stays automatic.
- Good: the `InboxTag` hoist is complete (remaining `sep-inbox-v1` hits are doc comments only), and the existing tests in `Tests/OnymIOSTests/` re-derive the formula independently, so they genuinely guard the refactor.

dev-onym (commented):
**REQUEST_CHANGES**

- `Packages/OnymSettings/Sources/OnymSettings/NotificationsSettingsView.swift:38` — the "How it works" copy claims the push server "holds only your inbox codes and an encrypted push token." The register payload (`PushRegistrationInteractor.swift:221-238`) also sends `userKey` — the current identity's Stellar public key, in the clear — plus a DeviceCheck attestation token and a signature over all of it. In a transport-honesty screen, "holds only" is a factual overclaim; the copy should name the identity key and attestation token, or the payload should stop carrying them.
- `Sources/OnymIOS/PushCoordinator.swift:67` + `IdentityModerationSigner.swift:12-22` — `userKeyID()`/`sign` resolve to whichever identity is *current*. Registrations refresh every 7 days, so successive registers for the same device token can be signed by different persona keys, letting the backend link persona keys to each other — a strictly larger correlation than the "all inbox codes belong to one device" trade documented at `PushRegistrationInteractor.swift:15-20` and in Settings. Either pin a single stable signing key for push or document/close this.
- `Sources/OnymIOS/PushAppDelegate.swift:31-40` — token ordering race: `replay` is read under the lock but yielded after unlocking, so a token APNs delivers in that window is yielded first and the older replayed token lands second. `updateToken` last-write-wins, so the interactor registers a stale token and wakes stop arriving. Yield the replay while holding the lock (or drop the replay if `latestToken` changed).
- `Sources/OnymIOS/PushCoordinator.swift:184` — `guard let latestTags, !latestRelays.isEmpty` silently drops relay-list changes to empty, so removing every relay leaves the backend watching the old relays indefinitely. Asymmetric with the deliberately-handled empty-tags case documented at :44-47; either propagate the empty relay set or say why it can't be.
- Test coverage: `Packages/OnymPush` has no test target contents at all, and nothing exercises `PushCoordinator` or `NotificationsSettingsFlow` — despite `notificationAuthorizationDenied` being injected at `PushCoordinator.swift:53-56` expressly "so the privacy-load-bearing … gate is unit-testable." The sibling pattern has `Tests/OnymIOSTests/BackupOperatorSettingsFlowTests.swift`; the revoke→disable gate, the denied-authorization snapback, and `registrationPending` are the cases to pin.
- `Packages/OnymFoundation/Sources/OnymFoundation/InboxTag.swift:12` — the hoist itself is clean and the six call sites match byte-for-byte, but the formula now has seven callers and no golden-vector test. One `#expect(InboxTag.derive(from: <fixed key>) == "<hex>")` guards every inbox route against a future edit here.
- `Packages/OnymChatsCore/Sources/OnymChatsCore/ChatReceiptSender.swift:1` — `import CryptoKit` is now unused (the two sibling files correctly dropped theirs).
- `project.yml:181-187` — the `aps-environment: development` caveat in the comment is real but inert here: the app target is `CODE_SIGN_STYLE: Automatic` (`project.yml:162`), so export rewrites it. Fine as-is; the comment slightly over-warns.

dev-onym (commented):
**COMMENT** — the privacy posture holds up (nothing reaches the backend until the opt-in: `reconcilePass` guards on `preferences.isEnabled`), the token-stream ordering claim checks out, and the InboxTag hoist is faithful; findings are UX truthfulness, robustness, and a total absence of tests.

- **No tests for any of it.** `Tests/OnymIOSTests` (136 files) gains nothing, yet this PR adds exactly the logic that wants unit tests: `PushCoordinator.cappedRelays` ordering/caps (`Sources/OnymIOS/PushCoordinator.swift:228`), the revoked-authorization disable gate that was made injectable *specifically* so it'd be testable (`PushCoordinator.swift:35`), the `NotificationsSettingsFlow` state machine, and the `DevicePushSigner` duplicate-item race. `DevicePushSigner(testNamespace:)` and `wipe()` exist for tests that don't exist (`Sources/OnymIOS/DevicePushSigner.swift:35,55`). The PR body's "app + unit-test targets" is also stale — `project.yml:255` already listed OnymPush from #311.
- **Revoked authorization flips the switch off with no explanation.** `appForegrounded()` → coordinator `disable()` → `refreshRegistrationState()` sets `isEnabled = false` but never `authorizationDenied` (`NotificationsSettingsFlow.swift:88-93`), so the user returning from system Settings — the trip this screen's own footnote sends them on — sees the toggle silently snap off and gets no "blocked in system Settings" note. The state exists; only the `enable()`-returned-false path sets it. Have `reconcileWithSystem`/`readIsEnabled` report *why*.
- **"Activating…" never clears while the screen stays open.** `setEnabled` calls `refreshRegistrationState()` immediately after `enable()` returns, and `enable()` only *schedules* the 2s-debounced register — so the fingerprint is always nil there and the pending footnote shows unconditionally, then goes stale for as long as the view is mounted (`NotificationsSettingsFlow.swift:82`, `NotificationsSettingsView.swift:44`). One poll/await after the debounce, or a coordinator-side signal, and the copy stops claiming unconfirmed after it's confirmed.
- **The system gate only catches `.denied`, not `.notDetermined`** (`PushCoordinator.swift:61-64`). Authorization reset out from under a persisted `push.enabled = true` leaves the app registering with APNs and the backend while no alert can ever display. `== .authorized` is the condition that matches the intent.
- **`.badge` is requested but never used** (`PushCoordinator.swift:124`): nothing sets a badge and `willPresent` returns `[.banner, .sound]`. Asking for a permission the app doesn't exercise sits oddly beside this PR's copy-honesty standard.
- **`cappedRelays` dedups and prioritizes on raw strings** (`PushCoordinator.swift:232,240`): a configured `wss://nostr.onym.app/` is neither recognized as the exempt default nor deduped against the bare form, so it can burn one of only four slots. Normalize before comparing.
- Nit: `Packages/OnymGroup/Sources/OnymGroup/CreateGroupInteractor.swift:1` — `import CryptoKit` is now dead (same cleanup as commit 30b90c7).
- Nit: `identitiesStream` (`IdentityRepository.swift:838`) and `snapshots` (`NostrRelaysRepository.swift:179`) both replay on subscribe, so the `currentIdentities()` / `currentEndpoints()` pre-reads at `PushCoordinator.swift:161,169` are redundant work, not a race fix.

PR #313, opened 2026-08-22
Tests for the bytes both sides must agree on, and the silences between

rinat-enikeev:
Final iOS PR of the push stack, on top of #312. 58 tests across the six push classes in OnymIOSTests, plus 2 InboxTag golden vectors in OnymFoundationTests — all passing on the iPhone 17 simulator.

- **`PushRegistrationPayloadTests`** — layout, domain separation, digest order-sensitivity/injectivity, absent-vs-empty token agreement, and the **cross-implementation fixtures**: hex produced by running this Swift implementation over pinned inputs. The backend's `payload.rs` (onymchat/onym-push#3) pins the identical strings, so byte drift on either side fails a build before it refuses every real signature in production.
- **`PushTokenEnvelopeTests`** — round-trip with the scheme (X25519 → HKDF `onym-push-token-v1` → AES-256-GCM) spelled out independently of the implementation under test, wrong-key refusal, malformed-key refusal, fresh randomness per seal.
- **`PushBackendClientTests`** — StubURLProtocol wire tests: camelCase/base64/RFC-3339 bodies, `{error, message}` refusals surfaced verbatim plus the `http_<status>` fallback for non-envelope bodies, fractional-seconds expiry decode, https-only guard including the empty-host case, registration-key decode.
- **`PushRegistrationInteractorTests`** — reconciliation where the assertion is as often "no call was made" as "one was": disabled ⇒ silence, unchanged fingerprint ⇒ silence, changed subscriptions ⇒ one replace-all call, failure ⇒ retried on next trigger, a 429 `capacity` refusal fully retryable, disable ⇒ unregister with the known token and no register queued after it, the set-of-pending-unregisters draining every debt (rotation + disable), degenerate expiry windows not churning, and the disable-vs-in-flight-register race ending unregistered.
- **`PushCoordinatorTests`** — the privacy-load-bearing authorization gate (cold-launch revoke ⇒ full disable, foreground revoke ⇒ full disable, granted ⇒ untouched, already-off ⇒ no-op) and the deterministic relay-cap truncation against the backend's per-request caps.
- **`NotificationsSettingsFlowTests`** — denied-authorization snapback, reentrant flips dropped, stale denial notes cleared, `registrationPending` following the probe, off-screen disable resync.
- **`InboxTagTests`** (OnymFoundationTests) — golden vector for the hoisted formula behind all seven inbox-tag call sites, plus a pin on the `sep-inbox-v1` domain-separation prefix itself.
- Shared test support: a lock-backed `RecorderBox` for values crossing from `@Sendable` handlers, and one `OneShotGate` replacing the two identical private gates.

### DIFF
--- a/Packages/OnymFoundation/Tests/OnymFoundationTests/InboxTagTests.swift
+++ b/Packages/OnymFoundation/Tests/OnymFoundationTests/InboxTagTests.swift
@@ -1,0 +1,24 @@
+import Foundation
+import XCTest
+@testable import OnymFoundation
+
+/// Golden vector for the one formula behind every inbox route. Seven
+/// call sites (senders, subscribers, the identity repository, push
+/// registration) derive tags through `InboxTag.derive`; this pin is
+/// what stands between them and a silent formula edit that would
+/// re-route every inbox at once. The hex was computed by an
+/// independent oracle (Python hashlib over
+/// `SHA256("sep-inbox-v1" || key)[..8]`) — regenerate it there, never
+/// by re-running the Swift under test.
+final class InboxTagTests: XCTestCase {
+    func testDeriveMatchesThePinnedVector() {
+        let key = Data(0..<32) // 000102…1f
+        XCTAssertEqual(InboxTag.derive(from: key), "fdee1b12aa915aa3")
+    }
+
+    func testDeriveOfEmptyKeyIsStillDomainSeparated() {
+        // SHA256("sep-inbox-v1") — the prefix alone. Pinned so the
+        // domain-separation string itself cannot drift unnoticed.
+        XCTAssertEqual(InboxTag.derive(from: Data()), "064cb2d524b4d3fe")
+    }
+}
--- a/Packages/OnymPush/Sources/OnymPush/StubPushBackendClient.swift
+++ b/Packages/OnymPush/Sources/OnymPush/StubPushBackendClient.swift
@@ -23,6 +23,10 @@ public actor StubPushBackendClient: PushBackendClient {
     public var registrationKeyAnswer: Data
     public var registerAnswer: Result<PushRegisterResponse, Error>
     public var unregisterAnswer: Result<Void, Error>
+    /// Awaited inside `register` before the request is recorded — a
+    /// test can hold a register "in flight" (e.g. behind a gate it
+    /// controls) while it interleaves a disable, then let it land.
+    public var onRegister: (@Sendable () async -> Void)?
 
     public init(
         registrationKeyAnswer: Data = StubPushBackendClient.defaultRegistrationKey,
@@ -40,6 +44,9 @@ public actor StubPushBackendClient: PushBackendClient {
     }
 
     public func register(_ request: PushRegisterRequest) async throws -> PushRegisterResponse {
+        if let onRegister {
+            await onRegister()
+        }
         registered.append(request)
         return try registerAnswer.get()
     }
@@ -61,7 +68,7 @@ public actor StubPushBackendClient: PushBackendClient {
         registrationKeyAnswer = key
     }
 
-    public func setUnregisterAnswer(_ answer: Result<Void, Error>) {
-        unregisterAnswer = answer
+    public func setOnRegister(_ hook: (@Sendable () async -> Void)?) {
+        onRegister = hook
     }
 }
--- a/Packages/OnymPush/Sources/OnymPush/StubPushBackendClient.swift
+++ b/Packages/OnymPush/Sources/OnymPush/StubPushBackendClient.swift
@@ -53,6 +53,10 @@ public actor StubPushBackendClient: PushBackendClient {
         registerAnswer = answer
     }
 
+    public func setUnregisterAnswer(_ answer: Result<Void, Error>) {
+        unregisterAnswer = answer
+    }
+
     public func setRegistrationKeyAnswer(_ key: Data) {
         registrationKeyAnswer = key
     }
--- a/Tests/OnymIOSTests/NotificationsSettingsFlowTests.swift
+++ b/Tests/OnymIOSTests/NotificationsSettingsFlowTests.swift
@@ -1,0 +1,140 @@
+import XCTest
+@testable import OnymSettings
+
+/// The Notifications screen's state machine: the toggle must never
+/// claim more than what actually happened — authorization denied snaps
+/// back, overlapping flips don't double-run, and "on" is distinguished
+/// from "accepted by the backend".
+@MainActor
+final class NotificationsSettingsFlowTests: XCTestCase {
+    func testDeniedAuthorizationSnapsBackAndSaysWhy() async {
+        let flow = NotificationsSettingsFlow(
+            isEnabled: false,
+            enable: { false },
+            disable: {}
+        )
+
+        await flow.setEnabled(true)
+
+        XCTAssertFalse(flow.isEnabled)
+        XCTAssertTrue(flow.authorizationDenied)
+        XCTAssertFalse(flow.isWorking)
+    }
+
+    func testGrantedAuthorizationEnables() async {
+        let flow = NotificationsSettingsFlow(
+            isEnabled: false,
+            enable: { true },
+            disable: {}
+        )
+
+        await flow.setEnabled(true)
+
+        XCTAssertTrue(flow.isEnabled)
+        XCTAssertFalse(flow.authorizationDenied)
+    }
+
+    /// A flip while the previous flip is still working is dropped —
+    /// two overlapping enables would prompt twice and register twice.
+    func testReentrantSetEnabledIsDropped() async {
+        let enableCalls = Counter()
+        let gate = AsyncGate()
+        let flow = NotificationsSettingsFlow(
+            isEnabled: false,
+            enable: {
+                await enableCalls.increment()
+                await gate.wait()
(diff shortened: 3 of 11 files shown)


dev-onym (commented):
**COMMENT** — fixtures verified correct; findings are coverage gaps, one vacuous assertion, and one overclaimed property.

I independently recomputed all four fixtures (length-prefix layout, both contexts, SHA-256 digest) from `PushRegistrationPayload.swift` in Python: all three payload hexes and the digest match byte-for-byte. The cross-implementation pinning is sound.

- `Tests/OnymIOSTests/PushRegistrationPayloadTests.swift:55` — the digest is **not** injective over subscription lists, so the test name and the "no two different field splits collide" comment (`:61`) overclaim. `digest` flattens to `(tag, relay)` pairs with no per-subscription boundary, so `[(t,[r1,r2])]` and `[(t,[r1]),(t,[r2])]` produce the identical digest `4c17a007…` (verified). Harmless today — the resulting watch set is the same — but it also means `PushRegistrationInteractor.fingerprint` (`PushRegistrationInteractor.swift:166`) treats a regrouping as "unchanged". Either prefix the subscription count/boundary on both sides, or narrow the name and document the collision class as intentional.
- `Tests/OnymIOSTests/PushRegistrationInteractorTests.swift:160` — vacuous: `preferences.setEnabled(false)` calls `clearRegistration()` itself (`PushPreferenceStore.swift:29`), so the fingerprint is already nil before `pushDisabled()` runs. The assertion passes even if `pushDisabled` stopped clearing. Clear the flag *after* asserting, or record a fingerprint and call `pushDisabled()` without flipping the preference.
- `Tests/OnymIOSTests/PushBackendClientTests.swift:29` — the register-body test uses `deviceToken: nil`, and synthesized `Codable` **omits** the key rather than emitting `null`. That's exactly the absent-token contract the payload tests pin on the signing side, but the wire side is unasserted. Add `XCTAssertNil(body?["deviceToken"])` plus a non-nil case, since serde on the Rust end needs the missing key to be accepted.
- `PushBackendClientTests.swift` covers `register` and `registration-key` only — `unregister` has no wire test, so `PushUnregisterRequest`'s shape, its `/v1/unregister` path, and the empty-200-body path through `send` are unvalidated.
- `PushRegistrationInteractorTests.swift:58` — nine 300 ms real sleeps (~3 s) is load-flaky on CI. Since `debounce` is already injected, consider polling for the expected call count with a deadline instead of a fixed sleep.
- Untested interactor branches: the `refreshInterval` re-register (`PushRegistrationInteractor.swift:102-106`) — the "stale sweep never reaps a live device" guarantee, reachable by injecting `refreshInterval: 0` or backdating via `recordRegistration(fingerprint:at:)`; `updateSubscriptions([])`, documented as meaningful (clears tags, stays registered); and `pushEnabled()`.
- `PushRegistrationPayloadTests.swift:111` — `testRegisterWithoutTokenFixtureMatchesTheBackend` duplicates `testAbsentAndEmptyDeviceTokenAgree`'s input; consider pinning the *empty-Data* variant here instead so both spellings are fixture-covered.
- `StubPushBackendClient.swift:18` — the fix is right (all-zero is a low-order point CryptoKit rejects), but generating a key in the default argument makes the stub's output non-deterministic per instance. A pinned valid 32-byte public key constant would keep stub behavior reproducible; tests that need to open the envelope set their own key regardless. Separately, the zero-key bug stayed invisible because `reconcile()`'s `catch` (`PushRegistrationInteractor.swift:133-137`) swallows everything — worth a test-only failure hook so silent-retry-forever is observable.

dev-onym (commented):
COMMENT — solid, high-signal tests; I independently recomputed every payload/digest fixture and all four match byte-for-byte. Notes below are test-quality, not production defects.

- `Tests/OnymIOSTests/PushCoordinatorTests.swift:67` — `defer { scene.cancel() }` doesn't await the task. `run()`'s task group is still observing `identitiesStream`/`relaysStream`/`PushAppDelegate.deviceTokens` when `tearDown` hits `keychain.wipeAll()` (:26) and removes the defaults domain. Cancellation isn't synchronous, so the orphaned reads race the wipe — a plausible flake and cross-test bleed. Hoist the task to a property and `await scene.value` (or `_ = await scene.result`) in `tearDown`.
- `Tests/OnymIOSTests/PushRegistrationInteractorTests.swift:235` — "a fresh interactor, same persisted state" isn't: `makeInteractor` builds a new `StubSigner()` (:23) with a fresh random Curve25519 key, so `second` unregisters under a *different* `userKey` than `first` registered with. The test would still pass if the fallback token were wrongly derived from signer identity. Take a `signer` parameter so both interactors share one.
- `PushRegistrationInteractorTests.swift:187` — trailing `preferences.setEnabled(false)` after the last assertion is dead. Related: this test deliberately calls `pushDisabled()` with `isEnabled` still `true`, a state the app never produces (the UI flips the preference first); the comment explains why, but worth an added assertion that no re-register followed, since `reconcilePass` would happily register again in that state.
- `PushBackendClientTests.swift:51,82` — `NSMutableDictionary` is written from the URLProtocol thread inside a `@Sendable` handler and read from the test thread. The `await` gives you a happens-after in practice, but the target is Swift 5 mode so nothing warns; a lock-backed `Recorder` box (the same shape as `FailureBox` at `PushRegistrationInteractorTests.swift:483`) would be both safe and typed.
- Coverage gaps in `URLSessionPushBackendClient`, given this PR is the coverage: the `rejection` fallback (non-`{error,message}` body → synthetic `http_<status>` code, `URLSessionPushBackendClient.swift:104`) is untested, as is the fractional-seconds decoder (`:114`) — the sole reason for the dual-formatter — and the `!host.isEmpty` half of the transport guard (`:68`, e.g. `https:///`). Three cheap tests.
- `PushBackendClientTests.swift:27` — `makeRegisterRequest` is `throws` but cannot throw; drop it and the `try` at four call sites.
- `NotificationsSettingsFlowTests.swift:145` / `PushRegistrationInteractorTests.swift:499` — `AsyncGate` and `TestGate` are the same one-shot gate written twice. `Tests/OnymIOSTests/Support/` already exists for exactly this.
- `StubPushBackendClient.swift:63-70` — the `setUnregisterAnswer` move is pure reordering churn in the diff; and `onRegister` is a `public var` *and* has a setter (matching the existing answer vars, so consistent, but the setter is redundant from outside the actor).
- PR body claims the 32-zero-byte stub-key fix is in this PR; it's already on `main` from #312 (`StubPushBackendClient.swift:8-18`). Body also says 22 tests; the diff adds ~34.

dev-onym (commented):
**COMMENT** — solid, high-value tests; the fixtures check out independently, but several assertions race the interactor's own bookkeeping.

- Verified independently (Python oracle, not the Swift under test): all four payload/digest fixtures (`PushRegistrationPayloadTests.swift:120,135,148,157,171`) and both InboxTag vectors (`InboxTagTests.swift:15,22`) reproduce exactly. Good cross-implementation pins.
- `Tests/OnymIOSTests/PushRegistrationInteractorTests.swift:173` — `waitUntil { registered.count == 1 }` is satisfied by the *stub's* append (`StubPushBackendClient.swift:50`), which happens before `client.register` returns and before `recordRegistration` runs (`PushRegistrationInteractor.swift:276`). So `XCTAssertNotNil(preferences.lastRegistrationFingerprint)` here can observe pre-record state. Same shape at `:243` (`lastRegisteredToken`) and `:395`. Fix: poll the preference itself, as `testRegisterRecordsExpiry` (`:290`) correctly does.
- `PushRegistrationInteractorTests.swift:181` and `:251` — `pushDisabled()` returns *without* draining when the pass slot is still held (`PushRegistrationInteractor.swift:146` → queue and return), and the preceding `waitUntil` can return while the register pass is still winding down. The immediate `XCTAssertEqual(unregistered.count, 1)` is therefore a timing window; `:214` and `:229` have the same shape. Wrap in `waitUntil`. (The in-flight test at `:556` gets this right.)
- `PushCoordinatorTests.swift:51` — `IdentityRepository` is built with the default `installMarker: .userDefaults`, so `reconcileFreshInstall` (`IdentityRepository.swift:742`) writes `app.onym.ios.identity.installMarker` into *standard* defaults, which `tearDown` (`:29`, which removes only the per-test suite) never cleans. Every other identity suite passes `installMarker: .inMemory(...)` explicitly; match that.
- `PushCoordinatorTests.swift:29` — `tearDown() async throws` omits `try await super.tearDown()`; the six other async tearDowns in this target call it.
- `Packages/OnymPush/Sources/OnymPush/StubPushBackendClient.swift:29,71` — a test-only `onRegister` hook lands in a shipped module's public API without the "tests only" caveat that `setOnReconcileFailure` carries (`PushRegistrationInteractor.swift:88`); worth the same comment. The `setUnregisterAnswer` relocation at `:63` is pure diff churn.
- `PushBackendClientTests.swift:181` — `XCTAssertNil(body?["subscriptions"])` is vacuously true: `PushUnregisterRequest` has no such field, so it pins nothing.
- `PushRegistrationInteractorTests.swift:78` — `waitUntil` sleeps after its last check, so it can fail with the condition true at the deadline; check once more before `XCTFail`.
- `PushRegistrationPayloadTests.swift:191` adds a seventh private `hex(_:)` copy in this target — the same duplication `HexTestSupport.swift` was created to stop, just in the encode direction.

PR #316, opened 2026-08-24
A package that holds only the colors, so another may hold different ones

rinat-enikeev:
First of five. Extracts the token layer so an adopter can ship Onym in their own colors.

`OnymTokens`, `OnymAccent` and `Color.dynamic` move out of `OnymBrand.swift` into a new zero-dependency `OnymDesignTokens` package. `OnymDesign` depends on it and re-exports it, so **all 758 call sites across the eight UI packages and the app compile untouched**. Nothing visible changes.

What it buys is a seam. An adopter copies the package, keeps the module name, changes the values, repoints one dependency. Nothing else in the graph knows.

The completeness check is the compiler: a replacement that forgets a token has no protocol to fail to conform to, only 758 call sites that no longer resolve.

`forSender` stays behind in `OnymDesign` — mapping a pubkey onto the palette is policy, not a value, and adopters should inherit it rather than reimplement it.

The seam is a **build-time** choice, not a runtime one; that trade is documented in the package README.

### DIFF
--- a/Packages/OnymDesign/Package.swift
+++ b/Packages/OnymDesign/Package.swift
@@ -7,7 +7,16 @@ let package = Package(
     products: [
         .library(name: "OnymDesign", targets: ["OnymDesign"])
     ],
+    dependencies: [
+        // The swappable token layer. Adopters who want a differently
+        // styled build repoint this one dependency at their own module
+        // named `OnymDesignTokens` — see that package's README.
+        .package(path: "../OnymDesignTokens")
+    ],
     targets: [
-        .target(name: "OnymDesign")
+        .target(
+            name: "OnymDesign",
+            dependencies: [.product(name: "OnymDesignTokens", package: "OnymDesignTokens")]
+        )
     ]
 )
--- a/Packages/OnymDesign/Sources/OnymDesign/OnymBrand.swift
+++ b/Packages/OnymDesign/Sources/OnymDesign/OnymBrand.swift
@@ -1,97 +1,11 @@
 import SwiftUI
 import UIKit
 
-// MARK: - Tokens
+@_exported import OnymDesignTokens
 
-/// Theme-adaptive design tokens for the Create Group flow. Each color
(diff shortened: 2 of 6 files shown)


dev-onym (commented):
**APPROVE** — clean pure-extraction refactor; the seam works as advertised.

- `Packages/OnymDesign/Sources/OnymDesign/OnymBrand.swift:4` — `@_exported import OnymDesignTokens` is what preserves every downstream call site untouched. All 63 files that `import OnymDesign` transitively see `OnymTokens` / `OnymAccent` / `Color.dynamic`. Note `@_exported` is an underscored (unofficial) attribute — reliable in practice but not stable API; worth flagging since this seam depends on it. Not a blocker.
- `Packages/OnymDesignTokens/Package.swift:5` — platform `.iOS("18.0")` matches `OnymDesign` (Package.swift:6), so no floor mismatch. Zero external deps, matches the README's promise.
- `Packages/OnymDesign/Sources/OnymDesign/OnymBrand.swift:8-32` — `forSender` correctly moved from an enum-static to an extension method on the re-imported `OnymAccent`; behaviour and signature unchanged, so the 63 call sites remain source-compatible. Docstring at :20-22 correctly captures the policy-vs-value split.
- `Packages/OnymDesignTokens/Sources/OnymDesignTokens/OnymAccent.swift:32-38` — private `rgb(_:)` helper here duplicates the `hex(_:_:)` helper in `OnymTokens.swift:55`. Pre-existing duplication carried over verbatim; not worth blocking on but a candidate for a follow-up trim.
- README (`Packages/OnymDesignTokens/README.md:34-38`) accurately documents the swap procedure. The "completeness by 758 call sites" argument holds — a replacement omitting a token fails compilation, and there's no protocol/registration surface that could silently accept a partial theme.
- Verified no Xcode `project.pbxproj` needs updating (SwiftPM-only workspace); the eight UI packages and app pick up tokens through `OnymDesign`'s existing dependency edge.

PR #317, opened 2026-08-24
Corners and letterforms become things a theme can decide

rinat-enikeev:
Second of five. Base: #316.

`OnymRadius` names ten corner steps for what they wrap — card, field, panel, hero, inset, control, badge, tile, chip, pill. `OnymType` puts the text and mono faces behind two functions. **164 corner literals and 464 font calls** now read from tokens; afterwards only one literal remains, and it is a drawing.

### The scale grew, deliberately

The design system names five radii. The app was spending fifteen, and only 33% of sites matched a system value. Where the two disagreed the app was usually right — 12pt is its single most common corner and the system has no word for it. So the scale fits the building rather than the drawing of it, and the design project will be updated to match.

### What actually moved

Sixteen surfaces changed shape, none by more than 4pt. Two of them were the same idea drawn twice — a square brand tile at 26pt in About and 22pt in recovery, one job at two radii. They share the `hero` token now.

The crown bar in `OnymGovIcon` keeps its literal: it rounds at a fraction of a drawing, not a corner of a surface.

`OnymType` deliberately carries **no size scale**. The 31 sizes the app spends are an accretion, not a system, and collapsing them changes every screen — a design decision, not a packaging one. Worth noting separately: `.system(size:)` does not respond to Dynamic Type, so accessibility text sizing is effectively off across 465 call sites. That is its own fix.

### Undeclared dependencies

Tokens now reach past `OnymDesign`, so eight packages and the app target declare the token package directly. Seven had been compiling on a dependency they never named — visible through the re-export, linking only because the app pulls everything in. Built standalone, the way CI builds them, they would not have linked.

`OnymOnboarding` keeps its no-UI-package rule: it uses `OnymType` and `OnymRadius` only, so it takes the zero-dependency token package rather than `OnymDesign`.


dev-onym (commented):
APPROVE — mechanical, per-site sweep with small semantic care; nits below.

- `Packages/OnymDesignTokens/Sources/OnymDesignTokens/OnymRadius.swift:6` — doc says "Nine steps" but ten are declared (card, field, panel, hero, inset, control, badge, tile, chip, pill). README:17 correctly says ten.
- `Packages/OnymDesignTokens/Sources/OnymDesignTokens/OnymRadius.swift:44` — `pill = .infinity` is safe via `shape(_:)` but a footgun for anyone who passes it straight to `RoundedRectangle(cornerRadius:)`. No current call site does this, but the doc note is easy to miss. Consider omitting the constant and exposing only `shape(_:)` for pills.
- `Packages/OnymGroup/Sources/OnymGroup/QRCodeScannerView.swift:3` and `Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift:3` — import `OnymDesign` but each only touches `OnymType`/`OnymRadius`. Given the PR's premise about declared dependencies, both should `import OnymDesignTokens` (both packages already depend on it).
- `Packages/OnymModerationUI/Sources/OnymModerationUI/MarkdownDocumentView.swift:352` — the 3pt-wide quote bar was `cornerRadius: 2` and now uses `OnymRadius.chip` (4). `chip` is documented as "tiny uppercase status chip" — semantically wrong role for a divider stripe, and it's a 2pt shape change on top. Consider a literal here on the same grounds the crown bar keeps its literal.
- `Packages/OnymRecovery/Package.resolved` and `Packages/OnymSettings/Package.resolved` (new files) — no other package checks in `Package.resolved`. Either add all of them or gitignore these two.
- `Packages/OnymIdentityUI/Sources/OnymIdentityUI/IdentitiesView.swift:264` — "Remove identity" destructive primary button uses `OnymRadius.field`; `card` fits the role better. Same value (14) so no visual diff.
- Font/radius mappings on the sixteen changed surfaces line up with the PR body's "≤4pt" claim from spot-checks (`ShareKeyView:31` 22→18, `IdentityDetailView:162` 18 stays, `ShareInviteView:170` 20→18). Fixed-size Dynamic Type behavior is preserved, matching the doc's explicit "changed the seam, not the sizing policy."

PR #318, opened 2026-08-24
The colors that were not listening

rinat-enikeev:
Third of five. Base: #317.

Ninety percent of the app already drew from tokens. The other ten percent — **85 sites** — were literals, and a literal does not hear a theme. An adopter swapping the token package would have found the settings banner still cream, the console still GitHub-dark, the icon tiles still Apple's blue. Restyled in most places and stubbornly Onym in the rest, which reads worse than not restyling at all.

### Three families collect them

| Family | What it holds |
| --- | --- |
| `OnymTile` | The nine icon-tile hues — a token family that had been sitting in a component file |
| `OnymTerminal` | The console treatment for self-host / deploy / relayer screens. Pinned dark on purpose: a terminal that turns white in light mode stops looking like a terminal |
| `OnymTint` | The pastels and the deep inks that read on them — hero tiles, the amber notice, governance badges |

`OnymTokens` gains four: `onTile`, which unlike `onAccent` does **not** flip with the appearance because a saturated tile carries its own contrast; `lightbox`, black because a viewfinder is black and not because the theme is; and `scrim` in both SwiftUI and UIKit forms, for the chat thread that draws media in layers.

### What moved on screen

Two things, both deliberate. The recovery quiz called SwiftUI's `Color.green` to mean "correct" and now calls the app's own success green, a little deeper in light mode. The key tile's gradient reaches for the accent purple rather than the system one.

### Known gap, deliberately left

The tints are still pinned to their light values, so an amber notice is still cream on a black screen. That was true before and this does not change it — six screens changing in dark mode is a design decision, not a packaging one. It is now a decision that costs one file.

Outside the token package, **no view holds a color literal any more**. The one exception is `Color.accentColor`, which comes from the app's asset catalog and is already adopter-controlled through a different door.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013z1YSTEHK4oxHTE9zabVyf

PR #319, opened 2026-08-24
The app and the drawing agree on what to call things

rinat-enikeev:
Fourth of five. Base: #318.

Eleven components existed on both sides under two names. The design system said `Card`, `Row`, `Chip`, `Footnote`, `LargeTitle`, `IconTile`, `SectionLabel`, `StepIndicator`, `PrimaryButton`, `TextButton`; the app said the same words with `Settings` in front — a prefix that stopped being true the moment these atoms started rendering chat, identity, moderation and recovery screens.

**474 references** now use the shared word. This makes the design project's "maps 1:1 to the `OnymDesign` SwiftUI package" claim true for the components that exist on both sides.

Three keep their prefix because there is nothing on the other side to agree with: `SettingsContentTile`, `SettingsRowDivider`, `SettingsQRCode`.

### On bare `Row`

The one name worth checking. Four types in the repo already declare a nested `Row` — in `PendingChatsFlow`, the backup client, the intro request log, and `ChatsView` — and a nested name shadows a top-level one inside its own scope. None of those four files render a design-system row, so nothing is shadowed today. It is the kind of thing that will be true right up until it isn't, so it is called out here rather than discovered later.

### Still not aligned

Eleven design-system components have no `OnymDesign` counterpart at all — `Screen`, `ScreenContent`, `NavBar`, `TabDock`, `GlassButton`, `ChatBubble`, `ChatComposer`, `TextField`, `Toggle`, `Badge`, `ListItem`. They exist in the app, open-coded in feature packages. Promoting them restructures `OnymChatsUI` and is deliberately not in this stack.

Note that `TextField` and `Toggle` will collide with SwiftUI's own types if promoted unprefixed — worth deciding before that work starts.

### DIFF
--- a/Packages/OnymBackupUI/Sources/OnymBackupUI/BackupEnrolmentView.swift
+++ b/Packages/OnymBackupUI/Sources/OnymBackupUI/BackupEnrolmentView.swift
@@ -113,8 +113,8 @@ public struct BackupEnrolmentView: View {
                     symbol: "clock.arrow.circlepath",
                     identifier: "backup.enrolment.schedule")
 
-                SettingsSectionLabel("WHAT THE OPERATOR PROMISES")
-                SettingsCard {
+                SectionLabel("WHAT THE OPERATOR PROMISES")
+                Card {
                     ForEach(Array(disclosure.items.enumerated()), id: \.element.id) { index, item in
                         VStack(alignment: .leading, spacing: 4) {
                             // `verbatim` throughout: these are the
@@ -140,7 +140,7 @@ public struct BackupEnrolmentView: View {
                     }
                 }
 
-                SettingsFootnote(
+                Footnote(
                     "These terms are pinned to every backup you make under them. If the operator publishes different terms later, backups stop until you have seen them.")
 
             }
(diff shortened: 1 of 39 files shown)


dev-onym (commented):
APPROVE — mechanical, symmetric rename (474/474 lines) matches the body's claim; only design-system usages are renamed and comments referencing old names are updated.

- `Packages/OnymDesign/SettingsDesign.swift:16-566` — all 10 public struct declarations renamed cleanly; `Card`, `Row`, `PrimaryButton`, `SectionLabel` generic constraints and `where Right == EmptyView` extensions still resolve.
- `Packages/OnymGroup/…/HandledIntroRequestLog.swift:101`, `OnymBackup/…/URLSessionBackupClient.swift:242`, `OnymInbox/…/PendingChatsFlow.swift:80`, `OnymChatsUI/…/ChatsView.swift:427` — four nested `Row` types confirmed; PR body already calls this out and none of those files import a design-system Row today, so no shadowing.
- No stale `Settings{Card,Row,Chip,Footnote,LargeTitle,IconTile,SectionLabel,StepIndicator,PrimaryButton,TextButton}` references remain in `.swift`.
- `Packages/OnymDesign/EXTRACTION-NOTES.md:60-105` and `Packages/OnymIdentityUI/EXTRACTION-NOTES.md:38-41` still list the old names — historical planning notes, but worth a follow-up sweep if these are kept current.
- Retained-prefix components (`SettingsContentTile`, `SettingsRowDivider`, `SettingsQRCode`) still referenced at `IdentityDetailView.swift:148,181`, `PrivacyEncryptionView.swift:83,90`, etc.; no accidental rename.
- `PR 3 supplies StepIndicator` comment update in `OnymOnboarding/…/OnboardingStepScaffold.swift:9` — kept accurate.

PR #320, opened 2026-08-24
Tests for the palette a stranger might replace, and the floors it must clear

rinat-enikeev:
Last of five. Base: #319.

Twenty tests across the two packages, plus the adopter documentation.

The **contract suite** is what an adopter runs against their own copy before wiring it in: every token resolves in both appearances, the surface ramp does not collapse, no two accents come out the same colour, the radius scale stays ordered, `pill` really is a capsule. The compiler already refuses an incomplete module; this makes the refusal one readable line instead of several hundred.

### The contrast suite found two things

**`OnymTint.orangeInk` measured 4.00:1** on its own pastel and 3.85:1 on the 16% tint — failing WCAG AA on both, the same way `amber` did once and for the same reason. It is `#BE4300` now, which clears 4.5:1 on each. A small visible change on the contract-detail and anchors screens.

**White glyphs on four of the nine icon tiles sit below 3:1** — amber 2.20, teal 2.48, orange 2.60, green 2.69. That is Apple's Settings palette drawn the way Settings draws it, and every tile sits beside a `Row` whose title says the same thing in words, so nothing is knowable only from the tile. `testOnTileStaysLegibleOnEveryTile` holds a **regression floor and says plainly it is not a conformance claim**. If a glyph ever becomes the only cue for something, the floor goes up and the palette comes down to meet it. Flagging explicitly in case you want that decided differently.

### A test of mine that was wrong before the code was

It asserted two particular pubkeys resolve to different accents — which, with six colours, is a coin that lands wrong one time in six. It caught a coincidence, not a property. The question worth asking is whether *every byte* of the key feeds the mapping, so an impostor cannot shop for a prefix that lands on their target's colour. It asks that now.

### CI

Both suites run the way the other package suites do, and `report-to-issue` parses their results.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -228,6 +228,35 @@ jobs:
             -derivedDataPath /tmp/onym-foundation-tests \
             -resultBundlePath /tmp/onym-foundation.xcresult
 
+      - name: xcodebuild test (OnymDesignTokens package)
+        # The token contract and the contrast floors. These gate the
+        # palette against a resync from the design system quietly
+        # reintroducing a colour that fails WCAG AA — which is exactly
+        # how the amber token went wrong once already. Same `always()`
+        # rationale as the other package test steps.
+        if: always()
+        working-directory: Packages/OnymDesignTokens
+        run: |
+          rm -rf /tmp/onym-designtokens-tests /tmp/onym-designtokens.xcresult
+          xcodebuild test \
+            -scheme OnymDesignTokens \
+            -destination "platform=iOS Simulator,name=${{ steps.sim.outputs.name }},OS=latest" \
+            -derivedDataPath /tmp/onym-designtokens-tests \
+            -resultBundlePath /tmp/onym-designtokens.xcresult
+
+      - name: xcodebuild test (OnymDesign package)
+        # Sender-to-accent mapping: determinism across launches and
+        # devices, and sensitivity to every byte of the pubkey.
+        if: always()
+        working-directory: Packages/OnymDesign
+        run: |
+          rm -rf /tmp/onym-design-tests /tmp/onym-design.xcresult
+          xcodebuild test \
+            -scheme OnymDesign \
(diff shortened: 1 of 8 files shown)


dev-onym (commented):
APPROVE

- `SenderAccentTests.swift:32-45` — `testEveryByteOfTheKeyFeedsTheMapping` is the right shape (any-change-moves-color at every position), a real upgrade over the deleted coincidence-catcher; iterating `alice.count` positions of a 32-char hex string is fine since ASCII == grapheme here.
- `SenderAccentTests.swift:87-90` — `XCTAssertNoThrow(forSender(""))` only guards Swift throws, not traps. The function presumably isn't throwing, so this reads as intent-only. Minor; a plain call would say the same thing.
- `TokenContract.swift:105-115` — accent-distinctness compares `%.3f` string keys, so accents that differ only in the 4th decimal would falsely register as identical. Given the current palette's separations this can't hit, but it's a slightly wrong ruler for the claim.
- `TokenContract.swift:145-151` — `testPillResolvesToACapsule` correctly checks the shape fills its frame rather than just asserting `.infinity`; nice.
- `ContrastTests.swift:76-86` — the 2.1 floor barely clears amber's measured 2.20. Documented explicitly as a regression guard, not conformance — reasonable, but any warmer amber would flip this red. Consider whether 2.0 gives useful headroom without losing the guard.
- `OnymTint.swift:40-44` — `orangeInk` change to `#BE4300` is justified by the failing measurements called out in the PR body, and the accompanying README entry documents the divergence alongside the amber one. Consistent.
- `release.yml:231-259` — both new test steps mirror the existing package-test pattern (`if: always()`, per-package derived data, xcresult artifacts, `report-to-issue` targets wired to `OnymDesignTokensTests`/`OnymDesignTests`). Target names match the `.testTarget` names in the two Package.swift files.
- README additions (`Packages/OnymDesignTokens/README.md:60-76, 95-121`) — adopter-run instructions and the known-shortfall paragraph are exactly what makes the swap story honest. The "these divergences… the contrast tests will fail if a resync undoes one" line closes the loop the PR opens.

PR #322, opened 2026-08-24
Type that answers to the person reading it

rinat-enikeev:
First of four. Makes `OnymType.font` and `OnymType.mono` grow and shrink with the reader's text-size setting.

Because #317 put every font call behind these two functions, that is **two function bodies rather than 470 call sites** — the reason the token seam was worth building.

`OnymType.fixed` joins them, for SF Symbols that are drawings rather than words. Nothing uses it yet; the layout PR decides what does.

### Two implementations that compile, read plausibly, and are wrong

Both were found by spiking rather than reasoning, and either could have shipped.

**1. `UIFontMetrics.scaledFont(for:)` rounds to whole points — at the default setting.**

```
scaledFont: 11.5pt -> 12.0pt at default
scaledFont: 12.5pt -> 13.0pt at default
scaledFont: 13.5pt -> 14.0pt at default
```

The app spends ~60 half-point sizes. That route ships a visible change to every reader who never touched the setting, in the name of helping the ones who did. Taking the metric as a **ratio** and applying it here keeps 13.5pt at exactly 13.5pt and grows it to 38.0pt at AX5.

**2. `scaledValue(for:)` with no trait collection silently returns 1.0 at every setting.**

It does not consult `UITraitCollection.current`. Passed `.current` explicitly it tracks properly, and SwiftUI does set `.current` while evaluating a view body — verified against a hosted view, ×2.82 at AX5. The version without it looks like working Dynamic Type support and delivers none.

### Nothing moves on screen yet

The type scales; the boxes it sits in do not. That is #323.

### DIFF
--- a/Packages/OnymDesignTokens/README.md
+++ b/Packages/OnymDesignTokens/README.md
@@ -18,15 +18,25 @@ touching a line of app code.
 | `OnymTerminal` | The pinned-dark console block: surface, deep, phosphor text, ink, two overlays |
 | `OnymTint` | Pastel fills and the deep inks that read on them — hero tiles, the amber notice, governance badges |
 | `OnymRadius` | 10 corner steps — `card` `field` `panel` `hero` `inset` `control` `badge` `tile` `chip` `pill` — plus `shape(_:)` |
-| `OnymType` | `font(size:weight:)` and `mono(size:weight:)` — the text and mono faces |
+| `OnymType` | `font(size:weight:)`, `mono(size:weight:)`, `fixed(size:weight:)`, `scale(_:)` — the text and mono faces, and the reader's size setting |
 | `Color.dynamic(light:dark:)` | Helper for declaring a color that follows the system trait collection |
 
 Radius steps are named for what they wrap, not what they measure. An
 adopter setting `inset` to 16 should not need to know it used to be 12.
 
 `OnymType` carries no size scale on purpose — sizes still arrive from
-the call site. Swapping the typeface is two function bodies; imposing a
-scale on 465 call sites is a design decision, not a packaging one.
+the call site. Swapping the typeface is three function bodies; imposing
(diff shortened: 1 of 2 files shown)


dev-onym (commented):
APPROVE

- `OnymType.swift:83-86` — Ratio via `scaledValue(for: 100)/100` is a sound workaround for `scaledFont(for:)` rounding whole points at default; reasoning in the PR body checks out.
- `OnymType.swift:83` — Defaulting `traits` to `.current` is correct: SwiftUI sets `UITraitCollection.current` during view body evaluation (Apple's documented behavior); the PR body verified 2.82× at AX5.
- No new tests added for the scaling itself. `TokenContract.swift:152-156` only covers face-distinctness. Consider adding `scale(UITraitCollection(preferredContentSizeCategory: .large)) == 1.0` and a monotonic-growth assertion across categories — this is the load-bearing invariant the PR body describes ("13.5pt at exactly 13.5pt at default"), and a regression here would silently ship the exact bug the PR is written to avoid.
- `OnymType.swift:88-90` — `scaled` is called on every font invocation and rebuilds `UIFontMetrics(forTextStyle: .body)` each time. Trivial cost; not worth changing, but flagging in case it shows up in a hot path.
- `OnymType.swift:83` — `UITraitCollection.current` is a main-thread concept; a `Font` computed off-main (e.g., in a background helper) would snapshot 1.0. All ~440 existing call sites are inside view bodies, so it's fine today; worth keeping in mind if adopters do otherwise.
- `OnymType.swift:65-67` — `fixed` has no call sites yet, as the PR notes; unused API is acceptable here because the follow-up layout PR is what will wire it up.
- Doc example at `OnymType.swift:20-22` references `scaled(_:)` which is `private`; adopters replacing the bodies inline in the enum can still call it, so the example works, but readers replacing at the module boundary should note it.

PR #323, opened 2026-08-24
Boxes that give, and the ones that shouldn't

rinat-enikeev:
Second of four. Base: #322.

The type scales now, so the layout has to let it. **Rendering the shared components at AX5 found four defects that the counting beforehand had pointed at almost none of.**

My pre-work measurement said "21 fixed heights, 27 line limits". Most of those turned out to be hairlines (`height: 1`), spacers, and `minHeight` on text editors — all already fine. The real breakage was somewhere none of my greps looked.

### 29 SF Symbols scaling inside fixed frames

A 12pt glyph in a 22pt circle, a 44pt one in a 96pt seal. Scaled by 2.82× they overflow their own boxes badly. These are drawings, not words, and take `OnymType.fixed`. The rule is narrow: a glyph the reader needs meaning from belongs in `font`, and its box has to give instead.

### Two buttons pinned around labels that outgrow them

52pt and 44pt heights around labels that reach 45pt and 41pt. They set a minimum now, and pad.

### 13 line limits lifted, 14 deliberately kept

Prose — group names, member names, subtitles — lifts at accessibility sizes via `onymLineLimit`. Keys and URLs keep `lineLimit(1)`: a truncated relay URL is kinder than sixty characters wrapped across a card, and a pubkey is for comparing, not reading. `Row` does both, so its title relaxes when it is a label and holds when it is mono.

### Tiles and title

Icon tiles stayed 30pt beside 46pt text, which reads as a bug rather than a choice. They track the text now, capped — a tile on the full curve would be 85pt of decoration crowding out what it decorates. The glyph is sized off whatever the tile settled on, so the two cannot disagree.

The large title broke mid-word: "Setting" above a lonely "s". Its minimum scale factor needed a line limit to work against — given unlimited lines SwiftUI wraps rather than shrinks, and wrapping was the thing being avoided.

### One limitation worth knowing

The fonts read the UIKit trait collection; the line limits read the SwiftUI environment. On device the system sets both and they agree. A SwiftUI-only `.environment(\.dynamicTypeSize, …)` override moves the second and not the first, so a test or preview has to set both. The app has **zero** previews and overrides dynamic type **nowhere**, so this costs nothing today — it would be the reason to move onto `@ScaledMetric` if that ever changes.

### Not decided here

At accessibility sizes rows keep tile and text side by side, leaving text ~240pt. Apple's pattern stacks the tile above the text at those sizes. That changes how every settings row looks for accessibility users, so it is a design call rather than a mechanical one — happy to add it if wanted.

### DIFF
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatEmptyStateView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatEmptyStateView.swift
@@ -91,7 +91,7 @@ struct ChatEmptyStateView: View {
     private func benefit(icon: String, title: LocalizedStringKey, detail: LocalizedStringKey) -> some View {
         HStack(alignment: .top, spacing: 12) {
             Image(systemName: icon)
-                .font(OnymType.font(size: 16, weight: .semibold))
+                .font(OnymType.fixed(size: 16, weight: .semibold))
                 .foregroundStyle(OnymAccent.blue.color)
                 .frame(width: 24, height: 22)
             VStack(alignment: .leading, spacing: 2) {
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/ChatMembersView.swift
@@ -262,7 +262,7 @@ struct ChatMembersView: View {
                         Text(group.name)
                             .font(OnymType.font(size: 18, weight: .semibold))
                             .foregroundStyle(OnymTokens.text)
-                            .lineLimit(1)
+                            .onymLineLimit(1)
                         Image(systemName: "pencil")
                             .font(OnymType.font(size: 13, weight: .semibold))
                             .foregroundStyle(OnymTokens.text3)
@@ -274,7 +274,7 @@ struct ChatMembersView: View {
                 Text(group.name)
                     .font(OnymType.font(size: 18, weight: .semibold))
                     .foregroundStyle(OnymTokens.text)
-                    .lineLimit(1)
+                    .onymLineLimit(1)
             }
         }
         .padding(.top, 16)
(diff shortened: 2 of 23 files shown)


dev-onym (commented):
APPROVE

- `Packages/OnymDesignTokens/Sources/OnymDesignTokens/OnymLineLimit.swift:36` — modifier is clean; the `relaxing` opt-out correctly threads through `Row` at `Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift:410` for mono keys.
- `Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift:28-46` — `@ScaledMetric` with `min(scaledSize, 46)` cap; glyph is sized off `size` (post-cap), so tile and glyph can't drift.
- `Packages/OnymDesign/Sources/OnymDesign/SettingsDesign.swift:169-170` — `lineLimit(2)` paired with `minimumScaleFactor(0.55)` is the right SwiftUI dance; the comment correctly notes wrapping wins over shrinking without a limit.
- `Packages/OnymGroup/Sources/OnymGroup/CreateGroupView.swift:132-142,158-165` — swapping fixed `height` for `minHeight` + vertical padding + `multilineTextAlignment(.center)` grows around the label as claimed. Only nit: the padding (14/12) plus text at rest still yields ~52/44pt, so no visual regression at default type.
- Consistency check on 8 files using `onymLineLimit` — all 4 packages (`OnymChatsUI`, `OnymGroup`, `OnymSettings`, `OnymDesign`) already depend on `OnymDesignTokens`, so no missing imports. `Sources/OnymIOS/OnboardingSteps.swift:1921` also has access via app target.
- `Packages/OnymGroup/Package.resolved` — newly tracked, but `Packages/OnymChatsCore`, `OnymSettings`, `OnymRecovery`, etc. all commit theirs; not gitignored. Consistent with repo convention.
- Non-blocking observation: several `Image(systemName:) .font(OnymType.font(...))` remain (e.g. `ChatMembersView.swift:266,394,436`, `CreateGroupView.swift:508` `xmark.circle.fill` clear button, `ChatThreadView.swift:795` `xmark` in a `.frame(width:24,height:24)`-style container). Some are meaningful glyphs that should scale; the xmark ones in constrained frames are worth a second look at AX5, but likely the "not decided here" set the PR alludes to.

PR #324, opened 2026-08-24
The chat thread joins the design system it was drawn from

rinat-enikeev:
Third of four. Base: #323.

**The premise of this PR was wrong and the audit corrected it.** Counting `UILabel` occurrences had suggested eighteen labels needed Dynamic Type turned on. Checking each one found twelve already scaled and only eight did not — the chat layer was in better shape than the count implied.

### What the count missed entirely

**None of the twenty font assignments went through the token layer.** Every one reached for `UIFont` directly, so the most-read screen in the app was the one place an adopter's typeface would never arrive. Nineteen route through `OnymType` now; the twentieth inherits from a sibling.

`OnymType` grows a UIKit half to make that possible: `uiFont`, `uiMono`, `uiMonoDigit`, `uiFixed`, `uiBody`.

### The same trap, a second time

These go through `UIFontMetrics.scaledFont` rather than the ratio the SwiftUI side uses, because that is what `adjustsFontForContentSizeCategory` tracks. And they pass `compatibleWith: .current`, for the same reason `scale()` has to:

> `scaledFont(for:)` alone ignores the current trait collection. **Measured: 12pt in, 12pt out, at the largest accessibility size.**

That is the second time this trap has appeared in this API, in the second of the two places it could. Both are now covered by tests in #325.

### Eight fixed, two deliberately not

A duration plate on a video thumbnail and an overflow count on a photo tile cannot grow 2.8× without leaving the artwork they sit on. The grid cannot give, so **the count is spoken instead of only drawn** — the overflow label carries an accessibility label now, which it did not before.

The voice message row pinned itself to 40pt around a duration that scales. Forty is a floor now.

### A correction worth reading

The composer sizes itself to `ceil(font.lineHeight) * 3`, and giving it a metrics-derived font in place of the text-style one left it two points taller than three lines of its own text. A test caught it. `OnymType.uiBody` exists for the places where the app *measures* text rather than merely drawing it.

My first explanation for this was wrong, and I want it on the record rather than quietly fixed: the two fonts do **not** carry different leading. They report identical metrics — 17pt, `lineHeight` 20.287, `ceil` 21 — and lay out differently anyway. Why is not established. What is known is that the arithmetic agrees with one of them and not the other, and which one. The doc comment and the test say exactly that and no more.

### DIFF
--- a/Packages/OnymChatsUI/Package.resolved
+++ b/Packages/OnymChatsUI/Package.resolved
@@ -1,0 +1,14 @@
+{
+  "pins" : [
+    {
+      "identity" : "onym-sdk-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/onymchat/onym-sdk-swift.git",
+      "state" : {
+        "revision" : "b24a3be023f7445777598ce280a84907b32f7452",
+        "version" : "0.0.2"
+      }
+    }
+  ],
+  "version" : 2
+}
--- a/Packages/OnymChatsUI/Sources/OnymChatsUI/AlbumGridView.swift
+++ b/Packages/OnymChatsUI/Sources/OnymChatsUI/AlbumGridView.swift
@@ -116,7 +116,10 @@ final class AlbumGridView: UIView {
             addSubview(playGlyph)
 
             overflowLabel.translatesAutoresizingMaskIntoConstraints = false
-            overflowLabel.font = .systemFont(ofSize: 22, weight: .semibold)
+            // Fixed: this sits centred on a photo tile in a grid of fixed
+            // cells. Growing it 2.8× would spill out of the tile. The
+            // count is announced instead — see accessibilityLabel below.
+            overflowLabel.font = OnymType.uiFixed(size: 22, weight: .semibold)
             overflowLabel.textColor = .white
             overflowLabel.textAlignment = .center
             overflowLabel.backgroundColor = OnymTokens.scrimUI.withAlphaComponent(0.45)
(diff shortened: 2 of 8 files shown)


dev-onym (commented):
COMMENT

- **AlbumGridView.swift:143 vs :159-162** — the `Tile` sets `isAccessibilityElement = true`, so VoiceOver treats it as a single element and never descends into subviews. Setting `overflowLabel.isAccessibilityElement = true` and `.accessibilityLabel = "\(n) more"` on a subview of an accessibility-element container has no effect; the count still won't be spoken. Fix by putting the count on the Tile's own `accessibilityLabel` when `overflowCount > 0` (e.g. "Photo, \(n) more") or by moving to `accessibilityElements`.

- Scaling-curve change is real and worth being explicit about: previously `caption1`/`caption2`/`footnote`/`subheadline` labels scaled along their own text-style curves via `UIFontMetrics(forTextStyle: .caption1)`; now they all scale along `.body` through `OnymType.uiFont` → `scaled(_:)` (OnymType.swift:139). At AX sizes the smaller-style curves grow more than body — expect the reply banner snippet, quote name/snippet, failure label, join-request caption, and system notice to all grow *less aggressively* than before at large settings. Matches the SwiftUI-side "one curve" rationale; just confirm this was intentional for the smaller styles too.

- `OnymType.uiFont` docstring (OnymType.swift:83-85) says every size passed today is whole. `uiFont(size: 13)` and similar pass through fine, but the mono/monoDigit sizes at 11/12/15 are also whole — good. If anyone later reaches for a half-point, `scaledFont` will silently round; the doc warns, no action needed.

- `Package.resolved` new file matches the pattern of the other seven packages in the repo — fine.

- Nit: `ChatInputPanelView.swift:280` swaps `.preferredFont(forTextStyle: .subheadline)` for `OnymType.uiFont(size: 15)`. `subheadline` is 15pt at default, so points-match, but same curve caveat as above.

Verdict: approvable after the AlbumGridView accessibility fix; everything else is either intentional-per-design or minor.

PR #325, opened 2026-08-24
Tests for the two traps, and one that could not be written honestly

rinat-enikeev:
Last of four. Base: #324.

Twenty-four tests across the two packages, guarding things that would otherwise fail silently at a setting no reviewer has switched on.

### Two exist because the obvious implementation is wrong

Both compile, both read plausibly, and neither is visible at the default text size:

- **`UIFontMetrics.scaledFont` rounds to whole points at the default setting.** 11.5pt comes back 12pt. Sixty half-point sizes would shift for readers who never touched anything.
- **`scaledValue` with no trait collection answers 1.0 at every setting.** It looks exactly like working Dynamic Type support and delivers none.

Both now fail loudly if someone reaches for the simpler-looking call.

### Layout tests

These began as a throwaway harness for looking at the render, and earned a place by finding four defects that counting call sites had pointed at none of. They assert the settings screen grows with the reader, a tile grows without becoming 85pt of decoration, and nothing escapes its width instead of wrapping.

### One test is deliberately absent

That `LargeTitle` shrinks a long word rather than breaking it mid-word is real, and was confirmed by looking at the render. It could not be asserted honestly: `minimumScaleFactor` shrinks the wrapped case too, so a one-line title and a two-line one land close enough in height that any threshold between them is a number chosen to make the test pass. The absence is written down where the test would be. A brittle test that agrees with the truth today is worth less than a note saying why there isn't one — snapshot testing would settle it, and the repo has no harness to hang that on.

### Two things removed rather than kept green

- A contract test resolved a SwiftUI `Font`'s point size through a helper that **always returned nil**, so it compared `-1` to `-1` and passed without measuring anything.
- A test asserting `uiBody` lays out to the height its own metrics promise turned out to assert something **false** — the promise holds in the composer's configuration, not in a bare text view. What survives asserts only the font's identity and points at the composer's own test for the behaviour, which is where it was caught to begin with.

CI already runs both package suites from the token stack, so these are picked up without further wiring.

### DIFF
--- a/Packages/OnymDesign/Tests/OnymDesignTests/LayoutGrowthTests.swift
+++ b/Packages/OnymDesign/Tests/OnymDesignTests/LayoutGrowthTests.swift
@@ -1,0 +1,111 @@
+import SwiftUI
+import UIKit
+import XCTest
+@testable import OnymDesign
+
+/// Whether the shared components actually grow when the reader asks
+/// them to.
+///
+/// This began as a throwaway harness for looking at the layout pass, and
+/// earned a place by finding four defects that counting call sites had
+/// pointed at none of: glyphs overflowing fixed boxes, buttons pinned
+/// around labels that outgrew them, tiles stranded at 30pt beside 46pt
+/// text, and a title breaking mid-word into "Setting" over a lonely "s".
+///
+/// On device the system sets both the UIKit trait collection — which the
+/// fonts read — and the SwiftUI environment, which the line limits read.
+/// A test has to set both, or the two disagree and it measures a state
+/// no reader is ever in.
+@MainActor
+final class LayoutGrowthTests: XCTestCase {
+
+    private struct Sample: View {
+        var body: some View {
+            VStack(alignment: .leading, spacing: 0) {
+                LargeTitle("Settings")
+                SectionLabel("TRANSPORT")
+                Card {
+                    Row(title: "Relayers",
+                        subtitle: "wss://relay.onym.app",
+                        subtitleMono: true) {
+                        IconTile(symbol: "antenna.radiowaves.left.and.right",
+                                 bg: OnymTile.indigo)
+                    }
+                    Row(title: "Notifications and alerts",
+                        subtitle: "Wake this device when a message arrives",
+                        last: true) {
+                        IconTile(symbol: "bell.fill", bg: OnymTile.red)
+                    }
+                }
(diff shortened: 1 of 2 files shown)


dev-onym (commented):
APPROVE

- `Packages/OnymDesignTokens/Tests/OnymDesignTokensTests/DynamicTypeTests.swift:41` — the default-setting guard against `scale()` rounding is a strong assertion; it catches the exact `UIFontMetrics.scaledFont` regression the module was written to dodge, and the enumerated half-point sizes match the app's real spend.
- `DynamicTypeTests.swift:57` — the ratio-monotonicity check plus `< 1.0` at extraSmall and `> 2.0` at AX5 is the cleanest possible witness that `scaledValue(for:compatibleWith:)` is being used correctly rather than the silently-1.0 form.
- `DynamicTypeTests.swift:77` — `drawnHeight` sensibly builds `Font` inside `performAsCurrent`, so scaling is baked in before `ImageRenderer` sees it; independent of whether the renderer itself respects the trait.
- `DynamicTypeTests.swift:63` — `XCTAssertNotEqual(scaled, scaled.rounded(), accuracy: 0.0001)` is unusual but correct (asserts distance > tolerance); worth a one-line comment for the next reader.
- `DynamicTypeTests.swift:139` — asserting only font *identity* for `uiBody`, and pointing at `ChatInputPanelViewTests.test_textGrows_capsAtMaxLineCount` for the layout behaviour, is honest given the container-vs-bare-view mismatch documented in `OnymType.swift:100`.
- `Packages/OnymDesign/Tests/OnymDesignTests/LayoutGrowthTests.swift:56` — combining `UITraitCollection.performAsCurrent` with `.environment(\.dynamicTypeSize, …)` is the correct dual-set; single-side setups routinely measure a state no reader is in.
- `LayoutGrowthTests.swift:74` — `1.8×` growth threshold is a heuristic but comfortably below expected (~2.7× given ~2.82 body multiplier and small fixed padding); low flake risk.
- `LayoutGrowthTests.swift:87` — 47pt cap check aligns with `IconTile.size = min(scaledSize, 46)` at `SettingsDesign.swift:29`.
- Deliberate-absence note at `LayoutGrowthTests.swift:100` and the removed-tests rationale in the PR body are the right call — a brittle snapshot-substitute test would age worse than the comment.
- Minor: PR body says "twenty-four tests" but the diff adds 12 (10 + 2). Not blocking.

